PR #27, opened 2020-03-09
Change oauth example to use supported version

tylerauerbeck:
This changes the oauth example CR to use the dex repo and version that the original PR was merged into.

### DIFF
--- a/examples/argocd-oauth.yaml
+++ b/examples/argocd-oauth.yaml
@@ -6,6 +6,6 @@ metadata:
     example: defaults
 spec:
   dex:
-    image: quay.io/dexidp/dex 
-    version: v2.22.0
+    image: quay.io/redhat-cop/dex 
+    version: v2.22.0-openshift
     openShiftOAuth: true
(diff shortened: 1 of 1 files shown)


tylerauerbeck:
Probably want to put the breaks on this as it looks like there are some issues using the dex image for this right now.

jmckind:
Thanks for the PR, keeping an eye on this as well

tylerauerbeck:
@jmckind Now that Andy has moved this to a more formal location (and until we can find a better integration plan with the upstream dex folks) would it make sense to point this to quay.io/redhat-cop/dex:v2.22.0-openshift ?

jmckind:
@tylerauerbeck That makes sense, we just need to test it against that new repo and I'm good with that change. The oauth example changed a little bit sinde this PR was submitted.

tylerauerbeck:
Sounds good. Will take this for a spin now and I'll ping you again once I confirm things are in good shape.

jmckind:
@tylerauerbeck If you want to update this PR with the new information, I will merge it. I think we said the following was the correct image and tag?

```
  dex:
    image: quay.io/redhat-cop/dex
    version: v2.22.0-openshift
    openShiftOAuth: true
```

tylerauerbeck:
@jmckind Updated this to what you mentioned above.

PR #70, opened 2020-04-29
Changed passthrough insecure policy to redirect

sabre1041:
Changed passthrough insecure policy to redirect


jmckind:
Thanks for this. Per our slack conversation, this is a good general option for most. Ultimately I would like to add the ability to pass through the `TLSConfig` options that one would set on a Route directly to allow total control over this.

PR #81, opened 2020-05-13
Updating to new version of the dex image

christianh814:
The dex Image moved to the Red Hat COP Repo. I believe @sabre1041 moved it a bit ago.

### DIFF
--- a/docs/reference/argocd.md
+++ b/docs/reference/argocd.md
@@ -160,9 +160,9 @@ metadata:
     example: openshift-oauth
 spec:
   dex:
-    image: quay.io/ablock/dex
+    image: quay.io/redhat-cop/dex
     openShiftOAuth: true
-    version: openshift-connector
+    version: v2.22.0-openshift
   rbac:
     defaultPolicy: 'role:readonly'
     policy: |


mcanevet:
Why can't we use the official v2.22.0 image? The OpenShift connector has been merged in this version. I'm using it with `openShiftOAuth: true` and it works fine.

tylerauerbeck:
@mcanevet What version of ArgoCD are you running. There were some breakages between versions of dex & argo-cd that were causing some issues. I believe from what I saw in the commits, this _should_ be fixed as of argo-cd v1.5.3? But we may want to give it a few additional releases before we bump this back to upstream (as this has been a bit up and down up to this point)

alexmt:
Dex v2.23.0 has an issue: https://github.com/dexidp/dex/issues/1695, so we had to use v2.22.0.  Any Argo CD version is compatible with v2.22.0 so it should be safe to use it.

mcanevet:
Yes, I'm using ArgoCD 1.5.2 with Dex 2.22.0 and it looks like everything works well so far.

sabre1041:
The OpenShift connector has a bug in 2.22.0 where the user is not properly mapped into ArgoCD. On the user information page, it will appear as a blank user

jmckind:
I am going to merge this for now as we know it works and at a minimum, it serves as documentation of overriding the image/tag if needed.

PR #86, opened 2020-05-26
Remove context: minikube

sbuvaneshkumar:
We are creating minikube with the context name argocd, but while setting context we're giving minikube/argocd.
I understand that if the user didn't specify the cluster name, the default name is used - minikube.

But I think it is better go as per the flow. 
It might create confusion (especially for the one, who simply copy paste from the doc and execute :)).


jmckind:
Thanks for the contribution @sbuvaneshkumar, this look good to me.

Issue #84, opened 2020-05-14
Add the ability to customize the OpenShift route that is created


tylerauerbeck:
Currently the operator provides the ability to customize the Ingress object that gets created, however right now the route setting in the operator is pretty vanilla. It would nice to be able to adjust some of the specifics of the Route object that gets created in a similar manner: `InsecureEdgeTerminationPolicy`, `host`, etc.

tylerauerbeck:
Actually looks like the settings I need are already present, but looks like some enhancement would be needed for some of the additional configs (TLS, etc.) per convo in https://github.com/argoproj-labs/argocd-operator/pull/70

Issue #74, opened 2020-04-29
Add Support for Custom TLSConfig on Routes


jmckind:
The operator should allow a user to specify the TLS configuration for each Route that the operator manages.

eldis80:
Hi. I was just checking an issue raised by our team. We'd like the argocd-server route to show same self-signed certificate as the routes we are using for other applications on OpenShift. So this would be a very welcome feature.

garyedwards:
Hey it would be nice to have this for ingress as well.  Use of annotations such as `cert-manager.io/cluster-issuer: letsencrypt-staging` for each `host` that is set (server & grpc).

garyedwards:
It feels like ingress is almost working except `spec > tls > hosts` is using the name from ArgoCD `metadata > name` rather then then using the `host` as set and used by http:

``` yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    cert-manager.io/cluster-issuer: letsencrypt-staging
    kubernetes.io/ingress.class: traefik
  creationTimestamp: "2020-05-14T15:59:10Z"
  generation: 1
  labels:
    app.kubernetes.io/name: example-argocd
    app.kubernetes.io/part-of: argocd
    example: ingress
  name: example-argocd
  namespace: my-argocd-operator
  ownerReferences:
  - apiVersion: argoproj.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: ArgoCD
    name: example-argocd
    uid: 35df8564-6f5f-49d8-874c-d2cee895b588
  resourceVersion: "97592"
  selfLink: /apis/extensions/v1beta1/namespaces/my-argocd-operator/ingresses/example-argocd
  uid: 432325f9-894e-46b7-85dd-651c46eff6af
spec:
  rules:
  - host: argocd.example.com
    http:
      paths:
      - backend:
          serviceName: example-argocd-server
          servicePort: http
        path: /
  tls:
  - hosts:
    - example-argocd
    secretName: argocd-secret
status:
  loadBalancer: {}
```

jmckind:
@eldis80 As I have been thinking about all of this more I wanted to see if your particular case, have you tried the `Server.Insecure` flag?

``` yaml
apiVersion: argoproj.io/v1alpha1
kind: ArgoCD
metadata:
  name: example-argocd
spec:
  server:
    insecure: true
    route: true
```

[...]

PR #91, opened 2020-05-27
Allow Override of Ingress and Route Resources

jmckind:
This PR aims to address a few issues (examples are #74 and #84) that have been raised regarding the ability to fully customize the Ingress and Route resources that are managed by the operator.

This introduces a breaking change to the ArgoCD spec but allows for much more flexibility.

#### Ingress

Here is an example of the current spec for managing Ingress.

``` yaml
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: ingress
spec:
  ingress:
    annotations:
      kubernetes.io/ingress.class: nginx
      nginx.ingress.kubernetes.io/rewrite-target: /static/$2
      cert-manager.io/cluster-issuer: letsencrypt
    path: /testpath
  server:
    ingress: true
    insecure: true
```

There is no way to override the TLS and other options for the Ingress resource. The proposal is to expose the IngresSpec directly to allow the ability to override all aspects of the Ingress.
 
Here is an example of using the proposed spec to explicitly set the defaults for the server Ingress resource.

``` yaml
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: ingress
spec:
  server:
    insecure: true
    ingress:
      enabled: true
      annotations:
        kubernetes.io/ingress.class: nginx
        nginx.ingress.kubernetes.io/rewrite-target: /static/$2
        cert-manager.io/cluster-issuer: letsencrypt
      spec:
        rules:
        - host: example-argocd
          http:
            paths:
            - backend:
                serviceName: example-argocd-server
                servicePort: http
              path: /
              pathType: ImplementationSpecific
        tls:
        - hosts:
          - example-argocd
          secretName: argocd-secret
```
The `Ingress` property on the ArgoCD spec would go away.

The `Ingress.Spec` field on the `Server`, `Grafana` and `Prometheus` properties of the ArgoCD resource would convert from simple boolean values to allow for full control over the Ingress resources for those components. `Enabled` flags are added to toggle the feature on/off as needed.

One issue that this brings up is that it is now much more verbose if someone just wants to change the path for the Ingress.

#### Route

The current ArgoCD spec currently offers even less for customizing Routes.

The same pattern will be applied to OpenShift Route resources. Here is an example of using the proposed spec to explicitly set the defaults for the server Route resource.

``` yaml
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: route
spec:
  server:
    route:
      enabled: true
      annotations:
        cert-manager.io/cluster-issuer: letsencrypt
      spec:
        host: example-argocd-server-argocd.apps.cluster.example.com
        port:
          targetPort: https
        tls:
          insecureEdgeTerminationPolicy: Redirect
          termination: passthrough
        to:
          kind: Service
          name: example-argocd-server
          weight: 100
        wildcardPolicy: None
```


jmckind:
@eldis80, @garyedwards, @tylerauerbeck and anyone else interested. Would this proposed change work for you?

garyedwards:
Hi @jmckind this proposal would defiantly allow me to do what I want. I would however echo your point about how much more verbose this solution would be.  As a work around at the moment I create the ingress manually and point it at the appropriate services and this is probably the correct thing to do if you want to make a lot of customisations.

The proposed change to the spec effectively allows for all of this to be in the in the CRD.  I am no expert on operator patterns but would question if this would be best practice regarding how much is abstracted etc.  Perhaps reusing the existing `host` variable for tls and allowing for annotations would be enough.  I think the paths should be dealt with by the operator internally.

It would defiantly work for me though and I really appreciate your work on this.

jmckind:
I am going to close this in favor of individual PRs for the Ingress and Route. The two cannot be treated exactly the same and simply exposing the underlying Spec puts too much on the end user of the operator when defining the ArgoCD custom resource.

Issue #83, opened 2020-05-14
Ability to utilize repository.credentials in initialRepositories


tylerauerbeck:
Right now when you add a repository via `initialRepositories` we have to specify credentials or keys for each individual repository that we would like to use. As of ArgoCD v1.4.X, we have the ability to specify a key to be used at a root of a repo (i.e org/project) level with `repository.credentials`. Not sure what the design of this would look like at this point, but I think it would be beneficial to be able to support both. 

Happy to help add functionality here, but wanted to get others input here first.

jmckind:
@tylerauerbeck I am looking through the [Repository Credentials](https://argoproj.github.io/argo-cd/operator-manual/declarative-setup/#repository-credentials) section of the Argo docs and I am trying to work though how to best support configuring this. I agree though that it would be nice to have the operator support this.

jmckind:
Support for this has been added in #105 and will go out in the next release.

PR #105, opened 2020-06-11
Add support for using `repository.credentials`

jacobsee:
Adds support for using the `repository.credentials` key, also called "credential templates" in Argo - very much implemented in the same way that `InitialRepositories` is.


jacobsee:
Addresses #83

jmckind:
Thanks for the PR @jacobsee! I will have a look this week.

PR #111, opened 2020-06-29
Mount repo server token

### LINKED ISSUE #110: Allow mounting of ServiceAccountToken / specifying of ServiceAccount in Repo Server
Currently the repo server component has `automountServiceAccountToken` set to false. When attempting to use the `lookup` function with helm, this causes us some issues because it has no idea how to communicate with the cluster. It would be nice to be able to override this if this was desired and also to specify the serviceAccount that I would want to use here.

tylerauerbeck:
This adds the ability to specify mounting the serviceaccount token to the repo-server pod as well as specifying the serviceaccount it should run as. This can be done by adding the following to the ArgoCD CR:

```
...
spec:
  repo:
    mountsatoken: true
    serviceaccount: argocd-application-controller
...
```

Resolves #110

### DIFF
--- a/docs/reference/argocd.md
+++ b/docs/reference/argocd.md
@@ -633,6 +633,8 @@ The following properties are available for configuring the Repo server component
 Name | Default | Description
 --- | --- | ---
 Resources | [Empty] | The container compute resources.
+MountSAToken | false | Whether the ServiceAccount token should be mounted to the repo-server pod.
+ServiceAccount | "" | The name of the ServiceAccount to use with the repo-server pod.
 
 ### Repo Example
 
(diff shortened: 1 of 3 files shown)


jmckind:
LGTM

bmhkb4:
These are not working for me.  It never updates the corresponding deployment with either value.  Any ideas?

`repo:
    mountsatoken: true
    serviceaccount: vault-secret-read
...    
...
...
`
Openshift: v4.10
Openshift Gitops Operator: v1.5.5
ArgoCD Vault Plugin: v1.12

bmhkb4:
```
securityContext:
  runAsUser: 999
```
is also not configurable from argocd CR

Issue #112, opened 2020-06-30
Missing 0.0.10 package in argocd-operator/deploy/olm-catalog/argocd-operator/


flickerfly:
I'm running my own catalog and still using the older folder based bundle format. I went to argocd-operator/deploy/olm-catalog/argocd-operator/ to grab that and found that version 0.0.10 isn't available in there. Could we get that generated?

jmckind:
Hey @flickerfly! I am working on that right now. :)

jmckind:
Like I mentioned in the [PR](https://github.com/argoproj-labs/argocd-operator/pull/113), I have rolled forward to v0.0.11 and it has been released along with the OLM artifacts. I will be adding the PRs to get v0.0.11 included in the Operator Hubs as well.

PR #113, opened 2020-06-30
Add OLM Artifacts for Operator v0.0.10

jmckind:
This PR addresses #112 and adds the OLM artifacts for the Argo CD operator v0.0.10.

### DIFF
--- a/deploy/catalog_source.yaml
+++ b/deploy/catalog_source.yaml
@@ -4,6 +4,6 @@ metadata:
   name: argocd-catalog
 spec:
   sourceType: grpc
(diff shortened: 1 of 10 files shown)


jmckind:
I have identified an issue with the operator at v0.0.10 during OLM upgrades when an existing Argo CD cluster is present. This only affects OLM upgrades and not other aspects of v0.0.10. Manual installations of v0.0.10 are not affected. I am going to roll this to v0.0.11 so there will not be a v0.0.10 version in the Operator Hub, it will go from v0.0.9 -> v0.0.11.

PR #114, opened 2020-07-01
Fix the route resources example

soharaki:
This PR fixes the example manifest of the OpenShift route resources.


jmckind:
Thanks for the PR @soharaki. However, the route spec has changed in v0.0.10, so the example is correct. Please see the [v0.0.10 release notes](https://github.com/argoproj-labs/argocd-operator/releases/tag/v0.0.10) for more information.

PR #123, opened 2020-07-07
Added kustomization in root dir referencing the deploy kustomization

jdbohrman:
I have added this commit to enable others to reference Kustomize from the root directory. You can use my fork as an example by running `kustomize build https://github.com/jdbohrman/argocd-operator`.


jmckind:
Thanks for your contribution @jdbohrman!

Issue #124, opened 2020-07-08
Argocd failed to deploy (openshift - argocd 0.0.11)


Numblesix:
Hello,

we are using OpenShift 4.4 deploying ArgoCD using the Operator fails starting from 0.0.11 because of this:
```
  - dependents:
    - group: rbac.authorization.k8s.io
      kind: PolicyRule
      message: namespaced rule:{"verbs":["get"],"apiGroups":[""],"resources":["endpoints"]}
      status: NotSatisfied
      version: v1beta1
    group: ""
    kind: ServiceAccount
    message: Policy rule not satisfied for service account
    name: argocd-redis-ha
    status: PresentNotSatisfied
    version: v1
```

Numblesix:
When checking the created Rolebindings i couldnt find the needed rolebinding 
```
➜  ~ oc get rolebindings
NAME                                            AGE
argocd-operator.v0.0.11-59498cfd4d-867fc7d49d   27m
argocd-operator.v0.0.11-5c4f7fbff8-6d77f974ff   27m
argocd-operator.v0.0.11-5c678c7876-5b6ccc9858   27m
argocd-operator.v0.0.11-79c565f65-65f799c44f    27m
argocd-operator.v0.0.11-7fb6c5c758-84475f47d    27m
system:deployers                                29m
system:image-builders                           29m
system:image-pullers                            29m
```
```
➜  ~ oc get rolebindings argocd-operator.v0.0.11-59498cfd4d-867fc7d49d -o yaml         
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
..
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: argocd-operator.v0.0.11-59498cfd4d
subjects:
- kind: ServiceAccount
  name: argocd-application-controller
  namespace: argocd
```
```
➜  ~ oc get rolebindings -o yaml argocd-operator.v0.0.11-5c4f7fbff8-6d77f974ff
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
...
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: argocd-operator.v0.0.11-5c4f7fbff8
subjects:
- kind: ServiceAccount
  name: argocd-server
  namespace: argocd
```
```
➜  ~ oc get rolebindings -o yaml argocd-operator.v0.0.11-5c678c7876-5b6ccc9858                                             
[...]

PR #128, opened 2020-07-13
Share ServiceAcount for Redis Components

jmckind:
This PR addresses #124 as a workaround for the upstream OLM issue that prevents two or more ServiceAccounts from having Roles that are duplicates. This will eventually get fixed in OLM but for now the Redis StatefulSet and HA Proxy deployment can use the same ServiceAccount, as the permissions required are the same. This can be changed in the future if a need arises to have different Roles for the two components.


Numblesix:
When will this version be pushed :) ?

PR #134, opened 2020-07-20
Update InitialSSHKnownHosts handling

### LINKED ISSUE #82: initialSSHKnownHosts overrides default SSHKnownHosts provided by ArgoCD
I believe this is currently working as designed, but I'd like to see if anyone has similar thoughts. Right now upstream ArgoCD provides the initial SSHKnownHosts for github, bitbucket, gitlab, etc. when you install the system. The current functionality of this operator overrides/removes them if you provide your own `initialSSHKnownHosts` to the operator. What I would like to see is something along the lines of:

- inject the default hosts as the upstream does
- append my custom list to these default hosts so that I don't have to "carry' along the default ones myself
- provide a key that does allow users to remove those default hosts in the case that they don't want them included

I believe this would look something like:

```
initialSSHKnownHosts:
  default-hosts: false
  keys: |
    bitbucket.org ssh-rsa AAAAB3NzaC...
    github.com ssh-rsa AAAAB3NzaC...
```

In this case, the "load" is much lighter for folks who want to opt out versus those who want to also include the default hosts (i.e. a single key versus numerous known hosts entries.

tylerauerbeck:
resolves #82 

This modifies `initialSSHKnownHosts` to have two different keys:

```yaml
initialSSHKnownHosts:
  excludedefaulthosts: false
  keys: |
    my-git.org: alkjasldkfjalskdfj
```

This allows users to either include/exclude the default ssh hosts that are included with ArgoCD by default. Previously if you wanted to provide your own set of keys, it would overwrite these default keys and you would be responsible for adding them back in if you wanted them.

Now the default behavior is if you don't exclude them, your specified keys will just be appended to the default keys.

### DIFF
--- a/pkg/controller/argocd/configmap.go
+++ b/pkg/controller/argocd/configmap.go
@@ -30,17 +30,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-const (
-	defaultKnownHosts = `bitbucket.org ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAubiN81eDcafrgMeLzaFPsw2kNvEcqTKl/VqLat/MaB33pZy0y3rJZtnqwR2qOOvbwKZYKiEO1O6VqNEBxKvJJelCq0dTXWT5pbO2gDXC6h6QDXCaHo6pOHGPUy+YBaGQRGuSusMEASYiWunYN0vCAI8QaXnWMXNMdFP3jHAJH0eDsoiGnLPBlBp4TNm6rYI74nMzgz3B9IikW4WVK+dc8KZJZWYjAuORU3jc1c/NPskD2ASinf8v3xnfXeukU0sJ5N6m5E8VLjObPEO+mN2t/FZTMZLiFqPWc/ALSqnMnnhwrNi2rbfg/rd/IpL8Le3pSBne8+seeFVBoGqzHM9yXw==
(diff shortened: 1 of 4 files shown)


pkg/controller/argocd/configmap.go
```diff
@@ -30,17 +30,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-const (
-	defaultKnownHosts = `bitbucket.org ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAubiN81eDcafrgMeLzaFPsw2kNvEcqTKl/VqLat/MaB33pZy0y3rJZtnqwR2qOOvbwKZYKiEO1O6VqNEBxKvJJelCq0dTXWT5pbO2gDXC6h6QDXCaHo6pOHGPUy+YBaGQRGuSusMEASYiWunYN0vCAI8QaXnWMXNMdFP3jHAJH0eDsoiGnLPBlBp4TNm6rYI74nMzgz3B9IikW4WVK+dc8KZJZWYjAuORU3jc1c/NPskD2ASinf8v3xnfXeukU0sJ5N6m5E8VLjObPEO+mN2t/FZTMZLiFqPWc/ALSqnMnnhwrNi2rbfg/rd/IpL8Le3pSBne8+seeFVBoGqzHM9yXw==
```
jmckind, line 34: Nice catch removing this old constant that was no longer in use.

PR #139, opened 2020-07-29
Release Operator v0.0.13

jmckind:
This PR updates the operator to v0.0.12, which includes the following.

- Updates to Server TLS Secret Should Sync With Argo Secret
- Add Properties for Redis HAProxy Component to control image/tag
- Allow users to either include/exclude the default ssh hosts that are included with ArgoCD by default


geoL86:
@jmckind When we can expect on a PR for community-operators hub?

PR #140, opened 2020-07-29
Update example with how to enable the route

christianh814:
Update the example on how to enable the route using the "new way"


bigkevmcd:
LGTM based on https://github.com/argoproj-labs/argocd-operator/blob/master/pkg/apis/argoproj/v1alpha1/argocd_types.go#L121 and https://github.com/argoproj-labs/argocd-operator/blob/master/pkg/apis/argoproj/v1alpha1/argocd_types.go#L248

PR #151, opened 2020-08-28
Fix for not updating when changing the ArgoCD resource.

bigkevmcd:
This changes the behaviour of the route reconciliation to allow it to patch the route objects when the Insecure flag changes.

This also adds a simple test for this behaviour.


jmckind:
LGTM

PR #161, opened 2020-09-25
Enable all-namespaces mode & cluster config

sbose78:
### Will not be merged, will be split into multiple PRs

**Why**

* The roles/rolebindings/service accounts responsible for powering an ArgoCD instance were not not created per instance of `ArgoCD` CR. This prevents the user from being able to install the ArgoCD control plane in namespaces where the operator workload itself wasn't deployed.

* The permissions given to argocd does not enable to sync cluster configuration like users/groups/volumes.

**What**
* Ensure namespace-scaped roles/rolebindings/service accounts for the Argo control plane deployments are created and reconciled for *every instance* of `ArgoCD`

* Ensure cluster-scoped objects are created per operand of `ArgoCD` so that they can be managed individually when needed.( not too picky on this though - I am having second thoughts on this )

* Support cluster config if `ARGOCD_CLUSTER_CONFIG_ENABLED` and `ARGOCD_CLUSTER_CONFIG_NAMESPACES` is set to set/populated.

* Enable "all-namespaces" OLM mode.

```
kind: ArgoCD
metadata:
  name: example
  namespace: sbose
spec:
  managementScope:
    cluster: true
```

**How**

- [x] Remove the ArgoCD control plane specific roles/rolebindings/service accounts from the operator control plane.
- [x] Add reconciliation logic for namespace-scoped permission objects.
- [x] Add reconciliation logic for cluster-scoped permission objects.
- [ ] End-to-end tests ( and unit tests ? ) 
- [ ] Update OLM bundle

*Try this!*

Login to a cluster and change to any namespace in your terminal

$ kubectl apply -f deploy/argo-cd/
$ kubectl apply -f deploy/crds/
$ operator-sdk run --local


pkg/controller/argocd/role.go
```diff
@@ -0,0 +1,86 @@
...
+	argoprojv1a1 "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
+	"github.com/argoproj-labs/argocd-operator/pkg/common"
+	v1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// newRoleWithName creates a new ServiceAccount with the given name for the given ArgCD.
+func newRoleWithName(name string, cr *argoprojv1a1.ArgoCD) *v1.Role {
```
sbose78, line 15: copied code, can be made better.

pkg/controller/argocd/rolebinding.go
```diff
@@ -0,0 +1,91 @@
...
+	argoprojv1a1 "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
+	"github.com/argoproj-labs/argocd-operator/pkg/common"
+	v1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// newRoleBinding returns a new RoleBinding instance.
+func newRoleBinding(cr *argoprojv1a1.ArgoCD) *v1.RoleBinding {
```
sbose78, line 15: copied code, can be improved.

pkg/controller/argocd/argocd_controller.go
```diff
@@ -64,7 +66,8 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileArgoCD{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	kc, _ := k8s.NewForConfig(mgr.GetConfig())
```
sbose78, line 69: wish to figure out a way to avoid adding a second client..

pkg/controller/argocd/util.go
```diff
@@ -682,6 +693,14 @@ func watchResources(c controller.Controller) error {
 		return err
 	}
 
+	if err := watchOwnedResource(c, &v1.Role{}); err != nil {
+		return err
```
bigkevmcd, line 697: This is using go 1.14 so we could improve the errors here?

pkg/controller/argocd/argocd_controller.go
iam-veeramalla: return statement(&ReconcileArgoCD) and return type(reconcile.Reconciler) are different.

pkg/controller/argocd/policy_rule.go
```diff
@@ -0,0 +1,341 @@
+package argocd
```
iam-veeramalla, line 1: Looks like this file is a exact duplicate of policyrule.go , can we remove one ?

deploy/crds/argoproj.io_argocds_crd.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
iam-veeramalla, line 1: apiextensions.k8s.io/v1 does not support fields like validation, it should be replaced with schema field.

deploy/crds/argoproj.io_argocds_crd.yaml
```diff
@@ -164,844 +164,854 @@ spec:
...
-                          associated with an Ingress.
-                        properties:
-                          hosts:
-                            description: Hosts are a list of hosts included in the
-                              TLS certificate. The values in this list must match
-                              the name/s used in the tlsSecret. Defaults to the wildcard
-                              host setting for the loadbalancer controller fulfilling
-                              this Ingress, if left unspecified.
-                            items:
+                      tls:
```
iam-veeramalla, line 167: The new changes looks to have broken the indentation in many places. Can we fix them ?

error: error validating "deploy/crds/argoproj.io_argocds_crd.yaml": error validating data: [ValidationError(CustomResourceDefinition.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.grafana.properties.ingress.properties.type): invalid type for io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.JSONSchemaProps: got "string", expected "map", ValidationError(CustomResourceDefinition.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.grafana.properties.ingress): unknown field "resources" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.JSONSchemaProps, ValidationError(CustomResourceDefinition.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.grafana.properties.ingress): unknown field "route" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.JSONSchemaProps, ValidationError(CustomResourceDefinition.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.grafana.properties.ingress): unknown field "size" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.JSONSchemaProps, ValidationError(CustomResourceDefinition.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.grafana.properties.ingress): unknown field "version" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.JSONSchemaProps, ValidationError(CustomResourceDefinition.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.grafana.properties.required): invalid type for io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.JSONSchemaProps: got "array", expected "map", ValidationError(CustomResourceDefinition.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.grafana.properties.type): invalid type for io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.JSONSchemaProps: got "string", expected "map"]; if you choose to ignore these errors, turn validation off with --validate=false

iam-veeramalla (changes requested):

sbose78:
Note, I'm not planning on getting this merged.

pkg/controller/argocd/role.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	return nil
+}
+
+// reconcileClusterRole
+func (r *ReconcileArgoCD) reconcileClusterRole(name string, policyRules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD) error {
+	rbacClient := r.kc.RbacV1()
+
+	role, err := rbacClient.ClusterRoles().Get(context.TODO(), generateResourceName(name, cr), metav1.GetOptions{})
```
iam-veeramalla, line 126: we need to replace generateResourceName with `name` as generateResourceName is only used to generate local role name not cluster role name.

// code that is working for me
role, err := rbacClient.ClusterRoles().Get(context.TODO(), name, metav1.GetOptions{})

pkg/controller/argocd/role.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	if err := r.reconcileRole(redisHa, policyRuleForRedisHa(), cr); err != nil {
+		return err
+	}
+
+	rules := []v1.PolicyRule{}
+
+	if cr.Spec.ManagementScope.Cluster != nil && *cr.Spec.ManagementScope.Cluster {
+		if os.Getenv("ARGOCD_CLUSTER_CONFIG_ENABLED") == "true" &&
```
iam-veeramalla, line 104: Should the customer add this to argocd CR along with cr.Spec.ManagementScope.Cluster filed?

sbose78: Sorry, had missed this :(

If we add this to the CRD, users with low privileges would be able to spin up an ArgoCD for privilege escalation. 


Therefore, I suggest, supporting only one "ARGOCD_CLUSTER_CONFIG_NAMESPACES".

If set with the list of namespaces, use it. 
If unset, it implies that the admin has disabled cluster config argocd altogether/

wtam2018:
Closing this PR as it has been addressed by #227 #225 #234

PR #167, opened 2020-09-30
Support cert-utils-operator with the CA generation.

### LINKED ISSUE #166: Support for cert-utils-operator
https://github.com/argoproj-labs/argocd-operator/blob/2f087b7f879c0b5469faeaa26b7c771b834f5d55/pkg/controller/argocd/secret.go#L85

We are attempting to setup a route a with re-encrypt so that we can use certificates generated by cert-manager with our custom enterprise CA. When adding the annotation **cert-utils-operator.redhat-cop.io/destinationCA-from-secret: "argocd-devops-ca"** cert-utlis is expecting the secret to have a key named ca.crt, but currently the key is being named tls.crt. Would it be possible to either rename the existing key to ca.crt, or create an additional key with the same value named ca.crt?

bigkevmcd:
This adds support for putting the CA generated into a key that can be picked up
by cert-utils-operator.

Fixes #166

### DIFF
--- a/pkg/controller/argocd/secret.go
+++ b/pkg/controller/argocd/secret.go
@@ -81,9 +81,11 @@ func newCASecret(cr *argoprojv1a1.ArgoCD) (*corev1.Secret, error) {
 		return nil, err
 	}
 
+	// This puts both ca.crt and tls.crt into the secret.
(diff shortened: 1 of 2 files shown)


jmckind:
LGTM

Issue #122, opened 2020-07-07
configmap argocd-tls-certs-cm does not get populated with cr.spec.tls.initialCerts (v0.0.9)


ThomasVitt:
Hello, 
we want to insert a root CA certificate into argocd repo server in order to connect to our private Git Repos without the "insecure" option. According to the documentation (https://argocd-operator.readthedocs.io/en/latest/reference/argocd/#tls-options) we are using the following spec: 

```yaml
spec:
  tls:
    initialCerts:
      root-ca.example.com: |-
        -----BEGIN CERTIFICATE-----
        MIIGbzCCBFegAwIBAgIQeM4J2V9oU5dCeqHWP/0dXzANBgkqhkiG9w0BAQsFADBt
        ...
        -----END CERTIFICATE-----
```

But the configmap "argocd-tls-certs-cm" contains no data after configuring the argocd object: 

```
[user@host ~]$ oc get cm argocd-tls-certs-cm
NAME                  DATA   AGE
argocd-tls-certs-cm   0      15m
```

Is it a bug or am I doing something wrong? 
The documentation says something about "The following properties are available for configuring the Grafana component.". But I think this is not for grafana, it's supposed to be for the repo-server, isn't it?

Thanks!

geoL86:
@ThomasVitt Hi, it works for me (v0.12.0) and your yaml looks working.

jdziedzic:
I have found that deleting the config-map and allowing the operator to re-create it resolves this. However, the server containers don't seem to pickup the change until they are rolled. Perhaps there is a watch missing on the config-map for changes to initialCerts? Or was the intention that this value would only be set at initial cluster creation and not updated after the ArgoCD cluster is up?

lsambolino:
Hello, it seems we're facing the same issue in our OpenShift cluster.
We have written the ArgoCD yaml as the following but the ConfigMap is then cleared to 0 bytes just few seconds later:
`apiVersion: argoproj.io/v1alpha1
kind: ArgoCD
metadata:
  creationTimestamp: "2021-11-03T17:54:27Z"
  finalizers:
  - argoproj.io/finalizer
  generation: 345
  name: openshift-gitops
  namespace: openshift-gitops
  resourceVersion: "366676868"
  uid: 81e50515-67e8-4b9c-9115-d2fce79fb89f
spec:
  applicationSet:
    resources:
      limits:
        cpu: "2"
        memory: 1Gi
      requests:
        cpu: 250m
[...]

PR #168, opened 2020-09-30
Fix reconciliation of argocd-tls-certs-cm.

bigkevmcd:
This fixes a reconciliation bug, where the TLS cert configmap wasn't updated on subsequent reconciliations.

Fixes: #122


bigkevmcd:
I'm going to try and switch these branches to using testenv.

bigkevmcd:
Switching to testenv is tricky because of the amount of reconciliation :-|

jmckind:
LGTM

jmckind:
@bigkevmcd Looks good but a couple more test conflicts here as well if you could review. Thanks!

PR #176, opened 2020-10-06
Changes to support ArgoCD v1.7.7.

### LINKED ISSUE #148: Add Support for Argo CD v1.7+
Argo CD v1.7.1 is the  current release and the operator should deploy this version by default.

bigkevmcd:
* Fix the repo-server deployment
   This updates the repo server to have an emptyDir mounted /app/config/gpg/keys

   This comes from this change https://github.com/argoproj/argo-cd/pull/4136/files

 * Fix for argocd-controller.
   This is arguably a bug in ArgoCD it's parsing an empty string into a
   clearing of the map for resource customizations.

Add go-cmp to the dependencies.

This should fix #148

### DIFF
--- a/pkg/common/defaults.go
+++ b/pkg/common/defaults.go
@@ -31,7 +31,7 @@ const (
 	ArgoCDDefaultArgoImage = "argoproj/argocd"
 
 	// ArgoCDDefaultArgoVersion is the Argo CD container image digest to use when version not specified.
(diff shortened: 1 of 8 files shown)


jmckind (approved):

jmckind:
@bigkevmcd This looks good to me but can you have a look at the conflict in the deployment test?

PR #187, opened 2020-10-14
Fix issue with replacing the resource configuration

bigkevmcd:
Fix for not updating the mount volumes when a deployment is changed.

This checks that the volumes and volume mounts are replaced if they differ.


jmckind:
LGTM

PR #190, opened 2020-10-20
Fixed port-forward command

afritzler:
The example described the in the docs isn't working without adding `--namespace argocd` to the `port-forward` directive.


bigkevmcd (approved):
Looks good, thank you for the fix.

PR #193, opened 2020-10-22
Add support for creating an empty gpg-keys map.

bigkevmcd:
This adds support for creating an empty gpg-keys configmap.


jmckind (approved):
LGTM

PR #194, opened 2020-10-23
Add support for proxies in the repo-server deployment.

bigkevmcd:
This adds support for copying the proxy configuration from the running operator process into the deployments created by the operator.

It's case insensitive, and copies over the `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` fields.

### DIFF
--- a/pkg/controller/argocd/deployment.go
+++ b/pkg/controller/argocd/deployment.go
@@ -17,8 +17,10 @@ package argocd
 import (
 	"context"
 	"fmt"
+	"os"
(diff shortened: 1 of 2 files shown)


jmckind (approved):
LGTM

dirien:
Hi, are there any updates regarding the release date of the 0.0.15 version of the operator? We use this on our internal OCP platform and not able to set the proxy is killing us to use argoc :(

@rehbergtr, could be a blocker for you too!

PR #195, opened 2020-10-28
Fixed wrong `TLS` key example for server route

xt99:
As per https://github.com/argoproj-labs/argocd-operator/blob/master/pkg/apis/argoproj/v1alpha1/argocd_types.go#L262, TLS key for `ArgoCDRouteSpec` should be lowercased (`tls` instead of `TLS`)


jmckind:
LGTM, thanks for the update @xt99.

PR #198, opened 2020-11-06
Add support for getting the images from the environment.

bigkevmcd:
This adds functionality for fetching the images from the configured environment variables for the operator process.

* ARGOCD_DEX_IMAGE
* ARGOCD_IMAGE
* ARGOCD_REDIS_HA_IMAGE
* ARGOCD_REDIS_IMAGE
* ARGOCD_GRAFANA_IMAGE

The resolution order for these images:

1. From the ArgoCD object configuration (there are spec entries for each image).
2. From the environment variables above.
3. From the hard-coded values in pkg/common/defaults.go

### DIFF
--- a/pkg/common/keys.go
+++ b/pkg/common/keys.go
@@ -160,4 +160,24 @@ const (
 
 	// ArgoCDKeyUsersAnonymousEnabled is the configuration key for anonymous user access.
 	ArgoCDKeyUsersAnonymousEnabled = "users.anonymous.enabled"
+
+	// ArgoCDDexImageEnvName is the environment variable used to get the image
+	// to used for the Dex container.
+	ArgoCDDexImageEnvName = "ARGOCD_DEX_IMAGE"
+
+	// ArgoCDImageEnvName is the environment variable used to get the image
+	// to used for the argocd container.
+	ArgoCDImageEnvName = "ARGOCD_IMAGE"
+
+	// ArgoCDRedisHAImageEnvName is the environment variable used to get the image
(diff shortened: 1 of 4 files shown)


sbose78 (approved):

pkg/common/keys.go
```diff
@@ -160,4 +160,24 @@ const (
 
 	// ArgoCDKeyUsersAnonymousEnabled is the configuration key for anonymous user access.
 	ArgoCDKeyUsersAnonymousEnabled = "users.anonymous.enabled"
+
+	// ArgoCDDexImageEnvName is the environment variable used to get the image
+	// to used for the Dex container.
+	ArgoCDDexImageEnvName = "ARGOCD_DEX_IMAGE"
```
wtam2018, line 166: Do we need to document these environment variables possibly in docs/reference/argocd.md?

bigkevmcd: Good point, I guess I'd originally assumed we'd do that as part of the image declarations in the yaml, but we can do it here.

How do you think we should call this out, with a table at the start, with a list of keys (within the CRD) and their equivalent environment-variables? Or should each image (they are spread out through the document) indicate that the YAML overrides the environment variable?

wtam2018: I like the latter a bit more.   For example, https://argocd-operator.readthedocs.io/en/latest/reference/argocd/#dex-options here we can add documentation in the description of `Image` to state something like the default is overridden by `ARGOCD_DEX_IMAGE`  environment variable.    I think it is also helpful to document how/where to set an environment.

PR #199, opened 2020-11-10
Create workload permission objects per instance of ArgoCD CR

sbose78:
Broke away from https://github.com/argoproj-labs/argocd-operator/pull/161

### **Why**

The roles/rolebindings/service accounts responsible for powering an ArgoCD instance were not not being created per instance of ArgoCD CR. This prevented the user from being able to install the ArgoCD control plane in namespaces where the operator workload itself wasn't deployed.

With this change, you could create an `ArgoCD` CR in any namespace and get yourself an ArgoCD instance!
 

### **What**

- All roles, rolebindings, clusterrolebindings and SAs associated with the control plane objects would be owned by the ArgoCD CR

- Only the "read all" cluster role for server and application-controller would be created/managed outside the operator ( by OLM ) since they wouldn't necessarily differ per installation of argocd

### **Next PR**
* Release ArgoCD in all-namespaces mode.

### **Future PRs**
Support setting up of cluster config ClusterRole


pkg/common/annotations.go
```diff
@@ -0,0 +1,11 @@
+package common
+
+const (
+	// AnnotationName is the annotation on child resources that specifies which ArgoCD instance
+	// name a specific object is associated with
+	AnnotationName = "argocds.argoproj.io/name"
```
wtam2018, line 6: Using the plural form of argocd (argocds) seems odd to me.   Is it intentional?

sbose78: The CRD is already named that way :) 

```
apiVersion: apiextensions.k8s.io/v1beta1
metadata:
  name: argocds.argoproj.io
spec:
  group: argoproj.io
  names:
    kind: ArgoCD
    listKind: ArgoCDList
    plural: argocds
```

pkg/controller/argocd/util.go
```diff
@@ -639,14 +640,23 @@ func labelsForCluster(cr *argoprojv1a1.ArgoCD) map[string]string {
 	return labels
 }
 
+// annotationsForCluster returns the labels for all cluster resources.
+func annotationsForCluster(cr *argoprojv1a1.ArgoCD) map[string]string {
+	labels := argoutil.DefaultAnnotations(cr)
+	for key, val := range cr.ObjectMeta.Annotations {
+		labels[key] = val
+	}
+	return labels
```
wtam2018, line 649:
```suggestion
// annotationsForCluster returns the annotations for all cluster resources.
func annotationsForCluster(cr *argoprojv1a1.ArgoCD) map[string]string {
	annotations := argoutil.DefaultAnnotations(cr)
	for key, val := range cr.ObjectMeta.Annotations {
		annotations[key] = val
	}
	return annotations
```

wtam2018:
/lgtm

@sbose78 Note: this is a conflicting file.

wtam2018:
@sbose78 please rebase.


Conflicting files
pkg/controller/argocd/util.go

sbose78:
Sure William!

pkg/controller/argocd/rolebinding.go
```diff
@@ -0,0 +1,141 @@
...
+	roleBinding := newRoleBindingWithname(name, cr)
+
+	// fetch existing rolebinding by name
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: roleBinding.Name, Namespace: cr.Namespace}, roleBinding)
+	roleBindingExists := true
+	if err != nil {
+		if errors.IsNotFound(err) {
+			roleBindingExists = false
+			roleBinding = newRoleBindingWithname(name, cr)
+		} else {
```
bigkevmcd, line 76: You could invert this code and drop the `else`.

pkg/controller/argocd/rolebinding.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	roleBinding.RoleRef = v1.RoleRef{
+		APIGroup: v1.GroupName,
+		Kind:     "Role",
+		Name:     role.Name,
+	}
+
+	controllerutil.SetControllerReference(cr, roleBinding, r.scheme)
+	if roleBindingExists {
+		err = r.client.Update(context.TODO(), roleBinding)
```
bigkevmcd, line 96: You could simplify this to: (and there's another case below with the same idea).

```Go
if roleBindingExists {
		return r.client.Update(context.TODO(), roleBinding)
}
return r.client.Create(context.TODO(), roleBinding)
```

pkg/controller/argocd/rolebinding.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// fetch existing rolebinding by name
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: roleBinding.Name}, roleBinding)
+	roleBindingExists := true
+	if err != nil {
+		if errors.IsNotFound(err) {
+			roleBindingExists = false
+			roleBinding = newClusterRoleBindingWithname(name, cr)
+		} else {
+			return err
```
bigkevmcd, line 116: See above for the same issue, inversion would simplify the code.

pkg/controller/argocd/rolebinding_test.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+func TestReconcileArgoCD_reconcileClusterRoleBinding(t *testing.T) {
+	logf.SetLogger(logf.ZapLogger(true))
+	a := makeTestArgoCD()
+	r := makeTestReconciler(t, a)
+
+	assertNoError(t, r.reconcileClusterRoleBinding("x", &v1.ClusterRole{}, &corev1.ServiceAccount{}, a))
+
+	// TODO: Compare contents of the object
```
bigkevmcd, line 28: This bit feels like it still needs to be done?

pkg/controller/argocd/service_account.go
```diff
@@ -95,3 +125,61 @@ func (r *ReconcileArgoCD) reconcileDexServiceAccount(cr *argoprojv1a1.ArgoCD) er
 
 	return r.client.Update(context.TODO(), sa)
 }
+
+func (r *ReconcileArgoCD) reconcileServiceAccountClusterPermissions(name string, cr *argoprojv1a1.ArgoCD) error {
+	var role *v1.ClusterRole
+	var sa *corev1.ServiceAccount
+	var err error
+
+	// setup role for
```
bigkevmcd, line 134: This comment is incomplete?

pkg/controller/argocd/service_account.go
```diff
@@ -95,3 +125,61 @@ func (r *ReconcileArgoCD) reconcileDexServiceAccount(cr *argoprojv1a1.ArgoCD) er
...
+
+	return r.reconcileClusterRoleBinding(name, role, sa, cr)
+}
+
+func (r *ReconcileArgoCD) reconcileServiceAccountPermissions(name string, rules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD) error {
+	var role *v1.Role
+	var sa *corev1.ServiceAccount
+	var err error
+
+	// setup role for
```
bigkevmcd, line 152: Copy paste from above, with the same incomplete comment?

pkg/controller/argocd/service_account_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+func TestReconcileArgoCD_reconcileServiceAccountPermissions(t *testing.T) {
+	logf.SetLogger(logf.ZapLogger(true))
+	a := makeTestArgoCD()
+	r := makeTestReconciler(t, a)
+	_, err := r.reconcileRole("x", []v1.PolicyRule{}, a)
+	assertNoError(t, err)
+
+	assertNoError(t, r.reconcileServiceAccountPermissions("x", []v1.PolicyRule{}, a))
+}
```
bigkevmcd, line 32: Shouldn't this make some assertions about what was reconciled?

Also, let's ensure that if it's rereconciled, it continues to work?

bigkevmcd (changes requested):
There are a number of places where bare `err` is returned, which could possibly use some explanation of the failure `fmt.Errorf`.

chetan-rns (commented):
LGTM. Tried it locally and was able to create multiple instances in different namespaces

chetan-rns (approved):

PR #209, opened 2020-12-03
Fix typos and remove dead code

wtam2018:
Fix typos and remove dead code

### DIFF
--- a/pkg/controller/argocd/util.go
+++ b/pkg/controller/argocd/util.go
@@ -49,7 +49,7 @@ type DexConnector struct {
 	Type   string                 `yaml:"type"`
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/argocd/util.go
```diff
@@ -581,7 +563,7 @@ func (r *ReconcileArgoCD) reconcileResources(cr *argoprojv1a1.ArgoCD) error {
 		return err
 	}
 
-	log.Info("reconciling deployments")
+	log.Info("reconciling `deployment`s")
```
bigkevmcd, line 566: I don't think this one is quite right.

wtam2018: Fixed.  I honestly don't know how that change was made.

bigkevmcd (approved):

PR #216, opened 2020-12-11
Allow disable dex

bigkevmcd:
This adds functionality for disabling Dex, defaulting to enabled, you would need to explicitly disable it.

```yaml
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: basic
spec:
  dex:
    disabled: true
```


sbose78:
Thank you, 
This PR makes it an opt-out, instead of an opt-in. Since the boolean defaults to `false`, existing dex-with-openshift users should see no impact. I'm OK with this :) 

I had something different in mind. _Don't deploy dex unless the user has a .spec.dex section specified in the CR explicitly._ Is this feasible ?

PR #218, opened 2020-12-14
Support disabling the DEX server.

bigkevmcd:
This adds support for an env-var `DISABLE_DEX` which cause the deployment for Dex to be skipped, and no config to be set in the config map for Dex.

### DIFF
--- a/pkg/controller/argocd/configmap.go
+++ b/pkg/controller/argocd/configmap.go
@@ -319,7 +319,6 @@ func (r *ReconcileArgoCD) reconcileArgoConfigMap(cr *argoprojv1a1.ArgoCD) error 
 	cm.Data[common.ArgoCDKeyApplicationInstanceLabelKey] = getApplicationInstanceLabelKey(cr)
 	cm.Data[common.ArgoCDKeyConfigManagementPlugins] = getConfigManagementPlugins(cr)
 	cm.Data[common.ArgoCDKeyAdminEnabled] = fmt.Sprintf("%t", !cr.Spec.DisableAdmin)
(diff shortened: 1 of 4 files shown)


sbose78:
Looks good, did you try bringing up an ArgoCD instance with disabled=true ?

bigkevmcd:
Yes, and indeed it doesn't create a dex server, or configure the configmap:

```shell
$ kubectl get pods
NAME                                                     READY   STATUS    RESTARTS   AGE
argocd-operator-9f7d8f595-hr5mt                          1/1     Running   0          2m53s
example-argocd-application-controller-7896758995-ggx8n   1/1     Running   0          100s
example-argocd-redis-5cc8b58ff9-zfgwz                    1/1     Running   0          100s
example-argocd-repo-server-7c955bb447-ptp8b              1/1     Running   0          100s
example-argocd-server-6c67cc9c8b-jpt2j                   1/1     Running   0          100s
```

```shell
$ k get cm/argocd-cm -o yaml | grep dex
$
``

**Review by sbose78 — APPROVED** [review]

**Comment by Morriz** [comment]
Can this be re-opened @bigkevmcd @sbose78? This is NOT working:
```
kind: Subscription
metadata:
  name: argocd-operator
spec:
  # channel: alpha
  name: argocd-operator
  # source: argocd-catalog
  source: operatorhubio-catalog
  sourceNamespace: olm
  config:
    env:
      - name: DISABLE_DEX
        value: "true"
      - name: ARGOCD_CLUSTER_CONFIG_NAMESPACES
        value: argocd
```

PR #224, opened 2021-01-05
fix: Manage cluster roles from the controller

chetan-rns:
The cluster roles created by the OLM will have a non-deterministic name(eg: argocd-operator.0.0.1-argocd-xxxx) and cannot be extracted for creating the cluster role bindings. This PR will add support for creating these cluster roles from the controller with standard names.


pkg/controller/argocd/util.go
```diff
@@ -701,10 +701,15 @@ func watchResources(c controller.Controller, clusterRoleBindingMapper handler.To
 	handlerClusterRoleBinding := &handler.EnqueueRequestsFromMapFunc{
 		ToRequests: clusterRoleBindingMapper,
 	}
+
 	if err := c.Watch(&source.Kind{Type: &v1.ClusterRoleBinding{}}, handlerClusterRoleBinding); err != nil {
 		return err
 	}
 
+	if err := c.Watch(&source.Kind{Type: &v1.ClusterRole{}}, handlerClusterRoleBinding); err != nil {
```
chetan-rns, line 709: Even though these cluster roles are common for all argocd instances the controller will have to watch these resources.  Here I've used the existing mapper function.

iam-veeramalla: Clusterroles should be different for each instance.

pkg/controller/argocd/role.go
```diff
@@ -50,20 +49,14 @@ func newRoleWithName(name string, cr *argoprojv1a1.ArgoCD) *v1.Role {
...
-	}
-
-	for _, clusterRole := range clusterRoleList.Items {
-		if strings.Contains(clusterRole.Name, name) {
-			return &clusterRole, nil
-		}
+func newClusterRole(name string, cr *argoprojv1a1.ArgoCD) *v1.ClusterRole {
+	return &v1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        name,
```
iam-veeramalla, line 55: clusterole.ObjectMeta.Name should be assigned with "cr.Name + "-" + argoComponentName". This will ensure that each argocd instance is created with a different clusterrole.

chetan-rns: These cluster roles will be the same for every instance of Argo CD. Hence the names are not prefixed with `cr.Name` and the cluster roles are shared. But creating separate cluster roles will allow teams to configure their ArgoCD instance without conflicts.

chetan-rns: I've pushed a commit to create separate cluster roles for each instance

pkg/controller/argocd/route_test.go
```diff
@@ -40,8 +39,6 @@ func TestReconcileRouteSetsInsecure(t *testing.T) {
 		},
 	}
 
-	createClusterRoles(t, r.client)
-
 	_, err := r.Reconcile(req)
 	assertNoError(t, err)
```
bigkevmcd, line 43: if you're going to use `gotest/assert` then this should be `assert.NilError` ?

pkg/controller/argocd/role.go
```diff
@@ -90,3 +83,29 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
+		if !errors.IsNotFound(err) {
+			return nil, fmt.Errorf("Failed to reconcile the cluster role for the service account associated with %s : %s", name, err)
+		}
+		clusterRoleExists = false
+		clusterRole = newClusterRole(name, cr)
+	}
+
+	clusterRole.Rules = policyRules
+
+	controllerutil.SetControllerReference(cr, clusterRole, r.scheme)
```
bigkevmcd, line 103: What if `SetControllerReference` returns an error?

pkg/controller/argocd/role.go
```diff
@@ -90,3 +83,29 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
+	if clusterRoleExists {
```
bigkevmcd, line 104: This can be simplified to eliminate the else.

pkg/controller/argocd/role.go
```diff
@@ -90,3 +83,29 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
+
+func (r *ReconcileArgoCD) reconcileClusterRole(name string, policyRules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD) (*v1.ClusterRole, error) {
+
+	clusterRole := newClusterRole(name, cr)
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: clusterRole.Name}, clusterRole)
+	clusterRoleExists := true
+
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return nil, fmt.Errorf("Failed to reconcile the cluster role for the service account associated with %s : %s", name, err)
```
bigkevmcd, line 95:
```suggestion
			return nil, fmt.Errorf("failed to reconcile the cluster role for the service account associated with %s : %s", name, err)
```

pkg/controller/argocd/role_test.go
```diff
@@ -37,20 +38,27 @@ func TestReconcileArgoCD_reconcileRole(t *testing.T) {
...
-	controllerClusterRole := "argocd-application-controller"
-	clusterRole, err = r.getClusterRole(controllerClusterRole)
-	assertNoError(t, err)
-	assert.Equal(t, controllerClusterRole, clusterRole.Name)
+	expectedName := fmt.Sprintf("%s-%s", a.Name, workloadIdentifier)
+	reconciledClusterRole := &v1.ClusterRole{}
+	assertNoError(t, r.client.Get(context.TODO(), types.NamespacedName{Name: expectedName}, reconciledClusterRole))
+	assert.DeepEqual(t, expectedRules, reconciledClusterRole.Rules)
+
+	// undersirable change.
```
bigkevmcd, line 56: typo

```suggestion
	// undesirable change.
```

But, maybe it could really explain what it's changing/why it's changing it?

pkg/controller/argocd/role.go
```diff
@@ -90,3 +83,29 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
 	return role, err
 }
+
+func (r *ReconcileArgoCD) reconcileClusterRole(name string, policyRules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD) (*v1.ClusterRole, error) {
+
+	clusterRole := newClusterRole(name, cr)
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: clusterRole.Name}, clusterRole)
+	clusterRoleExists := true
+
+	if err != nil {
```
bigkevmcd, line 93: I think you can simplify this logic a bit?

bigkevmcd (changes requested):

sbose78:
We'll probably need a custom mapper for watching clusterroles https://github.com/argoproj-labs/argocd-operator/blob/master/pkg/controller/argocd/custommapper.go#L10

chetan-rns:
@sbose78 The existing mapper will work for cluster roles as well. I've just updated the name to make it generic

bigkevmcd (approved):
I'm ok with this change, my concern is that the permissions are pretty permissive, and because they're being watched, there's no way to reduce them, as they'd be reconciled back to the defaults.

PR #225, opened 2021-01-05
Expose API/attribute to create ArgoCD for cluster config

iam-veeramalla:

Support cluster config if cr.Spec.ManagementScope.Cluster is set to true and cr.Spec.ManagementScope.Namespaces is provided

```
kind: ArgoCD
metadata:
  creationTimestamp: null
  name: argocd-clusterconfig-bar
  namespace: bar
spec:
  resourceExclusions: |
    - apiGroups:
      - tekton.dev
      clusters:
      - '*'
      kinds:
      - TaskRun
      - PipelineRun
  server:
    route:
      enabled: true
  managementScope:
    cluster: true
    namespaces:
    - bar
    - foo 
```


pkg/controller/argocd/rolebinding.go
```diff
@@ -14,20 +14,25 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-// newRoleBinding returns a new RoleBinding instance.
-func newClusterRoleBinding(cr *argoprojv1a1.ArgoCD) *v1.ClusterRoleBinding {
+const (
+	openshiftConfigNamespace = "openshift-config"
```
bigkevmcd, line 18: This isn't used.

pkg/controller/argocd/rolebinding.go
```diff
@@ -131,3 +176,43 @@ func (r *ReconcileArgoCD) reconcileClusterRoleBinding(name string, role *v1.Clus
...
+			Namespace: cr.Namespace,
+		},
+	}
+	roleBinding.RoleRef = v1.RoleRef{
+		APIGroup: v1.GroupName,
+		Kind:     "ClusterRole",
+		Name:     "admin",
+	}
+
+	if roleBindingExists {
```
bigkevmcd, line 212: This code can be simplified.

pkg/controller/argocd/testing.go
```diff
@@ -33,6 +33,8 @@ const (
 	testArgoCDName = "argocd"
 )
 
+var clusterConfig = true
```
bigkevmcd, line 36: This isn't used outside of `makeTestArgoCDForClusterConfig`

bigkevmcd (changes requested):
The tests don't seem to be passing for me locally?

```
$ go test . | grep FAIL
--- FAIL: TestReconcileArgoCD_Reconcile (0.00s)
--- FAIL: TestReconcileArgoCD_reconcileClusterRole (0.00s)
--- FAIL: TestReconcileArgoCDClusterConfig_reconcileClusterRole (0.00s)
--- FAIL: TestReconcileArgoCD_reconcileRoleBinding (0.00s)
--- FAIL: TestReconcileArgoCD_reconcileClusterRoleBinding (0.00s)
--- FAIL: TestReconcileRouteSetsInsecure (0.00s)
--- FAIL: TestReconcileRouteUnsetsInsecure (0.00s)
--- FAIL: TestReconcileArgoCD_reconcileServiceAccountPermissions (0.00s)
--- FAIL: TestReconcileArgoCD_reconcileServiceAccountClusterPermissions (0.00s)
FAIL
FAIL	github.com/argoproj-labs/argocd-operator/pkg/controller/argocd	0.538s
FAIL
```

pkg/controller/argocd/rolebinding.go
```diff
@@ -132,3 +161,43 @@ func (r *ReconcileArgoCD) reconcileClusterRoleBinding(name string, role *v1.Clus
...
 }
+
+// reconcileArgoApplier ensures that a specific rolebinding is present in a managedNamespace
+func (r *ReconcileArgoCD) reconcileArgoApplier(controlPlaneServiceAccount string, cr *argoprojv1a1.ArgoCD, managedNamespace string) error {
+	roleBinding := newRoleBindingWithname(controlPlaneServiceAccount, cr)
+
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: roleBinding.Name, Namespace: cr.Namespace}, roleBinding)
+	roleBindingExists := true
+	if err != nil {
+		if errors.IsNotFound(err) {
```
bigkevmcd, line 172: This code could be simplified by inverting this check.

pkg/controller/argocd/rolebinding_test.go
```diff
@@ -67,9 +64,33 @@ func TestReconcileArgoCD_reconcileClusterRoleBinding(t *testing.T) {
 	// try reconciling it again to ensure undesirable changes are overwritten
 	assert.NilError(t, r.reconcileClusterRoleBinding(workloadIdentifier, expectedClusterRole, expectedServiceAccount, a))
 
-	clusterRoleBinding = &v1.ClusterRoleBinding{}
+	clusterRoleBinding = &rbacv1.ClusterRoleBinding{}
 	assert.NilError(t, r.client.Get(context.TODO(), types.NamespacedName{Name: expectedName}, clusterRoleBinding))
+}
+
+func TestReconcileArgoCD_reconcileRoleBinding_with_extensions(t *testing.T) {
+	logf.SetLogger(logf.ZapLogger(true))
```
bigkevmcd, line 72: It's not clear what extensions are being tested here?

pkg/controller/argocd/util.go
```diff
@@ -566,6 +566,16 @@ func (r *ReconcileArgoCD) reconcileResources(cr *argoprojv1a1.ArgoCD) error {
 		return err
 	}
 
+	log.Info("reconciling roles")
+	if _, err := r.reconcileRoles(cr); err != nil {
+		return err
+	}
+
+	log.Info("reconcling rolebindings")
```
bigkevmcd, line 574: s/reconcling/reconciling/

pkg/reconciler/openshift/openshift.go
```diff
@@ -0,0 +1,25 @@
...
+	v1 "k8s.io/api/rbac/v1"
+
+	"github.com/argoproj-labs/argocd-operator/pkg/controller/argocd"
+)
+
+func init() {
+	argocd.Register("reconciler", reconcilerHook)
+}
+
+func reconcilerHook(cr *argoprojv1alpha1.ArgoCD, v interface{}) error {
```
bigkevmcd, line 14: There are no tests for the reconcilerHook?

We should isolate all the openshift specific code to this package, starting with the GVK extensions.

pkg/reconciler/openshift/openshift.go
```diff
@@ -0,0 +1,25 @@
+package openshift
+
+import (
+	argoprojv1alpha1 "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
+	v1 "k8s.io/api/rbac/v1"
```
bigkevmcd, line 5: Let's import this as `rbacv1`?

bigkevmcd (changes requested):
This needs a fair bit of reworking.

hack/env.sh
```diff
@@ -17,6 +17,7 @@
 # Environment variables used in development. This file is sourced in the other scripts.
 
 # General vars
+export ARGOCD_BUILD_ARGS="--go-build-args -tags=openshift"
```
bigkevmcd, line 20: This basically means that every build would include the openshift tags?

bigkevmcd: This is still an issue here?

pkg/apis/argoproj/v1alpha1/argocd_types.go
```diff
@@ -316,9 +316,20 @@ type ArgoCDServerServiceSpec struct {
 	Type corev1.ServiceType `json:"type"`
 }
 
+// ArgoCDScope defines the scope of resources that ArgoCD would be managing
+type ArgoCDScope struct {
+	// Cluster when enabled lets ArgoCD manage cluster resources.
+	Cluster    bool     `json:"cluster,omitempty"`
+	Namespaces []string `json:"namespaces,omitempty"`
+}
```
bigkevmcd, line 324: Can you add a comment for the Namespaces field?

pkg/controller/argocd/register.go
```diff
@@ -0,0 +1,33 @@
+package argocd
```
bigkevmcd, line 1: Let's rename this to hooks.go?

pkg/controller/argocd/register_test.go
```diff
@@ -0,0 +1,111 @@
+package argocd
```
bigkevmcd, line 1: See above about renaming this file?

pkg/controller/argocd/register_test.go
```diff
@@ -0,0 +1,111 @@
...
+	logf.SetLogger(logf.ZapLogger(true))
+	a := makeTestArgoCDForClusterConfig()
+
+	Register(testDeploymentHook)
+
+	testDeployment := makeTestDeployment()
+	if err := applyReconcilerHook(a, testDeployment); err != nil {
+		t.Fatalf("failed to apply reconciler hook: %#v\n", err)
+	}
+	var expectedReplicas int32 = 3
```
bigkevmcd, line 51: You could simplify this by deferencing `testDeployment.Spec.Replicas` on the next line?

pkg/controller/argocd/register_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Error should effect the Deployment Resource
+	testDeployment := makeTestDeployment()
+	err := applyReconcilerHook(a, testDeployment)
+	expectedError := fmt.Sprintf("Error returned from %s for %s", "testErrorHook", testDeployment.Name)
+	assert.Error(t, err, expectedError, applyReconcilerHook(a, testDeployment))
+}
+
+func stubDefaultHookFunc(t *testing.T) func() {
+	origDefaultHooksFunc := hooks
```
bigkevmcd, line 106: You're not using `t` in this function?

Also, how about we name this `resetHooks` ?

I'd also be inclined to do something like `hooks = []Hook{}` which would mean that you have a known state for the tests`

pkg/controller/argocd/role.go
```diff
@@ -60,34 +47,57 @@ func newClusterRole(name string, rules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD)
...
+
+// reconcileRole
+func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD) (*v1.Role, error) {
+
+	role := newRole(name, policyRules, cr)
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: role.Name, Namespace: cr.Namespace}, role)
 	if err != nil {
 		if !errors.IsNotFound(err) {
-			return nil, fmt.Errorf("Failed to reconcile the role for the service account associated with %s : %s", name, err)
+			return nil, err
```
bigkevmcd, line 87: I think this should be restored, note that the error message should start with a lower-case F ;-)

pkg/controller/argocd/role_test.go
```diff
@@ -28,11 +28,11 @@ func TestReconcileArgoCD_reconcileRole(t *testing.T) {
 	assert.NilError(t, r.client.Get(context.TODO(), types.NamespacedName{Name: expectedName, Namespace: a.Namespace}, reconciledRole))
 	assert.DeepEqual(t, expectedRules, reconciledRole.Rules)
 
-	// undersirable change.
+	// update reconciledRole with undesirable policy rules
```
bigkevmcd, line 31: What's undesirable about it?

pkg/controller/argocd/role_test.go
```diff
@@ -28,11 +28,11 @@ func TestReconcileArgoCD_reconcileRole(t *testing.T) {
...
 	reconciledRole.Rules = policyRuleForRedisHa()
 	assert.NilError(t, r.client.Update(context.TODO(), reconciledRole))
 
-	// overwrite it.
+	// Check if the undesirable policy rules are overwritten by the reconciler
```
bigkevmcd, line 35: See above?

pkg/controller/argocd/rolebinding.go
```diff
@@ -61,7 +60,37 @@ func newRoleBindingWithname(name string, cr *argoprojv1a1.ArgoCD) *v1.RoleBindin
...
 }
 
-func (r *ReconcileArgoCD) reconcileRoleBinding(name string, role *v1.Role, sa *corev1.ServiceAccount, cr *argoprojv1a1.ArgoCD) error {
+// reconcileRoleBindings will ensure that all ArgoCD RoleBindings are configured.
+func (r *ReconcileArgoCD) reconcileRoleBindings(cr *argoprojv1a1.ArgoCD) error {
+	if err := r.reconcileRoleBinding(applicationController, policyRuleForApplicationController(), cr); err != nil {
+		return err
+	}
+	if err := r.reconcileRoleBinding(dexServer, policyRuleForDexServer(), cr); err != nil {
+		return err
```
bigkevmcd, line 69: Let's wrap these with messages, to make it easier to identify where the error occurred?

pkg/reconciler/openshift/openshift_test.go
```diff
@@ -0,0 +1,81 @@
...
+	logf.SetLogger(logf.ZapLogger(true))
+	a := makeTestArgoCDForClusterConfig()
+	testClusterRole := &v1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      a.Name + "-" + testApplicationController,
+			Namespace: a.Namespace,
+		},
+		Rules: makeTestPolicyRules(),
+	}
+	if err := reconcilerHook(a, testClusterRole); err != nil {
```
bigkevmcd, line 24: I'm suspect this should use `assert.NilError` (there are a number of places where this could be used).

pkg/reconciler/openshift/openshift_test.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	assert.DeepEqual(t, makeTestPolicyRules(), testNotApplicableClusterRole.Rules)
+}
+
+func TestReconcileArgoCD_testDeployment(t *testing.T) {
+
+	logf.SetLogger(logf.ZapLogger(true))
+	a := makeTestArgoCDForClusterConfig()
+
+	testDeployment := makeTestDeployment()
```
bigkevmcd, line 76: For this, I'd probably simplify it to a corev1.Namespace, less typing :-)

bigkevmcd (changes requested):
Looking not too bad, some small tidyups and we should be able to land this.

hack/env.sh
bigkevmcd: I don't think we should do this here, as it means that the environment will always build for openshift.

pkg/apis/argoproj/v1alpha1/argocd_types.go
bigkevmcd: This still needs to be done.

pkg/controller/argocd/hooks.go
```diff
@@ -0,0 +1,33 @@
...
+
+// Register adds a modifier for updating resources during reconciliation.
+func Register(h Hook) {
+	mutex.Lock()
+	defer mutex.Unlock()
+	hooks = append(hooks, h)
+}
+
+func applyReconcilerHook(cr *argoprojv1alpha1.ArgoCD, i interface{}) error {
+	for _, v := range hooks {
```
bigkevmcd, line 26: I appreciate that it's unlikely to happen in real-world use, but it's best-practice to lock and unlock the mutex in here.

pkg/controller/argocd/hooks_test.go
```diff
@@ -0,0 +1,111 @@
...
+	case *appsv1.Deployment:
+		err := fmt.Errorf("Error returned from %s for %s", "testErrorHook", o.Name)
+		return err
+	}
+	return nil
+}
+
+func TestReconcileArgoCD_testDeploymentHook(t *testing.T) {
+	defer resetHooks()()
+	logf.SetLogger(logf.ZapLogger(true))
```
bigkevmcd, line 42: logf isn't used in the tests?

pkg/controller/argocd/hooks_test.go
```diff
@@ -0,0 +1,111 @@
...
+	a := makeTestArgoCDForClusterConfig()
```
bigkevmcd, line 43: You could simplify this to just `makeTestArgoCD()` the tests don't require the cluster config.

pkg/controller/argocd/role_test.go
```diff
@@ -48,17 +49,43 @@ func TestReconcileArgoCD_reconcileClusterRole(t *testing.T) {
...
 
-	expectedName := fmt.Sprintf("%s-%s", a.Name, workloadIdentifier)
 	reconciledClusterRole := &v1.ClusterRole{}
-	assert.NilError(t, r.client.Get(context.TODO(), types.NamespacedName{Name: expectedName}, reconciledClusterRole))
+	clusterRoleName := generateResourceName(workloadIdentifier, a)
+	assert.NilError(t, r.client.Get(context.TODO(), types.NamespacedName{Name: clusterRoleName}, reconciledClusterRole))
 	assert.DeepEqual(t, expectedRules, reconciledClusterRole.Rules)
 
-	// Update the ClusterRole with undesirable policy rules
+	// undersirable change.
```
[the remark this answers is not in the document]
bigkevmcd: This is still an issue

pkg/controller/argocd/rolebinding.go
bigkevmcd: I think we should wrap these with messages.

pkg/controller/argocd/testing.go
```diff
@@ -60,6 +64,87 @@ func makeTestArgoCD(opts ...argoCDOpt) *argoprojv1alpha1.ArgoCD {
 	return a
 }
 
+func makeTestArgoCDForClusterConfig(opts ...argoCDOpt) *argoprojv1alpha1.ArgoCD {
```
bigkevmcd, line 67: You could remove the opts, you're not using them?

bigkevmcd (changes requested):
Still needs some work.

pkg/controller/argocd/hooks_test.go
iam-veeramalla: I see that all the tests are  using this statement,  do u think we should get rid of it ?

deploy/crds/argoproj.io_argocds_crd.yaml
```diff
@@ -10,998 +10,1016 @@ spec:
     plural: argocds
     singular: argocd
   scope: Namespaced
-  subresources:
```
wtam2018, line 13: Can you please remove formatting changes and only add changes relevant to the new features?   It is impossible to review and has a higher risk to introduce bugs.

hack/build.sh
```diff
@@ -22,4 +22,4 @@ HACK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 source ${HACK_DIR}/env.sh
 
 echo "Building image ${ARGOCD_OPERATOR_IMAGE}"
-${OPERATOR_SDK} build ${ARGOCD_OPERATOR_IMAGE} --image-builder ${ARGOCD_OPERATOR_IMAGE_BUILDER}
+${OPERATOR_SDK} build ${ARGOCD_OPERATOR_IMAGE} ${ARGOCD_BUILD_ARGS} --image-builder ${ARGOCD_OPERATOR_IMAGE_BUILDER}
```
wtam2018, line 25: Should we document what ARGOCD_BUILD_ARGS is for and what can be passed?

pkg/apis/argoproj/v1alpha1/argocd_types.go
```diff
@@ -316,9 +316,21 @@ type ArgoCDServerServiceSpec struct {
 	Type corev1.ServiceType `json:"type"`
 }
 
+// ArgoCDScope defines the scope of resources that ArgoCD would be managing
+type ArgoCDScope struct {
+	// Cluster when enabled lets ArgoCD manage cluster resources.
+	Cluster bool `json:"cluster,omitempty"`
```
wtam2018, line 322: How about rename to ClusterConfigEnabled since it is a bool?

hack/build.sh
bigkevmcd: You can pass anything to the operator-sdk you like.

deploy/crds/argoproj.io_argocds_crd.yaml
iam-veeramalla: changed the required and re-ran operator-sdk generate crds

pkg/apis/argoproj/v1alpha1/argocd_types.go
iam-veeramalla: Made the required change.

hack/build.sh
wtam2018: But there is no words on the intend to pass "openshift".

bigkevmcd: Yeah, the whole point of using build tags is that we don't _need_ to and by default the openshift specific bits won't be included?

We could probably even remove the openshift.go from the cmd package (which uses the build tags) and leave it to the gitops-operator to do the registration (by importing).

wtam2018: I like the latter.  Let's leave the OpenShift specific to `gitops-operator` when possible.

pkg/controller/argocd/policyrule.go
```diff
@@ -3,7 +3,6 @@ package argocd
 import v1 "k8s.io/api/rbac/v1"
 
 func policyRuleForApplicationController() []v1.PolicyRule {
-
```
wtam2018, line 6: Please undo this file.   There is no real changes.

pkg/controller/argocd/rolebinding.go
```diff
@@ -41,10 +41,9 @@ func newClusterRoleBindingWithname(name string, cr *argoprojv1a1.ArgoCD) *v1.Clu
 func newRoleBinding(cr *argoprojv1a1.ArgoCD) *v1.RoleBinding {
 	return &v1.RoleBinding{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        cr.Name,
-			Labels:      labelsForCluster(cr),
-			Annotations: annotationsForCluster(cr),
```
wtam2018, line 46: Q: We are dropping annotations on role binding.   I just want to make sure it is intentional.

pkg/controller/argocd/rolebinding.go
```diff
@@ -25,10 +25,10 @@ func newClusterRoleBinding(cr *argoprojv1a1.ArgoCD) *v1.ClusterRoleBinding {
 	}
 }
 
-// newRoleBindingWithname creates a new RoleBinding with the given name for the given ArgCD.
+// newClusterRoleBindingWithname creates a new ClusterRoleBinding with the given name for the given ArgCD.
 func newClusterRoleBindingWithname(name string, cr *argoprojv1a1.ArgoCD) *v1.ClusterRoleBinding {
-	roleBinding := newClusterRoleBinding(cr)
-	roleBinding.ObjectMeta.Name = fmt.Sprintf("%s-%s", cr.Name, name)
+	roleBinding := newClusterRoleBinding(name, cr)
+	roleBinding.Name = "cluster-" + generateResourceName(name, cr)
```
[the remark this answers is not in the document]
iam-veeramalla: This is just to differentiate it with rolebinding. 
ClusterRoleBinding - 	 "cluster-" + generateResourceName(name, cr)
RoleBinding - generateResourceName(name, cr)

hack/build.sh
bigkevmcd: This is complicated tho', we do want to shift the openshift routing out of the core reconciliation.

pkg/controller/argocd/rolebinding.go
iam-veeramalla: This is coming from the draft PR 
https://github.com/argoproj-labs/argocd-operator/pull/161

and I assume it is intentionally removed, may be they are not needed for rolebinding resources. I have added them for the clusterrolebindings.

wtam2018: @sbose78 Is the removing the annotations above intentional?

sbose78: Honestly, I can't seem to remember why I had removed it :)

iam-veeramalla: Added it back @sbose78 @wtam2018

hack/build.sh
iam-veeramalla: I added a documentation(comment) which explains the behaviour of ARGOCD_BUILD_ARGS. I will resolve the conversation if it looks good.

hack/build.sh
```diff
@@ -22,4 +22,6 @@ HACK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
+# export ${ARGOCD_BUILD_ARGS}="--go-build-args "-tags openshift"" to add openshift package to the build process that
+# registers a reconciler hook which modifies the reconciler to create ArgoCD for Cluster Config.
```
bigkevmcd, line 27: This comment likely won't age well, but it's probably fine.

bigkevmcd (approved):
I'm happy enough with this.

pkg/apis/argoproj/v1alpha1/argocd_types.go
```diff
@@ -316,9 +316,21 @@ type ArgoCDServerServiceSpec struct {
 	Type corev1.ServiceType `json:"type"`
 }
 
+// ArgoCDScope defines the scope of resources that ArgoCD would be managing
+type ArgoCDScope struct {
+	// When enabled lets ArgoCD manage cluster resources.
+	ClusterConfigEnabled bool `json:"clusterconfigenabled,omitempty"`
```
wtam2018, line 322: I wonder if we can remove `ClusterConfigEnabled` all together.    If `Namespacce` is not empty, this feature is enabled.   If we remove `ClusterConfigEnabled`, we should rename `Namespace` to something like `ClusterConfiguredNamespaces`.

pkg/reconciler/openshift/testing.go
```diff
@@ -0,0 +1,86 @@
+package openshift
+
```
wtam2018, line 2: You may want to rename this file to openshift_test.go so that it only gets built by `go test`.

pkg/controller/argocd/rolebinding.go
```diff
@@ -132,3 +162,40 @@ func (r *ReconcileArgoCD) reconcileClusterRoleBinding(name string, role *v1.Clus
...
+		}
+	}
+	roleBinding.Subjects = []v1.Subject{
+		{
+			Kind:      v1.ServiceAccountKind,
+			Name:      controlPlaneServiceAccount,
+			Namespace: cr.Namespace,
+		},
+	}
+	roleBinding.RoleRef = v1.RoleRef{
```
wtam2018, line 191: @sbose78 I think this can be a problem to run in OSD.

pkg/reconciler/openshift/testing.go
iam-veeramalla: renamed the file

pkg/apis/argoproj/v1alpha1/argocd_types.go
iam-veeramalla: Makes sense, Updated with a new push

pkg/reconciler/openshift/testing.go
bigkevmcd: I should point out that this is pretty much a zero-cost thing, the logic from that file won't end up in the final binary.

And naming it "testing.go" is actually pretty common, for examples https://github.com/hashicorp/nomad/blob/master/client/testing.go https://github.com/hashicorp/nomad/blob/master/command/agent/consul/testing.go https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/testing/testing.go

PR #234, opened 2021-01-25
Set env var to configure ArgoCD for Cluster Config

iam-veeramalla:
This commit Modifies the behaviour of https://github.com/argoproj-labs/argocd-operator/pull/225. 

What?
Cluster admin should set a env variable "ARGOCD_CLUSTER_CONFIG_NAMESPACES" with a list of namespaces(comma separated) where ArgoCD instance can manage cluster config resources. Users should request admin to get their namespaces added to the ARGOCD_CLUSTER_CONFIG_NAMESPACES env variable.

User cannot request for a ArgoCD instance without such privileges in  ARGOCD_CLUSTER_CONFIG_NAMESPACES. In such cases, corresponding namespace should be removed from the ARGOCD_CLUSTER_CONFIG_NAMESPACES env variable.

### DIFF
--- a/pkg/reconciler/openshift/clusterconfig_test.go
+++ b/pkg/reconciler/openshift/clusterconfig_test.go
@@ -1,6 +1,8 @@
 package openshift
 
 import (
+	"os"
+
 	argoprojv1alpha1 "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -12,6 +14,7 @@ const (
 	testNamespace             = "argocd"
 	testArgoCDName            = "argocd"
 	testApplicationController = "argocd-application-controller"
+	testDummyNameSpace        = "dummy"
 )
 
 func makeTestPolicyRules() []rbacv1.PolicyRule {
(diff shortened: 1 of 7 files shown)


pkg/reconciler/openshift/clusterconfig_test.go
```diff
@@ -31,21 +34,26 @@ func makeTestPolicyRules() []rbacv1.PolicyRule {
...
-				ClusterConfigNamespaces: allowedNamespaces,
-			},
+			Namespace: testDummyNameSpace,
 		},
 	}
 	return a
 }
 
+func setClusterConfigNamespaces() {
+	os.Setenv("ARGOCD_CLUSTER_CONFIG_NAMESPACES", "argocd,foo,bar")
```
bigkevmcd, line 54: Couple of issues with this, you're setting it to the same value in each test, so you could set it once, but that's not what I'd do, I'd add a t.Cleanup function that could do `os.Unsetenv` _or_ restore the original value (probably the former).

But, there's not enough testing around the parsing of environment variable...
Try testing it with `argocd, foo, bar` for  example.

bigkevmcd (changes requested):

pkg/reconciler/openshift/clusterconfig_test.go
iam-veeramalla: I thought about this `argocd,  foo,  bar`, I was confused if I need to handle the spaces or should I say its not provided as per documentation.

pkg/reconciler/openshift/openshift.go
```diff
@@ -15,7 +18,7 @@ func reconcilerHook(cr *argoprojv1alpha1.ArgoCD, v interface{}) error {
 	switch o := v.(type) {
 	case *rbacv1.ClusterRole:
 		if o.ObjectMeta.Name == cr.ObjectMeta.Name+"-argocd-application-controller" {
-			if len(cr.Spec.ManagementScope.ClusterConfigNamespaces) > 0 {
```
sbose78, line 18: Let's keep the API knob, I think. Any reason you think we should not have it?

.spec.managementScope.cluster = true|false

sbose78: Also,

```
cr.ObjectMeta.Name+"-argocd-application-controller"
```

This doesn't give us very unique names, does it 🤔 . 

Maybe something like this could help us ?

```
cr.ObjectMeta.Name+"-"+cr.ObjectMeta.Namespace+"-argocd-application-controller"
```

pkg/reconciler/openshift/openshift_test.go
```diff
@@ -64,15 +64,31 @@ func TestReconcileArgoCD_testDeployment(t *testing.T) {
 	assert.NilError(t, reconcilerHook(a, testDeployment))
 }
 
+func TestReconcileArgoCD_notInClusterConfigNamespaces(t *testing.T) {
+
+	a := makeTestArgoCD()
+	testClusterRole := &rbacv1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      a.Name + "-" + testApplicationController,
+			Namespace: a.Namespace,
```
sbose78, line 73: The ClusterRole is cluster-scoped and not namespace-scoped.

pkg/reconciler/openshift/openshift.go
iam-veeramalla: > ec.managementScope.cluster = true|false

I think this will provide an inconvenient advantage to the user. Users can request admins to add their namespace to the ARGOCD_CLUSER_CONFIG_NAMESPACES and may not bother about getting it removed if they dont need it in future. Also, I was thinking, why would a user request to add his namespace in ARGOCD_CLUSER_CONFIG_NAMESPACES and not add spec.managementScope.cluster = true in his CR.

sbose78: The admin would whitelist in the env var.
The user would then request for it in the CR. The request will be honored only if that namespace is whitelisted.

sbose78: We may choose to skip having the knob in the CRD as of now.

pkg/reconciler/openshift/openshift_test.go
iam-veeramalla: True ..

pkg/reconciler/openshift/clusterconfig_test.go
iam-veeramalla: added a new function trimWhitespaces() to handle this.

bigkevmcd (approved):
Thanks for the changes, looks good to me.

pkg/reconciler/openshift/openshift.go
bigkevmcd: @sbose78 Wouldn't this limit the length of the object name/namespace combination very severely, as you've already used 31 characters in the suffix and separator?

Issue #223, opened 2021-01-04
ConfigMap url


lcc3108:
Hi, I use argocd-operator with kubernetes and istio and custom Dex as shown below yaml
```yaml
apiVersion: argoproj.io/v1alpha1
kind: ArgoCD
metadata:
  name: argocd
  namespace: argocd
spec:
  oidcConfig: >
    name: my-dex
    issuer: DEX_FQDN
    clientID: id
    clientSecret: secret
    requestedScopes: ["openid", "profile", "email"]
  server:
    host: ARGOCD_FQDN
    insecure: true
  usersAnonymousEnabled: true
```
However, redirection url expected ARGO_FQDN, but ServiceName came out when I log-in to Custom Dex.

I find return server.host only if Ingress is set or Route is set

I am using Istio and do not want to be complicated by creating objects such as Ingress

I think function (getArgoServerURI OR getDexOAuthRedirectURI) return ARGO_FQDN when server.host value is set

I apologize for my poor English.

Thanks

bigkevmcd:
Hello @lcc3108 if I'm understanding this issue correctly, you're configuring the `spec.server.host` and would like that to be the login URL provided to Dex, but because there's no ingress or route, it's not working?

lcc3108:
Yes. When I'm using Istio and I'm not using Ingress and Route, can't login dex.


When 
server.host set ARGOCD_FQDN,
route and ingress not set

argo-cm
```
data:
  application.instanceLabelKey: ""
  configManagementPlugins: ""
  dex.config: ""
  ga.anonymizeusers: "false"
  ga.trackingid: ""
  help.chatText: ""
  help.chatUrl: ""
  kustomize.buildOptions: ""
  oidc.config: |
    name: my-dex
    issuer: https://dex.FQDN
    clientID: argocd
    clientSecret: argocd
    requestedScopes: ["openid", "profile", "email", "groups"]
  repository.credentials: ""
  resource.exclusions: ""
  resource.inclusions: ""
  statusbadge.enabled: "false"
  url: https://argocd-server
  users.anonymous.enabled: "false"
```

![image](https://user-images.githubusercontent.com/38449269/103884533-14860f00-5122-11eb-9ded-54fc1970cc3f.png)

buehlmann:
[...]

PR #236, opened 2021-01-29
[WIP] Allow user to provide custom external hostname for argo-server

shubhamagarwal19:
This PR fixes the issue -> https://github.com/argoproj-labs/argocd-operator/issues/223

Allows user to provide an external hostname as redirect-url to argocd-server without enabling ingress/route. 
When user explicitly sets the field `server.host` with a value and chooses not to enable ingress/route for it, the `url` in `argocd-cm` (argocd config map) should still be populated with the external hostname provided.

### DIFF
--- a/pkg/controller/argocd/util.go
+++ b/pkg/controller/argocd/util.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
 	"text/template"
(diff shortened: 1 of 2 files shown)


pkg/controller/argocd/util.go
```diff
@@ -186,6 +192,10 @@ func (r *ReconcileArgoCD) getArgoServerURI(cr *argoprojv1a1.ArgoCD) string {
 		}
 	}
 
+	if _, err := url.ParseRequestURI(host); err == nil {
+		return host
```
bigkevmcd, line 196: We should probably log something here to warn that we're ignoring the URL, otherwise it might be a bit confusing that a "valid" looking YAML didn't work, and no error messages?

shubhamagarwal19: like we discussed, i think it's OK to not have this check, since it says hostname rather than a URL. So we don't need to have this check !! 
worst case if a user provides a URL rather than a FQDN, url in the config map would end up something like `https://https://xyz.com`

bigkevmcd (approved):
Nice work, looks good to me.

PR #237, opened 2021-01-29
Update route API varaible from openshift package

chetan-rns:
This PR moves the logic to set the route availability variable to openshift package

### DIFF
--- a/pkg/reconciler/openshift/openshift.go
+++ b/pkg/reconciler/openshift/openshift.go
@@ -1,17 +1,20 @@
 package openshift
 
 import (
+	"log"
 	"os"
 	"strings"
 
 	argoprojv1alpha1 "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
-	rbacv1 "k8s.io/api/rbac/v1"
-
 	"github.com/argoproj-labs/argocd-operator/pkg/controller/argocd"
+	rbacv1 "k8s.io/api/rbac/v1"
 )
 
 func init() {
 	argocd.Register(reconcilerHook)
+	if err := argocd.VerifyRouteAPI(); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/reconciler/openshift/openshift.go
```diff
@@ -1,17 +1,20 @@
...
 	argoprojv1alpha1 "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
-	rbacv1 "k8s.io/api/rbac/v1"
-
 	"github.com/argoproj-labs/argocd-operator/pkg/controller/argocd"
+	rbacv1 "k8s.io/api/rbac/v1"
 )
 
 func init() {
 	argocd.Register(reconcilerHook)
+	if err := argocd.VerifyRouteAPI(); err != nil {
```
wtam2018, line 15: `routeAPIFound` won't be `true` unless operator is buit with `-tags=openshift`  Are you sure about this operator *is* compiled with `-tags=openshift` ?  Otherwise, the community ArgoCD operator won't have OpenShift Route supports.

pkg/reconciler/openshift/openshift.go
```diff
@@ -1,17 +1,20 @@
...
+		log.Fatal(err)
+	}
 }
```
wtam2018, line 19: @bigkevmcd This module should really live downstream.   As long as `argocd.Register() enables a way to registry hooks.   It is downstream's job to register reconciler hooks.  They are not something that upstream cares about or needs to maintain.

bigkevmcd: @wtam2018 I don't disagree, the separation between "gitops operator" and "argocd-operator" is not very clear, and if we're not planning on releasing the "argocd-operator" we should be clearer in our separation.

PR #238, opened 2021-02-01
Added openshift hook to use redhat specific redis image args

shubhamagarwal19:
Added hook to update the redis deployment spec for openshift cluster with redhat specifc redis image args.

### DIFF
--- a/pkg/controller/argocd/deployment.go
+++ b/pkg/controller/argocd/deployment.go
@@ -669,6 +669,10 @@ func (r *ReconcileArgoCD) reconcileRedisDeployment(cr *argoprojv1a1.ArgoCD) erro
 		Env:       proxyEnvVars(),
 	}}
 
+	if err := applyReconcilerHook(cr, deploy); err != nil {
+		return err
+	}
+
 	existing := newDeploymentWithSuffix("redis", "redis", cr)
 	if argoutil.IsObjectFound(r.client, cr.Namespace, existing.Name, existing) {
 		if cr.Spec.HA.Enabled {
--- a/pkg/controller/argocd/deployment_test.go
+++ b/pkg/controller/argocd/deployment_test.go
@@ -906,23 +906,25 @@ func TestReconcileArgoCD_reconcileRedisDeployment(t *testing.T) {
 	// tests reconciler hook for redis deployment
 	cr := makeTestArgoCD()
 	r := makeTestReconciler(t, cr)
-	Register(testReconcilerHook)
+
+	defer resetHooks()()
+	Register(testDeploymentHook)
 
 	assert.NilError(t, r.reconcileRedisDeployment(cr))
 	d := &appsv1.Deployment{}
-	assert.NilError(t, r.client.Get(context.TODO(), types.NamespacedName{Name: cr.Name+"-redis", Namespace: cr.Namespace}, d))
-	assert.DeepEqual(t, int32(4), *d.Spec.Replicas)
+	assert.NilError(t, r.client.Get(context.TODO(), types.NamespacedName{Name: cr.Name + "-redis", Namespace: cr.Namespace}, d))
+	assert.DeepEqual(t, int32(3), *d.Spec.Replicas)
 }
 
-func testReconcilerHook(cr *argoprojv1alpha1.ArgoCD, v interface{}) error {
-	switch o := v.(type) {
-	case *appsv1.Deployment:
-		var x int32 = 4
-		if o.ObjectMeta.Name == cr.ObjectMeta.Name+"-redis" {
-			o.Spec.Replicas = &x
-		}
-	}
-	return nil
+func TestReconcileArgoCD_reconcileRedisDeployment_with_error(t *testing.T) {
+	// tests reconciler hook for redis deployment
+	cr := makeTestArgoCD()
+	r := makeTestReconciler(t, cr)
+
+	defer resetHooks()()
+	Register(testErrorHook)
+
+	assert.Error(t, r.reconcileRedisDeployment(cr), "this is a test error")
 }
 
 func restoreEnv(t *testing.T) {
@@ -1019,4 +1021,3 @@ func controllerProcessors(n int32) argoCDOpt {
 		a.Spec.Controller.Processors.Status = n
 	}
 }
-
--- a/pkg/controller/argocd/deployment_test.go
+++ b/pkg/controller/argocd/deployment_test.go
@@ -902,6 +902,29 @@ func TestReconcileArgoCD_reconcileServerDeploymentChangedToInsecure(t *testing.T
 	}
 }
 
+func TestReconcileArgoCD_reconcileRedisDeployment(t *testing.T) {
+	// tests reconciler hook for redis deployment
+	cr := makeTestArgoCD()
+	r := makeTestReconciler(t, cr)
+	Register(testReconcilerHook)
+
+	assert.NilError(t, r.reconcileRedisDeployment(cr))
+	d := &appsv1.Deployment{}
+	assert.NilError(t, r.client.Get(context.TODO(), types.NamespacedName{Name: cr.Name+"-redis", Namespace: cr.Namespace}, d))
+	assert.DeepEqual(t, int32(4), *d.Spec.Replicas)
+}
+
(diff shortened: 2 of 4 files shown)


pkg/controller/argocd/deployment.go
```diff
@@ -669,6 +669,10 @@ func (r *ReconcileArgoCD) reconcileRedisDeployment(cr *argoprojv1a1.ArgoCD) erro
 		Env:       proxyEnvVars(),
 	}}
 
+	if err := applyReconcilerHook(cr, deploy); err != nil {
```
iam-veeramalla, line 672: Should we add a test to verify this behaviour ?

pkg/reconciler/openshift/openshift_test.go
```diff
@@ -107,3 +107,14 @@ func TestAllowedNamespaces(t *testing.T) {
...
+func TestReconcileArgoCD_reconcileRedisDeployment(t *testing.T) {
+	a := makeTestArgoCD()
+	testDeployment := makeTestDeployment()
+	testDeployment.ObjectMeta.Name = a.Name + "-" + "redis"
+
+	want := append([]string{"redis-server"}, testDeployment.Spec.Template.Spec.Containers[0].Args ...)
+
+	assert.NilError(t, reconcilerHook(a, testDeployment))
+	assert.DeepEqual(t, testDeployment.Spec.Template.Spec.Containers[0].Args, want)
+}
```
iam-veeramalla, line 120: Can you add more tests here ?

pkg/reconciler/openshift/openshift.go
```diff
@@ -22,6 +24,12 @@ func reconcilerHook(cr *argoprojv1alpha1.ArgoCD, v interface{}) error {
 				o.Rules = append(o.Rules, policyRulesForClusterConfig()...)
 			}
 		}
+	case *appsv1.Deployment:
+		if o.ObjectMeta.Name == cr.Name+"-redis" {
+			o.Spec.Template.Spec.Containers = []corev1.Container{{
+				Args: append([]string{"redis-server"}, o.Spec.Template.Spec.Containers[0].Args...),
```
wtam2018, line 30: The argument list of Red Hat Redis looks like this:      
```
      - args:
        - redis-server
        - --save
        - ""
        - --appendonly
        - "no"
        - --protected-mode
        - "no"
```

wtam2018: The order does not matter.  I guess you can have `--protected-mode no` in front of `--save`.

wtam2018: You can test with `registry.redhat.io/rhel8/redis-5` image.

shubhamagarwal19: So what we are doing here is inserting `redis-server` arg in the existing set of args defined for stock redis. This makes our arg list look something like this 
```
      - args:
        - redis-server
        - --save
        - ""
        - --appendonly
        - "no"
 ```
Do we need turn off the protected mode explicitly? since we didn't do this for the stock redis !!

Comment (wtam2018) [reply pkg/reconciler/openshift/openshift.go:30]: I think we do.   The stock image runs with protected mode by default.

Comment (wtam2018) [reply pkg/reconciler/openshift/openshift.go:30]: @shubhamagarwal19 @iam-veeramalla @bigkevmcd Can you explain the go mechanism to "compile out"  the openshift specific package by default?

**Comment by bigkevmcd** [comment]
@wtam2018 The operator's main package uses https://golang.org/cmd/go/#hdr-Build_constraints but, it doesn't need to, as anything that imports the code can also explicitly import the package.

File: pkg/reconciler/openshift/openshift.go (hunk 3 above)
Comment (bigkevmcd) [reply pkg/reconciler/openshift/openshift.go:30]: @wtam2018 I'm not sure if you're familiar or not with Go build tags, but you can read about them here https://golang.org/pkg/go/build/#hdr-Build_Constraints

The way the extension functions (hooks) work is that they register on import from the init function, and the build tags for the [manager](https://github.com/argoproj-labs/argocd-operator/blob/master/cmd/manager/openshift.go) only trigger the import when the openshift tags are provided.

This will have no impact on the gitops-operator, because it doesn't use the same `main` function, _but_ it can import the same package to register the hooks, and because it's openshift specific, it doesn't need conditional compilation tags, but could easily be configured with them.

File: pkg/controller/argocd/deployment.go (hunk 1 above)
Comment (shubhamagarwal19) [reply pkg/controller/argocd/deployment.go:672]: added a test to verify the behaviour.

File: pkg/reconciler/openshift/openshift_test.go (hunk 2 above)
Comment (shubhamagarwal19) [reply pkg/reconciler/openshift/openshift_test.go:126]: added another test case to verify that deployment hook is only executed for redis deployment.

File: pkg/reconciler/openshift/openshift.go (hunk 3 above)
Comment (shubhamagarwal19) [reply pkg/reconciler/openshift/openshift.go]: @wtam2018, i verified the changes using the image `registry.redhat.io/rhscl/redis-5-rhel7:latest` without specifying `--protected-mode` arg. works fine !!

File: pkg/controller/argocd/deployment_test.go [hunk 4]
```diff
@@ -902,6 +902,30 @@ func TestReconcileArgoCD_reconcileServerDeploymentChangedToInsecure(t *testing.T
...
+	// tests reconciler hook for redis deployment
+	cr := makeTestArgoCD()
+	r := makeTestReconciler(t, cr)
+	Register(testReconcilerHook)
+	var expected int32 = 4
+
+	assert.NilError(t, r.reconcileRedisDeployment(cr))
+	d := &appsv1.Deployment{}
+	assert.NilError(t, r.client.Get(context.TODO(), types.NamespacedName{Name: cr.Name+"-redis", Namespace: cr.Namespace}, d))
+	assert.DeepEqual(t, &expected, d.Spec.Replicas)
```
bigkevmcd, line 915: You could simplify this to...

```go
assert.DeepEqual(t, int32(4), *d.Spec.Replicas)
```

pkg/controller/argocd/deployment.go
bigkevmcd: The current version `6087379c3dbd275e48110b0be0d59431613c6959` doesn't seem to have this change?

pkg/controller/argocd/deployment_test.go
```diff
@@ -902,6 +902,30 @@ func TestReconcileArgoCD_reconcileServerDeploymentChangedToInsecure(t *testing.T
...
+	switch o := v.(type) {
+	case *appsv1.Deployment:
+		var x int32 = 4
+		if o.ObjectMeta.Name == cr.ObjectMeta.Name+"-redis" {
+			o.Spec.Replicas = &x
+		}
+	}
+	return nil
+}
+
```
bigkevmcd, line 928: You could reuse the `testDeploymentHook` function here instead of creating another function?

bigkevmcd (changes requested):

pkg/controller/argocd/deployment.go
shubhamagarwal19: yeah it doesn't, it is a part of previous commit. This PR has 2 commits! `22523cb1babcae613531be61180e87bef47ab8b1`

pkg/controller/argocd/deployment_test.go
shubhamagarwal19: `testDeploymentHook` would work for any deployment, i wanted to keep this test hook specifically for Redis Deployment.

pkg/controller/argocd/deployment_test.go
shubhamagarwal19: done !! thanks

pkg/reconciler/openshift/openshift.go
shubhamagarwal19: Added `protected-mode=no` to the list of args for redhat based redis since stock redis turns the `protected-mode` off by default.

pkg/controller/argocd/deployment_test.go
shubhamagarwal19: using `testDeploymentHook` as we don't need redis specific name change only need to verify the behaviour

bigkevmcd (approved):

PR #241, opened 2021-02-04
fixed reconciliation of redis stateful sets

shubhamagarwal19:
This PR fixes the issues with reconciliation logic of redis stateful sets.


bigkevmcd (approved):
Nice bug fix!

Thanks

PR #242, opened 2021-02-05
fix: do not deploy service for dex when disabled

shubhamagarwal19:
Current reconciliation logic does not check if Dex is enabled/disabled and creates a service for it.
Added a check to verify that Dex is enabled before creating a service for it.


iam-veeramalla:
While this change is perfect, I was wondering if this should be handled in GITOPS operator ?

shubhamagarwal19:
> While this change is perfect, I was wondering if this should be handled in GITOPS operator ?

I think (correct me if i am wrong), the idea of disabling Dex was not limited to GITOPS operator or ArgoCD on Openshift. A user might wanna use ArgoCD operator but disable Dex. 
PR for refrence -> https://github.com/argoproj-labs/argocd-operator/pull/218

bigkevmcd (approved):
Thanks for this, I missed it when I added the original code.

This reminds me that I think we should have a `reconcileDex` which would take care of the Deployment _and_ the Service at the same time, rather than split over the two disparate files.

PR #247, opened 2021-02-17
Remove cluster resources on instance deletion

chetan-rns:
Currently, the service accounts and cluster role resources are not cleaned up when the argocd instance is deleted. This PR introduces a finalizer and a pre-delete hook logic to delete the remaining resources. 

How to test:
1. Create a test namespace with the name `argocd-test`.
2. Run the operator locally.
```shell
operator-sdk run local --watch-namespace argocd-test
```
3. Create an instance of argocd in argocd-test. This will create cluster roles and bindings.
4. Delete the instance.
5. Observe the cluster roles, cluster role bindings, and service accounts getting deleted.

### DIFF
--- a/pkg/controller/argocd/argocd_controller.go
+++ b/pkg/controller/argocd/argocd_controller.go
@@ -90,13 +90,39 @@ func (r *ReconcileArgoCD) Reconcile(request reconcile.Request) (reconcile.Result
 		return reconcile.Result{}, err
 	}
 
-	if argocd.ObjectMeta.DeletionTimestamp.IsZero() {
-		if err := r.reconcileResources(argocd); err != nil {
-			// Error reconciling ArgoCD sub-resources - requeue the request.
+	if argocd.GetDeletionTimestamp() != nil {
+		if argocd.IsDeletionFinalizerPresent() {
+			if err := r.deleteClusterResources(argocd); err != nil {
+				return reconcile.Result{}, err
+			}
+
+			if err := r.removeDeletionFinalizer(argocd); err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+		return reconcile.Result{}, nil
+	}
+
+	if !argocd.IsDeletionFinalizerPresent() {
+		if err := r.addDeletionFinalizer(argocd); err != nil {
 			return reconcile.Result{}, err
 		}
 	}
 
+	if err := r.reconcileResources(argocd); err != nil {
+		// Error reconciling ArgoCD sub-resources - requeue the request.
+		return reconcile.Result{}, err
+	}
+
 	// Return and don't requeue
 	return reconcile.Result{}, nil
+}
+
+func containsString(slice []string, s string) bool {
+	for _, item := range slice {
+		if item == s {
+			return true
+		}
+	}
+	return false
 }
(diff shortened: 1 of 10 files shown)


shubhamagarwal19:
This PR https://github.com/argoproj-labs/argocd-operator/pull/248 would affect how the name is generated for cluster scoped resources !!

chetan-rns:
Thanks @shubhamagarwal19! I'll wait for the name change PR to go in and rebase my branch

chetan-rns:
@shubhamagarwal19 As suggested by @bigkevmcd, I've updated the PR to use labels and selectors instead of names. It can be merged independently

pkg/controller/argocd/role.go
```diff
@@ -116,3 +117,12 @@ func (r *ReconcileArgoCD) reconcileClusterRole(name string, policyRules []v1.Pol
 	applyReconcilerHook(cr, clusterRole)
 	return clusterRole, r.client.Update(context.TODO(), clusterRole)
 }
+
+func deleteClusterRoles(c client.Client, clusterRoleList *v1.ClusterRoleList) error {
+	for _, clusterRole := range clusterRoleList.Items {
+		if err := c.Delete(context.TODO(), &clusterRole); err != nil {
+			return err
```
bigkevmcd, line 124: Might want to return return `fmt.Errorf("failed to delete ClusterRole %q during cleanup: %w", clusterRole.Name, err)` in case the returned error doesn't indicate it.

pkg/controller/argocd/util.go
```diff
@@ -655,6 +659,67 @@ func (r *ReconcileArgoCD) reconcileResources(cr *argoprojv1a1.ArgoCD) error {
 	return nil
 }
 
+func (r *ReconcileArgoCD) deleteClusterResources(cr *argoprojv1a1.ArgoCD) error {
+	selector, err := argocdInstanceSelector(cr.Name)
+	if err != nil {
+		return err
```
bigkevmcd, line 665: See above, for the note about the error.

Something to consider, if the result of `filterObjectsBySelector` returned a useful enough error, then you could just return it, instead of rewrapping :-)

bigkevmcd (commented):
Looks great, thanks for the changes to use labels.

pkg/controller/argocd/argocd_controller.go
```diff
@@ -90,13 +90,44 @@ func (r *ReconcileArgoCD) Reconcile(request reconcile.Request) (reconcile.Result
...
+	if err := r.reconcileResources(argocd); err != nil {
+		// Error reconciling ArgoCD sub-resources - requeue the request.
+		return reconcile.Result{}, err
+	}
+
 	// Return and don't requeue
 	return reconcile.Result{}, nil
 }
+
+func containsString(slice []string, s string) bool {
```
wtam2018, line 126: this function is not used.

chetan-rns: I've removed the function

pkg/controller/argocd/util.go
chetan-rns: Updated the error message

pkg/controller/argocd/role.go
chetan-rns: Updated the error message

bigkevmcd (approved):

dewan-ahmed:
@sbose78 @chetan-rns  I validated this as part of https://issues.redhat.com/browse/GITOPS-574
logs: https://github.com/redhat-developer/gitops-operator/pull/54#issuecomment-784434111

PR #248, opened 2021-02-18
generate unique name for cluster scoped resources

shubhamagarwal19:
/kind bug

**[Issue]**
Currently the name generated for all resources follows the convention **"instance-name"-"service-name"**, which would work fine for namespace scoped resources.   

The problem occurs in cluster scoped resources.   Each argocd instance needs to reference to its copy of cluster resources.   We need to generate unique name for cluster resources.  Otherwise,  ArgoCD instance A cannot  have its cluster role that is different than ArgoCD instance B for example.

We need to generate unique name for cluster resources for 
@shubhamagarwal19 explained to me that this function is to generate unique name for cluster resources that are used by different argocd instances.   The fact that we need to generate "ClusterResourceName" can still sound strange.  Cluster resource names are re   You are generating a unique name for cluster resource.  IMHO renaming the function to "generateUniqueResourceName" is a better func name.

**[Proposal]**
Append namespace in cluster scoped resources. **"instance-name"-"namespace"-"service-name"**

**[Reproduce-Issue]**

- Using the current operator, create 2 ArgoCD instances in separate namespaces, with the **same instance name**, SAME being very important. 
- You would see that **Cluster Role** and **Role Binding** are only created for the first instance.
- You can confirm this with the **Service Account Mapping in the Cluster Role Binding**. Only the SA for the first instance would be mapped in the Cluster Role Binding. 

**[How-to-test-fix]**

- Again create 2 ArgoCD instances using the latest operator. 
- Make sure the cluster scoped resources (currently just **Cluster Role** and **Role Binding**) are created with the naming convention provided above for both instances respectively. 
- Both **Cluster Role Bindings** should have the mapping for Service Account for each ArgoCD instance.

### DIFF
--- a/pkg/controller/argocd/role.go
+++ b/pkg/controller/argocd/role.go
@@ -36,15 +36,15 @@ func generateResourceName(argoComponentName string, cr *argoprojv1a1.ArgoCD) str
 	return cr.Name + "-" + argoComponentName
 }
 
-// generateClusterResourceName generates unique names for cluster scoped resources
-func generateClusterResourceName(argoComponentName string, cr *argoprojv1a1.ArgoCD) string {
+// GenerateUniqueResourceName generates unique names for cluster scoped resources
+func GenerateUniqueResourceName(argoComponentName string, cr *argoprojv1a1.ArgoCD) string {
 	return cr.Name + "-" + cr.Namespace + "-" + argoComponentName
 }
 
 func newClusterRole(name string, rules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD) *v1.ClusterRole {
 	return &v1.ClusterRole{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        generateClusterResourceName(name, cr),
+			Name:        GenerateUniqueResourceName(name, cr),
 			Labels:      labelsForCluster(cr),
 			Annotations: annotationsForCluster(cr),
 		},
--- a/pkg/controller/argocd/role.go
+++ b/pkg/controller/argocd/role.go
@@ -36,10 +36,15 @@ func generateResourceName(argoComponentName string, cr *argoprojv1a1.ArgoCD) str
 	return cr.Name + "-" + argoComponentName
 }
 
+// generateClusterResourceName generates unique names for cluster scoped resources
+func generateClusterResourceName(argoComponentName string, cr *argoprojv1a1.ArgoCD) string {
+	return cr.Name + "-" + cr.Namespace + "-" + argoComponentName
+}
+
 func newClusterRole(name string, rules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD) *v1.ClusterRole {
 	return &v1.ClusterRole{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        generateResourceName(name, cr),
+			Name:        generateClusterResourceName(name, cr),
 			Labels:      labelsForCluster(cr),
 			Annotations: annotationsForCluster(cr),
 		},
--- a/pkg/reconciler/openshift/openshift.go
+++ b/pkg/reconciler/openshift/openshift.go
@@ -18,7 +18,7 @@ func init() {
 func reconcilerHook(cr *argoprojv1alpha1.ArgoCD, v interface{}) error {
 	switch o := v.(type) {
 	case *rbacv1.ClusterRole:
-		if o.ObjectMeta.Name == cr.ObjectMeta.Name + cr.ObjectMeta.Namespace + "-argocd-application-controller" {
+		if o.ObjectMeta.Name == argocd.GenerateUniqueResourceName("argocd-application-controller", cr) {
 			if allowedNamespace(cr.ObjectMeta.Namespace, os.Getenv("ARGOCD_CLUSTER_CONFIG_NAMESPACES")) {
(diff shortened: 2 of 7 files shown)


iam-veeramalla:
LGTM

chetan-rns:
@shubhamagarwal19 Should we get rid of the prefix `cluster-` for clusterrolebindings? I don't see this prefix added for clusterroles

shubhamagarwal19:
> @shubhamagarwal19 Should we get rid of the prefix `cluster-` for clusterrolebindings? I don't see this prefix added for clusterroles

Thats a good point !! now that we have `namespace` added to the cluster resource name we can get rid of `cluster-` prefix

pkg/controller/argocd/role.go
```diff
@@ -36,10 +36,15 @@ func generateResourceName(argoComponentName string, cr *argoprojv1a1.ArgoCD) str
 	return cr.Name + "-" + argoComponentName
 }
 
+// generateClusterResourceName generates unique names for cluster scoped resources
+func generateClusterResourceName(argoComponentName string, cr *argoprojv1a1.ArgoCD) string {
+	return cr.Name + "-" + cr.Namespace + "-" + argoComponentName
```
wtam2018, line 41: I think <namespace>-<name>-<comp-name> naming scheme can more logically reflect the nmaing hirearchy.

wtam2018:
@shubhamagarwal19  Please assess any impact to the proposed name changes.

shubhamagarwal19:
> @shubhamagarwal19 Please assess any impact to the proposed name changes. That is, anything is affected by this change and anything needs to be adjusted accordingly.

@wtam2018, did look into if this change would impact anything and all i could think of was the upgrade scenario, where the upgraded GitOps Operator would deploy new cluster roles/role-bindings without cleaning up the existing ones. 
But @sbose78, suggested it is OK, since we are not promising cleanup on upgrade for TP.

pkg/controller/argocd/role.go
```diff
@@ -36,10 +36,15 @@ func generateResourceName(argoComponentName string, cr *argoprojv1a1.ArgoCD) str
 	return cr.Name + "-" + argoComponentName
 }
 
+// generateClusterResourceName generates unique names for cluster scoped resources
+func generateClusterResourceName(argoComponentName string, cr *argoprojv1a1.ArgoCD) string {
```
wtam2018, line 40: Here is where I got confused.   The resource is actually not a cluster scoped resource.  It is a namespace scoped resource, right?   Can you please fix the comment and rename the function to something like "generateUniqueResourceName".

pkg/reconciler/openshift/openshift.go
```diff
@@ -18,7 +18,7 @@ func init() {
 func reconcilerHook(cr *argoprojv1alpha1.ArgoCD, v interface{}) error {
 	switch o := v.(type) {
 	case *rbacv1.ClusterRole:
-		if o.ObjectMeta.Name == cr.ObjectMeta.Name+"-argocd-application-controller" {
+		if o.ObjectMeta.Name == cr.ObjectMeta.Name + cr.ObjectMeta.Namespace + "-argocd-application-controller" {
```
wtam2018, line 21: Please call the "generateClusterResourceName"

wtam2018 (commented):
Please add "how to test" including expected result.

pkg/reconciler/openshift/openshift.go
shubhamagarwal19: @wtam2018, is it OK if we make this change as part of refactoring? We could probably create exported functions `GenerateResourceName` and `GenerateClusterResourceName` in the utils package and use them across the code rather than changing it just here?

chetan-rns:
@wtam2018 @shubhamagarwal19 This approach may introduce weird names like `argocd-cluster-openshift-gitops-argocd-application-controller`. I think `generateName:` would help us here to create unique names. But I'm not sure about the complications of using generateName

pkg/reconciler/openshift/openshift.go
shubhamagarwal19: Done !!!

shubhamagarwal19:
> @wtam2018 @shubhamagarwal19 This approach may introduce weird names like `argocd-cluster-openshift-gitops-argocd-application-controller`. I think `generateName:` would help us here to create unique names. But I'm not sure about the complications of using generateName

cc: @sbose78, What do you think about this ?

wtam2018:
@shubhamagarwal19 the steps to test is a bit vague.   Can you put more detailed steps?   E.g. what cluster resources should be checked?

shubhamagarwal19:
> @shubhamagarwal19 the steps to test is a bit vague. Can you put more detailed steps? E.g. what cluster resources should be checked?

I have highlighted the resources that are needed to be verified and how to be verified also provided both **[Reproduce-Issue]** and **[How-to-test-fix]** in the description. 
I can maybe add more detailed steps in the Jira :)

pkg/reconciler/openshift/openshift.go
shubhamagarwal19: Also, updated the function name from `GenerateClusterResourceName` to `GenerateUniqueResourceName`, as that makes more sense.

wtam2018 (approved):
LGTM
@shubhamagarwal19  Thanks!

PR #249, opened 2021-02-19
make dex init container resilient to restarts

bakito:
Issue: the dec init container is stuck in a crash loop if the binary is aready in use.

Resolution: do not overwrite existing files as implemented in argocd PR https://github.com/argoproj/argo-cd/pull/3136


wtam2018 (commented):
LGTM

bigkevmcd (approved):

PR #254, opened 2021-02-24
Provide admin access to ArgoCD application-controller for local ns

shubhamagarwal19:
This PR adds a hook that provides the `argocd-application-controller` to have `admin` privileges only for the namespace it is installed in and not cluster-wide for Openshift deployments. 

Note: A custom role for `argocd-application-controller` with limited permissions will still be created but for openshift deployments, `Cluster Role - admin` will be used for creating the `RoleBinding`

### DIFF
--- a/pkg/controller/argocd/hooks_test.go
+++ b/pkg/controller/argocd/hooks_test.go
@@ -29,6 +29,14 @@ func testClusterRoleHook(cr *argoprojv1alpha1.ArgoCD, v interface{}) error {
 	return nil
 }
 
+func testRoleBindingHook(cr *argoprojv1alpha1.ArgoCD, v interface{}) error {
+	switch o := v.(type) {
+	case *v1.RoleBinding:
+		o.RoleRef.Name = "test-admin-role"
+	}
+	return nil
+}
+
 func testErrorHook(cr *argoprojv1alpha1.ArgoCD, v interface{}) error {
 	return errMsg
 }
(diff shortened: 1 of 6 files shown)


pkg/controller/argocd/hooks_test.go
```diff
@@ -29,6 +29,14 @@ func testClusterRoleHook(cr *argoprojv1alpha1.ArgoCD, v interface{}) error {
 	return nil
 }
 
+func testRoleBindingHook(cr *argoprojv1alpha1.ArgoCD, v interface{}) error {
+	switch o := v.(type) {
+	case *v1.RoleBinding:
+		o.RoleRef.Name = "admin"
```
bigkevmcd, line 35: I'd tend to make this more obvious that it's _not_ the default value.

e.g. something like `just-a-test-admin` way it's clearer in the _test_ that it's definitely applying the change.

shubhamagarwal19: updated it to `test-admin-role` !!

bigkevmcd (approved):
I'm _ok_ with this, with the shift to unique names, this will break, because the name won't be so clear-cut, and you'll need to look at labels, but I think it's probably easier to fix this everywhere when the labels are sorted.

bigkevmcd:
> I'm _ok_ with this, with the shift to unique names, this will break, because the name won't be so clear-cut, and you'll need to look at labels, but I think it's probably easier to fix this everywhere when the labels are sorted.

PR #257, opened 2021-02-25
Fix dex clientID with unit test

wtam2018:
Add unit test for @sabre1041's fix

### DIFF
--- a/pkg/controller/argocd/configmap_test.go
+++ b/pkg/controller/argocd/configmap_test.go
@@ -22,9 +22,12 @@ import (
 
 	argoprojv1alpha1 "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
 	"github.com/argoproj-labs/argocd-operator/pkg/common"
+	"github.com/argoproj-labs/argocd-operator/pkg/controller/argoutil"
 	"github.com/google/go-cmp/cmp"
+	"gopkg.in/yaml.v2"
 	"gotest.tools/assert"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/controller/argocd/configmap_test.go
```diff
@@ -140,6 +143,51 @@ func TestReconcileArgoCD_reconcileArgoConfigMap_withDisableAdmin(t *testing.T) {
...
+
+	dex, ok := cm.Data["dex.config"]
+	if !ok {
+		t.Fatal("reconcileArgoConfigMap with dex failed")
+	}
+
+	m := make(map[string]interface{})
+	err = yaml.Unmarshal([]byte(dex), &m)
+	if err != nil {
+		t.Fatalf("unmarshal '%q' failed: %v", dex, err)
```
bigkevmcd, line 180: Nit, you don't need `'%q'` here, `%q` will quote the value.

I suspect this could also be done with `assert.NilError` it has functionality for formatting a string.

pkg/controller/argocd/configmap_test.go
```diff
@@ -140,6 +143,51 @@ func TestReconcileArgoCD_reconcileArgoConfigMap_withDisableAdmin(t *testing.T) {
...
+	assert.NilError(t, err)
+
+	dex, ok := cm.Data["dex.config"]
+	if !ok {
+		t.Fatal("reconcileArgoConfigMap with dex failed")
+	}
+
+	m := make(map[string]interface{})
+	err = yaml.Unmarshal([]byte(dex), &m)
+	if err != nil {
```
bigkevmcd, line 179: Minor nit, you could use something like

```assert.NilError(t, err, fmt.Sprintf("failed to unmarshal %s", dex))` here

bigkevmcd (approved):
Looks ok to me.

PR #265, opened 2021-03-10
OpenShift hooks for HA

jopit:

### DIFF
--- a/docs/reference/argocd.md
+++ b/docs/reference/argocd.md
@@ -337,7 +337,7 @@ The following properties are available for configuring High Availability for the
 Name | Default | Description
 --- | --- | ---
 Enabled | `false` | Toggle High Availability support globally for Argo CD.
-RedisProxyImage | `haproxy` | The Redis HAProxy container image. This overrides the `ARGOCD_REDIS_HA_IMAGE`.
+RedisProxyImage | `haproxy` | The Redis HAProxy container image. This overrides the `ARGOCD_REDIS_HA_PROXY_IMAGE`environment variable.
 RedisProxyVersion | `2.0.4` | The tag to use for the Redis HAProxy container image.
 
 ### HA Example
--- a/pkg/common/keys.go
+++ b/pkg/common/keys.go
@@ -176,8 +176,12 @@ const (
 	// to used for the argocd container.
 	ArgoCDImageEnvName = "ARGOCD_IMAGE"
 
+	// ArgoCDRedisHAProxyImageEnvName is the environment variable used to get the image
+	// to used for the Redis HA Proxy container.
+	ArgoCDRedisHAProxyImageEnvName = "ARGOCD_REDIS_HA_PROXY_IMAGE"
+
 	// ArgoCDRedisHAImageEnvName is the environment variable used to get the image
-	// to used for the Redis HA Proxy image.
+	// to used for the the Redis container in HA mode.
 	ArgoCDRedisHAImageEnvName = "ARGOCD_REDIS_HA_IMAGE"
 
 	// ArgoCDRedisImageEnvName is the environment variable used to get the image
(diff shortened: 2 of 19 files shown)


pkg/common/keys.go
```diff
@@ -176,8 +176,12 @@ const (
 	// to used for the argocd container.
 	ArgoCDImageEnvName = "ARGOCD_IMAGE"
 
+	// ArgoCDRedisHAProxyImageEnvName is the environment variable used to get the image
+	// to used for the Redis HA Proxy container.
+	ArgoCDRedisHAProxyImageEnvName = "ARGOCD_REDIS_HA_PROXY_IMAGE"
```
wtam2018, line 181: Please update the doc in dcos/reference/argcd.md

pkg/controller/argocd/policyrule.go
```diff
@@ -92,6 +92,20 @@ func policyRuleForRedisHa() []v1.PolicyRule {
 				"get",
 			},
 		},
+		{
```
wtam2018, line 95: Can these be added by reconciler hook?

jopit: It can't easily be added by the reconciler hook.  The hook expects an instance of the ArgoCD CR as a parameter, which we don't have in this function.  Even if we passed `nil` into the hook for the ArgoCD instance, there is no way to tell what the PolicyRule is for.  We would have to blindly assume that its for Redis HA.  This doesn't seem very clean.  I think to do it cleanly, we'd need another kind of hook into which we can pass additional information to identify what the passed object is for.  I don't think it's worth it.

For the corresponding change in ArgoCD itself, these values were added directly:
https://github.com/argoproj/argo-cd/pull/4800
https://github.com/argoproj/argo-cd/pull/5078

pkg/common/keys.go
jopit: done

pkg/controller/argocd/policyrule.go
wtam2018: Thanks for looking into it.  I was thinking since we have created the reconciler hook to isolate OpenShift specific code, we probably should use it to put all OpenShift roles in it.  I think we can pass ArgoCD CR to policyRuleForRedisHa() and we can add another argument to the hook func to idenifty the intent of the resource.  I agree that the code can get more complicated.  WDYT?

BTW, the reference to the corresponding change in ArgoCD is not quite the same situation as in here.  The OpenShift roles are added to an overlay specific to OpenShift.  We could generate an install yaml without OpenShift roles but it isn't needed.

docs/reference/argocd.md
```diff
@@ -337,7 +337,7 @@ The following properties are available for configuring High Availability for the
 Name | Default | Description
 --- | --- | ---
 Enabled | `false` | Toggle High Availability support globally for Argo CD.
-RedisProxyImage | `haproxy` | The Redis HAProxy container image. This overrides the `ARGOCD_REDIS_HA_IMAGE`.
```
wtam2018, line 340: We are not removing `ARGOCD_REDIS_HA_IMAGE`, right?

jopit: No, it's still there.

wtam2018: Can you please double check?  It looks like it has been replaced from my end.

jopit: I didn't make any changes to where `ARGOCD_REDIS_HA_IMAGE` is used.  The doc above was wrong, `ARGOCD_REDIS_HA_IMAGE` was never used to specify the haproxy image, it was always used to specify the Redis image to use when in HA mode.  I had to introduce a new environment variable to allow the haproxy image to be specified.

The `ARGOCD_REDIS_HA_IMAGE` env var is used in the `getRedisHAContainerImage` function in  `pkg/controller/argocd/util.go`, which this PR does not modify

There is no way to override the Redis image or the Redis HA image in the ArgoCD CR

wtam2018 (approved):
LGTM
thanks @jopit !

PR #269, opened 2021-03-22
fix: add repository credentials from CR to the existing argocd config…

chetan-rns:
Updating the `RepositoryCredentials` field should update the existing `argocd-cm` config map by adding the repository's credentials. This would eliminate the need for updating the config map manually.

How to Test:
1. Install the operator by creating a new catalog source
2. Create an Argo CD CR
3. Configure the Argo CD CR with your private repo info
```yaml
repositoryCredentials: |
    url: https://github.com/<org>/<private-repo>    // add your private repo URL
    usernameSecret:
        name: my-secret
        key: username
    passwordSecret:
        name: my-secret
        key: password
``` 
4. Check if the argocd-cm config map is updated with the above config
5. Check if Argo CD is able to deploy resources from private repo

### DIFF
--- a/docs/reference/argocd.md
+++ b/docs/reference/argocd.md
@@ -486,7 +486,7 @@ spec:
 
 Git repository credential templates to configure Argo CD to use upon creation of the cluster.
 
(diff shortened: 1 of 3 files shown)


wtam2018 (approved):
LGTM
@chetan-rns thanks!
Can you please add how to test your fix in description?

PR #305, opened 2021-04-30
fix: add compute resources to init containers

### LINKED ISSUE #304: Pods with init containers fail to start when installed into a namespace with resource quotas
**Describe the bug**
The Dex and HAProxy deployments and the Application Controller stateful set (when import is configured in the CR) define init containers, but do not specify the required compute resources.  This will cause the pods to fail to start when installed into a namespace that has resource quotas set.

**To Reproduce**
Install ArgoCD into a namespace that has resource quotas set, using a CR that specifies compute resources for the components.
The Dex pod will not start.

**Expected behavior**
All pods should start

jopit:
**What type of PR is this?**

/kind bug

This PR adds compute resources to definition of several of the ArgoCD init containers.  Without this, the associated pods will fail to start if ArgoCD is installed into a namespace which has resource quotas set.

Fixes #304 

To test, define a namespace that has resource quotas set, then create an ArgoCD CR that defines compute resources for its components, and install into the namespace.  All pods should start successfully.

### DIFF
--- a/pkg/controller/argocd/deployment.go
+++ b/pkg/controller/argocd/deployment.go
@@ -349,6 +349,7 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoprojv1a1.ArgoCD) error 
 		Image:           getArgoContainerImage(cr),
 		ImagePullPolicy: corev1.PullAlways,
 		Name:            "copyutil",
+		Resources: getDexResources(cr),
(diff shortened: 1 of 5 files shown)


wtam2018:
Is it possible to add some unit tests?

jopit:
@wtam2018 I've added some unit tests.

wtam2018 (approved):
LGTM
thanks @jopit

PR #314, opened 2021-05-12
fix: do not create Cluster Roles for non cluster config ArgoCD instance

shubhamagarwal19:
**What type of PR is this?**

> /kind bug
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

Fixes #?


pkg/controller/argocd/policyrule.go
```diff
@@ -8,6 +8,19 @@ import (
 func policyRuleForApplicationController() []v1.PolicyRule {
 
 	return []v1.PolicyRule{
+		{
+			APIGroups: []string{
+				"*",
```
wtam2018, line 13: Can you explain why this is added?

pkg/controller/argocd/policyrule.go
```diff
@@ -126,6 +139,17 @@ func policyRuleForDexServer() []v1.PolicyRule {
 
 func policyRuleForServer() []v1.PolicyRule {
 	return []v1.PolicyRule{
+		{
+			APIGroups: []string{
```
wtam2018, line 143: And pls explain why this is added.

pkg/controller/argocd/policyrule.go
shubhamagarwal19: These permissions were added as part of the ClusterRole (https://github.com/argoproj-labs/argocd-operator/pull/314/files#diff-e7925228b00ce545bae87b56e0167a16b69ae670b1ab4fda380089c88b766900L180), but now we don't have a cluster role associated with the service account of the application controller. 
So i have moved these permissions to the Role associated with the service account. This provides the application controller to read all the resources deployed within the same namespace.

pkg/controller/argocd/policyrule.go
shubhamagarwal19: For the same reason as described for the `application-controller`

pkg/controller/argocd/policyrule.go
wtam2018: @chetan-rns Do we need to move the following here?
```
{
			NonResourceURLs: []string{
				"*",
			},
			Verbs: []string{
				"get",
				"list",
			},
		},
    ```

File: pkg/controller/argocd/policyrule.go (hunk 2 above)
Comment (wtam2018) [reply pkg/controller/argocd/policyrule.go:143]: Do we need to move this here?
```
{
			APIGroups: []string{
				"",
			},
			Resources: []string{
				"pods",
				"pods/log",
			},
			Verbs: []string{
				"get",
			},
		},
        ```

wtam2018: Also, we should merge the rules of the same api group.

pkg/controller/argocd/role.go
```diff
@@ -107,20 +107,30 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
 }
 
 func (r *ReconcileArgoCD) reconcileClusterRole(name string, policyRules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD) (*v1.ClusterRole, error) {
-
 	clusterRole := newClusterRole(name, policyRules, cr)
-	err := r.client.Get(context.TODO(), types.NamespacedName{Name: clusterRole.Name}, clusterRole)
+	clusterRole.Rules = policyRules
```
wtam2018, line 111: this line is redundant and not needed.

pkg/controller/argocd/role.go
```diff
@@ -107,20 +107,30 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
 
-	clusterRole.Rules = policyRules
-	applyReconcilerHook(cr, clusterRole, "")
-	return clusterRole, r.client.Update(context.TODO(), clusterRole)
+	if len(clusterRole.Rules) == 0 {
+		return nil, r.client.Delete(context.TODO(), existingClusterRole)
+	}
+
+	existingClusterRole.Rules = policyRules
+	applyReconcilerHook(cr, existingClusterRole, "")
```
wtam2018, line 132: `applyReconcilerHook` is called the second time to policyRules.  Openshift policy rules will get appended again.

pkg/controller/argocd/rolebinding.go
```diff
@@ -157,6 +157,15 @@ func (r *ReconcileArgoCD) reconcileClusterRoleBinding(name string, role *v1.Clus
 		roleBinding = newClusterRoleBindingWithname(name, cr)
```
wtam2018, line 157: This like can be deleted.  (I know you didn't add it.)

pkg/controller/argocd/policyrule.go
shubhamagarwal19: `NonResourceURL` rules can only be added to a ClusterRole, since they are cluster scoped. So we cannot add them to a Role.

pkg/controller/argocd/service_account.go
```diff
@@ -132,14 +132,15 @@ func (r *ReconcileArgoCD) reconcileServiceAccountClusterPermissions(name string,
 	var sa *corev1.ServiceAccount
 	var err error
 
-	if role, err = r.reconcileClusterRole(name, rules, cr); err != nil {
+	sa, err = r.reconcileServiceAccount(name, cr)
```
wtam2018, line 135: Just curious why switching the order of these two reconcile func calls?

pkg/controller/argocd/role.go
shubhamagarwal19: `ReconcileClusterRole` is a generic function. I know right now the way we are calling it seems redundant but in the future, there can be a possibility where we need to create ClusterRoles and reconcile them. So it is for that.

pkg/controller/argocd/rolebinding.go
shubhamagarwal19: I think it is not redundant. When we do a `GET` on `clusterRoleBinding`, if it doesn't exist, it populates the `roleBinding` object with an empty struct, so we need to reinitialize it.

pkg/reconciler/openshift/openshift.go
```diff
@@ -133,6 +133,28 @@ func getArgsForRedhatHaRedisInitContainer() []string {
 // policyRulesForClusterConfig defines rules for cluster config.
 func policyRulesForClusterConfig() []rbacv1.PolicyRule {
 	return []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{
+				"*",
```
wtam2018, line 138: With this, we can get rid of `"compliance.openshift.io",` rule

.gitignore
```diff
@@ -79,3 +79,4 @@ tags
 .vscode/*
 .history
 # End of https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
+vendor/
```
wtam2018, line 82: please revert

pkg/controller/argocd/service_account.go
shubhamagarwal19: Now the returned `Role` can also be `nil`. So I wanted to perform those checks after the ServiceAccount is created. No specific reason other than that !!

pkg/controller/argocd/policyrule.go
shubhamagarwal19: We don't need to add this, since it has the permissions to read all the resources in all API groups within the namespace. So the existing rule takes care of this!

pkg/controller/argocd/role.go
shubhamagarwal19: `applyReconcilerHook` is being called on different objects. But I agree we don't need to call it again. I can directly substitute the rules!! Thanks, I'll update this!

pkg/controller/argocd/rolebinding.go
wtam2018, line 157: This like can be deleted. (I know you didn't add it.)

wtam2018: Also, it looks like `reconcileArgoApplier()` is a dead func.

shubhamagarwal19: > This like can be deleted. (I know you didn't add it.)

I think we need this. 
When we do a GET on clusterRoleBinding and if it doesn't exist, it populates the `roleBinding` object with an empty struct, so we need to reinitialize it.

pkg/controller/argocd/role.go
wtam2018: I think we should remove it if it is redundant and not needed.

pkg/controller/argocd/role.go
wtam2018: Different objects (clusterRole and existingClusterRole) but both objects embed the same policyRules if I read this right.  So, policyRules ends up with 2 x injected rules.

pkg/controller/argocd/rolebinding.go
wtam2018: name and cr do not change in the second call `newClusterRoleBindingWithname(name, cr).   `roleBinding` simply get assinged again without anyone touching.  The second call is clearly not needed.  But, if you are not comfortable removing, then let's leave it.

wtam2018: Nevermind.  I missed the RoleBinding passed in to the GET.

pkg/controller/argocd/role.go
shubhamagarwal19: Done !!! fixed it

pkg/reconciler/openshift/openshift.go
shubhamagarwal19: Removed this rule !! Thanks!

pkg/controller/argocd/rolebinding.go
shubhamagarwal19: removed `reconcileArgoApplier()` since it wasn't being called anywhere !!

pkg/controller/argocd/role.go
shubhamagarwal19: I think it's still needed because the definition of the function is to append the policy rules to the cluster role provided. Right now we are not passing any rules and only rely on the rules provided by the reconciler hook. 
But according to the definition of the function `reconcileClusterRole`, we should try and append the policy rules provided, irrespective of what is being provided.

pkg/controller/argocd/policyrule.go
```diff
@@ -8,6 +8,19 @@ import (
 func policyRuleForApplicationController() []v1.PolicyRule {
 
 	return []v1.PolicyRule{
+		{
```
wtam2018, line 11: On openshift, controller sa already has cluster admin https://github.com/argoproj-labs/argocd-operator/blob/master/pkg/reconciler/openshift/openshift.go#L48.   We should clean this up a bit to remove the delicated rules.  Here is the cluster admin role rules.  https://gist.githubusercontent.com/wtam2018/41f82ccee071cc3dc0f78b2d49805344/raw/ea09491d662747e16c57d7d43cebb7284fd092bb/gistfile1.txt

pkg/controller/argocd/role.go
wtam2018: The line above it (L110) creates the `clusterRole` object with `policyRules` as an argument.  So, the `Rules` field of `clusterRole` already has `policyRules` assigned.  If you remove the line `clusterRole.Rules = policyRules` what problem does it cause?    I am missing something?

pkg/controller/argocd/policyrule.go
shubhamagarwal19: This hook (https://github.com/argoproj-labs/argocd-operator/blob/master/pkg/reconciler/openshift/openshift.go#L48) binds the Cluster Role `admin` to a role binding which provides the `application-controller` `admin` access to the namespace it is installed in but not cluster-wide. 
while the rule defined here is for any ArgoCD instance irrespective of the cluster i.e Openshift or Non-Openshift will provide the access to get/list all resources for all API groups. 
Although I do agree, these roles are all over the place and we do need to look into it.

pkg/controller/argocd/role.go
shubhamagarwal19: ohhh sorry, I misunderstood your comment, Yes, that makes total sense, this line is redundant and I should remove it. Thanks !

shubhamagarwal19: Thanks for the comment, fixed it !!

pkg/controller/argocd/policyrule.go
wtam2018: @shubhamagarwal19 My point is  that on openshift hook, we already create cluster admin role (yes namespace scoped RoleBinding) bound to controller sa, policyRuleForApplicationController() is adding rules that already exist in cluster admin roles and we should remove duplicates here in policyRuleForApplicationController().

shubhamagarwal19: Gottcha, but what I mean is that the rules defined in `policyRuleForApplicationController()` are for any ArgoCD instance irrespective of the Cluster it is deployed on. 
The hooks are defined for OpenShift specific deployments and yes, it will have the admin privileges for the namespace, but if anyone is using the ArgoCD operator on a non Openshift Cluster, we would need these rules. 
Am i missing something here ?

wtam2018: You are right.  I thought `policyRuleForApplicationController` was invoked by the reconcileHook.  Nevermind then.

wtam2018: We should move the openshift hook to openshift operator as a future cleanup.

wtam2018 (approved):
LGTM
thanks @shubhamagarwal19 !

snorwin:
is it possible to release this fix soon?

shubhamagarwal19:
> is it possible to release this fix soon?

Hey @snorwin, we are waiting for a couple of more fixes to go in and then release a new version!!

Issue #312, opened 2021-05-10
Automatically install and configure Keycloak SSO for ArgoCD instance


wtam2018:
**Is your feature request related to a problem? Please describe.**
As an admin, I want to enable KeyCloak on my Argo CD instance by add the following stanza in the CR.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: ArgoCD

spec:
  sso:
    provider: keycloak
```
Argocd operator should create all configurations necessary to enable SSO for my instance.

As an admin, I want to disable KeyCloak SSO by removing sso provider in my CR.  All keycloak configurations and resources should be removed from cluster.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

iam-veeramalla:
Keycloak can automatically installed and configured to use.
```
apiVersion: argoproj.io/v1alpha1
kind: ArgoCD

spec:
  sso:
    provider: keycloak
```

PR #316, opened 2021-05-21
feat: Configure Keycloak SSO for Argo CD Instance(#312)

iam-veeramalla:

**What type of PR is this?**
> /kind enhancement

#312 

Yes

#312 

Fixes #?
#312 

1. Create a new version of operator from this commit and deploy the catalog source. Install the operator from operator hub.
2.  Create a new argo cd CR with `.spec.sso.provider: keycloak`
```
kind: ArgoCD
metadata:
  name: example-argocd
spec:
  sso:
    provider: keycloak
  server:
    route:
     enabled: true
```


cmd/manager/main.go
```diff
@@ -34,6 +34,8 @@ import (
 	"github.com/argoproj-labs/argocd-operator/version"
 
 	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	oappsv1 "github.com/openshift/api/apps/v1"
```
wtam2018, line 37: I'd rename the aliase to just `appsv1`.

pkg/controller/argocd/keycloak_client.go
```diff
@@ -0,0 +1,194 @@
...
+
+type httpclient struct {
+	requester requester
+	URL       string
+	token     string
+}
+
+// Creates a new realm for managing identity brokering between ArgoCD and openshift
+// using openshift-v4 as identity provider.
+func create(cfg *keycloakConfig) (string, error) {
```
wtam2018, line 30: rename function name to createRealm()?

pkg/apis/argoproj/v1alpha1/argocd_types.go
```diff
@@ -354,6 +354,11 @@ const (
 type ArgoCDSSOSpec struct {
 	// Provider installs and configures the given SSO Provider with Argo CD.
 	Provider SSOProviderType `json:"provider,omitempty"`
+	// InsecureMode should be set to true in the below cases.
+	// 1. Keycloak service is not accesible by service name due to Network policies.
+	// 2. Operator is run locally for development purpose and default Ingress Controller which generates self-signed certificates
+	// is used as the CA authority.
+	InsecureMode bool `json:"insecureMode,omitempty"`
```
wtam2018, line 361: `InsecureMode` is basically to turn off cert validation in the tls connecction between the keycloak api server, right?   I think we can just name the field `VerifyTLS` and default it to true.

pkg/controller/argocd/sso.go
```diff
@@ -45,19 +93,66 @@ func (r *ReconcileArgoCD) reconcileSSO(cr *argoprojv1a1.ArgoCD) error {
...
+				Namespace: templateInstanceRef.Namespace}, &existingTemplateInstance)
 			if err != nil && errors.IsNotFound(err) {
-				log.Info("Creating a new keycloak template instance")
+				log.Info(fmt.Sprintf("Template API found, Installing keycloak using openshift templates for ArgoCD %s in namespace %s",
+					cr.Name, cr.Namespace))
 				err = r.client.Create(context.TODO(), templateInstanceRef)
 				if err != nil {
 					return err
 				}
 			}
```
wtam2018, line 110: We need to handle if `Get` failure (other than IsNotFound())

pkg/controller/argocd/sso.go
```diff
@@ -45,19 +93,66 @@ func (r *ReconcileArgoCD) reconcileSSO(cr *argoprojv1a1.ArgoCD) error {
...
+
```
wtam2018, line 111: Hmmm ... we are not doing anything with `existingTemplateInstance`.

pkg/controller/argocd/sso.go
```diff
@@ -45,19 +93,66 @@ func (r *ReconcileArgoCD) reconcileSSO(cr *argoprojv1a1.ArgoCD) error {
...
+			existingDC := &oappsv1.DeploymentConfig{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      defaultKeycloakIdentifier,
+					Namespace: cr.Namespace,
+				},
+			}
+			r.client.Get(context.TODO(), types.NamespacedName{Name: existingDC.Name, Namespace: existingDC.Namespace}, existingDC)
```
wtam2018, line 118: We don't bother to error check?

pkg/controller/argocd/sso.go
```diff
@@ -16,15 +16,63 @@ package argocd
...
 )
 
-var templateAPIFound = false
+const (
+	// SuccessResonse is returned when a realm is created in keycloak.
+	successResonse = "201 Created"
+	// ExpectedReplicas is used to identify the keycloak running status.
+	expectedReplicas int32 = 1
+	// ServingCertSecretName is a secret that holds the service certificate.
+	servingCertSecretName = "sso-x509-https-secret"
```
wtam2018, line 36: Please search "sso-x509-https-secret" in keycloak.go and replace them with this constants.

pkg/controller/argocd/sso.go
```diff
@@ -16,15 +16,63 @@ package argocd
...
 	template "github.com/openshift/api/template/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
-var templateAPIFound = false
+const (
+	// SuccessResonse is returned when a realm is created in keycloak.
+	successResonse = "201 Created"
```
wtam2018, line 32: successResponse?   please fix typo.

pkg/apis/argoproj/v1alpha1/argocd_types.go
iam-veeramalla: The reason behind naming it as InsecureMode is to provide an insight that it should mostly not be used. I believe customers usually have their clusters setup with certificates signed by a valid CA. So it should not be used even if the operator is run using the operator-sdk outside the cluster for development purpose. 

Please suggest.

pkg/controller/argocd/keycloak.go
```diff
@@ -287,26 +327,28 @@ func getKeycloakRouteTemplate(ns string) *routev1.Route {
 	}
 }
 
-func newKeycloakTemplateInstance(ns string) (*template.TemplateInstance, error) {
-	tpl, err := newKeycloakTemplate(ns)
+func newKeycloakTemplateInstance(cr *argoprojv1a1.ArgoCD) (*template.TemplateInstance, error) {
+	tpl, err := newKeycloakTemplate(cr)
 	if err != nil {
 		return &template.TemplateInstance{}, err
```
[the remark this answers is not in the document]
iam-veeramalla: my bad !! Thanks :)

pkg/controller/argocd/sso.go
iam-veeramalla: ack ..

pkg/controller/argocd/sso.go
iam-veeramalla: Template Instance takes some time to create the deployment config, Till then reconciller will not be able to find the Deployment Config.

Should we handle the NotFoundError as log message ?  something like 
`TemplateInstance not found or in the creation state`

pkg/controller/argocd/sso.go
iam-veeramalla: ack , thank you :)

pkg/controller/argocd/sso.go
iam-veeramalla: ack, thank you :)

cmd/manager/main.go
iam-veeramalla: sure :)

pkg/apis/argoproj/v1alpha1/argocd_types.go
wtam2018: People know tlsverify should not be disabled in production.   We already use  `TLSVerify` some other struct in the CR.   `InsecureMode` and the comment there do not tell me what it does.   Someone will have to look into the code to figure out.   In the end, it is just a TLSVerify flag plus time sent on looking at the code to see what it does.

iam-veeramalla: ack ... made changes accordingly :)

pkg/controller/argocd/sso.go
iam-veeramalla: ack .. Thanks :)

pkg/controller/argocd/sso.go
iam-veeramalla: ack .. Thanks :)

pkg/controller/argocd/keycloak_client.go
```diff
@@ -0,0 +1,194 @@
...
+	c := &http.Client{Transport: transport}
+	return c, nil
+}
+
+// createTLSConfig constructs and returns a TLS Config with a root CA read
+// from the serverCert param if present, or a permissive config which
+// is insecure otherwise.
+// An Insecure config is returned also when .spec.SSO.insecure is set to true.
+func createTLSConfig(serverCert []byte, verifyTLS bool) (*tls.Config, error) {
+	if serverCert == nil || !verifyTLS {
```
wtam2018, line 151: I think we should error out if serverCert is nil.  In other words, server cert is required although it can be signed signed.

wtam2018 (approved):

PR #318, opened 2021-05-21
fix: permissions associated with non cluster ArgoCD instance

shubhamagarwal19:
**What type of PR is this?**
> /kind bug

This PR fixes the permissions associated with the non-cluster ArgoCD instance. 
- provides all api groups privileges to the application-controller sa (non-cluster) argocd for the namespace it is deployed in.  
- provides"read" privileges to the all apis/resources to server sa (non-cluster) argocd for the namespace it is deployed in.
- changes the default scope of the (non-cluster) argocd instance from all namespaces to the namespace it is deployed in.  NOTE: The operator only checks if a secret of the "in-cluster` exists. if it does, then it doesn't do anything. If there's no secret with the name `in-cluster`, it creates one with the namespaces only contains the namespace it is deployed in.  Users can manually update the secret either using the k8s apis or ArgoCD console.  Name of the secret: `<argocd-instance-name>-default-cluster-config`

This PR also brings feature parity of "cluster" argocd CD to community operator.   Previously, the `ARGOCD_CLUSTER_CONFIG_NAMESPACES` environment variable is only honored by the operator when it is compiled with `openshift` flag.   This behavior is enabled also to the community operator with this PR.  The `ARGOCD_CLUSTER_CONFIG_NAMESPACES` environment variable on the operator workload can be set to include a list of namespaces.  Argocd instances created on these namespaces will have cluster scope permission.

Fixes #?

- Use the latest build with the changes and deploy it on the cluster.
- Create an argocd instance for non-cluster config. 
- It should have `admin` privileges for the namespace it is deployed in. Should be able to deploy apps in the same namespace. 

Changes specific to Openshift - 
- The scope of the Argocd instance should be `namespace scoped`. 
  - To verify that, login to the Argocd console.
  - Go to, `Manage` -> `Clusters` -> `in-cluster` -> Under `General` section check the namespaces (it should be the namespace in which Argocd Instance is deployed)
- Similarly, the ArgoCD instance for cluster config should have the `namespaces` as empty. That is, the scope of the instance is cluster scoped. 
- Provide non-cluster config argocd instance deployed in `foo` namespace, the permissions to deploy apps in `bar` namespace by creating role/role_bindings. The deployment should fail.
 - Go to, `Manage` -> `Clusters` -> `in-cluster` -> Under `General` section add `bar` namespace to the `namespaces`. So it should look like `namespaces: foo,bar`. Deployment should now be successfully completed.


pkg/controller/argocd/rolebinding.go
```diff
@@ -124,12 +124,6 @@ func (r *ReconcileArgoCD) reconcileRoleBinding(name string, rules []v1.PolicyRul
 		Name:     role.Name,
 	}
 
-	if name == applicationController {
-		if err := applyReconcilerHook(cr, roleBinding, ""); err != nil {
-			return err
-		}
-	}
-
```
jannfis, line 132: Why did we remove this?

pkg/common/keys.go
```diff
@@ -194,4 +194,10 @@ const (
 
 	// ArgoCDDeletionFinalizer is a finalizer to implement pre-delete hooks
 	ArgoCDDeletionFinalizer = "argoproj.io/finalizer"
+
+	// ArgoCDDefaultServer is the default server address
+	ArgoCDDefaultServer = "https://kubernetes.default.svc"
+
+	// ArgoCDClusterSecretLabel is needed for cluster secrets
+	ArgoCDClusterSecretLabel = "argocd.argoproj.io/secret-type"
```
jannfis, line 202: I think this constant should be named `ArgoCDSecretTypeLabel` instead (it will be used for more than just `cluster` secrets).

pkg/controller/argocd/secret.go
```diff
@@ -379,6 +385,38 @@ func (r *ReconcileArgoCD) reconcileGrafanaSecret(cr *argoprojv1a1.ArgoCD) error
...
+
+	existingSecret := &corev1.Secret{}
+	if argoutil.IsObjectFound(r.client, secret.Namespace, secret.Name, existingSecret) {
+		if reflect.DeepEqual(secret.Data["namespaces"], existingSecret.Data["namespaces"]) {
+			// do nothing
+			return nil
+		}
+		existingSecret.Data["namespaces"] = secret.Data["namespaces"]
+		return r.client.Update(context.TODO(), existingSecret)
+	}
```
jannfis, line 415: Actually, this would prevent users from extending the scope of the Argo CD instance, right? So, it will prevent users from manually adding namespaces to be managed, but instead always reconcile the namespace back to the name of the namespace the Argo CD instance is running in. I think we _should_ let users override this, if they really would like to.

pkg/controller/argocd/secret.go
```diff
@@ -379,6 +385,38 @@ func (r *ReconcileArgoCD) reconcileGrafanaSecret(cr *argoprojv1a1.ArgoCD) error
...
+	})
+
+	secret.Data = map[string][]byte{
+		"config":     dataBytes,
+		"name":       []byte("in-cluster"),
+		"server":     []byte(common.ArgoCDDefaultServer),
+		"namespaces": []byte(cr.Namespace),
+	}
+
+	applyReconcilerHook(cr, secret, "")
```
jannfis, line 405: What will this actually do?

pkg/controller/argocd/secret.go
```diff
@@ -379,6 +385,38 @@ func (r *ReconcileArgoCD) reconcileGrafanaSecret(cr *argoprojv1a1.ArgoCD) error
 	return r.client.Create(context.TODO(), secret)
 }
 
+// reconcileClusterPermissionsSecret ensures ArgoCD instance is namespace-scoped
+func (r *ReconcileArgoCD) reconcileClusterPermissionsSecret(cr *argoprojv1a1.ArgoCD) error {
```
jannfis, line 389: I'm thinking this method should handle a scenario where user already has created a cluster secret for the `in-cluster` cluster, but with another name. Should we look for secrets in the current namespace with a label `argocd.argoproj.io/secret-type` and `.data.name` value of `in-cluster`?

I'm not quite sure how Argo CD would behave when there are two cluster with the same value of `.data.name`. Probably, the behaviour would be undefined and it would pick randomly the first one returned by the K8s API.

pkg/controller/argocd/secret.go
```diff
@@ -379,6 +385,38 @@ func (r *ReconcileArgoCD) reconcileGrafanaSecret(cr *argoprojv1a1.ArgoCD) error
...
+	secret := argoutil.NewSecretWithSuffix(cr.ObjectMeta, "namespaces")
```
jannfis, line 390: I think the we should use a suffix of `cluster-in-cluster` instead of `namespaces` here. Because the secret actually defines the connection to the cluster, and the namespace configuration is just a side-effect :)

jannfis (commented):
Thanks for coming up so quickly with a fix, @shubhamagarwal19 !

I have a few comments, please see below.

pkg/controller/argocd/rolebinding.go
shubhamagarwal19: Earlier we were applying the reconciler hook for openshift specific deployments. This would bind the openshift `Cluster Role admin` to the application controller SA. now we don't need this, since we are creating a custom admin role binding the instance with that.

pkg/controller/argocd/secret.go
shubhamagarwal19: Yeah, this is what i was confused with. But if we are providing an alternative that would create a new Secret, should we be allowing the user to update this manually?

pkg/controller/argocd/secret.go
shubhamagarwal19: This reconciler hook would be applied if the ArgoCD instance is a cluster config instance(openshift specific). For those instances, `namespaces` would be empty, which would allow the instance to access cluster scoped resources.

pkg/controller/argocd/secret.go
shubhamagarwal19: makes sense, i'll try this scenario out. Thanks !!

pkg/controller/argocd/secret.go
jannfis: So we'll have a hook in GitOps Operator that sets `.data.namespaces` to the empty string?

shubhamagarwal19: Yeah, basically it just deletes the key. 
https://github.com/argoproj-labs/argocd-operator/pull/318/files#diff-464eb8ec36b8f4f45d63589aa754c7933c5db014ec7185c15fa38198718d7303R63

pkg/common/keys.go
shubhamagarwal19: done !

pkg/controller/argocd/secret.go
shubhamagarwal19: done !

pkg/controller/argocd/secret.go
shubhamagarwal19: done!

pkg/controller/argocd/secret.go
shubhamagarwal19: done !

pkg/controller/argocd/secret.go
```diff
@@ -379,6 +386,50 @@ func (r *ReconcileArgoCD) reconcileGrafanaSecret(cr *argoprojv1a1.ArgoCD) error
...
+	})
+
+	secret.Data = map[string][]byte{
+		"config":     dataBytes,
+		"name":       []byte("in-cluster"),
+		"server":     []byte(common.ArgoCDDefaultServer),
+		"namespaces": []byte(cr.Namespace),
+	}
+
+	applyReconcilerHook(cr, secret, "")
```
wtam2018, line 406:
```suggestion
	if err := applyReconcilerHook(cr, secret, ""); err != nil {
		return err
	}
```

pkg/controller/argocd/secret.go
```diff
@@ -379,6 +386,50 @@ func (r *ReconcileArgoCD) reconcileGrafanaSecret(cr *argoprojv1a1.ArgoCD) error
...
+		return err
+	}
+
+	secrets := clusterSecrets.Items
+
+	for _, s := range secrets {
+		// check if cluster secret with name in-cluster exists
+		// do nothing if exists.
+		if string(s.Data["name"]) == "in-cluster" {
+			return nil
```
wtam2018, line 426: If in-cluster secret exists, we are not updating the "namespaces" field.  I thought we needed to reconcile the namespaces field to add argocd's namespace.

pkg/controller/argocd/secret.go
```diff
@@ -379,6 +386,50 @@ func (r *ReconcileArgoCD) reconcileGrafanaSecret(cr *argoprojv1a1.ArgoCD) error
...
+	if err := r.client.List(context.TODO(), clusterSecrets, opts); err != nil {
+		return err
+	}
+
+	secrets := clusterSecrets.Items
+
+	for _, s := range secrets {
+		// check if cluster secret with name in-cluster exists
+		// do nothing if exists.
+		if string(s.Data["name"]) == "in-cluster" {
```
wtam2018, line 425: It is better to find by server address than by name because the name can changed and does not have to be "in-cluster".

pkg/controller/argocd/secret.go
shubhamagarwal19: we don't want to do that right now, since we want the users to be able to manually update the Secret if they wish to grant additional permissions to the ArgoCD instance. 
So for that, i have added a `NOTE` in the description.

pkg/controller/argocd/secret.go
shubhamagarwal19: No i think the point here is to not have multiple secrets for `in-cluster` configuration. The name has to be `in-cluster` AFAIK. This actually is not the name of the secret, it is infact the name of the cluster configuration, which i am assume will always be `in-cluster`. 
@jannfis can maybe provide a better insight on this!

pkg/controller/argocd/secret.go
shubhamagarwal19: done! Thanks

pkg/controller/argocd/secret.go
jannfis: @wtam2018 has a good point there, the user may change the symbolic name of the default cluster from `in-cluster` to something else. The constant here is the server address, `https://kubernetes.default.svc`.

shubhamagarwal19: Thanks @jannfis @wtam2018 !! fixed it !

pkg/controller/argocd/secret.go
```diff
@@ -379,6 +388,57 @@ func (r *ReconcileArgoCD) reconcileGrafanaSecret(cr *argoprojv1a1.ArgoCD) error
 	return r.client.Create(context.TODO(), secret)
 }
 
+// reconcileClusterPermissionsSecret ensures ArgoCD instance is namespace-scoped
+func (r *ReconcileArgoCD) reconcileClusterPermissionsSecret(cr *argoprojv1a1.ArgoCD) error {
+	// do not create the secret if the key is not set.
+	if _, ok := os.LookupEnv("ARGOCD_CLUSTER_CONFIG_NAMESPACES"); !ok {
+		return nil
+	}
```
jannfis, line 396: I believe this environment variable will always be set. Shouldn't we rather make sure that we return when the current namespace is not in the list of namespaces exposed via this env var?

For example, if `ARGOCD_CLUSTER_CONFIG_NAMESPACES` is set to `openshift-gitops, argocd-cluster`, then we return with a no-op here if we are reconciling in either `openshift-gitops` or `argocd-cluster` namespace.

jannfis: Ah, hum, I might be mistaken here. So the intent here is that we only do this if the operator is being part of OpenShift GitOps, and that's when the variable will be set, correct?

shubhamagarwal19: Yes right !! We are only using this check to distinguish between Community Operator and Openshift Gitop OPerator operations !!

sbose78: > I believe this environment variable will always be set. Shouldn't we rather make sure that we return when the current namespace is not in the list of namespaces exposed via this env var?

Yes.

We shouldn't have to inject any special downstream logic here. Whether "ARGOCD_CLUSTER_CONFIG_NAMESPACES" has "foo" or "openshift-gitops", it shouldn't matter.

pkg/controller/argocd/secret.go
wtam2018: This patch is supposed to fix existing argocd instances.   We should not require users to manually edit the secret the exising argoco instances.   BTW, please make sure you test the upgrade path.  If the namespaces is empty, it is likely that the users have not edit the secret.  In this case, we should add the namespace there.  Otherwise, if the namespaces is not empty, we can leave it.

pkg/reconciler/openshift/openshift.go
```diff
@@ -83,6 +87,24 @@ func getPolicyRuleForRedisHa() rbacv1.PolicyRule {
 	}
 }
 
+func getPolicyRuleForApplicationController() []rbacv1.PolicyRule {
+	return []rbacv1.PolicyRule{
```
wtam2018, line 91: I don't see why this is openshift specific and can't just go in policyRuleForApplicationController().

Also, I tested the upgrade and rbac is not fixed and existing instances still have permission problems after grading.

pkg/controller/argocd/secret.go
wtam2018: @jannfis 
I created an namespace scoped argocd instance using v1.1.1.   I went to setup and view my in-cluster cluster and did not change anything.   I then upgraded my operator.   My argocd instance is still nu-usuable.   One of the reasons is the namespaces is not patched.

pkg/reconciler/openshift/openshift.go
shubhamagarwal19: By the default the rule provided in `policyRuleForApplicationController()` is for the community operator which is to provide the `"*"` admin access to all api groups and resources. 
But we don't want that for our Openshift Deployments. So we aggregate read permissions for all resources with Cluster Role Admin. This rule provides read access to all api groups and resources.

pkg/controller/argocd/secret.go
jannfis: With this logic, and the environment _not_ set at all, the operator will currently:

* Set up namespace-scoped permissions for the `argocd-application-controller`
* _Not_ create a cluster secret to make the Argo CD instance namespace-scoped

resulting in an error.

pkg/reconciler/openshift/openshift.go
wtam2018: Why community operator grant omni permission?

amitkrout:
AFAIU the CI does not comprises e2e test. CI only build and push the freshly compiled operator image through travis [1] for each pr we trigger, plus UTs verification through GitHub Actions [2]. I think we can introduce some kind of CI infra that helps us to verify e2e test for each pr we trigger. WDYT ?

I have one suggestion to implement OpenShift CI test infra [3] which is powerful and provides the CI infrastructure to run e2e test against a freshly compiled operator image. 

Here it is a well documented steps on how to test operators built with the operator SDK and deploy through OLM - https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/

[1] https://github.com/argoproj-labs/argocd-operator/blob/master/.travis.yml
[2] https://github.com/argoproj-labs/argocd-operator/blob/master/.github/workflows/go.yml
[3] https://github.com/openshift/release

ping @sbose78 @wtam2018 @jannfis

pkg/controller/argocd/role.go
```diff
@@ -85,6 +86,21 @@ func (r *ReconcileArgoCD) reconcileRoles(cr *argoprojv1a1.ArgoCD) (role *v1.Role
 
 // reconcileRole
 func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD) (*v1.Role, error) {
+	if name == applicationController {
+		if err := applyReconcilerHook(cr, &policyRules, "policyRulesForApplicationController"); err != nil {
+			return nil, err
+		}
+		if !reflect.DeepEqual(policyRules, policyRuleForApplicationController()) {
+			admin := &v1.ClusterRole{}
+			if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "admin"}, admin); err != nil {
```
wtam2018, line 95: clusterrole admin is openshift specific.  It should be handled in the openshift hook.

pkg/controller/argocd/rolebinding.go
```diff
@@ -124,19 +126,22 @@ func (r *ReconcileArgoCD) reconcileRoleBinding(name string, rules []v1.PolicyRul
...
-		return r.client.Update(context.TODO(), roleBinding)
+
+		// if the RoleRef changes, delete the existing role binding and create a new one
+		if !reflect.DeepEqual(roleBinding.RoleRef, existingRoleBinding.RoleRef) {
+			if err = r.client.Delete(context.TODO(), existingRoleBinding); err != nil {
+				return err
+			}
+		} else {
+			existingRoleBinding.Subjects = roleBinding.Subjects
+			return r.client.Update(context.TODO(), existingRoleBinding)
```
wtam2018, line 143: If there is no changes, we should update it.

pkg/controller/argocd/role.go
shubhamagarwal19: moved this to openshift hooks!!

pkg/controller/argocd/rolebinding.go
shubhamagarwal19: if there is no change, k8s api would handle it and won't update anything !! if there's a change it'll update the object.

amitkrout:
@shubhamagarwal19 Can you please share the test scenarios you want QE to review and verify.

pkg/controller/argocd/rolebinding.go
wtam2018: This logic looks odd.  We treat RoleRef differently.  We delete/re-recreate existingRoleBinding if RoleRef has changed but we just re-assign Subjects and Update the existingRoleBinding.   Why is the difference?

pkg/controller/argocd/rolebinding.go
```diff
@@ -124,19 +126,22 @@ func (r *ReconcileArgoCD) reconcileRoleBinding(name string, rules []v1.PolicyRul
 		Name:     role.Name,
 	}
 
-	if name == applicationController {
-		if err := applyReconcilerHook(cr, roleBinding, ""); err != nil {
-			return err
-		}
-	}
-
 	controllerutil.SetControllerReference(cr, roleBinding, r.scheme)
```
wtam2018, line 129: A new `roleBinding` only gets created if it does not exist already and `roleRef` has changed.

pkg/controller/argocd/rolebinding_test.go
```diff
@@ -86,17 +86,3 @@ func TestReconcileArgoCD_reconcileClusterRoleBinding(t *testing.T) {
 	clusterRoleBinding = &rbacv1.ClusterRoleBinding{}
 	assert.NilError(t, r.client.Get(context.TODO(), types.NamespacedName{Name: expectedName}, clusterRoleBinding))
 }
-
-func TestReconcileArgoCD_reconcileRoleBinding_application_controller(t *testing.T) {
```
wtam2018, line 90: Please add back this test (make any adjustment necessary).

pkg/controller/argocd/secret.go
wtam2018: Retested successfully with latest commit.

amitkrout:
> AFAIU the CI does not comprises e2e test. CI only build and push the freshly compiled operator image through travis [1] for each pr we trigger, plus UTs verification through GitHub Actions [2]. I think we can introduce some kind of CI infra that helps us to verify e2e test for each pr we trigger. WDYT ?
> 
> I have one suggestion to implement OpenShift CI test infra [3] which is powerful and provides the CI infrastructure to run e2e test against a freshly compiled operator image.
> 
> Here it is a well documented steps on how to test operators built with the operator SDK and deploy through OLM - https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/
> 
> [1] https://github.com/argoproj-labs/argocd-operator/blob/master/.travis.yml
> [2] https://github.com/argoproj-labs/argocd-operator/blob/master/.github/workflows/go.yml
> [3] https://github.com/openshift/release
> 
> ping @sbose78 @wtam2018 @jannfis

I will work on this post GitOps Operator 1.1.2 release.

pkg/controller/argocd/rolebinding.go
shubhamagarwal19: This would add the Controller reference to the `newRoleBinding` object only which gets created in L147, but it makes sense to move it just above the line where the role is actually getting created. 
Done !!

pkg/controller/argocd/rolebinding_test.go
shubhamagarwal19: This test was added to test if the hooks were executed when a new role binding was being created but now we have removed the hook itself so we don't need this test case at all. 
We don't call `applyReconcilerHook` while reconciling the `roleBinding` so i think it is not needed anymore.

wtam2018 (approved):

PR #326, opened 2021-05-27
feat: Enable service CA for reencrypt routes to argocd-server on OpenShift

### LINKED ISSUE #235: Changing route termination to reencrypt fails
Hello,

First of all, thank you for providing such a nice tool and operator.

I run ArgoCD operator 0.0.14 on top of OpenShift 4.6. And I have recently installed a custom default ingress certificate for the routers [1].
Until then I was running with the usual "self-signed" browser warnings with the simple configuration:
```
  server:
    route:
      enabled: true
```

I wanted the ArgoCD route to benefit from this default certificate and thus tried to re-encrypt the connection. So I changed the route configuration in the CRD:
```
  server:
    route:
      enabled: true
      tls:
        termination: reencrypt
```
Unfortunately, it failed to handle the request properly and replied with the usual OpenShift "Application is not available" page.
The error page was served with the new trusted default certificate though :)

I played with the various route fields but I haven't found a configuration that works. 

As a side-note, I also gave a try to let the internal connection unencrypted and rely on edge termination (even though fully encrypted seems nicer). I thought I could achieve this with insecure: true and termination: edge combination but it also failed (I assume because of #217 and #220)

Any hint at what I am missing?
Thanks!

[1] https://docs.openshift.com/container-platform/4.6/security/certificates/replacing-default-ingress-certificate.html

jannfis:

**What type of PR is this?**

/kind enhancement

This PR introduces functionality to properly enable the `reencrypt` feature for the OpenShift route to `argocd-server` service. It leverages the OpenShift service CA to request a TLS certificate and key to be injected into the `argocd-server-tls` secret, so that the route controller can successfully validate the TLS connection between the controller and the service.

It will be automatically setup if the route to the server is configured with `reencrypt` termination policy.

Fixes #235 

To test this, you will need to use a recent build of Argo CD, e.g. applying this CR into a namespace of your choice:

```yaml
kind: ArgoCD
metadata:
  name: argocd
spec:
  image: quay.io/argoproj/argocd
  server:
    route:
      enabled: true
      tls:
        termination: reencrypt
```

**On a virgin cluster**

1. Apply CRDs

```shell
make install
```
2. Run operator locally: `make run`
3. Apply above CR to a namespace of your choice
4. Wait until everything has been reconciled and all Argo CD workloads are up & running
5. Verify that `argocd-server` service has been annotated correctly with the `service.beta.openshift.io/serving-cert-secret-name` annotation and a value of `argocd-server-tls`:

```shell
$ oc get service argocd-server -o jsonpath='{.metadata.annotations}'
{"service.alpha.openshift.io/serving-cert-signed-by":"openshift-service-serving-signer@1622106000","service.beta.openshift.io/serving-cert-secret-name":"argocd-server-tls","service.beta.openshift.io/serving-cert-signed-by":"openshift-service-serving-signer@1622106000"}
```

6. Verify that `argocd-server-tls` secret has been created in the namespace:

```shell
$ oc get secret argocd-server-tls
NAME                TYPE                DATA   AGE
argocd-server-tls   kubernetes.io/tls   2      76s
```

7. Verify that the route has been created with `reencrypt` policy:

```shell
$ oc get route argocd-server 
NAME            HOST/PORT                                                                           PATH   SERVICES        PORT    TERMINATION   WILDCARD
argocd-server   argocd-server-argocd.apps.whatever.host.example.com          argocd-server   https   reencrypt     None
```

8. Copy the URL from the route's HOST/PORT field and use it to connect to your Argo CD. You most likely get a browser warning about insecure certificate, ignore it. 

9. If the Argo CD login screen appears, test was successful.


docs/usage/routes.md
```diff
@@ -45,3 +45,78 @@ To get the password for the admin user:
...
+
+To enable `reencrypt` mode, you can use the following configuration:
+
+```yaml
+spec:
+  server:
+    route:
+      enabled: true
+      tls:
+        termination: passthrough
```
iam-veeramalla, line 120: This should be reencrypt :)

docs/usage/routes.md
```diff
@@ -45,3 +45,78 @@ To get the password for the admin user:
...
+```
```
iam-veeramalla, line 121: I think it is better to recommend 
insecureEdgeTerminationPolicy: Redirect

docs/usage/routes.md
```diff
@@ -45,3 +45,78 @@ To get the password for the admin user:
...
+For this to work, the route controller needs to be able to validate the Argo CD server's TLS
+certificate, otherwise the request will fail.
+
+If you enable `reencrypt` mode in an OCP cluster, the Operator will request a valid TLS
+certificate for the `argocd-server` service from OpenShift's Service CA, which is sufficient
+for satisfying the validation constraints of the route controller. The Service CA will issue
+this certificate to a secret named `argocd-server-tls` in the operand's namespace if it does
+not yet exist.
+
+When you later chose to switch back to another TLS termination policy, you should manually
```
iam-veeramalla, line 109: We don't want to handle this in the code ? I think we already have the watchers set for openshift routes, if we want to handle this, we can add a predicate.

Is there any major harm if the user forgets to delete the secret after switching the termination from re-encrypt to something else?

jannfis: I'm not sure whether it would be the correct thing to delete the secret, because we never created it in the first place - it was created by the Service CA operator. But maybe it would make sense to delete it from the Operator, when we see that Auto TLS has been disabled and we can safely assume that the secret was created by the Service CA, and not by the user.

docs/usage/routes.md
jannfis: Thanks, fixed!

docs/usage/routes.md
jannfis: Thanks, added!

iam-veeramalla (approved):
LGTM !! As discussed, Feel free to add a switch to disable OpenShift service CA in this or a new PR.

PR #338, opened 2021-06-14
fix: Properly setup edge termination for server route

jannfis:

**What type of PR is this?**

> /kind bug

This PR properly sets up the route of the server for edge termination. It changes behavior as follows:

1. The default route termination policy will now be `edge`, regardless of `.spec.Server.Insecure` setting
2. If the policy is `edge` (either _implicit_ or _explicit_), `argocd-server` workload will be started with `--insecure` flag

**ATTN**

This is a slightly breaking change, as it changes the default route termination policy from `passthrough` to `edge`. 

We do believe that it's better to have `edge` termination as the default, instead of `passthrough`, as `passthrough` by default uses a self-signed certificate on the Argo CD server endpoint. `edge` will terminate at the cluster, and route traffic within the cluster as plain text to Argo CD components.

The behavior previous to this PR is:
* If `.spec.server.route.enabled`  is set to `true`, and `.spec.server.route.tls` is not set, `passthrough` termination will be the default, serving a self-signed TLS certificate
* User has to set `.spec.server.insecure: true` to enable `edge` termination policy (`.spec.server.insecure` was the leading knob)

The behavior after this PR will be:
* Changing `.spec.server.insecure` manually will be discouraged, it will have no effect on route configuration
* If `.spec.server.route.enabled`  is set to `true`, and `.spec.server.route.tls` is not set, `edge` termination will be the default and the server workload will be _implicitly_ set to run _with_ `--insecure`
*  If `.spec.server.route.enabled`  is set to `true`, and `.spec.server.route.tls.termination` is set to either `passthrough` or `reencrypt`, the server workload will be _implicitly_ set to run _without_ `--insecure`

Fixes #?

1. Create `ArgoCD` with a configuration that enables the route, e.g.
```yaml
kind: ArgoCD
metadata:
  name: argocd
spec:
  server:
    route:
      enabled: true
```
2. Have the operator reconcile the CR
3. Check that a route has been created, and that `edge` termination policy is set and the target port is set to `http` (*not* `https`): `oc get route argocd-server`
4. Check that the `argocd-server` deployment has been configured with the `--insecure` flag: `oc describe deployment argocd-server`
4. Navigate to the URL of the route with your browser
5. Argo CD login page should appear

Then, change the route's TLS properties to enable `passthrough` mode:

1. `oc patch argocd argocd --type merge --patch '{"spec":{"server":{"route":{"tls":{"termination": "passthrough"}}}}}''`
2. Check the route has been changed to `passthrough` policy and target port has changed to `https`: `oc get route argocd-server`

Then, switch back to `edge` mode:

1. `oc patch argocd argocd --type merge --patch '{"spec":{"server":{"route":{"tls":{"termination": "edge"}}}}}''`
2. Check the route has been changed to `edge` policy, target port has changed to `http`: `oc get route argocd-server`
3. Check that the `argocd-server` deployment has been re-configured with the `--insecure` flag: `oc describe deployment argocd-server`

### DIFF
--- a/docs/reference/argocd.md
+++ b/docs/reference/argocd.md
@@ -942,6 +942,13 @@ Path | `/` | The path for the Route.
 TLS | [Object] | The TLSConfig for the Route.
 WildcardPolicy| `None` | The wildcard policy for the Route. Can be one of `Subdomain` or `None`.
 
+If the route is enabled, it will use a default termination policy of `edge`
+for handling requests. With edge termination, the `Server.Insecure` setting
+will be automatically set to `true`. This will terminate TLS at the edge of
+your cluster, but the communication between the route and Argo CD will be
+unencrypted. You can change the way TLS is handled by defining an alternate
+TLS termination policy for the server route.
+
 ### Server Example
 
 The following example shows all properties set to the default values.
--- a/pkg/controller/argocd/deployment.go
+++ b/pkg/controller/argocd/deployment.go
@@ -202,7 +202,8 @@ func getArgoServerCommand(cr *argoprojv1a1.ArgoCD) []string {
 	cmd := make([]string, 0)
 	cmd = append(cmd, "argocd-server")
 
-	if getArgoServerInsecure(cr) {
+	// The --insecure flag is implicit for when using route with edge termination
+	if getArgoServerInsecure(cr) || serverRouteIsEdgeTermination(cr) {
 		cmd = append(cmd, "--insecure")
 	}
 
--- a/pkg/controller/argocd/deployment_test.go
+++ b/pkg/controller/argocd/deployment_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/argoproj-labs/argocd-operator/pkg/common"
+	v1 "github.com/openshift/api/route/v1"
 	"gotest.tools/assert"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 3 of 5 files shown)


iam-veeramalla:
Thanks @jannfis , PR looks great. I am convinced with making `--insecure`(starting argocd server in insecure mode) as default for argocd-server when the route termination is `Edge`.

I am trying to understand the advantages of choosing Edge as default termination over Passthrough. Are we saying, we believe No certificate is better than Self signed certificate ? Are there no downsides of starting ArgoCD in `--insecure` mode by default?

jannfis:
I have rebased #326 to latest master, now that we have Argo CD v2.1 as the default image to be installed.

Once that is merged, we could make `reencrypt` the default. User accessing the Argo CD route would then be presented the cluster's route certificate, which hopefully will be valid for them. TLS between router and Argo CD service would be encrypted and validated using the cluster's service CA. This would make the complete experience end-to-end protected using TLS.

WDYT?

iam-veeramalla:
> I have rebased #326 to latest master, now that we have Argo CD v2.1 as the default image to be installed.
> 
> Once that is merged, we could make `reencrypt` the default. User accessing the Argo CD route would then be presented the cluster's route certificate, which hopefully will be valid for them. TLS between router and Argo CD service would be encrypted and validated using the cluster's service CA. This would make the complete experience end-to-end protected using TLS.
> 
> WDYT?

Makes sense.

wtam2018:
@jannfis  please resolve conflicts.

iam-veeramalla:
> @jannfis please resolve conflicts.

@wtam2018 This PR can be closed as we can use re-encrypt as default with Argo CD v2.1 (Please refer #326 ). The part of the PR which starts Argo CD server with `--insecure` flag may just be needed. However, imho I think ask the user to add the `--insecure` flag by himself so that the user understands that choose to start Argo CD server in an insecure way.

wtam2018:
Closing this issue as we are in favor of move secure re-encrypt mode over edge mode.

PR #346, opened 2021-06-23
allow operator to manage permissions based on namespace labels

### LINKED ISSUE #352: Automating K8s RBAC for Argo CD instances

Proposed by @jannfis 

Currently, the integration pattern for the Operator is the following

* Use the out-of-the-box (OOTB) Argo CD instance to manage some cluster resources
* Use a dedicated Argo CD instance to manage an application in the instance’s target namespace

When a user creates a new Argo CD instance with a target namespace of e.g. “myapp”, the operator will automatically create Kubernetes RBAC permissions in the target namespace in the form of a Role and a RoleBinding for the ServiceAccount used by the Argo CD instance’s application controller. 

This is fine if the target namespace is the only namespace being managed by the Argo CD instance. However, users also want to use this instance to manage multiple namespaces in the cluster. The current recommendation is to create the Role and RoleBinding objects manually. This can be cumbersome to set up, complicated to manage and should be automated.

### Overview

We should give users an easy way to configure namespaces that should be managed by a given Argo CD instance.

I envision this by using a label on any namespace:
`argocd.argoproj.io/managed-by: <name of Argo CD instance>`

Each namespace should be only manageable by a single Argo CD instance, but one Argo CD instance will manage N namespaces, effectively establishing a 1:n relationship.

This pattern will work best with people creating new projects (namespaces) manually on their cluster. They can simply give a specific Argo CD instance named “example” permissions to the namespace “myproj” by issuing the following command:

`$ kubectl label namespace myproj argocd.argoproj.io/managed-by=example`

Likewise, revoking permissions is as simple as running

`$ kubectl label namespace myproj argocd.argoproj.io/managed-by-`

By leveraging project templates, this could be already automated at project creation time.

Given a scenario where the Argo CD instance has permissions to create Namespace resources in the cluster, and such a Namespace resource exists in the application’s manifests, it is not clear yet whether the approach will work. Argo CD will create the namespace from the resource, possibly with the `managed-by` annotation, as the first action of the initial sync and continue with the other resources. However, the Operator might not have created appropriate RBAC resources yet, and the sync will fail due to permission denied. I believe the auto-retry feature on sync will succeed the sync on the next retry, however, this has to be proven.

### RBAC resources managed by the operator

The Role and RoleBinding resources created by the Operator should be clearly indicated as such. The Operator must explicitly set the ownerReferences on the resources, and the proposal is to use an additional label for easier indexing and retrieval of the resources. The proposed label is the same as for the Namespace resource:

    `argocd.argoproj.io/managed-by: <name of Argo CD instance>`

The `ownerReferences` will instruct the Garbage Collector of Kubernetes to remove the resources whenever the Operator’s Operand is uninstalled from the cluster, so that a proper cleanup is done automatically.

In the MVP implementation of this proposal, the Role should give namespace admin permissions. Going forward, we can think of ways to provide templates for Roles that will get reconciled, so that permissions can be further locked down.
Reconciliation process
The  Operator will perform reconciliation of appropriate Role and RoleBinding namespaces that are annotated as proposed above.

The Operator will establish a watch on Namespace resources, and if it sees such a label is being added, modified or removed, reconcile appropriate Cluster and ClusterRole resources into that namespace.

The Operator knows which instances of Argo CD currently exist, and it also has the appropriate permissions for reconciling Cluster and ClusterRole resources everywhere.

When a namespace is annotated with the `argocd.argoproj.io/managed-by: <value> `label, the  Operator will:

- Check for existing Role and RoleBinding resources in the namespace that carry a label of `argocd.argoproj.io/managed-by`, and if it matches the value of the annotation with the same name set on the Namespace

- If values differ between the `argocd.argoproj.io/managed-by` on the RBAC resources and the Namespace resource, the operator will delete the existing Role and RoleBinding resources from the Namespace

- If values are the same, the Operator will reconcile the existing Role and RoleBinding resources to the version as deemed correct by the Operator

- If no such Role and RoleBinding with a label `argocd.argoproj.io/managed-by` are found, the Operator will create them

### Further possible enhancements

Permission levels / templates
Another annotation, e.g. `argocd.argoproj.io/permission-level`: <value>, could be specified on the namespace to indicate what kind of permissions Argo CD application controller should receive for this namespace.

Using this annotation, we could also provide some predefined permission levels additional to “admin” permissions, and also let the user define their own permission templates in the operand, which they can refer to from the annotation.

**Additional context**

shubhamagarwal19:
**What type of PR is this?**
> /kind enhancement

This PR addresses the need to manually provide the permissions to ArgoCD instance for managing a new namespace. This feature would allow the user to simply label a namespace with the label `argocd.argoproj.io/managed-by: <namespace of argocd>` and the operator would create the required permissions. 

Fixes  #352

// TODO

- [x] Add Unit Tests

### DIFF
--- a/pkg/common/keys.go
+++ b/pkg/common/keys.go
@@ -201,6 +201,6 @@ const (
 	// ArgoCDSecretTypeLabel is needed for cluster secrets
 	ArgoCDSecretTypeLabel = "argocd.argoproj.io/secret-type"
 
-	// ArgoCDManagedNamespaceLabel is needed to identify namespace managed by an instance on ArgoCD
-	ArgoCDManagedNamespaceLabel = "argocd.argoproj.io/managed-by"
+	// ArgoCDManagedByLabel is needed to identify namespace managed by an instance on ArgoCD
+	ArgoCDManagedByLabel = "argocd.argoproj.io/managed-by"
 )
--- a/pkg/common/keys.go
+++ b/pkg/common/keys.go
@@ -200,4 +200,7 @@ const (
 
 	// ArgoCDSecretTypeLabel is needed for cluster secrets
 	ArgoCDSecretTypeLabel = "argocd.argoproj.io/secret-type"
+
+	// ArgoCDManagedNamespaceLabel is needed to identify namespace managed by an instance on ArgoCD
+	ArgoCDManagedNamespaceLabel = "argocd.argoproj.io/managed-by"
 )
--- a/pkg/controller/argocd/argocd_controller.go
+++ b/pkg/controller/argocd/argocd_controller.go
@@ -17,16 +17,18 @@ package argocd
 import (
 	"context"
 	"fmt"
+	argoproj "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
+	"github.com/argoproj-labs/argocd-operator/pkg/common"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
-
-	argoproj "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
 )
 
 // blank assignment to verify that ReconcileArgoCD implements reconcile.Reconciler
@@ -57,7 +59,7 @@ func add(mgr manager.Manager, r *ReconcileArgoCD) error {
 	}
 
 	// Register watches for all controller resources
-	if err := watchResources(c, r.clusterResourceMapper, r.tlsSecretMapper); err != nil {
+	if err := watchResources(c, r.clusterResourceMapper, r.tlsSecretMapper, r.namespaceResourceMapper); err != nil {
 		return err
 	}
 
@@ -92,6 +94,18 @@ func (r *ReconcileArgoCD) Reconcile(request reconcile.Request) (reconcile.Result
 		}
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
+	}
+
+	namespace := &corev1.Namespace{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: request.Namespace}, namespace); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if val, ok := namespace.Labels[common.ArgoCDManagedNamespaceLabel]; !ok || val != argocd.Namespace {
+		namespace.Labels[common.ArgoCDManagedNamespaceLabel] = argocd.Namespace
+		if err = r.client.Update(context.TODO(), namespace); err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
 	if argocd.GetDeletionTimestamp() != nil {
--- a/pkg/controller/argocd/argocd_controller.go
+++ b/pkg/controller/argocd/argocd_controller.go
@@ -102,6 +102,9 @@ func (r *ReconcileArgoCD) Reconcile(request reconcile.Request) (reconcile.Result
 	}
 
 	if val, ok := namespace.Labels[common.ArgoCDManagedNamespaceLabel]; !ok || val != argocd.Namespace {
+		if namespace.Labels == nil {
+			namespace.Labels = make(map[string]string)
+		}
 		namespace.Labels[common.ArgoCDManagedNamespaceLabel] = argocd.Namespace
 		if err = r.client.Update(context.TODO(), namespace); err != nil {
 			return reconcile.Result{}, err
--- a/pkg/controller/argocd/argocd_controller.go
+++ b/pkg/controller/argocd/argocd_controller.go
@@ -17,6 +17,7 @@ package argocd
 import (
 	"context"
 	"fmt"
+
 	argoproj "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
 	"github.com/argoproj-labs/argocd-operator/pkg/common"
 
@@ -96,25 +97,14 @@ func (r *ReconcileArgoCD) Reconcile(request reconcile.Request) (reconcile.Result
 		return reconcile.Result{}, err
 	}
 
-	namespace := &corev1.Namespace{}
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: request.Namespace}, namespace); err != nil {
-		return reconcile.Result{}, err
-	}
-
-	if val, ok := namespace.Labels[common.ArgoCDManagedNamespaceLabel]; !ok || val != argocd.Namespace {
-		if namespace.Labels == nil {
-			namespace.Labels = make(map[string]string)
-		}
-		namespace.Labels[common.ArgoCDManagedNamespaceLabel] = argocd.Namespace
-		if err = r.client.Update(context.TODO(), namespace); err != nil {
-			return reconcile.Result{}, err
-		}
-	}
-
 	if argocd.GetDeletionTimestamp() != nil {
 		if argocd.IsDeletionFinalizerPresent() {
 			if err := r.deleteClusterResources(argocd); err != nil {
 				return reconcile.Result{}, fmt.Errorf("failed to delete ClusterResources: %w", err)
+			}
+
+			if err := r.removeManagedByLabelFromNamespace(argocd.Namespace); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to remove label from namespace[%v], error: %w", argocd.Namespace, err)
 			}
 
 			if err := r.removeDeletionFinalizer(argocd); err != nil {
@@ -126,6 +116,21 @@ func (r *ReconcileArgoCD) Reconcile(request reconcile.Request) (reconcile.Result
 
 	if !argocd.IsDeletionFinalizerPresent() {
 		if err := r.addDeletionFinalizer(argocd); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	namespace := &corev1.Namespace{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: request.Namespace}, namespace); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if val, ok := namespace.Labels[common.ArgoCDManagedByLabel]; !ok || val != argocd.Namespace {
+		if namespace.Labels == nil {
+			namespace.Labels = make(map[string]string)
+		}
+		namespace.Labels[common.ArgoCDManagedByLabel] = argocd.Namespace
+		if err = r.client.Update(context.TODO(), namespace); err != nil {
 			return reconcile.Result{}, err
 		}
 	}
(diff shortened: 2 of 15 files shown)


pkg/controller/argocd/role.go
```diff
@@ -55,19 +57,19 @@ func newClusterRole(name string, rules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD)
 
 // reconcileRoles will ensure that all ArgoCD Service Accounts are configured.
 func (r *ReconcileArgoCD) reconcileRoles(cr *argoprojv1a1.ArgoCD) (role *v1.Role, err error) {
```
wtam2018, line 59: This function probably should not return `role`.

pkg/controller/argocd/role.go
```diff
@@ -83,30 +85,54 @@ func (r *ReconcileArgoCD) reconcileRoles(cr *argoprojv1a1.ArgoCD) (role *v1.Role
...
 
-	if name == dexServer && isDexDisabled() {
-		// Delete any existing Role created for Dex
-		return role, r.client.Delete(context.TODO(), role)
+		if name == dexServer && isDexDisabled() {
+			// Delete any existing Role created for Dex
+			if err := r.client.Delete(context.TODO(), role); err != nil {
+				return nil, err
+			}
+			roles = append(roles, role)
```
wtam2018, line 126: Doing an `append` here does not look right.

pkg/controller/argocd/secret.go
```diff
@@ -423,12 +440,22 @@ func (r *ReconcileArgoCD) reconcileClusterPermissionsSecret(cr *argoprojv1a1.Arg
...
 			} else {
-				return nil
+				ns := strings.Split(string(s.Data["namespaces"]), ",")
+				for _, n := range namespaces {
+					if !containsString(ns, strings.TrimSpace(n)) {
+						ns = append(ns, strings.TrimSpace(n))
+					}
+					sort.Strings(ns)
+					s.Data["namespaces"] = []byte(strings.Join(ns, ","))
+					return r.client.Update(context.TODO(), &s)
```
wtam2018, line 457: I don't think we want to return yet.  Also, we should only need to call `Update` once outside of the for-loop.

wtam2018 (changes requested):
I think you forgot about writing unit tests.

pkg/controller/argocd/role.go
shubhamagarwal19: These roles are needed when we reconcile `RoleBindings`.

pkg/controller/argocd/secret.go
shubhamagarwal19: Technically, there would always be only one instance of the Secret for the default cluster server, but if there are multiple instances (which is highly unlikely) we would update the first available instance. 
Do you think we should be trying to update all the secrets if there are multiple instances?

pkg/controller/argocd/custommapper.go
```diff
@@ -96,3 +97,30 @@ func (r *ReconcileArgoCD) tlsSecretMapper(o handler.MapObject) []reconcile.Reque
 
 	return result
 }
+
+func (r *ReconcileArgoCD) namespaceResourceMapper(o handler.MapObject) []reconcile.Request {
+	var result = []reconcile.Request{}
+
+	labels := o.Meta.GetLabels()
+	if v, ok := labels[common.ArgoCDManagedNamespaceLabel]; ok {
+		argocds := &argoprojv1alpha1.ArgoCDList{}
```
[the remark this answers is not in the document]
shubhamagarwal19: According to the design it is was decided to use the <source namespace> as the value for the label to identify the ArgoCD instance. 
But it does seem like a better idea to use an identifier `namespace/name`. That way i wouldn't need to list ArgoCD instances, i can directly query for the particular instance. @jannfis @sbose78 thoughts?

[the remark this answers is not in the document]
jannfis: That sounds tempting, but we need to take care not to hit the 63 characters limit for labels here. Names of namespaces can become pretty long already (they max out at 63 chars as well). So if someone has a long namespace name, this may not work.

pkg/controller/argocd/argocd_controller.go
```diff
@@ -94,6 +96,21 @@ func (r *ReconcileArgoCD) Reconcile(request reconcile.Request) (reconcile.Result
 		return reconcile.Result{}, err
 	}
 
+	namespace := &corev1.Namespace{}
```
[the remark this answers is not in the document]
jannfis: I think it's up to the user to remove the label, because they put it there initially?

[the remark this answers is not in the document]
jannfis: Err. Oops. OK, I see now :smile: We put the label on the namespace where Argo CD is installed to. I agree that upon uninstallation of the Operand, that one label should get removed as well. Maybe in the finalizer logic for `ArgoCD` resource?

pkg/controller/argocd/custommapper.go
```diff
@@ -96,3 +97,30 @@ func (r *ReconcileArgoCD) tlsSecretMapper(o handler.MapObject) []reconcile.Reque
 
 	return result
 }
+
+func (r *ReconcileArgoCD) namespaceResourceMapper(o handler.MapObject) []reconcile.Request {
```
jannfis, line 101: Can you please put some comments on this function, describing what it does and why?

pkg/controller/argocd/role.go
jannfis: I am confused about this method as well. It basically never returns any role, if I interpret the code correctly.

if `r.reconcileRole()` returns `err != nil`, `role` would be `nil` as well, right? So basically, all it ever returns is `nil` for `role` return value. :thinking:

pkg/controller/argocd/role.go
```diff
@@ -83,30 +85,54 @@ func (r *ReconcileArgoCD) reconcileRoles(cr *argoprojv1a1.ArgoCD) (role *v1.Role
 }
 
 // reconcileRole
-func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD) (*v1.Role, error) {
-	role := newRole(name, policyRules, cr)
-	if err := applyReconcilerHook(cr, role, ""); err != nil {
+func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD) ([]*v1.Role, error) {
```
jannfis, line 88: I think we really need some documentation within this method. Why is any role created, or not. This is a critical path function. Please help others who need to modify this in the future to understand _why_ something is done the way it is done. Don't assume any context knowledge.

pkg/controller/argocd/rolebinding.go
```diff
@@ -83,68 +83,79 @@ func (r *ReconcileArgoCD) reconcileRoleBindings(cr *argoprojv1a1.ArgoCD) error {
 }
 
 func (r *ReconcileArgoCD) reconcileRoleBinding(name string, rules []v1.PolicyRule, cr *argoprojv1a1.ArgoCD) error {
```
jannfis, line 85: Same here as with `reconcileRole()`. Please document it. Help others understanding it.

pkg/controller/argocd/secret.go
```diff
@@ -423,12 +440,22 @@ func (r *ReconcileArgoCD) reconcileClusterPermissionsSecret(cr *argoprojv1a1.Arg
...
 	for _, s := range clusterSecrets.Items {
 		// check if cluster secret with default server address exists
-		// do nothing if exists.
+		// update the list of namespaces if value differs.
 		if string(s.Data["server"]) == common.ArgoCDDefaultServer {
 			if clusterConfigInstance {
-				r.client.Delete(context.TODO(), &s)
+				if err := r.client.Delete(context.TODO(), &s); err != nil {
+					return err
+				}
```
jannfis, line 448: I was just wondering whether we want to only delete `namespaces` key from the secret here, instead of completely deleting the secret. User may have created secret to change other properties, such as cluster shard index. But let's make this another PR.

pkg/controller/argocd/secret.go
jannfis: We do assume (and support) only one ArgoCD instance per namespace, so I think it's ok.

jannfis (changes requested):
Thanks @shubhamagarwal19 

I'm addressing some minor issues below. PTAL.

pkg/controller/argocd/role.go
shubhamagarwal19: `reconcileRole` returns a list of roles for all namespaces which is managed by a single ArgoCD instance. So for example if `foo and bar` is managed by an instance of ArgoCD, then it would return a list of roles for `application-controller/argocd-server ... etc` for each namespace. 
This list of roles is then used by `reconcileRolebinding` to create roleBindings for each role. 

Although i see this as an issue, this would create issues while trying to remove the rolebindings say for `DEX` if it gets disabled.

pkg/controller/argocd/secret.go
shubhamagarwal19: Could there be any other configuration in the secret that could affect the functioning of a `ClusterConfig ArgoCD instance` ?

pkg/controller/argocd/custommapper.go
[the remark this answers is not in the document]
sbose78: That one extra API Call should be fine :) the lesser the user does, the more accurate he will be.

pkg/common/keys.go
```diff
@@ -200,4 +200,7 @@ const (
 
 	// ArgoCDSecretTypeLabel is needed for cluster secrets
 	ArgoCDSecretTypeLabel = "argocd.argoproj.io/secret-type"
+
+	// ArgoCDManagedNamespaceLabel is needed to identify namespace managed by an instance on ArgoCD
+	ArgoCDManagedNamespaceLabel = "argocd.argoproj.io/managed-by"
```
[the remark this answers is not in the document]
shubhamagarwal19: done! thanks

pkg/controller/argocd/argocd_controller.go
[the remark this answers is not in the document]
shubhamagarwal19: Thanks, @wtam2018 @jannfis!! As suggested, added the code to delete the label from the namespace in the `finalizer logic` !

pkg/controller/argocd/role.go
shubhamagarwal19: removed it!! Thanks, I had added this for something that I had in my mind, but that wouldn't work!!

pkg/controller/argocd/custommapper.go
shubhamagarwal19: added the comments! Thanks

pkg/controller/argocd/rolebinding.go
shubhamagarwal19: added comments!

pkg/controller/argocd/role.go
shubhamagarwal19: We actually don't need the list of `roles` when we are calling the `reconcileRole` explicitly and just need to make sure that the role was created without any errors!! 
But when we call `reconcileRole` from within the `reconcileRoleBindings`, we need the list of roles that were created which would eventually be used to create the `RoleBindings`!

This would make more sense when we refactor the ArgoCD Operator code, which wouldn't reconcile the `Role/RoleBinding/ServiceAccount` individually, rather, it would call a function something like `reconcileArgoCDPermissions` and within the function, it would reconcile all the resources in the order `ServiceAccount -> Roles -> RoleBindings`

pkg/controller/argocd/role.go
shubhamagarwal19: Added comments!

shubhamagarwal19:
> I think you forgot about writing unit tests.

Done!! added tests!

PR #307, opened 2021-05-03
feat: Install keycloak for openshift using OpenShift templates

iam-veeramalla:
Signed-off-by: iam-veeramalla <abhishek.veeramalla@gmail.com>

**What type of PR is this?**
> /kind enhancement

**What does this PR do / why we need it**:
This PR adds `sso provider` in CR.   The reconciler installs Keycloack/RHSSO using OpenShift Templates.   This PR does not configure.  This PR also does not delete installed Keycloak.

**Have you updated the necessary documentation?**
Added

**Which issue(s) this PR fixes**:

Fixes #?
This PR address installing keycloak for openshift of #312

**How to test changes / Special notes to the reviewer**:
Deploy the example argocd CR
```
apiVersion: argoproj.io/v1alpha1
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: route
spec:
  sso:
    provider: keycloak
  server:
    route:
      enabled: true
```

or add the below section to your existing argocd CR Spec
```
  sso:
    provider: keycloak
```

* Get the SSO hostname or url. It can only be accessed using https://<hostname>

```
oc get route | grep sso
NAME                    HOST/PORT                                                                                  PATH   SERVICES                PORT    TERMINATION            WILDCARD
sso                     sso-default.apps.ci-ln-0ckf2vb-d5d6b.origin-ci-int-aws.dev.rhcloud.com                            sso                     <all>   reencrypt              None
```

* Get User and Password from the SSO pod name
```
oc exec sso-1-<> -- "env" | grep SSO_ADMIN_USERNAME
oc exec sso-1-<> -- "env" | grep SSO_ADMIN_PASSWORD
```

PR #349, opened 2021-06-25
Add Keycloak Support for Disconnected/Air-gapped OpenShift Cluster

iam-veeramalla:
**What type of PR is this?**
> /kind enhancement

This PR adds support for using Keycloak as SSO for ArgoCD in Air-gapped OpenShift clusters.

This is a follow-up PR for  #316 and #307 

NA

This is a candidate build, changes can be tested in the downstream using RedHat internal build system


pkg/controller/argocd/keycloak.go
```diff
@@ -228,18 +242,6 @@ func getKeycloakDeploymentConfigTemplate(cr *argoprojv1a1.ArgoCD) *appsv1.Deploy
 				},
 			},
 			Triggers: appsv1.DeploymentTriggerPolicies{
-				appsv1.DeploymentTriggerPolicy{
```
iam-veeramalla, line 231: Keycloak deployment config **was** configured with "Image Change" and "Config Change" triggers. 

Deployment Config when configured with both "Image Change" and "Config Change", "Image Config" takes the precedence.

Keycloak/RHSSO image streams are not available OOTB on Disconnected OpenShift clusters because of which the replication controller does not come up. It also does not make much sense to have an "Image Change" trigger in this case as we dont want to create a new Keycloak pod for every Image stream tag update.

https://docs.openshift.com/container-platform/3.5/dev_guide/deployments/basic_deployment_operations.html#triggers

We can also try using a Deployment instead of Deployment Config in **future** but we have to give over some advantages of Deployment config like 
https://developers.redhat.com/blog/2019/09/20/using-red-hat-openshift-image-streams-with-kubernetes-deployments#benefits_of_image_streams

docs/usage/keycloak.md
```diff
@@ -77,6 +77,46 @@ kubectl -n argocd exec keycloak-1-2sjcl -- "env" | grep SSO_ADMIN_PASSWORD
...
+In a [disconnected](https://access.redhat.com/documentation/en-us/red_hat_openshift_container_storage/4.7/html/planning_your_deployment/disconnected-environment_rhocs) cluster, Keycloak communicates with OpenShift Oauth Server through proxy. Below are some additional steps that needs to followed to get Keycloak integrated with OpenShift Oauth Login.
+
+### Login to the Keycloak Pod
+
+```bash
+oc exec -it dc/keycloak -n argocd -- /bin/bash
+```
+
+### Run JBoss Cli command
+
```
wtam2018, line 91: Where do I get this script?

pkg/controller/argocd/keycloak.go
```diff
@@ -101,7 +117,7 @@ func getKeycloakContainer() corev1.Container {
 			{Name: "SSO_SERVICE_USERNAME", Value: "${SSO_SERVICE_USERNAME}"},
 			{Name: "SSO_SERVICE_PASSWORD", Value: "${SSO_SERVICE_PASSWORD}"},
 		},
-		Image:           "",
+		Image:           getKeycloakContainerImage(),
```
wtam2018, line 120: Please inherit proxy env variables (`proxyEnvVars()`) from the operator (see `reconcileDexDeployment()` for example)

pkg/controller/argocd/keycloak.go
```diff
@@ -540,7 +542,7 @@ func createRealmConfig(cfg *keycloakConfig) ([]byte, error) {
 				DisplayName: "Login with OpenShift",
 				ProviderID:  "openshift-v4",
 				Config: map[string]string{
-					"baseUrl":      fmt.Sprintf("https://%s", "kubernetes.default.svc"),
+					"baseUrl":      fmt.Sprintf("https://%s", "kubernetes.default.svc.cluster.local"),
```
wtam2018, line 545: Why not just "https://kubernetes.default.svc.cluster.local"?

wtam2018 (changes requested):

pkg/controller/argocd/keycloak.go
iam-veeramalla: Hi William, Adding proxyEnvVars() does not work with Keycloak. We need to add the Proxy settings instead to standalone-openshift.xml 

I have documented the steps on how to achieve this. 
https://github.com/argoproj-labs/argocd-operator/blob/c25532c31d4efc0a2258ffa005aecf14f3103981/docs/usage/keycloak.md#additional-steps-for-disconnected-openshift-clusters

pkg/controller/argocd/keycloak.go
```diff
@@ -46,8 +47,23 @@ var (
...
+// getKeycloakContainerImage will return the container image for the Keycloak.
+//
+// There are two possible options for configuring the image, and this is the
+// order of preference.
+//
+// 1. From the Environment, this looks for the `ARGOCD_KEYCLOAK_IMAGE` field and uses
+// that.
+// 2. the default is configured in common.ArgoCDKeycloakVersion and
+// common.ArgoCDKeycloakImageName.
+func getKeycloakContainerImage() string {
```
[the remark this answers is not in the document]
iam-veeramalla: In case of Keycloak, I think we should not be providing an option for `ImageName` in CR. RHSSO is running a very old version of Keycloak, If we provide this option in CR, we are opening it up for Users to provide an Image of their choice which might not work as latest versions of Keycloak has changed Drastically.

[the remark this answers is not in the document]
wtam2018: We do this for all other images.  The age of the version is not a factor to justify going against the pattern.   You can say the same to dex or any other images configurable in CR can open up for users to use a bad/incompatible images.   Making a variant for keycloak implementation means more difficult to maintain the code.

pkg/controller/argocd/keycloak.go
wtam2018: We inherit proxy env variables in all workloads.  The question is should we.   I understand that jboss add another layer to control policy to go out to proxy or not.   In case, there are other apps/processes running in the keycloak pod but on the jboss stack.  Having the proxy env var can make sure that the cluster proxy policy can be applied.

iam-veeramalla: RHSSO containers are dependent on "standalone-openshift.xml" for proxy mappings. 

Unfortunately the presence of "proxy env vars" somehow does not allow RHSSO container to read the entries in the "standalone-openshift.xml".  I have tested using both "proxy env vars" and "standalone-openshift.xml" but that does not work.

pkg/controller/argocd/keycloak.go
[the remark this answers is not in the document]
iam-veeramalla: acknowledge

pkg/controller/argocd/keycloak.go
iam-veeramalla: acknowledge

pkg/controller/argocd/keycloak.go
wtam2018: If the presence of proxy env vars causes the request to OpenShift IdP not to go through proxy, it means that the proxy env vars DO have effect and the NO_PROXY env var causes the request (with the mapped/proxy address) not to go through the proxy.  That means the mapped/proxy address is found on NO_PROXY list.   It also tells me that if NO_PROXY were correctly configured (so that the request of the unmapped/IdP address would go through proxy), this mapping procedures (https://github.com/argoproj-labs/argocd-operator/blob/c25532c31d4efc0a2258ffa005aecf14f3103981/docs/usage/keycloak.md#additional-steps-for-disconnected-openshift-clusters) would not be needed.

wtam2018: Requiring to manually perform the JBoss proxy mapping can be a problem for lot of customers.   Remember, you have to redo it when the pod is restarted.  :-(

iam-veeramalla: 1. By setting both proxy env_vars and proxy-mappings in the standalone-openshift.xml, I can login to the Keycloak Pod and run the curl command to OAuth Server route and get the response back. I used the verbose option to confirm that the request is using "HTTPS_Proxy" and not "No_Proxy". Which means the HTTPS_Proxy env_var is well configured and NO_Proxy is not used.

2. Irrespective of misconfiguration of proxy env_vars, Configuring "Standalone-openshift.xml" is a must thing as per the Keycloak documentation and Keycloak folks.

I am equally confused on why adding proxy env_vars is causing an issue in this case but I am sure we cannot get rid of the manual steps to add the proxy_mappings in the "standalone-openshift.yaml"

I have the Image ready, we can manually add the proxy env_vars and test it.

iam-veeramalla: > Requiring to manually perform the JBoss proxy mapping can be a problem for lot of customers. Remember, you have to redo it when the pod is restarted. :-(

Totally agree on this , Keycloak team says this is the only way till they release v2.

wtam2018 (approved):
LGTM
thanks @iam-veeramalla

PR #353, opened 2021-07-06
feat: Remove `admin` permissions for ArgoCD instance when namespace label is removed

shubhamagarwal19:
**What type of PR is this?**
/kind enhancement

This PR provides the ability to remove the permissions to manage a namespace from an argocd instance when the Label `argocd.argoproj.io/managed-by` is removed. 

Fixes #?

- Deploy the latest Operator.
- Create an ArgoCD instance.
- Create a new namespace with the label `argocd.argoproj.io/managed-by: <argocd-instance-namespace>`
- Make sure the ArgoCD instance has the permissions to manage the namespace by deploying an application or verifying if the correct roles/role_bindings were created.
- Manually remove the label `argocd.argoproj.io/managed-by: <argocd-instance-namespace>` from the namespace.
- Make sure the created roles/rolebindings were deleted. 
- Commit some change to the github repo and trigger the sync. It should fail.

### DIFF
--- a/pkg/controller/argocd/util.go
+++ b/pkg/controller/argocd/util.go
@@ -1075,7 +1075,11 @@ func namespaceFilterPredicate() predicate.Predicate {
 			// This checks if the old meta had the label, if it did, delete the RBACs for the namespace
 			// which were created when the label was added to the namespace.
 			if ns, ok := e.MetaOld.GetLabels()[common.ArgoCDManagedByLabel]; ok && ns != "" {
-				if err := deleteRBACsForNamespace(ns, e.MetaOld.GetName()); err != nil {
+				k8sClient, err := initK8sClient()
+				if err != nil {
+					return false
+				}
+				if err := deleteRBACsForNamespace(ns, e.MetaOld.GetName(), k8sClient); err != nil {
 					log.Error(err, fmt.Sprintf("failed to delete RBACs for namespace: %s", e.MetaOld.GetName()))
 				} else {
 					log.Info(fmt.Sprintf("Successfully removed the RBACs for namespace: %s", e.MetaOld.GetName()))
@@ -1087,19 +1091,8 @@ func namespaceFilterPredicate() predicate.Predicate {
 }
 
 // deleteRBACsForNamespace deletes the RBACs when the label from the namespace is removed.
-func deleteRBACsForNamespace(ownerNS, sourceNS string) error {
+func deleteRBACsForNamespace(ownerNS, sourceNS string, k8sClient kubernetes.Interface) error {
 	log.Info(fmt.Sprintf("Removing the RBACs created for the namespace: %s", sourceNS))
-	cfg, err := config.GetConfig()
-	if err != nil {
-		log.Error(err, "unable to get k8s config")
-		return err
-	}
-
-	k8sClient, err := kubernetes.NewForConfig(cfg)
-	if err != nil {
-		log.Error(err, "unable to create k8s client")
-		return err
-	}
 
 	// List all the roles created for ArgoCD using the label selector
 	labelSelector := metav1.LabelSelector{MatchLabels: map[string]string{common.ArgoCDKeyPartOf: common.ArgoCDAppName}}
@@ -1164,4 +1157,20 @@ func deleteRBACsForNamespace(ownerNS, sourceNS string) error {
 		}
 	}
 	return nil
+}
+
+func initK8sClient() (*kubernetes.Clientset, error) {
+	cfg, err := config.GetConfig()
+	if err != nil {
+		log.Error(err, "unable to get k8s config")
+		return nil, err
+	}
+
+	k8sClient, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		log.Error(err, "unable to create k8s client")
+		return nil, err
+	}
+
+	return k8sClient, nil
 }
--- a/pkg/controller/argocd/util.go
+++ b/pkg/controller/argocd/util.go
@@ -1091,13 +1091,13 @@ func deleteRBACsForNamespace(ownerNS, sourceNS string) error {
 	log.Info(fmt.Sprintf("Removing the RBACs created for the namespace: %s", sourceNS))
 	cfg, err := config.GetConfig()
 	if err != nil {
-		log.Error(err, fmt.Sprintf("unable to get k8s config"))
+		log.Error(err, "unable to get k8s config")
 		return err
 	}
 
 	k8sClient, err := kubernetes.NewForConfig(cfg)
 	if err != nil {
-		log.Error(err, fmt.Sprintf("unable to create k8s client"))
+		log.Error(err, "unable to create k8s client")
 		return err
 	}
 
@@ -1140,6 +1140,9 @@ func deleteRBACsForNamespace(ownerNS, sourceNS string) error {
 		return err
 	}
 	for _, secret := range secrets.Items {
+		if string(secret.Data["server"]) != common.ArgoCDDefaultServer {
+			continue
+		}
 		if namespaces, ok := secret.Data["namespaces"]; ok {
 			namespaceList := strings.Split(string(namespaces), ",")
 			var result []string
--- a/pkg/controller/argocd/util.go
+++ b/pkg/controller/argocd/util.go
@@ -19,9 +19,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"k8s.io/apimachinery/pkg/types"
 	"os"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"text/template"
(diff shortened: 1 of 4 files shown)


pkg/controller/argocd/util.go
```diff
@@ -1067,7 +1070,82 @@ func namespaceFilterPredicate() predicate.Predicate {
...
 	}
 }
+
+func deleteRBACsForNamespace(ownerNS, sourceNS string) error {
+	log.Info(fmt.Sprintf("Removing the RBACs created for the namespace: %s", sourceNS))
+	cfg, err := rest.InClusterConfig()
+	if err != nil {
+		log.Error(err, fmt.Sprintf("unable to get k8s config"))
+		return err
+	}
```
jannfis, line 1090: Will this work when the Operator is run locally? We had that issue before I think.

jannfis (commented):
This code is on a very critical path and has _nil_ explanation of what it's doing. Please put out in comments what it is doing, and why it is doing that.

pkg/controller/argocd/util.go
shubhamagarwal19: Thanks Jann!! Completely forgot about that. Fixed it!! Also added comments

pkg/controller/argocd/util.go
```diff
@@ -1064,10 +1067,98 @@ func containsString(arr []string, s string) bool {
...
+	for _, roleBinding := range roleBindings.Items {
+		err = k8sClient.RbacV1().RoleBindings(sourceNS).Delete(context.TODO(), roleBinding.Name, metav1.DeleteOptions{})
+		if err != nil {
+			log.Error(err, fmt.Sprintf("failed to delete role binding for namespace: %s", sourceNS))
+		}
+	}
+
+	// Get the cluster secret used for configuring ArgoCD
+	labelSelector = metav1.LabelSelector{MatchLabels: map[string]string{common.ArgoCDSecretTypeLabel: "cluster"}}
+	secrets, err := k8sClient.CoreV1().Secrets(ownerNS).List(context.TODO(), metav1.ListOptions{LabelSelector: labels.Set(labelSelector.MatchLabels).String()})
```
jannfis, line 1137: I think it's better to get the secret by name. User might have configured remote clusters (which end up having a `cluster` secret-type label as well). Now think of the situation where a remote cluster has a similar namespace to the one we are going to remove from the in-cluster secret -> oops.

pkg/controller/argocd/util.go
```diff
@@ -1064,10 +1067,98 @@ func containsString(arr []string, s string) bool {
...
+	// List all the roles created for ArgoCD using the label selector
+	labelSelector := metav1.LabelSelector{MatchLabels: map[string]string{common.ArgoCDKeyPartOf: common.ArgoCDAppName}}
+	roles, err := k8sClient.RbacV1().Roles(sourceNS).List(context.TODO(), metav1.ListOptions{LabelSelector: labels.Set(labelSelector.MatchLabels).String()})
+	if err != nil {
+		log.Error(err, fmt.Sprintf("failed to list roles for namespace: %s", sourceNS))
+		return err
+	}
+
+	// Delete all the retrieved roles
+	for _, role := range roles.Items {
```
jannfis, line 1113: We do set `ownerReferences` to the roles and rolebindings we create, right? I think it would make sense to check whether we are the owner, before deleting the role. I can imagine a scenario where an operator or even a user adds the part-of label with a value of `argocd` manually, and then this role or rolebinding would get removed by accident.

pkg/controller/argocd/util.go
```diff
@@ -1064,10 +1067,98 @@ func containsString(arr []string, s string) bool {
...
+
+	// List all the roles bindings created for ArgoCD using the label selector
+	roleBindings, err := k8sClient.RbacV1().RoleBindings(sourceNS).List(context.TODO(), metav1.ListOptions{LabelSelector: labels.Set(labelSelector.MatchLabels).String()})
+	if err != nil {
+		log.Error(err, fmt.Sprintf("failed to list role bindings for namespace: %s", sourceNS))
+		return err
+	}
+
+	// Delete all the retrieved role bindings
+	for _, roleBinding := range roleBindings.Items {
```
jannfis, line 1128: Same as above for roles.

jannfis (changes requested):
Thank you @shubhamagarwal19. LGTM so far, only have a few concerns. Please see below.

pkg/controller/argocd/util.go
```diff
@@ -1064,10 +1067,98 @@ func containsString(arr []string, s string) bool {
...
+		return err
+	}
+	for _, secret := range secrets.Items {
+		if namespaces, ok := secret.Data["namespaces"]; ok {
+			namespaceList := strings.Split(string(namespaces), ",")
+			var result []string
+
+			for _, n := range namespaceList {
+				// remove the namespace from the list of namespaces
+				if strings.TrimSpace(n) == sourceNS {
```
wtam2018, line 1149: If this is the only namespace on the list and we remove it, we end up with an empty namespaces slice which means all namespaces.  Will the (non-default) argocd instance in this namespace have problem to create/sync applications in the same namespace (because it tries to access all namespaces)?

pkg/controller/argocd/util.go
```diff
@@ -1064,10 +1067,98 @@ func containsString(arr []string, s string) bool {
...
+			}
+			// Update the secret with the updated list of namespaces
+			if _, err = k8sClient.CoreV1().Secrets(ownerNS).Update(context.TODO(), &secret, metav1.UpdateOptions{}); err != nil {
+				log.Error(err, fmt.Sprintf("failed to update cluster permission secret for namespace: %s", ownerNS))
+				return err
+			}
+		}
+	}
+	return nil
+}
```
wtam2018, line 1164: How about adding some unit tests?

pkg/controller/argocd/util.go
shubhamagarwal19: Maybe we can do the cluster address check ? if it is `https://kubernetes.default.svc` or not ? 
We can't do a name check because there can be a case where user has already created the secret with the configuration using the manual workaround we provided earlier. So we can't be sure of the name ? 
Also, we are doing a similar check while creating the secret as well, so i think cluster address check should be fine?

pkg/controller/argocd/util.go
shubhamagarwal19: The implementation ensures that the secret always contains the namespace where argocd instance is deployed, i.e if ArgoCD was installed in `foo` namespace, the secret would always have `foo` namespace in the list of namespaces until the ArgoCD instance is marked as a `CLUSTER_CONFIG_INSTANCE`, then it would remove the list of namespaces from the secret.

shubhamagarwal19: so even if a user tries to remove the label or update the label to a different value, reconciler will make sure that the label is added again.

pkg/controller/argocd/util.go
shubhamagarwal19: Yep, even i had thought of using the `managed-by` label with the value as `argocd-instance-name`, but we don't have the ArgoCD instance details. So i went with `part-of` label instead. 
Let me check if i can somehow retireve the ArgoCD instance name!

pkg/controller/argocd/util.go
shubhamagarwal19: Yeah, i tried writing the UTs for it, but it is calling multiple functions like `config.GetConfig()` and `kubernetes.NewForConfig()`, which i am not sure how to mock. 
I also checked other pieces of code which is calling the same function, that doesn't have any UTs either which i could reference!
So, still exploring on this!

pkg/controller/argocd/util.go
jannfis: I think it'd be sufficient for now to check for the `ownerReferences` being an `ArgoCD` resource.

pkg/controller/argocd/util.go
jannfis: Yes, checking for secret being for `https://kubernetes.default.svc` (or rather, the appropriate constant) would be sufficient.

shubhamagarwal19: done!

pkg/controller/argocd/util.go
shubhamagarwal19: As discussed! we can't have a cross-namespace owner reference when the owner itself is namespace scoped. 
So, the resources created by the operator don't have the owner references which could be evaluated. 

Will add a story to track this for the future!

jannfis: Thanks, and you're correct with your finding about cross-namespace ownerReferences. The risk is for edge cases only, so a future implementation is fine for me.

jannfis (approved):
Thanks, LGTM now.

Please add unit tests as @wtam2018 suggested before merging.

pkg/controller/argocd/util.go
shubhamagarwal19: Added the unit test cases !!

pkg/controller/argocd/util.go
```diff
@@ -1064,10 +1067,110 @@ func containsString(arr []string, s string) bool {
...
+
+func initK8sClient() (*kubernetes.Clientset, error) {
```
jannfis, line 1162: This seems like a dupe of https://github.com/argoproj-labs/argocd-operator/blob/98a64dadb9b1687302d8de81cfe68fb06af2404e/pkg/reconciler/openshift/openshift.go#L346

It makes sense to move above method to a dedicated package and make it public.

PR #360, opened 2021-07-10
feat: Add ability to apply additional labels to generated routes

### LINKED ISSUE #359: Enable the addition of labels to the automatically created routes
We use router shards and need to add a "shard={shardname}" label to each of our routes in order for them to be exposed properly. We can currently add annotations, but it would be useful to be able to add labels as well.

tylerauerbeck:
**What type of PR is this?**
/kind enhancement

This PR enables the ability to add additional labels to a generated route. This can be useful in a number of scenarios including router sharding, etc.

Fixes #359 

This can be tested by building the operator from the PR branch and deploying an ArgoCD instance with the following CR:

```yaml
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: route
spec:
  grafana:
    enabled: true
    route:
      enabled: true
  prometheus:
    enabled: true
    route:
      enabled: true
  server:
    route:
      enabled: true
      labels:
        shard: my-shard
```

You can see from the above CR that different labels can be applied to each route (server, prometheus, grafana) individually and that applying a label to one does not apply it to them all. The PR also makes sure that the default labels that are applied across all ArgoCD objects are kept in place.

### DIFF
--- a/docs/reference/argocd.md
+++ b/docs/reference/argocd.md
@@ -290,6 +290,7 @@ Name | Default | Description
 --- | --- | ---
 Annotations | [Empty] | The map of annotations to use for the Ingress resource.
 Enabled | `false` | Toggle creation of an Ingress resource.
+Labels | [Empty] | The map of labels to add to the Route.
 Path | `/` | Path to use for Ingress resources.
 TLS | [Empty] | TLS configuration for the Ingress.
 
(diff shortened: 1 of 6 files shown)


docs/reference/argocd.md
```diff
@@ -290,6 +290,7 @@ Name | Default | Description
 --- | --- | ---
 Annotations | [Empty] | The map of annotations to use for the Ingress resource.
 Enabled | `false` | Toggle creation of an Ingress resource.
+Labels | [Empty] | The map of labels to add to the Route.
```
wtam2018, line 293: We should move it down to `Route Options`.

wtam2018:
@tylerauerbeck thanks for the PR.  Do you mind to add a unit test?

tylerauerbeck:
@wtam2018 Good catch. I'll get the docs updated and add a unit test for this.

tylerauerbeck:
@wtam2018 This should be in good shape now.

wtam2018 (approved):
LGTM
Thanks!

PR #364, opened 2021-07-12
fix: Allow reconciliation of resource overrides

jannfis:

**What type of PR is this?**

/kind bug

Perform reconciliation of deployments and statefulsets when resource requests or limits have changed.

Fixes #?

* Create ArgoCD CR
* Set resource specifications on `controller`, `repo` or `server` components
* Before this fix: Neither application controller, repo server or server objects will get reconciled (keep old resource spec)
* After this fix: All three workloads should get updated resource spec from ArgoCD CR

### DIFF
--- a/pkg/controller/argocd/deployment.go
+++ b/pkg/controller/argocd/deployment.go
@@ -409,6 +409,11 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoprojv1a1.ArgoCD) error 
 			changed = true
 		}
 
+		if !reflect.DeepEqual(deploy.Spec.Template.Spec.Containers[0].Resources, existing.Spec.Template.Spec.Containers[0].Resources) {
+			existing.Spec.Template.Spec.Containers[0].Resources = deploy.Spec.Template.Spec.Containers[0].Resources
+			changed = true
+		}
+
 		if changed {
 			return r.client.Update(context.TODO(), existing)
 		}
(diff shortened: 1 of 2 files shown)


wtam2018:
We also need to fix keycloak workloads.  See keycloak.go  (We should probably move some of the deployment stuff to deployment.go).  It looks like resource request/limits are not read from CR.

iam-veeramalla:
> We also need to fix keycloak workloads. See keycloak.go (We should probably move some of the deployment stuff to deployment.go). It looks like resource request/limits are not read from CR.

@wtam2018 Should we consider this change when we move out of OpenShift Templates and use the kubernetes manifests for Keycloak Installation ? I believe that is the correct time to make this fix.

P.S:
I have seen Keycloak does not come up with with anything less than the current request/limits

wtam2018:
I don't think we can wait.   Users may need to tweak these settings if sso enabled.

jannfis:
> We also need to fix keycloak workloads

Yes, I agree. But this should come with another PR imho.

jannfis:
> I have seen Keycloak does not come up with with anything less than the current request/limits

Valid concern, since Keycloak is a Java application which probably sets minimum JVM heap size as well and most likely is pretty hungry for RAM. Anyway, I agree with William that we should let users override the resource limits. I can think of users wanting to tweak it for CPU being available at scheduling time.

wtam2018 (approved):
LGTM
thanks!

PR #366, opened 2021-07-13
fix: Allow ArgoCD CR to override keycloak resource request/limit defaults

iam-veeramalla:
**What type of PR is this?**
> /kind bug

**What does this PR do / why we need it**:
Allow ArgoCD CR to override keycloak resource request/limit defaults

**How to test changes / Special notes to the reviewer**:
1. Run/Install the operator
2. Create ArgoCD instance with Keycloak (.spec.sso.provider = keycloak)
3. Update the SSO resource request/limit
4. Verify the changes

PR #367, opened 2021-07-14
fix: Keycloak fails to update OIDC config in argocd-cm intermittently

iam-veeramalla:
> /kind bug

#366  introduced a new bug which does not allow operator to update the keycloak related OIDC configuration intermittently. 

Reason for failure:
A user can choose to override Keycloak container resource requests/limits from Argo CD CR (Please refer #366 ). When a user does not provide Keycloak container resource requests/limits in Argo CD CR, operator overrides the default values and updates the Keycloak deployment config with new resource requirements.

#366 fails to get the new version(modified) deployment config to update the OIDC configuration. 

NA

1. Run the operator either locally or using operator hub.
2. Patch the Argo CD CR to enable Login with Keycloak.
```
spec:
   sso:
     provider: keycloak
```
3. Wait for the operator to reconcile keycloak changes. Keycloak may take a couple of minutes to start.
4. Intermittently you will see operator cannot update argocd-cm with OIDC config. which mean intermittently you will see  that "Login with Keycloak" option is not enabled in the previous build, Which is fixed.
5. on the Argo CD login page, you should see "Login Via Keycloak"

### DIFF
--- a/pkg/controller/argocd/keycloak.go
+++ b/pkg/controller/argocd/keycloak.go
@@ -673,7 +673,7 @@ func (r *ReconcileArgoCD) updateArgoCDConfiguration(cr *argoprojv1a1.ArgoCD, kRo
 	}
 
 	// Update ArgoCD instance for OIDC Config with Keycloakrealm URL
-	o, _ := yaml.Marshal(oidcConfig{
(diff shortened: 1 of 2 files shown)


wtam2018 (approved):
LGTM

PR #368, opened 2021-07-15
fix: delete existing RBACs when namespace label is updated

shubhamagarwal19:
**What type of PR is this?**
/kind bug

This PR fixes the issue when namespace label `argocd.argoproj.io/managed-by` is updated to a different value i.e from namespace `foo` -> `bar`. 
The issue being, RBACs that were created for "foo" does not get removed when the namespace label is updated, while it creates the RBACs for the new namespace (bar). 

Fixes #?
https://issues.redhat.com/browse/GITOPS-1180

- Deploy the latest Operator.
- deploy two argocd instances in the namespace `foo1` and `foo2`.
- create a namesapce `bar` with the label `argocd-argoproj.io/managed-by: foo1`.
- Make sure argocd instance in `foo1` can manage the resources in `bar` namespace.
- update the namespace label value from `foo1` -> `foo2`
- Make sure RBACs for `foo2` are created and RBACs for `foo1` gets deleted.

### DIFF
--- a/pkg/controller/argocd/util.go
+++ b/pkg/controller/argocd/util.go
@@ -1097,9 +1097,23 @@ func containsString(arr []string, s string) bool {
 func namespaceFilterPredicate() predicate.Predicate {
 	return predicate.Funcs{
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			// If ArgoCDManagedByLabel exists, return true.
-			// Event is then handled by the reconciler.
-			if _, ok := e.MetaNew.GetLabels()[common.ArgoCDManagedByLabel]; ok {
+			// This checks if ArgoCDManagedByLabel exists in newMeta, if exists then -
+			// 1. Check if oldMeta had the label or not? if no, return true
+			// 2. if yes, check if the old and new values are different, if yes,
+			// first deleteRBACs for the old value & return true.
+			// Event is then handled by the reconciler, which would create appropriate RBACs.
+			if valNew, ok := e.MetaNew.GetLabels()[common.ArgoCDManagedByLabel]; ok {
+				if valOld, ok := e.MetaOld.GetLabels()[common.ArgoCDManagedByLabel]; ok && valOld != valNew {
+					k8sClient, err := initK8sClient()
+					if err != nil {
+						return false
+					}
+					if err := deleteRBACsForNamespace(valOld, e.MetaOld.GetName(), k8sClient); err != nil {
+						log.Error(err, fmt.Sprintf("failed to delete RBACs for namespace: %s", e.MetaOld.GetName()))
+					} else {
+						log.Info(fmt.Sprintf("Successfully removed the RBACs for namespace: %s", e.MetaOld.GetName()))
+					}
+				}
 				return true
 			}
 			// This checks if the old meta had the label, if it did, delete the RBACs for the namespace


pkg/controller/argocd/util.go
```diff
@@ -1097,10 +1097,28 @@ func containsString(arr []string, s string) bool {
...
-			// Event is then handled by the reconciler.
-			if _, ok := e.MetaNew.GetLabels()[common.ArgoCDManagedByLabel]; ok {
-				return true
+			// This checks if ArgoCDManagedByLabel exists in newMeta, if exists then -
+			// 1. Check if oldMeta had the label or not? if no, return true
+			// 2. if yes, check if old and new value are the same, if yes, return true.
+			// 3. If old and new value are not the same, first deleteRBACs for old val, return true.
+			// Event is then handled by the reconciler, which would create appropriate RBACs.
+			if valNew, ok := e.MetaNew.GetLabels()[common.ArgoCDManagedByLabel]; ok {
+				if valOld, ok := e.MetaOld.GetLabels()[common.ArgoCDManagedByLabel]; !ok || valOld == "" {
```
iam-veeramalla, line 1106: Can this if , if-else and else block be modified into Just one if block ?

if valOld != valNew {
      // If old and new value are not the same, first deleteRBACs for old val, return true.
     // Event is then handled by the reconciler, which would create appropriate RBACs.
}

shubhamagarwal19: Thanks @iam-veeramalla, fixed it!!

jannfis:
Just a question: Creation of the new RBAC roles is handled somewhere down the road after deleting the old ones in case of update, right?

shubhamagarwal19:
> Just a question: Creation of the new RBAC roles is handled somewhere down the road after deleting the old ones in case of update, right?

Yep! after it's done deleting the RBACs for the older namespace, it would create them for the updated namespace value.

pkg/controller/argocd/util.go
```diff
@@ -1097,9 +1097,23 @@ func containsString(arr []string, s string) bool {
...
 	return predicate.Funcs{
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			// If ArgoCDManagedByLabel exists, return true.
-			// Event is then handled by the reconciler.
-			if _, ok := e.MetaNew.GetLabels()[common.ArgoCDManagedByLabel]; ok {
+			// This checks if ArgoCDManagedByLabel exists in newMeta, if exists then -
+			// 1. Check if oldMeta had the label or not? if no, return true
+			// 2. if yes, check if the old and new values are different, if yes,
+			// first deleteRBACs for the old value & return true.
+			// Event is then handled by the reconciler, which would create appropriate RBACs.
```
jannfis, line 1104: :+1: thanks for documenting!

jannfis (approved):
LGTM

pkg/controller/argocd/util.go
```diff
@@ -1097,9 +1097,23 @@ func containsString(arr []string, s string) bool {
...
+			if valNew, ok := e.MetaNew.GetLabels()[common.ArgoCDManagedByLabel]; ok {
+				if valOld, ok := e.MetaOld.GetLabels()[common.ArgoCDManagedByLabel]; ok && valOld != valNew {
+					k8sClient, err := initK8sClient()
+					if err != nil {
+						return false
+					}
+					if err := deleteRBACsForNamespace(valOld, e.MetaOld.GetName(), k8sClient); err != nil {
```
wtam2018, line 1111: Deleting RBAC inside of Predicate Func does not sound right.  Predicate Func is used to filter event.

PR #376, opened 2021-07-22
fix: ArgoCD server resources limit/request changes in ArgoCD CR is not reconcilied

iam-veeramalla:

**What type of PR is this?**
> /kind bug

Users can modify argocd server container resource limit/request in the ArgoCD CR.  The operator should apply/reconcile the changes to the cluster.   However, reconciler is not checking for changes in ArgoCD CR against the running state due to a bug.
This PR adds a check in the reconcile function to check for changes in ArgoCD CR on argocd server container resource limit/request against the cluster and set the "changed" flag accordingly.

NA

1. Create ArgoCD CR in the foo namespace.
2. Record the resource limit/request on the running argocd server pod 
2. Update the ArgoCD Server container resource request/limits to different values
3. Verify that the resource request/limits changes have been applied to the cluster not updated on the container.


wtam2018:
LGTM
thanks @iam-veeramalla !

wtam2018 (approved):

PR #378, opened 2021-07-22
fix: ApplicationSet controller Deployment reconcilliation update logic misses volume field

### LINKED ISSUE #377: ApplicationSet controller Deployment reconcilliation update logic misses volume field
**Describe the bug**
The ApplicationSet controller Deployment reconciliation logic works well on Create, but on update it fails to include the Volumes, causing an error in the logs like:
`
Deployment.apps \"applicationset-controller\" is invalid: [spec.template.spec.containers[0].volumeMounts[0].name: Not found: \"ssh-known-hosts\", spec.template.spec.containers[0].volumeMounts[1].name: Not found: \"tls-certs\", spec.template.spec.containers[0].volumeMounts[2].name: Not found: \"gpg-keys\", spec.template.spec.containers[0].volumeMounts[3].name: Not found: \"gpg-keyring\", spec.template.spec.containers[0].volumeMounts[4].name: Not found: \"argocd-repo-server-tls\"]"
`
In this case, the podspec containers volumemounts is correctly updated in the Deployment, but the `volumes` field of the Deployment is not, causing the above error.

This was reproduced in the upgrade scenario, where the old version of the ApplicationSet deployment would be missing these fields, causing the update logic to be invoked, and incorrectly doing a partial update of the Deployment (causing the above error).

jgwest:
**What type of PR is this?**
/kind bug

See [parent issue](https://github.com/argoproj-labs/argocd-operator/issues/377) for details.

This PR:
- Adds the fields to update logic which were previously missing.   Missing fields are volumes, serviceaccount name, and labels.
- Improves the existing Deployment create unit test, to handle these fields
- Adds a new test for updating an existing Deployment, which would catch the issue described in the original PR

Fixes #377

You can reproduce by:
1. Creating an ArgoCD with the appSet field defined
2. Waiting for it to reconcile the applicationset-controller Deploment
3. Ediit the applicationset-controller Deployment to remove the volume and volumeMount field. 
4. You should find the operator correctly reconciles it back to the expected value.

### DIFF
--- a/pkg/controller/argocd/applicationset.go
+++ b/pkg/controller/argocd/applicationset.go
@@ -156,10 +156,23 @@ func (r *ReconcileArgoCD) reconcileApplicationSetDeployment(cr *argoprojv1a1.Arg
 
 	if existing := newDeploymentWithSuffix("applicationset-controller", "controller", cr); argoutil.IsObjectFound(r.client, cr.Namespace, existing.Name, existing) {
 
-		// If the Deployment already exists, make sure the containers are up-to-date
-		actualContainers := existing.Spec.Template.Spec.Containers[0]
-		if !reflect.DeepEqual(actualContainers, podSpec.Containers) {
+		existingSpec := existing.Spec.Template.Spec
+
+		deploymentsDifferent := !reflect.DeepEqual(existingSpec.Containers[0], podSpec.Containers) ||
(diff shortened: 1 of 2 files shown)


wtam2018 (approved):
LGTM
thanks @jgwest !

PR #399, opened 2021-08-22
🐴 rename uid_entrypoint.sh -> entrypoint.sh 🐴

### LINKED ISSUE #398: uid_entrypoint.sh renamed to entrypoint.sh, breaks using v2.1.0
**Describe the bug**
```
argocd-repo-server-6dc987d959-q75xr   0/1     CreateContainerError
```
- because of hardcoded uid_entrypoint.sh which was renamed here

https://github.com/argoproj/argo-cd/commit/4ae1d876fd6c66ae665db996a0d2e9c43fdac7e6
https://github.com/argoproj/argo-cd/blob/master/entrypoint.sh

**To Reproduce**
install version 2.1.0 argocd

eformat:
**What type of PR is this?**
/kind bug

change to entrypoint.sh

Fixes #398 


tylerauerbeck:
@eformat This is a good catch. But doesn't this then break this for anything < 2.1.0? Maybe (at least for now and until the operator specifically doesn't support < 2.1.0) the logic should be something more along the lines of: 

```go
if cr.Spec.Version < 2.1.0 { 
    cmd = append(cmd,"uid_entrypoint.sh") 
} else {
  cmd = append(cmd,"entrypoint.sh")
}
```

Could also wrap that in a helper function to keep things a bit more clean like a number of other attributes are determined.

@wtam2018 @iam-veeramalla Not sure what the current process is. Do certain releases of the operator only support X number of versions? I think the above makes sense until < 2.1.0 aren't supported. Interested to see others thoughts here.

eformat:
@tylerauerbeck - yup, very good point. just built this tested locally OK. i couldn't immediately see a Version check upstream that would normalize versions (something like github.com/hashicorp/go-version), happy to change to use one though ..

jannfis:
Just as a side-note: With Argo CD v2.1.1, we do have a symlink `uid_entrypoint.sh -> entrypoint.sh` in the container to restore backwards compat. So, the only version the Operator would currently not support is v2.1.0, and I think that'd be fine.

eformat:
ah yup .. thanks @jannfis - seems fair.  will close this. for reference see https://github.com/argoproj/argo-cd/pull/7047 and
https://github.com/argoproj/argo-cd/blob/613db278a799c6fd7a822b251f22f68b3ab7533d/Dockerfile#L66

pearj:
Should this actually be closed and not merged?

in https://github.com/argoproj/argo-cd/pull/7047 it says that the symlink will be removed next release

does that v3 or v2.1.2 or v2.2? I’m guessing it’s the next major release?

PR #400, opened 2021-08-23
Add a hook to enable serving cert annotation to argocd server

chetan-rns:
**What type of PR is this?**
> /kind enhancement

This PR adds an OpenShift hook to add cert annotation to the argocd server service. OpenShift looks for this annotation and creates a secret in the same namespace with certs  

Fixes #?

### DIFF
--- a/pkg/controller/argocd/service.go
+++ b/pkg/controller/argocd/service.go
@@ -17,6 +17,7 @@ package argocd
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	argoprojv1a1 "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
 	"github.com/argoproj-labs/argocd-operator/pkg/common"
@@ -393,7 +394,17 @@ func (r *ReconcileArgoCD) reconcileServerMetricsService(cr *argoprojv1a1.ArgoCD)
 // reconcileServerService will ensure that the Service is present for the Argo CD server component.
 func (r *ReconcileArgoCD) reconcileServerService(cr *argoprojv1a1.ArgoCD) error {
 	svc := newServiceWithSuffix("server", "server", cr)
-	if argoutil.IsObjectFound(r.client, cr.Namespace, svc.Name, svc) {
+
+	if err := applyReconcilerHook(cr, svc, ""); err != nil {
+		return err
+	}
+
+	existingSvc := &corev1.Service{}
+	if argoutil.IsObjectFound(r.client, cr.Namespace, svc.Name, existingSvc) {
+		if !reflect.DeepEqual(existingSvc.Annotations, svc.Annotations) {
+			existingSvc.Annotations = mergeAnnotations(svc.Annotations, existingSvc.Annotations)
+			return r.client.Update(context.TODO(), existingSvc)
+		}
 		return nil // Service found, do nothing
 	}
 
@@ -467,4 +478,15 @@ func (r *ReconcileArgoCD) reconcileServices(cr *argoprojv1a1.ArgoCD) error {
 		return err
 	}
 	return nil
+}
+
+func mergeAnnotations(from, to map[string]string) map[string]string {
+	merged := map[string]string{}
+	for k, v := range to {
+		merged[k] = v
+	}
+	for k, v := range from {
+		merged[k] = v
+	}
+	return merged
 }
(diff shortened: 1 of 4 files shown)


iam-veeramalla:
Hi @chetan-rns , Can you please provide some details regarding the motivation behind adding this change ?

pkg/controller/argocd/service.go
```diff
@@ -393,7 +394,17 @@ func (r *ReconcileArgoCD) reconcileServerMetricsService(cr *argoprojv1a1.ArgoCD)
...
 	svc := newServiceWithSuffix("server", "server", cr)
-	if argoutil.IsObjectFound(r.client, cr.Namespace, svc.Name, svc) {
+
+	if err := applyReconcilerHook(cr, svc, ""); err != nil {
+		return err
+	}
+
+	existingSvc := &corev1.Service{}
+	if argoutil.IsObjectFound(r.client, cr.Namespace, svc.Name, existingSvc) {
+		if !reflect.DeepEqual(existingSvc.Annotations, svc.Annotations) {
```
shubhamagarwal19, line 404: If a user adds an annotation to the service, the operator would update the service on each reconcile since `svc.Annotations` would never be equal to `existingSvc.Annotations`. 
Can we perform a check which makes sure that the required annotations are present in the existing service object? and if not present, update it? what do you think?

chetan-rns: Makes sense. Will add a check

chetan-rns:
@iam-veeramalla @shubhamagarwal19 I've updated the description with more details

pkg/controller/argocd/service.go
chetan-rns: Added a check to update only if there are new annotations

pkg/controller/argocd/service.go
```diff
@@ -468,3 +478,19 @@ func (r *ReconcileArgoCD) reconcileServices(cr *argoprojv1a1.ArgoCD) error {
 	}
 	return nil
 }
+
+func mergeAnnotations(existing, svc *metav1.ObjectMeta) bool {
+	changed := false
+	if existing == nil {
```
shubhamagarwal19, line 484: shouldn't we be checking for `existing.Annotations == nil` rather than the objectMeta? I doubt ObjectMeta would be nil if there's an existing service object!

chetan-rns: Updated the condition!

shubhamagarwal19 (approved):
LGTM

iam-veeramalla:
@chetan-rns After the cluster generates a secret for your service, your PodSpec should mount it, and the Pod will run after it becomes available.

Did you verify that the `argocd-server` has started with the certificate that the `ingress-operator` has created for you ? I dont see the changes for that in this PR.

pkg/controller/argocd/service.go
```diff
@@ -393,7 +393,17 @@ func (r *ReconcileArgoCD) reconcileServerMetricsService(cr *argoprojv1a1.ArgoCD)
 // reconcileServerService will ensure that the Service is present for the Argo CD server component.
 func (r *ReconcileArgoCD) reconcileServerService(cr *argoprojv1a1.ArgoCD) error {
 	svc := newServiceWithSuffix("server", "server", cr)
-	if argoutil.IsObjectFound(r.client, cr.Namespace, svc.Name, svc) {
+
+	if err := applyReconcilerHook(cr, svc, ""); err != nil {
+		return err
+	}
+
+	existingSvc := &corev1.Service{}
```
iam-veeramalla, line 401: @chetan-rns After the cluster generates a secret for your service, your PodSpec should mount it, and the Pod will run after it becomes available.

Did you verify that the argocd-server has started with the certificate that the ingress-operator has created for you ? I dont see the changes for that in this PR.

iam-veeramalla (changes requested):

pkg/controller/argocd/service.go
chetan-rns: @iam-veeramalla We're trying to mount the generated secrets in a pod created by the downstream operator. Hence there are no mounts in the argocd-server cc @shubhamagarwal19

iam-veeramalla: @chetan-rns @shubhamagarwal19  any reason why you dont want to implement that here ?

chetan-rns: The pod where we want to mount the secrets is created by the downstream operator.

iam-veeramalla: as discussed we would need to mount the secret to enable TLS configuration for Argo CD Server

PR #402, opened 2021-08-25

shubhamagarwal19:
**What type of PR is this?**
> /kind bug

When the ArgoCD instance is deleted from the source namespace, operator would remove the `managed-by` label from all the namespaces that were managed by this particular argocd instance. 
The operator currently already has the logic to remove the RBACs from the namespaces from which the label is being removed, so it takes care of removing the RBACs as soon as the labels is removed. 
This ensures there are no redundant roles/role_bindings/labels are present for a namespace. 

> https://issues.redhat.com/browse/GITOPS-1228

TODO

### DIFF
--- a/pkg/controller/argocd/util.go
+++ b/pkg/controller/argocd/util.go
@@ -794,13 +794,10 @@ func (r *ReconcileArgoCD) deleteClusterResources(cr *argoprojv1a1.ArgoCD) error 
 
 func (r *ReconcileArgoCD) removeManagedByLabelFromNamespace(namespace string) error {
 	nsList := &corev1.NamespaceList{}
-	selector := labels.NewSelector()
-	requirement, err := labels.NewRequirement(common.ArgoCDManagedByLabel, selection.Equals, []string{namespace})
-	if err != nil {
-		return fmt.Errorf("failed to create a requirement for %w", err)
+	listOption := client.MatchingLabels{
+		common.ArgoCDManagedByLabel: namespace,
 	}
-	selector.Add(*requirement)
-	if err := r.client.List(context.TODO(), nsList, client.MatchingLabelsSelector{Selector: selector}); err != nil {
+	if err := r.client.List(context.TODO(), nsList, listOption); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/argocd/util.go
```diff
@@ -793,16 +793,29 @@ func (r *ReconcileArgoCD) deleteClusterResources(cr *argoprojv1a1.ArgoCD) error
...
-	if ns.Labels == nil {
-		return nil
+	for _, n := range nsList.Items {
+		ns := &corev1.Namespace{}
+		if err := r.client.Get(context.TODO(), types.NamespacedName{Name: n.Name}, ns); err != nil {
+			return err
+		}
+
+		if ns.Labels == nil {
+			return nil
```
jannfis, line 811: Shouldn't that be a `continue` instead?

pkg/controller/argocd/util.go
```diff
@@ -793,16 +793,29 @@ func (r *ReconcileArgoCD) deleteClusterResources(cr *argoprojv1a1.ArgoCD) error
...
+		}
+		delete(ns.Labels, common.ArgoCDManagedByLabel)
+		if err := r.client.Update(context.TODO(), ns); err != nil {
+			log.Error(err, fmt.Sprintf("failed to remove label from namespace [%s]", ns.Name))
+		}
```
jannfis, line 816: Do we want to test if that label still exists on the namespace, before running `delete()` and `Update()`?

jannfis (changes requested):
Thanks @shubhamagarwal19. One minor thing I see: I know we use a label selector on `managed-by` label for listing the namespaces, and _should_ expect the labels to be set on the namespaces returned. But meanwhile, maybe they have been changed externally so we may want to do some proper error handling.

Also, I think this change should not go to `master` but to `rhos-1.2` branch, due to the Operator SDK upgrade ongoing.

@wtam2018 Can we make the CI jobs run on the rhos-1.2 branch (and future release branches) as well?

pkg/controller/argocd/util.go
shubhamagarwal19: yes, thanks for pointing out!!

pkg/controller/argocd/util.go
shubhamagarwal19: done!!

shubhamagarwal19:
Closing this PR, created a new one for `rhos-v1.2`
https://github.com/argoproj-labs/argocd-operator/pull/404

PR #404, opened 2021-08-25

shubhamagarwal19:
**What type of PR is this?**
> /kind bug

When the ArgoCD instance is deleted from the source namespace, operator would remove the managed-by label from all the namespaces that were managed by this particular argocd instance.
The operator currently already has the logic to remove the RBACs from the namespaces from which the label is being removed, so it takes care of removing the RBACs as soon as the labels is removed.
This ensures there are no redundant roles/role_bindings/labels are present for a namespace.

> https://issues.redhat.com/browse/GITOPS-1228

BUILD for testing - create a catalog source using the following yaml:
```
kind: CatalogSource
metadata:
  name: argocd-operator-catalog
spec:
  sourceType: grpc
  image: quay.io/shuagarw/argocd-operator-registry:v1.2.1-1228
  displayName: Argo CD Operators
  publisher: Red Hat Developer
```

Steps to test - 
- Deploy the above provided build and install the argocd-operator
- Create an argocd instance in namespace `foo`. Wait for the instance to be in an available state.
- Create 2 namespaces `bar1` and `bar2`.
- Label both namespaces with the managed-by label - `argocd.argoproj.io/managed-by: foo`
- Make sure the correct Roles/RoleBindings were created for the namespaces. 
- Delete the ArgoCD instance in foo namespace. 
- Make sure the created Role/RoleBindings and Labels were removed from the namespaces.

### DIFF
--- a/pkg/controller/argocd/util.go
+++ b/pkg/controller/argocd/util.go
@@ -798,7 +798,7 @@ func (r *ReconcileArgoCD) deleteClusterResources(cr *argoprojv1a1.ArgoCD) error 
 	return nil
 }
 
-func (r *ReconcileArgoCD) removeManagedByLabelFromNamespace(namespace string) error {
+func (r *ReconcileArgoCD) removeManagedByLabelFromNamespaces(namespace string) error {
 	nsList := &corev1.NamespaceList{}
 	listOption := client.MatchingLabels{
 		common.ArgoCDManagedByLabel: namespace,
--- a/pkg/controller/argocd/util.go
+++ b/pkg/controller/argocd/util.go
@@ -799,16 +799,32 @@ func (r *ReconcileArgoCD) deleteClusterResources(cr *argoprojv1a1.ArgoCD) error 
 }
 
 func (r *ReconcileArgoCD) removeManagedByLabelFromNamespace(namespace string) error {
-	ns := &corev1.Namespace{}
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: namespace}, ns); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/controller/argocd/util.go
```diff
@@ -799,16 +799,33 @@ func (r *ReconcileArgoCD) deleteClusterResources(cr *argoprojv1a1.ArgoCD) error
 }
 
 func (r *ReconcileArgoCD) removeManagedByLabelFromNamespace(namespace string) error {
```
iam-veeramalla, line 801: should we rename this function to `removeManagedByLabelFromNamespaces` ?

pkg/controller/argocd/util_test.go
```diff
@@ -477,3 +478,54 @@ func TestDeleteRBACsForNamespace(t *testing.T) {
...
+	err = r.client.Create(context.TODO(), ns3)
+	assert.NilError(t, err)
+
+	err = r.removeManagedByLabelFromNamespace(a.Namespace)
+	assert.NilError(t, err)
+
+	nsList := &v1.NamespaceList{}
+	err = r.client.List(context.TODO(), nsList)
+	assert.NilError(t, err)
+	for _, n := range nsList.Items {
```
iam-veeramalla, line 522: 👍

pkg/controller/argocd/util.go
shubhamagarwal19: done! update the name of the function

iam-veeramalla:
Thanks @shubhamagarwal19 for the PR !!
/LGTM
/approve

PR #410, opened 2021-08-30
feat: add NodeSelector and Tolerations in argo crd

### LINKED ISSUE #342: Tolerations & nodeselectors in argocd CRD.
**Describe the bug**
I can't seem to find a way to add toleration's & node-selectors to the pods that the operator deploys, I can add toleration's to the operator pod but not sure how to do it for the other pods.

**To Reproduce**
Look at operator docs & API spec.
Write an argocd resource.

**Expected behavior**
A option in the argocd resource definition for toleration's & node-selector

**Screenshots**
N\A

**Additional context**
For now I patch the deployments with the toleration's & node-selectors after the pods have been deployed, they seem to get moved to the right nodes.
Since I'm using a cloud based Kubernetes cluster I'm unable to set the podtolerationrestriction admission controller.

saumeya:

**What type of PR is this?**
 /kind enhancement

This PR adds the field **nodePlacement** to the the argocd crd, which enables the user to add _nodeSelector_ and _tolerations_ on the workloads.

Fixes #342 

To test these changes run, `make install` `make run` and in the argocd cr yaml  _spec_, add the following -
spec:
  nodePlacement: 
    nodeSelector: 
      key1: value1
    tolerations: 
    - key: key1
      operator: Equal
      value: value1
      effect: NoSchedule
    - key: key1
      operator: Equal
      value: value1
      effect: NoExecute


iam-veeramalla:
@saumeya  please sign the CLA

controllers/argocd/deployment.go
```diff
@@ -1102,3 +1130,15 @@ func isDexDisabled() bool {
 	}
 	return false
 }
+
+// to update nodeSelector and tolerations in reconciler
+func updateNodePlacement(existing *appsv1.Deployment, deploy *appsv1.Deployment, changed *bool) {
+	if !reflect.DeepEqual(existing.Spec.Template.Spec.NodeSelector, deploy.Spec.Template.Spec.NodeSelector) {
+		existing.Spec.Template.Spec.NodeSelector = deploy.Spec.Template.Spec.NodeSelector
```
iam-veeramalla, line 1137: I remember we want to setup Node Affinity rather than Node selector as per @shubhamagarwal19 suggestion. So that we can implement `requiredDuringSchedulingIgnoredDuringExecution` and `preferredDuringSchedulingIgnoredDuringExecution`. Is there any change in the plan ?

https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity

saumeya: I did go through the node-affinity concept, it is different from nodeSelector, and the issue https://github.com/argoproj-labs/argocd-operator/issues/342 specifically asks for nodeSelector and Tolerations

iam-veeramalla:
Please add documentation as well :)

controllers/argocd/deployment.go
```diff
@@ -626,12 +635,17 @@ func (r *ReconcileArgoCD) reconcileRedisDeployment(cr *argoprojv1a1.ArgoCD) erro
 // reconcileRedisHAProxyDeployment will ensure the Deployment resource is present for the Redis HA Proxy component.
 func (r *ReconcileArgoCD) reconcileRedisHAProxyDeployment(cr *argoprojv1a1.ArgoCD) error {
 	deploy := newDeploymentWithSuffix("redis-ha-haproxy", "redis", cr)
+	nodePlacementDeploy := newDeploymentWithSuffix("redis-ha-haproxy", "redis", cr)
```
shubhamagarwal19, line 638: I think if there is no existing `redisHAProxyDeployment`, the new one is created without the `NodeSelector and Tolerations` since we are using the `deploy` object rather than `nodePlacementDeploy`.
I would recommend using only a single object i.e `deploy` and creating a new object `existing` to fetch the existing deployment. This would maintain consistency with other resources.

controllers/argocd/statefulset.go
```diff
@@ -85,6 +85,11 @@ func newStatefulSetWithSuffix(suffix string, component string, cr *argoprojv1a1.
 
 func (r *ReconcileArgoCD) reconcileRedisStatefulSet(cr *argoprojv1a1.ArgoCD) error {
 	ss := newStatefulSetWithSuffix("redis-ha-server", "redis", cr)
+	nodePlacementSs := newStatefulSetWithSuffix("redis-ha-server", "redis", cr)
```
shubhamagarwal19, line 88: same! the new stateful set that is being created will not have NodeSlector and Tolerations spec.

controllers/argocd/deployment.go
```diff
@@ -377,7 +377,10 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoprojv1a1.ArgoCD) error
 	if dexDisabled {
 		log.Info("reconciling for dex, but dex is disabled")
 	}
-
+	if cr.Spec.NodePlacement != nil {
```
shubhamagarwal19, line 380: Since we are adding the `NodePlacement` spec for all deployment types, is it OK if we add it within the `newDeploymentWithName` func?
Haven't evaluated this completely though, but it would make sense IMO. That way we don't have to explicitly add this for each deployment type.

shubhamagarwal19 (commented):
Thanks for the PR @saumeya, looks great. 
Just have some minor comments -

controllers/argocd/deployment.go
saumeya: Thanks a lot!!! will look at this right now!

saumeya:
> Please add documentation as well :)

Sure

controllers/argocd/deployment.go
saumeya: Thanks a lot for this, makes the code a lot more leaner :)

shubhamagarwal19 (approved):
LGTM

controllers/argocd/deployment_test.go
```diff
@@ -397,6 +397,56 @@ func Test_proxyEnvVars(t *testing.T) {
 	}
 }
 
+func TestReconcileArgoCD_reconcileRepoDeployment_nodePlacement(t *testing.T) {
```
iam-veeramalla, line 400: I feel this function can look more generic.

controllers/argocd/deployment.go
```diff
@@ -626,19 +630,25 @@ func (r *ReconcileArgoCD) reconcileRedisDeployment(cr *argoprojv1a1.ArgoCD) erro
 // reconcileRedisHAProxyDeployment will ensure the Deployment resource is present for the Redis HA Proxy component.
 func (r *ReconcileArgoCD) reconcileRedisHAProxyDeployment(cr *argoprojv1a1.ArgoCD) error {
 	deploy := newDeploymentWithSuffix("redis-ha-haproxy", "redis", cr)
-	if argoutil.IsObjectFound(r.Client, cr.Namespace, deploy.Name, deploy) {
+
+	existing := newDeploymentWithSuffix("redis-ha-haproxy", "redis", cr)
```
iam-veeramalla, line 634: This line looks same as the above line ?

controllers/argocd/statefulset.go
```diff
@@ -505,3 +511,15 @@ func (r *ReconcileArgoCD) triggerStatefulSetRollout(sts *appsv1.StatefulSet, key
 	sts.Spec.Template.ObjectMeta.Labels[key] = nowNano()
 	return r.Client.Update(context.TODO(), sts)
 }
+
+//to update nodeSelector and tolerations in reconciler
+func updateNodePlacementStateful(existing *appsv1.StatefulSet, ss *appsv1.StatefulSet, changed *bool) {
```
iam-veeramalla, line 516: Please add a test to verify the behavior of this function

controllers/argocd/deployment.go
saumeya: yes, need _existing_ to check if any current change has happened so that the reconciler can update with the latest changes, in other deployments this comparison is in the end, so the assignment of these variables is not next to each other

controllers/argocd/deployment_test.go
saumeya: could you elaborate, do you mean not just _repoDeployment_ but to add other as well ?

controllers/argocd/statefulset.go
saumeya: sure

controllers/argocd/deployment_test.go
iam-veeramalla: I feel this function makes it look that the feature is available only for repo server. Never mind, you can keep this.

iam-veeramalla:
Verified the changes using the below Argo CD CR. 
```
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: nodeplacement-example
spec:
  nodePlacement:
    nodeSelector:
      node-role.kubernetes.io/master: ''
    tolerations:
    - key: node-role.kubernetes.io/master
      operator: Equal
      value: ''
      effect: NoSchedule
```

As expected the workloads are scheduled on 3 master nodes.

controllers/argocd/statefulset.go
```diff
@@ -505,3 +511,15 @@ func (r *ReconcileArgoCD) triggerStatefulSetRollout(sts *appsv1.StatefulSet, key
...
+	if !reflect.DeepEqual(existing.Spec.Template.Spec.NodeSelector, ss.Spec.Template.Spec.NodeSelector) {
```
iam-veeramalla, line 517: 👍

iam-veeramalla (approved):
LGTM !! Thanks @saumeya

PR #413, opened 2021-08-31
fix: Only set ownerReferences for roles & rolebindings in CR's namespace

### LINKED ISSUE #408: Failures in calls to controllerutil.SetControllerReference when reconciling roles and rolebindings
**Describe the bug**

As part of the upgrade of the operator-sdk (PR #406), the linter reported errors in existing code where where the return value of calls to controllerutil.SetControllerReference were not being checked. After checks were added, the unit tests started failing which indicates an existing bug in the code.

In order to unblock the sdk upgrade, the checks were changed to log the error and continue.  This issue is being raised to get a proper fix for the problem.

In several places in `role.go` and `rolebinding.go`, the code tries to set an ArgoCD CR as the owner reference for objects which may potentially be either be cluster-scoped or in another namespace from the CR itself.  In both cases the attempt to set the owner reference will fail.

The locations where the problem occurs have been marked with `// TODO handle this error properly` comments

jannfis:
**What type of PR is this?**

/kind bug

This prevents setting ("illegal" cross-namespace) `ownerReferences` to role and rolebindings created from the `managed-by` feature. We only set up `ownerReferences` for resources in the same namespace as the ArgoCD CR.

Since this may change the behaviour of Argo CD's drift detection in certain situations (because we do apply all labels from the ArgoCD CR to the resources we create), a second PR that fixes this behavior will be coming soon.

Fixes #408 

* Create `ArgoCD` CR in namespace `ns-one`
* Create namespace `ns-two`
* Apply `managed-by` label to `ns-two`: `kubectl label ns ns-two argocd.argoproj.io/managed-by=ns-one`
* Observe no errors in the Operator's log
* Ensure that roles created in `ns-two` have no `ownerReferences` applied to them

### DIFF
--- a/controllers/argocd/role.go
+++ b/controllers/argocd/role.go
@@ -116,10 +116,12 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
 			if name == dexServer && isDexDisabled() {
 				continue // Dex is disabled, do nothing
 			}
-			if err = controllerutil.SetControllerReference(cr, role, r.Scheme); err != nil {
-				// TODO handle this error properly
-				// return nil, fmt.Errorf("failed to set ArgoCD CR \"%s\" as owner for role \"%s\": %s", cr.Name, role.Name, err)
-				log.Error(err, fmt.Sprintf("failed to set ArgoCD CR \"%s\" as owner for role \"%s\"", cr.Name, role.Name))
+
+			// Only set ownerReferences for roles in same namespace as ArgoCD CR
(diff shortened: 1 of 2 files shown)


jopit (changes requested):
There's one more location where `controlerutil.SetControllerReference` can fail, in `rolebinding.go` at around line 208.  It should be ok to just remote the attempt to set the owner reference, since the rolebinding in question is cluster-scoped.

jannfis:
Thanks @jopit, fixed.

jopit (approved):

PR #414, opened 2021-08-31
fix: Do not copy all labels from ArgoCD CR to resources we create

jannfis:
**What type of PR is this?**

/kind bug

Currently, when we create resources, we inherit all labels from the `ArgoCD` CR to those resources. If the `ArgoCD` CR was created by Argo CD itself from Git, it will have an instance label (`app.kubernetes.io/instance` by default), indicating that the resource is managed by that particular Argo CD instance/application. However, this label will also be inherited on the resources that the Operator creates, and that are not managed through Argo CD (but through the Operator). This confuses Argo CD, as it thinks it might be managing these resources.

IMHO, we don't need to copy the labels from the `ArgoCD` CR at all, and just set some default labels on the resources we create to be able to identify it.

Also, this PR includes a slight refactor of methods used for labeling and annotating our resources.

Fixes #?

* Create a YAML [0] containing `ArgoCD` spec, and have some labels in the metadata, for example
* Apply YAML to some namespace: `kubectl apply -n some-ns -f argocd-cr.yaml`
* Wait for reconciliation
* Inspect metadata of any resources created by the Operator. In none of these, the labels `foo: bar` or `bar: foo` should be visible. Only the default labels `app.kubernetes.io/name`, `app.kubernetes.io/part-of` and `app.kubernetes.io/managed-by` should be there.
  * `kubectl get -n some-ns secret argocd-secret -o jsonpath='{.metadata.labels}' `
  * `kubectl get -n some-ns secret argocd-default-cluster -o jsonpath='{.metadata.labels}' `
  * `kubectl get -n some-ns deployment argocd-server -o jsonpath='{.metadata.labels}' `

> [0] argocd-cr.yaml

```yaml
kind: ArgoCD
metadata:
  name: argocd
  labels:
    foo: bar
    bar: foo
spec:
  route:
     enabled: true
```

### DIFF
--- a/api/v1alpha1/argocd_types.go
+++ b/api/v1alpha1/argocd_types.go
@@ -672,3 +672,13 @@ func (argocd *ArgoCD) IsDeletionFinalizerPresent() bool {
 	}
 	return false
 }
+
+// ApplicationInstanceLabelKey returns either the custom application instance
+// label key if set, or the default value.
+func (a *ArgoCD) ApplicationInstanceLabelKey() string {
+	if a.Spec.ApplicationInstanceLabelKey != "" {
+		return a.Spec.ApplicationInstanceLabelKey
+	} else {
+		return common.ArgoCDDefaultApplicationInstanceLabelKey
+	}
+}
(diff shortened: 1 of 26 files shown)


jopit (approved):
LGTM

PR #417, opened 2021-09-01

shubhamagarwal19:
Cherry-picked the fix added to rhos-v1.2 branch -> https://github.com/argoproj-labs/argocd-operator/pull/404

### DIFF
--- a/controllers/argocd/util_test.go
+++ b/controllers/argocd/util_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/argoproj-labs/argocd-operator/common"
 	"github.com/argoproj-labs/argocd-operator/controllers/argoutil"
 
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	testclient "k8s.io/client-go/kubernetes/fake"
 )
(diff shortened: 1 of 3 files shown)


controllers/argocd/util_test.go
```diff
@@ -460,3 +461,54 @@ func TestDeleteRBACsForNamespace(t *testing.T) {
...
+	r := makeTestReconciler(t)
+
+	ns := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{
+		Name: "testNamespace",
+		Labels: map[string]string{
+			common.ArgoCDManagedByLabel: a.Namespace,
+		}},
+	}
+
+	err := r.client.Create(context.TODO(), ns)
```
jannfis, line 476: Client method became public with Operator SDK upgrade, please check.

jannfis (commented):
Thanks for the "back"-port to master, @shubhamagarwal19.

Just one thing, because the build & tests due to a unit test, please see below.

controllers/argocd/util_test.go
shubhamagarwal19: ahhh thanks .. missed it for the test files!

iam-veeramalla (approved):
LGTM, Thanks @shubhamagarwal19  !!

PR #426, opened 2021-09-02
chore: Enable GitHub workflows on release branches

iam-veeramalla:
**What type of PR is this?**
> /kind enhancement

Enable Github workflows for release branches


.github/workflows/ci-build.yaml
```diff
@@ -3,17 +3,17 @@ on:
   push:
     branches:
       - 'master'
-      - 'v[0-9]+.[0-9]+'
```
jopit, line 6: Don't we still need the workflows enabled for `vx.x.x` branches?  When we release version 0.1.0 of the operator, isn't its branch going to be `v0.1.0`, not `rhos-v0.1.0`?

jopit (changes requested):

.github/workflows/ci-build.yaml
jannfis: I think we may consider naming our release branches `release-X.Y`. IMHO, we don't need release branches for vX.Y.Z, only for vX.Y (vX.Y.Z will then be a tag within that release branch, because it is not supposed to change anymore).

Also, all this could be simplified using a glob match pattern instead of regular expressions:

* `release-*`
* `rhos-*`

iam-veeramalla: sure, let me change that.

iam-veeramalla: Made changes accordingly, Please verify if it looks good now.
@jannfis @jopit

jannfis (approved):
LGTM, thanks!

Please wait for @jopit's approval before merging, tho.

jopit (approved):

PR #429, opened 2021-09-02
chore: upgrade operator-sdk to v1.11.0

jopit:

**What type of PR is this?**

> /kind enhancement

There's been a change in the operator-sdk version 1.10.1 to not generate a file (`bundle/manifests/argocd-operator_v1_serviceaccount.yaml`) when running `make bundle`, and in version 1.11.0 the presence of that file will actually cause `make bundle` to error out. This PR is to remove that file and regenerate the bundle artifacts using operator-sdk version 1.11.0

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=argocd-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.9.0+git
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.10.0+git
 LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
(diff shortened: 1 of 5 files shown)


bundle.Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=argocd-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.9.0+git
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.10.0+git
```
iam-veeramalla, line 9: why does this say v1.10 when the sdk upgraded to v1.11 ?

iam-veeramalla (commented):
@jopit

bundle.Dockerfile
jopit: Those values are generated by the operator-sdk.  I don't know why they seem to be behind, but the original `operator-sdk-v1.9.0+git` was generated by version 1.10.0 of the operator-sdk, and I've seen it with other versions of the sdk, so it's not specific to the 1.11.0 version.

iam-veeramalla (approved):
Thanks @jopit !!

Issue #431, opened 2021-09-06
Add support to restrict login to optional list of groups using Dex


iam-veeramalla:
**Is your feature request related to a problem? Please describe.**
Dex allows the admin to restrict login to optional list of groups. Which means, only the users **who are part of any of the these groups** will be able to login into Argo CD.

**Describe the solution you'd like**
Enhance the Argo CD CR(specifically the Dex field) to support this.

**Describe alternatives you've considered**
NA

**Additional context**
https://dexidp.io/docs/connectors/openshift/#dex-configuration

PR #432, opened 2021-09-06
feat: Add support to restrict login to optional list of groups using Dex

iam-veeramalla:

**What type of PR is this?**
> /kind enhancement

Dex allows the admin to restrict login to optional list of groups. Which means, only the users **who are part of any of the these groups** will be able to login into Argo CD.

#431 

Fixes #?
#431 

1. Run the operator locally using `make run`
2. Submit the below Argo CD CR. 
```
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: route
spec:
  dex:
    openShiftOAuth: true
    groups:
     - foo
  server:
    route:
     enabled: true
```
3. Verify that only users part of the `foo` group would be able to Login into Argo CD. Try login using kube:admin user and you should see the below error.
Screenshot 2021-09-06 at 9 05 40 AM


jannfis (approved):
LGTM

PR #433, opened 2021-09-06
chore: Install CRDs as part of `make run` target

iam-veeramalla:

> /kind bug

`make run` target fails with below error when the Argo CD related CRDs are not installed on the cluster.  include the `install` target as part of the `make run`.

```
09-06T13:22:17.479+0530	ERROR	setup	problem running manager	{"error": "failed to wait for argocdexport caches to sync: no matches for kind \"ArgoCDExport\" in version \"argoproj.io/v1alpha1\""}
```

NA

On a fresh cluster, run `make run` and verify that operator does not fail with the below error.

```
09-06T13:22:17.479+0530	ERROR	setup	problem running manager	{"error": "failed to wait for argocdexport caches to sync: no matches for kind \"ArgoCDExport\" in version \"argoproj.io/v1alpha1\""}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,10 +93,7 @@ test: manifests generate fmt vet envtest ## Run tests.
 build: generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 
-install-crds: ## Install CRDs
-	kubectl apply -f config/crd/bases/
-
-run: manifests generate fmt vet install-crds ## Run a controller from your host.
+run: manifests generate fmt vet install ## Run a controller from your host.
 	go run ./main.go
 
 docker-build: test ## Build docker image with the manager.
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,10 @@ test: manifests generate fmt vet envtest ## Run tests.
 build: generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 
-run: manifests generate fmt vet ## Run a controller from your host.
+install-crds: ## Install CRDs
+	kubectl apply -f config/crd/bases/
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -93,7 +93,10 @@ test: manifests generate fmt vet envtest ## Run tests.
 build: generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 
-run: manifests generate fmt vet ## Run a controller from your host.
+install-crds: ## Install CRDs
+	-kubectl create -f config/crd/bases/
```
iam-veeramalla, line 97: `-` at the start of this line ignores any errors on this command.

This is added to ignore the errors like "CRDs already installed"

jannfis: And what about legitimate errors? Such as, cannot connect to cluster or permission denied? What if the CRDs were already installed, but required updates?

I would suggest to use `apply` instead of `create`, and to not ignore errors.

jannfis (commented):
LGTM with minor comment up for discussion

Makefile
iam-veeramalla: ah, so stupid of me. Thanks @jannfis

jannfis: Oh, and there is already `install` target that does install the CRDs into the cluster. Have you considered using that?

iam-veeramalla: Thanks for that, I did not realise. So, either we can include the `install` as part of `make run`(updated the PR) or simply close the PR and users would run `make install` and then `make run` ?
Please feel free to close or approve the PR :)

wtam2018: I would not alter the behavior of `make run`.   It is probably a good idea not re-apply the manifests in every run and let the install be explicit.   Let's close it.

PR #434, opened 2021-09-06
fix: Remove trailing newline from admin.password

jannfis:
**What type of PR is this?**

/kind bug

The Operator reconciles the admin password for Argo CD from the `<instance>-cluster` secret, which stores a base64 encoded version of the password in its `admin.password` field. Users tend to encode base64 data like this: `echo foobar | base64`. This also encodes the final newline from `echo` into the base64 data, which effectively makes the password useless (users will not be able to login).

This change removes any trailing newline from the decoded value in `admin.password`, so users won't shoot themselves in the foot.

Fixes #?

* Run Operator locally
* Create new Argo CD instance
* Create new password with trailing newline: `echo foobar | base64`
* Set new password by c&p from above: `oc patch secret argocd-cluster --type merge --patch '{"data":{"admin.password":"Zm9vYmFyCg=="}}'`
* Log into Argo CD using user "admin" and password "foobar"

### DIFF
--- a/controllers/argocd/secret.go
+++ b/controllers/argocd/secret.go
@@ -299,14 +299,17 @@ func (r *ReconcileArgoCD) reconcileExistingArgoSecret(cr *argoprojv1a1.ArgoCD, s
 	changed := false
 
 	if hasArgoAdminPasswordChanged(secret, clusterSecret) {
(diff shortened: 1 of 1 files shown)


iam-veeramalla (approved):
LGTM !! Thanks @jannfis

PR #437, opened 2021-09-09
feat: Allow setting custom environment in Argo CD workloads

### LINKED ISSUE #141: RFE: Allow setting environment variables via operator
For quite a few cases the option to supply environment variables via operator would be really useful. Especially for corporate proxies, but also secrets. 
I suggest something simple in the form of:
```yaml
environment:
  - name: VAR
    value: VAL
  - name: VAR2
    configmapRef:
      name: public
      key: key
  - name: VAR3
    secretKeyRef:
      name: secret
      key: key
```
Which, while verbose, should translate well to the actual deployments. I could imagine this as global for all deployments or service specific.

### LINKED ISSUE #201: Requirement: Add proxy function to support air-gap environment.
Support these ENVS:

- http_proxy
- https_proxy
- no_proxy

to connect Internet pages(such: github) via proxy.

### LINKED ISSUE #395: Ability add env to containers
Hi, we are using our kustomize, which needs to read environment variables in the argocd-repo-server container. Can you add the ability to define variables like we did in a non-operator setup?
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: argocd-repo-server
  namespace: argocd
spec:
  replicas: 5
  template:
    spec:
      containers:
      - name: argocd-repo-server
        env:
          - name: CLUSTER
            value: prod
```

I think it should look something like this.

```
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: repo
spec:
  repo:
    env: {}
```

jannfis:
**What type of PR is this?**

/kind enhancement

This allows users to specify custom environment variables for argocd-server, argocd-repo-server and argocd-controller workloads, which allows to set some configuration options that are not yet managed by the Operator.

It follows a simple paradigm for precendence:

1. If cluster has global HTTP proxy variables set, they will be set
2. Any variable specified by the user will be added, and possibly override the HTTP proxy vars set in (1) (to allow for custom HTTP proxy settings)
3. Any option in the Operand that actually sets an environment variable (e.g. .spec.repo.execTimeout) will override the same variable set by the user - Options in the Operand will always take precedence

*Additional note:*

I started refactoring the unit test assertion framework with this change, from `gotest.tools/assert` to `github.com/stretchr/testify` ([MIT license](https://github.com/stretchr/testify/blob/master/LICENSE)), because it just provides so much more functionality and is almost a drop-in replacement. I will start global refactoring on this soon, as well.

Also, some documentation update included.

Closes #141 
Closes #201 
Closes #395

* Set `.spec.Server.Env`, `.spec.Repo.Env` or `.spec.Controller.Env` to contain one or more values
* Wait for reconciliation and rollout of workload(s)
* Inspect workload that environment has been set


wtam2018:
Is doc update missing?

jannfis:
> Is doc update missing?

Indeed, forgot to push the commit to my branch. Thanks :)

controllers/argocd/deployment.go
```diff
@@ -801,9 +801,12 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argoprojv1a1.ArgoCD) error
 		deploy.Spec.Template.Spec.ServiceAccountName = cr.Spec.Repo.ServiceAccount
 	}
 
-	repoEnv := proxyEnvVars()
+	// Global proxy env vars go first
+	repoEnv := cr.Spec.Repo.Env
+	// Environment specified in the CR take precedence over everything else
+	repoEnv = argoutil.EnvMerge(repoEnv, proxyEnvVars(), false)
```
wtam2018, line 807: We want to be able to overide the proxyEnvVars (cluster proxy env vars) with proxy env vars defined in the CR.  So, the merge func call should look like:
```
	repoEnv = argoutil.EnvMerge(proxyEnvVars(), repoEnv, true)
```

jannfis: IMHO, it does exactly that:

1. Set env to what is specified in CR
2. Try to merge results of `proxyEnvVars()` into `repoEnv`, but do not override  existing entries (`override` is `false`)

wtam2018: OK.  You are right.  Both get the same result.

wtam2018:
LGTM
Thanks @jannfis

wtam2018 (approved):

SahilMutha:
Has this fix been released yet ?

PR #442, opened 2021-09-20
fix: log error and update status when both dex and keycloak are configured

### LINKED ISSUE #312: Automatically install and configure Keycloak SSO for ArgoCD instance

### LINKED ISSUE #447: Handle the case when both Keycloak and Dex are configured in the CR.
**Describe the bug**
Argo CD Operator currently supports two SSO providers `Dex` and `Keycloak`. `Keycloak` can be configured using the `.sso.provider: keycloak` and Dex can be configured using `.spec.dex`. This leads to a great chance of misconfiguration, A user can unknowingly or intentionally add both `Keycloak` and `Dex` to the CR.

This scenario should be handled. 

**To Reproduce**
1. Run the operator locally using `make install && make run`
2. Create an Argo CD CR with Keycloak as SSO provider
```
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: route
spec:
  sso:
    provider: keycloak
    verifyTLS: false
  server:
    route:
     enabled: tru
```
3.  Now add Dex as another SSO provider
```
.......
dex:
  openShiftOAuth: true
```
**Expected behavior**
Only one SSO provider should be allowed.

**Screenshots**

**Additional context**

ciiay:
**What type of PR is this?**

/kind enhancement

To indicate SSO misconfiguration when both Dex and keycloak are configured
1. Log an error message that a SSO provider is already configured.
2. Update the Argo CD CR status to reflect the same.

Fixes #447 

1. Configure both Dex and sso of keycloak in CR and apply it
2. Observe Error log of "duplicated SSO configuration"
3. Observe "ssoConfig" field in status shows "Only can install one SSO provider, please uninstall Dex or keycloak" in updated argocd instance CR


ciiay:
Hi @iam-veeramalla is it redundant to define error log in both `func reconcileSSO` and `func reconcileDexDeployment`? It shows same error log twice in log now. I think once `func reconcileDexDeployment` is migrated to `sso.go` we can remove one of them, just want to make sure this solution is fine for now. Thanks!

ciiay:
Updated Status in CR:

1. When only one SSO is configured.
Screen Shot 2021-09-21 at 12 39 51 PM

2. When both Dex and keycloak are confirgured.
Screen Shot 2021-09-21 at 12 39 34 PM

controllers/argocd/status.go
```diff
@@ -93,6 +98,21 @@ func (r *ReconcileArgoCD) reconcileStatusDex(cr *argoprojv1a1.ArgoCD) error {
 	return nil
 }
 
+// reconcileStatusSsoConfig will ensure that the SsoConfig status is updated for the given ArgoCD.
+func (r *ReconcileArgoCD) reconcileStatusSsoConfig(cr *argoprojv1a1.ArgoCD) error {
+	status := "Unknown"
+
+	if cr.Spec.SSO != nil && !reflect.DeepEqual(cr.Spec.Dex, argoprojv1a1.ArgoCDDexSpec{}) {
+		status = "Only can install one SSO provider, please uninstall Dex or keycloak"
```
jaideepr97, line 106: Can we rephrase this to say something like "Installation of multiple SSO providers simultaneously is not permitted. Please choose a single provider"

controllers/argocd/status.go
```diff
@@ -93,6 +98,21 @@ func (r *ReconcileArgoCD) reconcileStatusDex(cr *argoprojv1a1.ArgoCD) error {
 	return nil
 }
 
+// reconcileStatusSsoConfig will ensure that the SsoConfig status is updated for the given ArgoCD.
+func (r *ReconcileArgoCD) reconcileStatusSsoConfig(cr *argoprojv1a1.ArgoCD) error {
+	status := "Unknown"
```
jaideepr97, line 103: @iam-veeramalla @ciiay 

Is my understanding correct that in case there is no conflict, then user might see `dex/keycloak` status as available/running and `ssoConfig` as unknown at the same time?
If so, that could be a little confusing to the user

controllers/argocd/sso.go
```diff
@@ -104,6 +105,12 @@ func verifyTemplateAPI() error {
 
 func (r *ReconcileArgoCD) reconcileSSO(cr *argoprojv1a1.ArgoCD) error {
 	if cr.Spec.SSO.Provider == argoprojv1a1.SSOProviderTypeKeycloak {
+		if !reflect.DeepEqual(cr.Spec.Dex, argoprojv1a1.ArgoCDDexSpec{}) {
+			err := e.New("duplicated SSO configuration")
+			log.Error(err, fmt.Sprintf("Only can install one SSO provider, please uninstall Dex for ArgoCD %s in namespace %s",
```
jaideepr97, line 110: Can we consider rephrasing this to say something like "Installation of multiple SSO providers simultaneously is not permitted. Please choose a single provider for ArgoCD %s in namespace %s"?

If my understanding is correct, if we have 2 different error messages for dex and keycloak, then during the reconcilliation loop both errors might show up in the operator logs when dex and keycloak are configured together, which to the user may look like we are asking them to uninstall both of them to fix the error even if they only need to uninstall one of them
Maybe if we kept the error message generic in both places, it could be more clear to the user to just pick one of the providers and keep that one?
Let me know your thoughts @iam-veeramalla @ciiay

controllers/argocd/deployment.go
```diff
@@ -332,6 +333,11 @@ func (r *ReconcileArgoCD) reconcileDeployments(cr *argoprojv1a1.ArgoCD) error {
 
 // reconcileDexDeployment will ensure the Deployment resource is present for the ArgoCD Dex component.
 func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoprojv1a1.ArgoCD) error {
+	if cr.Spec.SSO != nil {
+		err := e.New("duplicated SSO configuration")
+		log.Error(err, fmt.Sprintf("Only can install one SSO provider, please uninstall keycloak for ArgoCD %s in namespace %s",
```
jaideepr97, line 338: Please see https://github.com/argoproj-labs/argocd-operator/pull/442/files#r714119303

controllers/argocd/status.go
ciiay: Hi @jaideepr97 , thanks for the comment. We only have `dex` field in the `status` before this change. `ssoConfig` was meant to tell user the error of misconfiguration of multiple sso providers. I thought about the same thing, but default value for all status fields are set to `Unknown` and I chose to make it consistent. Do you think another field name like `ssoMisconfig` would be better? Or change the default value to something else to indicate no error for `ssoConfig`?

controllers/argocd/sso.go
ciiay: I agree with the concern of different error logs could mislead the user to uninstall both sso providers. We can use a generic phrase as the one in the status. 
"Installation of multiple SSO providers simultaneously is not permitted" is very good to me. For the new phrase of "Please choose a single provider ...", I personally feel it's not as clear as "please uninstall A or B" as a new user. I prefer a straightforward instruction even though your version sounds more classic. Let's get more opinions from others :)

controllers/argocd/status.go
wtam2018: I would keep the ssoConfig status consistent with other statuses.  That is, I would stick with something like: Unknown/Success/Failed.

controllers/argocd/deployment.go
```diff
@@ -16,6 +16,7 @@ package argocd
 
 import (
 	"context"
+	e "errors"
```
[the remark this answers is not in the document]
ciiay: We don't need an alias `e` as you mentioned. @iam-veeramalla and I made it an alias `e` to keep it consistent with same check in `sso.go` as this `reconcileDexDeployment` function will be migrated into `sso.go` soon. Do you prefer to keep it simple like no alias for now?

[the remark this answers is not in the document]
wtam2018: I'm fine with it either way.  I usually don't use alias unless there are conflicts.

controllers/argocd/sso.go
ciiay: Got feedback from William, will go with "Installation of multiple SSO providers is not permitted.  Please choose a single provider for Argo CD %s in namespace %s." for both checks.

api/v1alpha1/argocd_types.go
```diff
@@ -642,6 +642,13 @@ type ArgoCDStatus struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="Dex",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
 	Dex string `json:"dex,omitempty"`
 
+	// SSOConfig is an indicator of misconfiguration where multiple SSO providers are configured
```
iam-veeramalla, line 645: The comment message is confusing. Does this sound good ?
"SSOConfig defines the status of SSO configuration"

controllers/argocd/status.go
iam-veeramalla: @ciiay Did we improve this behavior ?

controllers/argocd/status.go
```diff
@@ -93,6 +98,23 @@ func (r *ReconcileArgoCD) reconcileStatusDex(cr *argoprojv1a1.ArgoCD) error {
 	return nil
 }
 
+// reconcileStatusSSOConfig will ensure that the SSOConfig status is updated for the given ArgoCD.
+func (r *ReconcileArgoCD) reconcileStatusSSOConfig(cr *argoprojv1a1.ArgoCD) error {
+	status := "Unknown"
+
+	if cr.Spec.SSO != nil && !reflect.DeepEqual(cr.Spec.Dex, argoprojv1a1.ArgoCDDexSpec{}) {
+		status = "Failed"
+	} else if (cr.Spec.SSO != nil && reflect.DeepEqual(cr.Spec.Dex, argoprojv1a1.ArgoCDDexSpec{})) || (cr.Spec.SSO == nil && !reflect.DeepEqual(cr.Spec.Dex, argoprojv1a1.ArgoCDDexSpec{})) {
```
iam-veeramalla, line 107: Please comment what this condition is actually doing.

controllers/argocd/status.go
ciiay: @iam-veeramalla Thanks for the review. I didn't get what exact you mean but this function is updated with new status values of `Unknown/Success/Failed` and the condition of each value is as following.
`// Success: Only one SSO provider is configured in CR.`
`// Failed: More than one SSO providers are configure in CR.`
`// Unknown: For some reason the SSO configuration could not be obtained.`

api/v1alpha1/argocd_types.go
ciiay: Description for other status fields all start with"xxx is a simple, high-level summary of where the Argo CD ...", I tried to keep it consistent but couldn't find a better way to put it. "SSOConfig defines the status of SSO configuration" I think this way user might be confused like what does "status of SSO configuration" mean.

@wtam2018 @jannfis Any idea of how to put the description of this new status field?

wtam2018: I think @iam-veeramalla looks good.

jaideepr97: Just out of curiosity, we have only defined the meanings of the different states (success/failed etc) of `ssoConfig` internally in the code, right?

How would an average user know about this?

wtam2018:
@ciiay Please add some unit tests.

api/v1alpha1/argocd_types.go
ciiay: @jaideepr97 The generated manifest file has the description of definition of each state, I don't know how they can check it out without look into the source code though, if this is what you mean. I'm also adding comments to the `reconcileStatusSSOConfig` function for developer's reference.

jaideepr97: Yeah, I mean the average user is not going to look at code, they'll just see `ssoConfig: success/failure` in their CR 

Maybe we should document this somewhere in our release notes then? @wtam2018 @iam-veeramalla

ciiay: I agree that as a user if I read `status.ssoConfig: Failed` it's hard to tell what exactly went wrong... It could be I put wrong value for the sso configuration, or I missed the sso configuration with no detail explained.

bundle/manifests/argoproj.io_argocds.yaml
```diff
@@ -1632,6 +1632,12 @@ spec:
                   For some reason the state of the Argo CD server component could
                   not be obtained.'
                 type: string
+              ssoConfig:
+                description: 'SSOConfig defines the status of SSO configuration. Success:
```
iam-veeramalla, line 1636: we are actually not verifying if the SSO is success or failure ... we are just validating that the configuration provided in the CR is correct.

So lets change the options from `Sucess`, `Failure` and `unknown` to `Valid`, `Invalid` and `unknown` ?

controllers/argocd/sso.go
```diff
@@ -104,6 +105,13 @@ func verifyTemplateAPI() error {
 
 func (r *ReconcileArgoCD) reconcileSSO(cr *argoprojv1a1.ArgoCD) error {
 	if cr.Spec.SSO.Provider == argoprojv1a1.SSOProviderTypeKeycloak {
+		if !reflect.DeepEqual(cr.Spec.Dex, argoprojv1a1.ArgoCDDexSpec{}) {
```
iam-veeramalla, line 108: This condition wont work :( as the structure contains non-comparable objects. We need to break down it to something like this.

if cr.Spec.Dex.OpenShiftOAuth == true || cr.Spec.Dex.Config != "" {
......
}

api/v1alpha1/argocd_types.go
wtam2018: The error message with full details goes to the log, right?  We can also generate an event.

bundle/manifests/argoproj.io_argocds.yaml
wtam2018: The status field is "sso config" not "sso" though.

controllers/argocd/status.go
ciiay: Comment added, thanks for the reminder.

controllers/argocd/status.go
ciiay: Thanks for the recommendation. This is implemented as `Unknown/Success/Failed`.

controllers/argocd/status.go
ciiay: This is updated as we discussed, thanks for the review.

controllers/argocd/deployment.go
ciiay: Thanks for the review, this is updated as you suggested.

api/v1alpha1/argocd_types.go
```diff
@@ -639,6 +639,9 @@ type ArgoCDStatus struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="Dex",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
 	Dex string `json:"dex,omitempty"`
```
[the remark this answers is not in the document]
ciiay: Thanks for the reminder. I ran `make generate`, `make manifests` and `make bundle` for this PR.

controllers/argocd/deployment.go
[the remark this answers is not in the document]
ciiay: This logic got removed from `deployment.go`.

api/v1alpha1/argocd_types.go
ciiay: Thanks for the review, let's add the explanation to the release note.

controllers/argocd/deployment_test.go
```diff
@@ -293,6 +293,18 @@ func TestReconcileArgoCD_reconcileDexDeployment_removes_dex_when_disabled(t *tes
 		deployment))
 }
 
+// Test when both Dex and keycloak are configured
+func TestReconcileArgoCD_reconcileDexDeployment_with_keycloak_configured(t *testing.T) {
+	restoreEnv(t)
+	logf.SetLogger(ZapLogger(true))
+	a := makeTestArgoCDForKeycloakWithDex()
+
+	templateAPIFound = true
```
jaideepr97, line 302: Do we need this?

config/crd/bases/argoproj.io_argocds.yaml
```diff
@@ -1629,6 +1629,8 @@ spec:
                   For some reason the state of the Argo CD server component could
                   not be obtained.'
                 type: string
+              ssoConfig:
```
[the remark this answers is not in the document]
ciiay: Thanks for the reminder, generated files included in updated PR.

controllers/argocd/deployment_test.go
ciiay: Thanks for the comment, this unit test got removed in the latest push as changes in `deployment.go` got removed.

iam-veeramalla:
resolves #447 
cc: @ciiay

iam-veeramalla:
@ciiay please take a look at the failing tests.

iam-veeramalla (approved):
PR looks good to me, I verified the changes on my test cluster in most possible ways that I can think of. 
/LGTM. Thanks @ciiay, really appreciate.

iam-veeramalla:
Good to be merged if @wtam2018 and @jaideepr97 have no objections :)

wtam2018:
Thanks @ciiay !

api/v1alpha1/argocd_types.go
ciiay: @wtam2018 Yes, there is an error log says "Installation of multiple SSO providers is not permitted. Please choose a single provider for Argo CD %s in namespace %s."

ciiay:
@wtam2018 thanks for approving and merging this PR!

PR #449, opened 2021-09-30
fix: add nodePlacement check in applicationSet reconciler

### LINKED ISSUE #450: ApplicationSets, Keycloak deployments does not respect NodeSelector and Tolerations in the Argo CD CR.
**Describe the bug**
#410  adds support to NodeSelectors and Tolerations. When NodeSelectors and Tolerations are added to the Argo CD CR as described in the #410 all the workloads gets scheduled on the expected Node. 

However, ApplicationSets and Keycloak are exceptions. Both ApplicationSets and Keycloak are not installed OOTB, A user needs to update them in the CR explicitly. When they are explicitly mentioned in the CR, they do not seem to respect the NodeSelector and Tolerations in the Argo CD CR.

**To Reproduce**
Run the operator locally using `make install && make run`
Create the Argo CD CR with spec as shown below.
```
spec:
  applicationSet: {}
  sso:
    provider: keycloak
  nodePlacement:
    nodeSelector:
      abc: abc
    tolerations:
      - effect: NoSchedule
        key: abc
        value: abc
```

**Expected behavior**
All the workloads including AppSets and Keycloak should get scheduled on the requested Node.

**Additional context**
NA

saumeya:
**What type of PR is this?**
 /kind bug

This PR adds check for nodeSelector and tolerations in applicationSet reconciler so that the pod can reschedule on correct nodes

Fixes #450 

`make install`, `make run` and then in argoCD CR spec add , verify that all pods now have nodeSelectors and tolerations
```
spec:
  applicationSet: {}
  nodePlacement:
    nodeSelector:
      abc: abc
    tolerations:
      - effect: NoSchedule
        key: abc
        value: abc
```

### DIFF
--- a/controllers/argocd/applicationset.go
+++ b/controllers/argocd/applicationset.go
@@ -179,7 +179,9 @@ func (r *ReconcileArgoCD) reconcileApplicationSetDeployment(cr *argoprojv1a1.Arg
 			existingSpec.ServiceAccountName != podSpec.ServiceAccountName ||
 			!reflect.DeepEqual(existing.Labels, deploy.Labels) ||
 			!reflect.DeepEqual(existing.Spec.Template.Labels, deploy.Spec.Template.Labels) ||
-			!reflect.DeepEqual(existing.Spec.Selector, deploy.Spec.Selector)
+			!reflect.DeepEqual(existing.Spec.Selector, deploy.Spec.Selector) ||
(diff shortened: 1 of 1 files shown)


iam-veeramalla (approved):
LGTM, Thanks @saumeya !!

PR #458, opened 2021-10-15
Allow setting replicas for argocd-repo-server and argocd-server

reginapizza:
**What type of PR is this?**

> /kind enhancement

I want to be able to control the number of replicas for the argocd-server and argocd-repo-server workloads. As these workloads are stateless, the replica count can be increased to distribute the workload better among pods. However, if I have the horizontal autoscaler enabled on argocd-server, I want that to be prioritized over the number set for replicas.

- Go to Makefile and change the version # (just to something else) and change image base tag to IMAGE_TAG_BASE ?= quay.io/{your quay or docker username}/argocd-operator
- Add the Argo CD instance: 
```
kind: ArgoCD
metadata:
  name: argocd
spec: {}
EOF                 
```
- run `oc edit argocd argocd`
- under spec.server add the field `replicas: x` where x is your desired number of replicas
- under spec.repo add the field `replicas: x` where x is your desired number of replicas
argocdCR-replicas
- save your changes
- go to OpenShift console and under Administrator View, go to Workloads > Deployments and set namespace as default. You should be able to see `argocd-server` and `argocd-repo-server` deployments with the number of replicas you had set in the CR. 
argocdCR-replicas-console


reginapizza:
This is good for review now whenever one can be provided by someone @iam-veeramalla @jannfis @wtam2018 @shubhamagarwal19 
(Thanks in advance!)

jannfis:
Thanks @reginapizza! 

I have a question about why you had chosen to make the replica setting a unique configuration, as opposed to integrating it into `repo` and `server` configuration, i.e.

```yaml
spec:
  repo:
    replicas: X
  server:
    replicas: Y
```
I think this would be a somewhat more simple approach, having the setting where the users (or at least me) would expect it.

reginapizza:
@jaideepr97 @jannfis would you please re-review when you get a chance? Thanks!

jaideepr97:
LGTM 
Thanks @reginapizza !
/lgtm

api/v1alpha1/argocd_types.go
```diff
@@ -341,6 +341,9 @@ type ArgoCDRepoSpec struct {
 	// MountSAToken describes whether you would like to have the Repo server mount the service account token
 	MountSAToken bool `json:"mountsatoken,omitempty"`
 
+	// Replicas defines the number of replicas for argocd-repo-server
+	Replicas *int32 `json:"replicas,omitempty"`
```
wtam2018, line 345: Any reasons why this is a pointer?

common/defaults.go
```diff
@@ -36,6 +36,12 @@ const (
 	// ArgoCDDexServerComponent is the name of the Dex server control plane component
 	ArgoCDDexServerComponent = "argocd-dex-server"
 
+	// ArgoCDRepoServerReplicas is the default number of replicas for argocd-repo-server
+	ArgoCDRepoServerReplicas = int32(1)
+
+	// ArgoCDServerReplicas is the default number of replicas for argocd-repo-server
```
wtam2018, line 42: PLease fix comment.  It says argocd-repo-server.

wtam2018:
There is a settings to enable auto scaling of argo cd sever.  https://argocd-operator.readthedocs.io/en/latest/reference/argocd/#server-example.
Will this server replica count defeat the auto scaling or they entirely unrelated?
Please add user doc.

wtam2018:
Please add how did you test it.

jannfis:
> Will this server replica count defeat the auto scaling or they are entirely unrelated?

I think this is a good hint. When replicas are set explicitly for `server`, we should ignore HPA settings. Also, we should  enable HPA configuration for argocd-repo-server as well, because it makes more sense there than for argocd-server.

jannfis:
> When replicas are set explicitly for server, we should ignore HPA settings

Or actually, the other way round. Before reconciling `.spec.server.replicas`, we should check whether HPA has been configured. If HPA is configured, ignore `.spec.server.replicas` and leave it to HPA to figure out how many replicas there should exist.

api/v1alpha1/argocd_types.go
reginapizza: because the typed value could also be nil. Should it not be a pointer?

wtam2018: What is the effective number of replica when the value of this int32 pionter is nil?

reginapizza: Oh I see what you mean actually, previously I thought this field was optional but I realize now that even if you get rid of it it will still come back. 

However, we have a line that says `deploy.Spec.Replicas = getArgoCDRepoServerReplicas(cr)` where `deploy.Spec.Replicas` is of type `*int32`, but `getArgoCDRepoServerReplicas` would return a value of type `int32` if we removed the pointer. (same for argocd-server)

controllers/argocd/deployment.go
```diff
@@ -34,6 +34,28 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+// getArgoCDRepoServerReplicas will return the size value for the argocd-repo-server replica count.
```
wtam2018, line 37: Since this return value is fed into `deploy.Spec.Replicas` which is a not pointer and has a default value of 1, we should not have our own default.  In other words, we should just pass the cr.Spec.Repo.Replicas value to `deploy.Spec.Replicas`.  If it is nil, pass nil to `deploy.Spec.Replicas`.

controllers/argocd/deployment.go
```diff
@@ -34,6 +34,28 @@ import (
...
+	replicas := common.ArgoCDRepoServerReplicas
+	if cr.Spec.Repo.Replicas != nil {
+		if *cr.Spec.Repo.Replicas >= 0 && *cr.Spec.Repo.Replicas != replicas {
+			replicas = *cr.Spec.Repo.Replicas
+		}
+	}
+	return &replicas
+}
+
+// getArgoCDServerReplicas will return the size value for the argocd-server replica count.
```
wtam2018, line 48: Please see my comment in `getArgoCDRepoServerReplicas()`.  It applies to this function as well.

controllers/argocd/deployment_test.go
```diff
@@ -40,6 +40,44 @@ var (
 		"argocd-server"}
 )
```
wtam2018, line 42: Please add a test to verfiy that if ha is enabled, argocd server replica count in cr is ignored.

api/v1alpha1/argocd_types.go
wtam2018: Ok.  thanks.

docs/reference/argocd.md
```diff
@@ -821,6 +821,7 @@ LogLevel | info | The log level to be used by the ArgoCD Repo Server. Valid opti
 LogFormat | text | The log format to be used by the ArgoCD Repo Server. Valid options are text or json.
 ExecTimeout | 180 | Execution timeout in seconds for rendering tools (e.g. Helm, Kustomize)
 Env | [Empty] | Environment to set for the repository server workloads
+Replicas | 1 | The number of replicas for the ArgoCD Repo Server.
```
wtam2018, line 824: Default to nil.  If this value set, it must be >= 0 This value is passed to replicas in Deployment spec.

docs/reference/argocd.md
```diff
@@ -972,6 +974,7 @@ Name | Default | Description
 Host | example-argocd | The hostname to use for Ingress/Route resources.
 [Ingress](#server-ingress-options) | [Object] | Ingress configuration for the Argo CD Server component.
 Insecure | false | Toggles the insecure flag for Argo CD Server.
+Replicas | 1 | The number of replicas  for Argo CD Server. If Server Autoscale is enabled, the number of replicas defined here will be overriden by the number of replicas decided by the autoscaler.
```
wtam2018, line 977: Default to nil. If this value set, it must be >= 0.  This value is passed to replicas in Deployment spec.  If HA is enabled, this field is ignored.

controllers/argocd/deployment_test.go
reginapizza: @wtam I've added to check if autoscale is enabled, could you check if what I have now is sufficient?

controllers/argocd/deployment.go
```diff
@@ -34,6 +34,28 @@ import (
...
+func getArgoCDRepoServerReplicas(cr *argoprojv1a1.ArgoCD) *int32 {
+	replicas := int32(1)
+	if *cr.Spec.Repo.Replicas != 0 {
+		if *cr.Spec.Repo.Replicas >= 0 && *cr.Spec.Repo.Replicas != replicas {
+			replicas = *cr.Spec.Repo.Replicas
+		}
+	}
+	return &replicas
+}
+
```
wtam2018, line 47:
```
if cr.Spec.Repo.Replicas != nil && *cr.Spec.Repo.Replicas >= 0 {
    return cr.Spec.Repo.Replicas
}

return nil

File: controllers/argocd/deployment.go [hunk 9]
```diff
@@ -929,6 +951,8 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argoprojv1a1.ArgoCD) error
 
 	deploy.Spec.Template.Spec.Volumes = repoServerVolumes
 
+	deploy.Spec.Replicas = getArgoCDRepoServerReplicas(cr)
+
```
wtam2018, line 955:
```
if getArgoCDRepoServerReplicas(cr) != nil {
    deploy.Spec.Replicas = getArgoCDRepoServerReplicas(cr)
}
```

controllers/argocd/deployment.go
wtam2018: // getArgoCDRepoServerReplicas will return the size value for the argocd-repo-server replica count if it has been set in argocd CR.  Otherwise, nil is returned.

jannfis (approved):
LGTM, thanks @reginapizza !

PR #472, opened 2021-11-05
fix: reconcile argocd cr changes for resouce inclusions in argocd-cm config map

### LINKED ISSUE #471: reconcile resource inclusions from Argo CD CR
**Describe the bug**
Operator currently ignores changes to resource inclusions expressed in the argo-cd CR instead of reconciling them in the config map

**To Reproduce**
1. Deploy operator
2. Update the argo-cd CR resource inclusions
3. wait for config map to be reconciled
4. config map does not reflect changes in resource inclusions 

**Expected behavior**
changes made to argo-cd CR concerning resource inclusions should be reflected in the `argocd-cm` config map

**Screenshots**

**Additional context**
Original issue: https://issues.redhat.com/browse/GITOPS-1518
Slack thread : https://coreos.slack.com/archives/CMP95ST2N/p1635956090105300

jaideepr97:
…g map

**What type of PR is this?**
 /kind bug
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

Adds reconciliation logic for resourceInclusions from argocd CR to `argocd-cm` config map

Fixes #471 

1. Deploy operator
2. Update the argo-cd CR resource inclusions
3. wait for config map to be reconciled
4. config map should reflect changes in resource inclusions

### DIFF
--- a/controllers/argocd/configmap.go
+++ b/controllers/argocd/configmap.go
@@ -471,6 +471,11 @@ func (r *ReconcileArgoCD) reconcileExistingArgoConfigMap(cm *corev1.ConfigMap, c
 		changed = true
 	}
 
+	if cm.Data[common.ArgoCDKeyResourceInclusions] != cr.Spec.ResourceInclusions {
+		cm.Data[common.ArgoCDKeyResourceInclusions] = cr.Spec.ResourceInclusions
+		changed = true
+	}
+
 	uri := r.getArgoServerURI(cr)
 	if cm.Data[common.ArgoCDKeyServerURL] != uri {
 		cm.Data[common.ArgoCDKeyServerURL] = uri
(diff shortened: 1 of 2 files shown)


iam-veeramalla:
Hi @jaideepr97 , #471 and this PR looks to address two different problems. 

Issue reported in #471 looks more to me as "Operator does not read and act on the ResourceInclusions provided in the Argo CD CR" . Jira provided in the issue also talks about the same thing. However, I can see this logic already handled in the operator.
https://github.com/argoproj-labs/argocd-operator/blob/5699a4968efbc2406220a1c1b49fbdbb43c6e490/controllers/argocd/configmap.go#L342

jaideepr97:
@iam-veeramalla 

I think that line of code probably only gets called when the `argocd-cm` is being created at the beginning
when `resrouceInclusions` are added the CR after there is already an exisiting `argocd-cm` configmap [e.g through `oc edit argoCD`], code redirects to the function `reconcileExistingArgoConfigMap` which currently ignores changes to `resourceInclusions`

iam-veeramalla:
> @iam-veeramalla
> 
> I think that line of code probably only gets called when the `argocd-cm` is being created at the beginning when `resrouceInclusions` are added the CR after there is already an exisiting `argocd-cm` configmap [e.g through `oc edit argoCD`], code redirects to the function `reconcileExistingArgoConfigMap` which currently ignores changes to `resourceInclusions`

ok makes sense :)

PR #473, opened 2021-11-08
feat: Allow users to pass Argo CD server command args from Argo CD CR(#468)

### LINKED ISSUE #468: Add support for ArgoCD server `--rootpath` parameter
The ui being a single page app, I can't place the service behind an url with a different rootpath other than `/`.

The server block should allow additional command parameters  
and merge them with the default when creating the Deployment resource.

Having a specific entry `rootpath` would also work, but limits flexibility.

**Additional context**

iam-veeramalla:
**What type of PR is this?**
> /kind enhancement

Allows a user to pass extra arguments to Argo CD Server command.

Name | Default | Description
--- | --- | ---
CommandArgs | [Empty] | List of arguments that will be added to Argo CD server command.

Please note that the command arguments would be added to the list of default arguments
https://github.com/argoproj-labs/argocd-operator/blob/70e3eca3b048d0525b8c98048864d2c545d19181/controllers/argocd/deployment.go#L239

### Server Command Arguments Example

``` yaml
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: server
spec:
  server:
    CommandArgs:
      - --rootpath
      - /argocd
```

Fixes #468

[x] Document Updated

Fixes #468 

1. Run the operator using `make install run`
2. Create an Argo CD Instance as below.
``` yaml
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: server
spec:
  server:
    CommandArgs:
      - --rootpath
      - /argocd
```
3. Wait for the operator to reconcile.
4. Describe the Argo CD server deployment to notice the newly added command arguments.


13013SwagR (commented):
It seems to be missing a modification of  `deployment.go`
`getArgoServer()` from https://github.com/argoproj-labs/argocd-operator/blob/18afda802312a424d96079f881a8e7e9f9fd09f3/controllers/argocd/deployment.go#L207-L238
the code above should check for the route path and append the command if set
with something like
```go
cmd = append(cmd, "--rootpath")
cmd = append(cmd, getRoutePath(...)) // defaults to '/'
```

jannfis (approved):
LGTM

iam-veeramalla:
> It seems to be missing a modification of `deployment.go` `getArgoServer()` from
> 
> https://github.com/argoproj-labs/argocd-operator/blob/18afda802312a424d96079f881a8e7e9f9fd09f3/controllers/argocd/deployment.go#L207-L238
> 
> 
> the code above should check for the route path and append the command if set
> with something like
> ```go
> cmd = append(cmd, "--rootpath")
> cmd = append(cmd, getRoutePath(...)) // defaults to '/'
> ```

Thanks 🙏  .. Will update the PR accordingly.

wtam2018:
This PR seems to be obsolete.

13013SwagR:
This PR was to implement the `--root-path` parameter as it is defined in the operator interface, but just not implemented.
See #468 
 @wtam2018 why would this be obsolete?
 Maybe it was implemented somewhere else since ?

wtam2018 (changes requested):
--rootpath should not be specifically cased for for Route only

wtam2018:
Instead of adding a "one-off" fix for --rootpath, we probalby should come up with a general solution to allow users to pass start command arguments to repo server.

iam-veeramalla:
> Instead of adding a "one-off" fix for --rootpath, we probalby should come up with a general solution to allow users to pass start command arguments to repo server.

makes sense. probably in a configmap or somewhere. Also these entries should be merged with the ones created already by the operator.

iam-veeramalla:
> > Instead of adding a "one-off" fix for --rootpath, we probalby should come up with a general solution to allow users to pass start command arguments to repo server.
> 
> makes sense. probably in a configmap or somewhere. Also these entries should be merged with the ones created already by the operator.

After giving it a thought I feel Argo CD CRD is the best place for this.

docs/reference/argocd.md
```diff
@@ -1024,6 +1025,30 @@ Name | Default | Description
...
+``` yaml
+apiVersion: argoproj.io/v1alpha1
+kind: ArgoCD
+metadata:
+  name: example-argocd
+  labels:
+    example: server
+spec:
+  server:
+    CommandArgs:
```
jannfis, line 1047: should be `commandArgs` here I guess

docs/reference/argocd.md
```diff
@@ -1091,6 +1116,9 @@ spec:
           kind: Deployment
           name: example-argocd-server
         targetCPUUtilizationPercentage: 50
+    CommandArgs:
```
jannfis, line 1119: should be `commandArgs` here I guess

controllers/argocd/route.go
```diff
@@ -243,6 +243,11 @@ func (r *ReconcileArgoCD) reconcileServerRoute(cr *argoprojv1a1.ArgoCD) error {
 		route.Spec.Host = cr.Spec.Server.Host // TODO: What additional role needed for this?
 	}
 
+	// Allow override of the Path for the Route
+	if len(cr.Spec.Server.Route.Path) > 0 {
+		route.Spec.Path = cr.Spec.Server.Route.Path
+	}
```
jannfis, line 249: This is somewhat unrelated to the purpose of this PR, no?

api/v1alpha1/argocd_types.go
```diff
@@ -467,6 +467,9 @@ type ArgoCDServerSpec struct {
 
 	// Env lets you specify environment for API server pods
 	Env []corev1.EnvVar `json:"env,omitempty"`
+
+	// Command arguments for Argo CD server
+	CommandArgs []string `json:"commandArgs,omitempty"`
```
jannfis, line 472: I think in the Helm ecosystem (and other operators I've seen), such an option is usually called `extraCommandArgs`, because it will append command args to a list of arguments already constructed by the Operator.

controllers/argocd/deployment.go
```diff
@@ -266,6 +266,10 @@ func getArgoServerCommand(cr *argoprojv1a1.ArgoCD) []string {
 	cmd = append(cmd, "--logformat")
 	cmd = append(cmd, getLogFormat(cr.Spec.Server.LogFormat))
 
+	if len(cr.Spec.Server.CommandArgs) > 0 {
+		cmd = append(cmd, cr.Spec.Server.CommandArgs...)
+	}
```
jannfis, line 271: I wonder whether we should check if one of the extra arguments shadows an argument that has been set by the Operator previously, and possibly log a warning or info message in that case?

I'm not sure how command line parsing would behave if there is the same option specified twice, but with a different value. Which one would take precedence? The first (set by the operator) or the second (set by the user)?

We should at least find out and document that behavior, so that there'll be no nasty surprises.

docs/reference/argocd.md
```diff
@@ -1024,6 +1025,30 @@ Name | Default | Description
 Enabled | false | Toggle Autoscaling support globally for the Argo CD server component.
 HPA | [Object] | HorizontalPodAutoscaler options for the Argo CD Server component.
 
+### Server Command Arguments
+
+Allows a user to pass arguments to Argo CD Server command.
+
+Name | Default | Description
+--- | --- | ---
+CommandArgs | [Empty] | List of arguments that will be added to Argo CD server command.
```
jannfis, line 1034: Should make clear that it will append to the list of parameters already set by the Operator, and the behavior on override behavior.

jannfis (commented):
I have some concerns and nits. PTAL.

config/crd/bases/argoproj.io_argocds.yaml
```diff
@@ -5097,6 +5097,11 @@ spec:
                     required:
                     - enabled
                     type: object
+                  commandArgs:
+                    description: Command arguments for Argo CD server
```
jopit, line 5101: The description should maybe say something like `Arguments to be appended to the command arguments for the Argo CD server`

controllers/argocd/deployment_test.go
```diff
@@ -1077,6 +1077,82 @@ func TestReconcileArgoCD_reconcileServerDeployment(t *testing.T) {
...
+
+	assert.NoError(t, r.Client.Get(
+		context.TODO(),
+		types.NamespacedName{
+			Name:      "argocd-server",
+			Namespace: a.Namespace,
+		},
+		deployment))
+
+	cmd := append(baseCommand, []string{"--rootpath", "/argocd"}...)
```
jopit, line 1117: This can be written more simply as

```
cmd := append(baseCommand, "--rootpath", "/argocd")
```

controllers/argocd/deployment_test.go
```diff
@@ -1077,6 +1077,82 @@ func TestReconcileArgoCD_reconcileServerDeployment(t *testing.T) {
...
+	assert.NoError(t, r.reconcileServerDeployment(a))
+	assert.NoError(t, r.Client.Get(
+		context.TODO(),
+		types.NamespacedName{
+			Name:      "argocd-server",
+			Namespace: a.Namespace,
+		},
+		deployment))
+
+	cmd = append(cmd, []string{"--foo", "bar", "test"}...)
```
jopit, line 1138: This can be written more simply as

```
cmd = append(cmd, "--foo", "bar", "test")
```

jopit (changes requested):

controllers/argocd/deployment.go
iam-veeramalla: After giving it a thought, I feel the better approach would be to error out when one of the extra arguments shadows an argument that has been set by the Operator previously. This will avoid unexpected consequences.

jopit (approved):

wtam2018 (approved):

PR #477, opened 2021-11-09
feat: bump cronjob api to v1

jaideepr97:
**What type of PR is this?**

> /kind bug
> /kind cleanup
> /kind failing-test
 /kind enhancement
> /kind documentation
> /kind code-refactoring

batch/v1beta1 version of cronjob is set to be deprecated in k8s v1.25. We should upgrade it in our code base before it becomes too late and causes bigger problems as we have seen it can in the past 

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125

Fixes #?


iam-veeramalla (approved):
LGTM, Thanks @jaideepr97

PR #490, opened 2021-11-23
fix: fix broken argocd export

### LINKED ISSUE #485: operator v0.1.0 ArgoCDExport Resource stuck in Pending phase
creating an ArgoCDExport resource (after the ArogCD cluster is up and running) using v0.1.0 doesn't start a kubernetes job, it's just stuck in **Pending** state. Local and AWS backends tested.

image

The problem seems to be that it's looking for the service account "argocd-application-controller" while the operator creates "argocd-argocd-application-controller"

I think the issue is caused by this line?
https://github.com/argoproj-labs/argocd-operator/blob/0be1f6f89a0164e05f4702800f0ac40ce1c9d5c5/controllers/argocdexport/job.go#L177

jopit:

**What type of PR is this?**
> /kind bug

Fixes the export function of the argocd operator

There were three problems that required fixing

1. The name of the service account has changed.  It is now prefixed by the name of the argocd instance it belongs to.
2. The argocd command to invoke the export function has changed from `argocd-util export` to `argocd admin export`
3. On OpenShift, the SCC that OpenShift assigns to the export job pod has changed from `restricted` to `anyuid`.  The effect of this is that the pod no longer has permissions to write the export data to the PV.  The fix for this is to specify the correct `runAsUser`, `runaAsGroup` and `fsGroup` values for the export job pod.

Fixes #485

Install the operator, create an argocd instance, create an argocdexport resource in the same namespace, ensure that the job pod completes successfully and that the logs indicate all went well, for example
```
% kubectl logs -n argocd example-argocdexport-hzwtw 
exporting argo-cd
creating argo-cd backup
encrypting argo-cd backup
argo-cd export complete
```

### DIFF
--- a/build/util/util.sh
+++ b/build/util/util.sh
@@ -32,7 +32,7 @@ export_argocd () {
 
 create_backup () {
     echo "creating argo-cd backup"
-    argocd-util export > ${BACKUP_EXPORT_LOCATION}
+    argocd admin export > ${BACKUP_EXPORT_LOCATION}
 }
 
 encrypt_backup () {
@@ -148,7 +148,7 @@ decrypt_backup () {
 
 load_backup () {
     echo "loading argo-cd backup"
-    argocd-util import - < ${BACKUP_EXPORT_LOCATION}
+    argocd admin import - < ${BACKUP_EXPORT_LOCATION}
 }
 
 usage () {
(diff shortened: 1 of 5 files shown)


wtam2018:
Thanks for the PR.  Please add some tests.   Are we addressing the uid/permission issue on openshift in this PR?

jopit:
Hi @wtam2018, I've added an e2e test for this. And yes, this PR does address the uid/permission issue on openshift.

wtam2018 (approved):
LGTM 
Thanks @jopit

PR #493, opened 2021-11-25
fix: operator cannot install keycloak when resourcequota scope is not set (#492)

### LINKED ISSUE #492: operator fails to install keycloak when resourcequota scope is terminating or no scope is set.
**Describe the bug**
when creating ArgoCD instances in user workload projects and defining `.spec.sso.provider: keycloak` the operator is creating a `DeploymentConfig` which is leading to a stuck deploy Pod due to missing resources in the `DepolymentConfig` and Quotas enforcing requests and limits set.
Please make sure, that the `DeploymentConfig` has resources set in the strategy or please migrate to a `StatefulSet` like the RH-SSO is creating the keycloak to make it possible to run keycloak in environments where quotas are enforces also for NotTerminating Pods.

**To Reproduce**
1. Create a ns with `resourcequota`.
2. Do not set any scope to the `resourcequota`.
3. Install or run the operator locally using `make run`.
4. Install keycloak by updating the Argo CD spec -> `.sso.provider: keycloak`.
5. You can see that keycloak deployment fails as the `keycloak-deploy` pod does not have any resource requests/limits set.

**Expected behavior**
operator should install and configure keycloak for any `resourcequota` scope.

**Screenshots**
NA

**Additional context**
NA

iam-veeramalla:

**What type of PR is this?**
> /kind bug

Fixes #492 

NA

Fixes #492 

Please refer to the issue #492

### DIFF
--- a/controllers/argocd/keycloak.go
+++ b/controllers/argocd/keycloak.go
@@ -305,6 +305,16 @@ func getKeycloakDeploymentConfigTemplate(cr *argoprojv1a1.ArgoCD) *appsv1.Deploy
 			Selector: map[string]string{"deploymentConfig": "${APPLICATION_NAME}"},
 			Strategy: appsv1.DeploymentStrategy{
 				Type: "Recreate",
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceMemory: resourcev1.MustParse("256Mi"),
+						corev1.ResourceCPU:    resourcev1.MustParse("250m"),
+					},
+					Limits: corev1.ResourceList{
+						corev1.ResourceMemory: resourcev1.MustParse("512Mi"),
+						corev1.ResourceCPU:    resourcev1.MustParse("500m"),
+					},
+				},
 			},
 			Template: &corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
--- a/controllers/argocd/keycloak_test.go
+++ b/controllers/argocd/keycloak_test.go
@@ -133,7 +133,22 @@ func TestNewKeycloakTemplate_testDeploymentConfig(t *testing.T) {
 	dc := getKeycloakDeploymentConfigTemplate(a)
 
 	assert.Equal(t, dc.Spec.Replicas, fakeReplicas)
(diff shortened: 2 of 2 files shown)


iam-veeramalla:
We dont need any reconciliation logic or a place holder in the CR as the lifetime of the deployer pod is only till the keycloak service is up and running. 

Background: On an OCP cluster, We use DeploymentConfig to install keycloak. DeploymentConfig creates a special pod called as a deployer pod. The scope or a lifetime of a deployer pod is only till the actual keycloak pod(container) is up and running. Once the keycloak pod is running, the deployer pod moves in to the `completed` state. So it is a terminating pod(pod with a limited lifetime).

Usually in most of the cases, `resourcequota` resource is created with a scope set to `NotTerminating`, which means the resource requirements is only made mandatory to the pods that have unlimited lifetime, so `deployer pods` do not fall into this category. However, In some cases, customers have `resourcequota` set to scope `Terminating`, probably because their pods have a lifetime associated. 

To overcome this issue, This PR sets a minimum resource requirements to the deployer pod. I don't see any reason to modify the resource requirements of deployer pod. so, I did not provide any option in the CR to override the default values.

iam-veeramalla:
PTAL @jannfis

controllers/argocd/keycloak.go
```diff
@@ -305,6 +305,16 @@ func getKeycloakDeploymentConfigTemplate(cr *argoprojv1a1.ArgoCD) *appsv1.Deploy
...
 				Type: "Recreate",
+				Resources: corev1.ResourceRequirements{
+					Limits: corev1.ResourceList{
+						corev1.ResourceMemory: resourcev1.MustParse("256Mi"),
+						corev1.ResourceCPU:    resourcev1.MustParse("250m"),
+					},
+					Requests: corev1.ResourceList{
+						corev1.ResourceMemory: resourcev1.MustParse("512Mi"),
+						corev1.ResourceCPU:    resourcev1.MustParse("500m"),
+					},
```
jannfis, line 316: I think having requests higher than limits make no sense - they should be swapped.

iam-veeramalla: My bad, made the required changes. PTAL @jannfis

wtam2018 (approved):
LGTM

PR #495, opened 2021-11-26
When multiple envs are merged, the pods may be coordinated many times due to unsorted

muicoder:
When env is not sorted, pod-template-hash changes will cause it may be in a coordinated many times.

Looking forward to a better env sorting method!

**What type of PR is this?**
> /kind bug

https://github.com/argoproj-labs/argocd-operator/issues/496

### DIFF
--- a/controllers/argoutil/env.go
+++ b/controllers/argoutil/env.go
@@ -2,6 +2,7 @@ package argoutil
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	"sort"
 )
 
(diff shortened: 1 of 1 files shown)


iam-veeramalla:
Hi @muicoder , Hey there is already a similar PR #467 .. Thanks for the PR anyways :) . Will try to get one of these merged asap.

iam-veeramalla:
duplicate of #467 . One of these should be merged.

PR #514, opened 2021-12-13
add route/ingress URL to .status

### LINKED ISSUE #246: Add the ingress/route URL to `.status`
Add the ingress/route URL to `.status` in the ArgoCD CR. 

The same applies for general Ingress too:
https://github.com/argoproj-labs/argocd-operator/blob/master/pkg/apis/argoproj/v1alpha1/argocd_types.go#L418

This would enable users/tools to extract the URL for launching the ArgoCD UI.

reginapizza:
**What type of PR is this?**
 /kind enhancement

This would add a new field, `.host`, to `.status` of `ArgoCD`. When route or ingress is enabled (priority given to route) then the route will be displayed in the new URL field. 

When no URL exists from a route or ingress, the field will not be displayed. 

When on a non-OpenShift cluster (meaning the Route API is not available), if the user chooses to enable Route, they will only get a log saying that Routes are not available in non-OpenShift environments and to please use Ingresses instead. The state of the application controller and of the URL will not be affected. 

When the route or ingress is configured, but the corresponding controller has not yet set it up properly (i.e. is not in Ready state or does not propagate its URL), this is indicated in the Operand as well in the value of `.status.url` as Pending instead of the URL. Also, if `.status.url` is Pending, this affects the overall status for the Operand by making it Pending instead of Available.

Fixes #246 

Note: this was tested with an OpenShift cluster
1. Get cluster and log in.

2. In your cloned repo on this branch, go to Makefile and change the version # (just to something else) and change image base tag to `IMAGE_TAG_BASE ?= quay.io/{your quay or docker username}/argocd-operator`. Then run `make docker-build`, `make docker-push`and login to a cluster and then run `make deploy`
3. Add the Argo CD instance:
```
kind: ArgoCD
metadata:
  name: argocd
spec: {}
EOF   
```
4. Edit the ArgoCD instance so that ingress and route (or at least just route if you're on an OpenShift cluster) is enabled with `oc edit argocd argocd`. `spec.server` should look something like: 
```
  server:
    autoscale:
      enabled: false
    grpc:
      ingress:
        enabled: false
    ingress:
      enabled: true
    route:
      enabled: true
    service:
      type: ""
  tls:
    ca: {}
```
Save your changes. 
5. Wait a second for it to update, and then check back on your CR `oc get argocd argocd -o yaml`. You should see that the URL has been added to the `.status` field. 
```
status:
  applicationController: Running
  dex: Running
  phase: Available
  redis: Running
  repo: Running
  server: Running
  ssoConfig: Unknown
  host: argocd-server-default.apps.app-svc-4.8-120914.devcluster.openshift.com
```

### DIFF
--- a/api/v1alpha1/argocd_types.go
+++ b/api/v1alpha1/argocd_types.go
@@ -702,6 +702,9 @@ type ArgoCDStatus struct {
 
 	// RepoTLSChecksum contains the SHA256 checksum of the latest known state of tls.crt and tls.key in the argocd-repo-server-tls secret.
 	RepoTLSChecksum string `json:"repoTLSChecksum,omitempty"`
+
+	// Host is the hostname of the Ingress.
+	Host string `json:"host,omitempty"`
 }
 
 // ArgoCDTLSSpec defines the TLS options for ArgCD.
--- a/controllers/argocd/status.go
+++ b/controllers/argocd/status.go
@@ -17,6 +17,10 @@ package argocd
 import (
 	"context"
 	"reflect"
+	"strings"
+
+	routev1 "github.com/openshift/api/route/v1"
+	corev1 "k8s.io/api/core/v1"
 
 	argoprojv1a1 "github.com/argoproj-labs/argocd-operator/api/v1alpha1"
 	"github.com/argoproj-labs/argocd-operator/controllers/argoutil"
@@ -51,6 +55,11 @@ func (r *ReconcileArgoCD) reconcileStatus(cr *argoprojv1a1.ArgoCD) error {
 	if err := r.reconcileStatusServer(cr); err != nil {
 		return err
 	}
+
+	if err := r.reconcileStatusHost(cr); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -211,4 +220,71 @@ func (r *ReconcileArgoCD) reconcileStatusServer(cr *argoprojv1a1.ArgoCD) error {
 		return r.Client.Status().Update(context.TODO(), cr)
 	}
 	return nil
+}
+
+// reconcileStatusHost will ensure that the host status is updated for the given ArgoCD.
+func (r *ReconcileArgoCD) reconcileStatusHost(cr *argoprojv1a1.ArgoCD) error {
+	cr.Status.Host = ""
+	cr.Status.Phase = "Available"
+	if cr.Spec.Server.Route.Enabled {
+		if !IsRouteAPIAvailable() {
+			log.Info("Routes not available in non-OpenShift environments, please use Ingresses instead")
+			return nil
+		}
+		route := newRouteWithSuffix("server", cr)
+		if !argoutil.IsObjectFound(r.Client, cr.Namespace, route.Name, route) {
+			log.Info("argocd-server route requested but not found on cluster")
+			return nil
+		} else {
+			// status.ingress not available
+			if route.Status.Ingress == nil {
+				cr.Status.Host = ""
+				cr.Status.Phase = "Pending"
+			} else {
+				// conditions exist and type is RouteAdmitted
+				if len(route.Status.Ingress[0].Conditions) > 0 && route.Status.Ingress[0].Conditions[0].Type == routev1.RouteAdmitted {
+					if route.Status.Ingress[0].Conditions[0].Status == corev1.ConditionTrue {
+						cr.Status.Host = route.Status.Ingress[0].Host
+						cr.Status.Phase = "Available"
+					} else {
+						cr.Status.Host = ""
+						cr.Status.Phase = "Pending"
+					}
+				} else {
+					// no conditions are available
+					if route.Status.Ingress[0].Host != "" {
+						cr.Status.Host = route.Status.Ingress[0].Host
+						cr.Status.Phase = "Available"
+					} else {
+						cr.Status.Host = "Unavailable"
+						cr.Status.Phase = "Pending"
+					}
+				}
+			}
+		}
+	} else if cr.Spec.Server.Ingress.Enabled {
+		ingress := newIngressWithSuffix("server", cr)
+		if !argoutil.IsObjectFound(r.Client, cr.Namespace, ingress.Name, ingress) {
+			log.Info("argocd-server ingress requested but not found on cluster")
+			return nil
+		} else {
+			if !reflect.DeepEqual(ingress.Status.LoadBalancer, corev1.LoadBalancerStatus{}) && len(ingress.Status.LoadBalancer.Ingress) > 0 {
+				var s []string
+				var hosts string
+				for _, ingressElement := range ingress.Status.LoadBalancer.Ingress {
+					if ingressElement.Hostname != "" {
+						s = append(s, ingressElement.Hostname)
+						continue
+					} else if ingressElement.IP != "" {
+						s = append(s, ingressElement.IP)
+						continue
+					}
+				}
+				hosts = strings.Join(s, ", ")
+				cr.Status.Host = hosts
+				cr.Status.Phase = "Available"
+			}
+		}
+	}
+	return r.Client.Status().Update(context.TODO(), cr)
 }
--- a/deploy/olm-catalog/argocd-operator/0.2.0/argoproj.io_argocds.yaml
+++ b/deploy/olm-catalog/argocd-operator/0.2.0/argoproj.io_argocds.yaml
@@ -4354,6 +4354,9 @@ spec:
                   of the  Argo CD Dex component Pods had a failure. Unknown: For some
                   reason the state of the Argo CD Dex component could not be obtained.'
                 type: string
+              host:
+                description: Host is the hostname of the Ingress.
+                type: string
               phase:
                 description: 'Phase is a simple, high-level summary of where the ArgoCD
                   is in its lifecycle. There are five possible phase values: Pending:
--- a/bundle/manifests/argoproj.io_argocds.yaml
+++ b/bundle/manifests/argoproj.io_argocds.yaml
@@ -4354,6 +4354,9 @@ spec:
                   of the  Argo CD Dex component Pods had a failure. Unknown: For some
                   reason the state of the Argo CD Dex component could not be obtained.'
                 type: string
+              host:
+                description: Host is the hostname of the Ingress.
+                type: string
               phase:
                 description: 'Phase is a simple, high-level summary of where the ArgoCD
                   is in its lifecycle. There are five possible phase values: Pending:
--- a/config/crd/bases/argoproj.io_argocds.yaml
+++ b/config/crd/bases/argoproj.io_argocds.yaml
@@ -4356,6 +4356,9 @@ spec:
                   of the  Argo CD Dex component Pods had a failure. Unknown: For some
                   reason the state of the Argo CD Dex component could not be obtained.'
                 type: string
+              host:
+                description: Host is the hostname of the Ingress.
+                type: string
               phase:
                 description: 'Phase is a simple, high-level summary of where the ArgoCD
                   is in its lifecycle. There are five possible phase values: Pending:
--- a/controllers/argocd/status_test.go
+++ b/controllers/argocd/status_test.go
@@ -1,9 +1,17 @@
 package argocd
 
 import (
+	"context"
 	"testing"
 
+	argoprojv1alpha1 "github.com/argoproj-labs/argocd-operator/api/v1alpha1"
+
+	routev1 "github.com/openshift/api/route/v1"
 	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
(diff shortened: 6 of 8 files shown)


controllers/argocd/status.go
```diff
@@ -212,3 +216,40 @@ func (r *ReconcileArgoCD) reconcileStatusServer(cr *argoprojv1a1.ArgoCD) error {
...
 	return nil
 }
+
+// reconcileStatusURL will ensure that the URL status is updated for the given ArgoCD.
+func (r *ReconcileArgoCD) reconcileStatusURL(cr *argoprojv1a1.ArgoCD) error {
+
+	if cr.Spec.Server.Route.Enabled {
+		route := newRouteWithSuffix("server", cr)
+		if argoutil.IsObjectFound(r.Client, cr.Namespace, route.Name, route) {
+			if route.Spec.Host == "" {
```
iam-veeramalla, line 226: any reason for choosing `route.spec.Host` instead of `route.status.ingress.conditions.host` ?

controllers/argocd/status.go
```diff
@@ -212,3 +216,40 @@ func (r *ReconcileArgoCD) reconcileStatusServer(cr *argoprojv1a1.ArgoCD) error {
...
+				cr.Status.URL = route.Spec.Host
+				cr.Status.Phase = "Available"
+				return r.Client.Status().Update(context.TODO(), cr)
+			}
+		}
+		return nil
+	} else if cr.Spec.Server.Ingress.Enabled {
+		ingress := newIngressWithSuffix("server", cr)
+		if argoutil.IsObjectFound(r.Client, cr.Namespace, ingress.Name, ingress) {
+			if ingress.Spec.Rules[0].Host == "" {
```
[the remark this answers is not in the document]
iam-veeramalla: If you think this is a corner case, we can add a comment.

deploy/olm-catalog/argocd-operator/0.2.0/argoproj.io_argocds.yaml
```diff
@@ -4336,6 +4336,8 @@ spec:
                   SSO providers are configure in CR. Unknown: For some reason the
                   SSO configuration could not be obtained.'
                 type: string
+              url:
```
iam-veeramalla, line 4339: This file is obsolete, was it updated by the `make bundle` target ?

reginapizza: yes it was done automatically by `make bundle`, am I on the wrong version or something? Maybe I'm missing some newer changes and need to rebase?

controllers/argocd/status.go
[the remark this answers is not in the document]
reginapizza: I wasn't aware of that... I'm not sure if it's an edge case or not, do you have any idea? 
cc @jannfis

controllers/argocd/deployment.go
```diff
@@ -1092,9 +1088,7 @@ func (r *ReconcileArgoCD) reconcileServerDeployment(cr *argoprojv1a1.ArgoCD) err
 		},
 	}
 
-	if replicas := getArgoCDServerReplicas(cr); replicas != nil {
-		deploy.Spec.Replicas = replicas
-	}
+	deploy.Spec.Replicas = getArgoCDServerReplicas(cr)
```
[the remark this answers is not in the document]
reginapizza: my bad, I'm not sure how this got in here! Definitely should not be, I'll remove it

deploy/olm-catalog/argocd-operator/0.2.0/argoproj.io_argocds.yaml
iam-veeramalla: this file is obsolete. No need to worry w.r.t this PR. I will take care of removing it in a different PR.

controllers/argocd/status.go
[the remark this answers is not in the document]
iam-veeramalla: you can add a TODO section on this for now.

iam-veeramalla:
@reginapizza please run `make bundle` and resubmit the file `0.2.0/argoproj.io_argocds.yaml`, my bad that I asked you to uncommit that file.

jaideepr97:
would it be a good idea to mention that the created route/ingress URL can be retrieved from the ArgoCD instance's `.status.URL` field in https://argocd-operator.readthedocs.io/en/latest/usage/ingress/ and https://argocd-operator.readthedocs.io/en/latest/usage/routes/?

@reginapizza

jaideepr97:
On further thought, a couple of cases occurred to me that I'm wondering how/if we should handle:

1. If someone tries to set `.spec.server.route: enabled` in a k8s cluster, this might throw an error. Maybe we should add a check for the route API here and fail silently instead?

2. Do we want to allow users to be able to set both route and ingress to enabled? If Both route and ingress are enabled on Openshift, is there a possibility we could end up with 2 routes for the same server? (One created by the operator, the other created by the openshift router automatically from the ingress) In that case should we just allow the route created by the router?
Let me know what you think @reginapizza @iam-veeramalla

reginapizza:
@jaideepr97 I've updated my PR to address your first concern, but I'm still not sure about the second. @iam-veeramalla do you think that it's a concern?

controllers/argocd/status.go
```diff
@@ -212,3 +217,43 @@ func (r *ReconcileArgoCD) reconcileStatusServer(cr *argoprojv1a1.ArgoCD) error {
...
+					return r.Client.Status().Update(context.TODO(), cr)
+				} else {
+					cr.Status.URL = route.Spec.Host
+					cr.Status.Phase = "Available"
+					return r.Client.Status().Update(context.TODO(), cr)
+				}
+			}
+			return nil
+		}
+		log.Info("Routes not available in non-OpenShift environments, please use Ingresses instead")
```
jaideepr97, line 241: Could you put this in an else block? Otherwise this statement would get logged even when RouteAPI is available

iam-veeramalla: instead use 
if ! IsRouteAPIAvailable() {
       log.Info("Routes not available in non-OpenShift environments, please use Ingresses instead")
}

route := newRouteWithSuffix("server", cr)
.....
.......

iam-veeramalla: else block is not needed :)

jaideepr97: That sounds good too
as long as we return after logging

jaideepr97:
LGTM
Thanks @reginapizza !!

api/v1alpha1/argocd_types.go
```diff
@@ -702,6 +702,9 @@ type ArgoCDStatus struct {
 
 	// RepoTLSChecksum contains the SHA256 checksum of the latest known state of tls.crt and tls.key in the argocd-repo-server-tls secret.
 	RepoTLSChecksum string `json:"repoTLSChecksum,omitempty"`
+
+	// URL is the url for the hostname of the route/ingress
+	URL string `json:"url,omitempty"`
```
wtam2018, line 707: Please rename this variable to `Host` and update the comment.  It is not really a URL. 
```
Host is the hostname of the Route or Ingress.
```

controllers/argocd/status.go
```diff
@@ -212,3 +217,44 @@ func (r *ReconcileArgoCD) reconcileStatusServer(cr *argoprojv1a1.ArgoCD) error {
 	}
 	return nil
 }
+
+// reconcileStatusURL will ensure that the URL status is updated for the given ArgoCD.
+func (r *ReconcileArgoCD) reconcileStatusURL(cr *argoprojv1a1.ArgoCD) error {
```
wtam2018, line 222: Please rename the function name to `reconcileStatusHost`.

controllers/argocd/status.go
```diff
@@ -212,3 +217,44 @@ func (r *ReconcileArgoCD) reconcileStatusServer(cr *argoprojv1a1.ArgoCD) error {
...
+	if cr.Spec.Server.Route.Enabled {
+		if !IsRouteAPIAvailable() {
+			log.Info("Routes not available in non-OpenShift environments, please use Ingresses instead")
+			return nil
+		}
+		route := newRouteWithSuffix("server", cr)
+		if argoutil.IsObjectFound(r.Client, cr.Namespace, route.Name, route) {
+			if route.Status.Ingress == nil {
+				status := "Pending"
+				cr.Status.URL = status
```
wtam2018, line 233: Why are we assigning `status` to `URL`?

controllers/argocd/status.go
```diff
@@ -212,3 +217,44 @@ func (r *ReconcileArgoCD) reconcileStatusServer(cr *argoprojv1a1.ArgoCD) error {
...
+				cr.Status.Phase = status
+				return r.Client.Status().Update(context.TODO(), cr)
+			} else {
+				cr.Status.URL = route.Spec.Host
```
wtam2018, line 237: We should get the host from `route.Status.Ingress` which is a list of RouteIngresses. I think we should show all the ingresses.

controllers/argocd/status.go
```diff
@@ -212,3 +217,44 @@ func (r *ReconcileArgoCD) reconcileStatusServer(cr *argoprojv1a1.ArgoCD) error {
...
+	} else if cr.Spec.Server.Ingress.Enabled {
+		ingress := newIngressWithSuffix("server", cr)
+		if argoutil.IsObjectFound(r.Client, cr.Namespace, ingress.Name, ingress) {
+			if ingress.Status.LoadBalancer.Ingress == nil {
+				status := "Pending"
+				cr.Status.Host = status
+				cr.Status.Phase = status
+				return r.Client.Status().Update(context.TODO(), cr)
+			} else {
+				cr.Status.Host = ingress.Spec.Rules[0].Host
```
wtam2018, line 252: We should get hostnames from ingress.Status.LoadBalancer.LoadBalancerIngress which is a list.

api/v1alpha1/argocd_types.go
```diff
@@ -702,6 +702,9 @@ type ArgoCDStatus struct {
 
 	// RepoTLSChecksum contains the SHA256 checksum of the latest known state of tls.crt and tls.key in the argocd-repo-server-tls secret.
 	RepoTLSChecksum string `json:"repoTLSChecksum,omitempty"`
+
+	// Host is the hostname of the Ingress.
+	Host string `json:"url,omitempty"`
```
wtam2018, line 707:
```suggestion
	Host string `json:"host,omitempty"`
```

wtam2018:
Ingress and route can be enabled at the same time. However, the status does not show ingress if route is enabled.  WDYT @jaideepr97 @iam-veeramalla @reginapizza Should status be honest about showing ingress as well?

jaideepr97:
Per my understanding, on openshift environments routes are given more importance than ingresses (since a route is created automatically for each ingress) so I would say it makes sense to just show the route and document that the ingress is not shown in this case 

On a related note, if both route and ingress are enabled we might end up with 2 routes for the same server deployment 
Is this a case we want to handle?

wtam2018:
@jaideepr97 My understanding is the opposite. https://docs.openshift.com/container-platform/4.6/networking/routes/route-configuration.html#nw-ingress-creating-a-route-via-an-ingress_route-configuration You can annotate an Ingress and OpenShift can create a Route automatically, 
The question came up for the case where both ingress and route are legitimately enabled by the use in the Argo CD CR. We are only showing the status of Route.

jannfis (approved):
LGTM

controllers/argocd/status.go
```diff
@@ -212,3 +221,70 @@ func (r *ReconcileArgoCD) reconcileStatusServer(cr *argoprojv1a1.ArgoCD) error {
...
+				}
+			}
+		}
+	} else if cr.Spec.Server.Ingress.Enabled {
+		ingress := newIngressWithSuffix("server", cr)
+		if !argoutil.IsObjectFound(r.Client, cr.Namespace, ingress.Name, ingress) {
+			log.Info("argocd-server ingress requested but not found on cluster")
+			return nil
+		} else {
+			if !reflect.DeepEqual(ingress.Status.LoadBalancer, corev1.LoadBalancerStatus{}) && len(ingress.Status.LoadBalancer.Ingress) > 0 {
```
wtam2018, line 271: @reginapizza @jannfis I can''t figure out why is the DeepEqual() is needed.

wtam2018:
@reginapizza please update how to test o include testing the ingress only case.

reginapizza:
@wtam2018 I've updated the instructions on how to test ingresses with it!

controllers/argocd/status.go
reginapizza: would `if !reflect.ValueOf(ingress.Status.LoadBalancer).IsZero() && len(ingress.Status.LoadBalancer.Ingress) > 0 {` be better?

wtam2018: Yes, `IsZero` a bit clean and faster then `DeepCopy` and `&& len(ingress.Status.LoadBalancer.Ingress) > 0` can be eliminated.

jaideepr97: I think we should keep the `len(ingress.Status.LoadBalancer.Ingress) > 0` as a sanity check, to make sure code only executes if we have at least 1 proper ingress hostname/ip to work with

iam-veeramalla:
Known Issue:
@ishitasequeira and I have noticed that this feature does not work as expected when Argo CD server is exposed using an Ingress resource(instead of route) on an OpenShift cluster. Please refer to the [issue](https://github.com/argoproj-labs/argocd-operator/issues/620) or [JIRA](https://issues.redhat.com/browse/GITOPS-1920) for more details.

PR #518, opened 2021-12-14
feat: Support for custom roles using env variables

chetan-rns:

**What type of PR is this?**
> /kind enhancement

1. Introduces the env variables `CONTROLLER_CLUSTER_ROLE` and `SERVER_CLUSTER_ROLE` for specifying the cluster roles for application controller and server respectively.
2. If these env variables are found, the operator will use these roles in the bindings and delete the existing default roles.
3. These roles will be applied to all the namespaces where Argo CD is running or managed by Argo CD using the managed-by label

1. Run the operator locally and create multiple Argo CD instances
2. Add the env variables `CONTROLLER_CLUSTER_ROLE` and `SERVER_CLUSTER_ROLE` 
3. Check if the default roles are deleted in all the namespaces
4. Check if the role bindings have an updated role ref i.e. it should refer to a cluster role instead of a role

### DIFF
--- a/controllers/argocd/role.go
+++ b/controllers/argocd/role.go
@@ -104,11 +104,7 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
 
 	// create policy rules for each namespace
 	for _, namespace := range namespaces.Items {
-		var useCustomRole bool
-		if customRole := getCustomRoleName(name); customRole != "" {
-			useCustomRole = true
-		}
-
+		customRole := getCustomRoleName(name)
 		role := newRole(name, policyRules, cr)
 		if err := applyReconcilerHook(cr, role, ""); err != nil {
 			return nil, err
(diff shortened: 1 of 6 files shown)


controllers/argocd/role.go
```diff
@@ -104,6 +104,11 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
 
 	// create policy rules for each namespace
 	for _, namespace := range namespaces.Items {
+		var useCustomRole bool
+		if customRole := getCustomRoleName(name); customRole != "" {
+			useCustomRole = true
+		}
```
jannfis, line 110: Any special reason for having the `useCustomRole` boolean, and not just using check for `customRole != ""` below?

chetan-rns: It's not required. I will get rid of it

iam-veeramalla (changes requested):
LGTM !! Thanks @chetan-rns 

Please add a document page to explain the usage and significance.

chetan-rns:
@iam-veeramalla I have added a section in the doc for custom roles. PTAL

wtam2018 (approved):
LGTM

iam-veeramalla:
@chetan-rns , we missed to add the new documentation page entry in the file
https://github.com/argoproj-labs/argocd-operator/blob/master/mkdocs.yml

can you fix this by creating a PR ?

PR #530, opened 2021-12-24
Filter managed namespaces only once per reconciliation

chetan-rns:
**What type of PR is this?**

/kind code-refactoring

Filter the list of managed namespaces only once during each reconciliation and reuse them at multiple places. Currently, we list the namespaces while reconciling roles and role bindings. Since RBAC is managed for each argocd component, the operator filters the namespaces more than eight times per reconciliation. 

Fixes #?


wtam2018 (approved):
LGTM
Thanks @chetan-rns

PR #532, opened 2022-01-04
feat: bump Argo CD version to 2.2.2

iam-veeramalla:
Signed-off-by: iam-veeramalla abhishek.veeramalla@gmail.com

What type of PR is this?

/kind enhancement

What does this PR do / why we need it:
upgrades Argo CD to v2.2.2

Have you updated the necessary documentation?
NA

How to test changes / Special notes to the reviewer:


chetan-rns:
@iam-veeramalla Should we update the CRDs for `Application` and `AppProject`?

iam-veeramalla:
> @iam-veeramalla Should we update the CRDs for `Application` and `AppProject`?

Thats a good one. Let me go and check :)

wtam2018:
Please update.
build/util/Dockerfile
docs/reference/argocd.md
go.mod/go.sum

controllers/argocd/configmap.go
```diff
@@ -680,15 +680,17 @@ func (r *ReconcileArgoCD) reconcileSSHKnownHosts(cr *argoprojv1a1.ArgoCD) error
 // reconcileTLSCerts will ensure that the ArgoCD TLS Certs ConfigMap is present.
 func (r *ReconcileArgoCD) reconcileTLSCerts(cr *argoprojv1a1.ArgoCD) error {
 	cm := newConfigMapWithName(common.ArgoCDTLSCertsConfigMapName, cr)
-	cm.Data = getInitialTLSCerts(cr)
```
wtam2018, line 683: where are the changes in configmap.go coming from?

wtam2018: There are changes to reconciler logic.  What amount of testing does it come with?

iam-veeramalla: I am equally surprised. I dont know why this unit test was failing. 

The issue is coming from isObjectFound method but I dont see any changes made. 
https://github.com/argoproj-labs/argocd-operator/blob/2da8dac3bba7f0fbe52cd390dd10120d5545da34/controllers/argoutil/resource.go#L86

iam-veeramalla: To be honest there is no change in the logic. It's mostly refactoring. So according to me no additional testing is needed.

wtam2018: Which unit test was failing?
The change of logic/behavior is that now it only sets controller owner reference when object does not exist (or when a new cm is created).

controllers/argocd/deployment.go
```diff
@@ -962,6 +962,11 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argoprojv1a1.ArgoCD) error
 		desiredImage := getRepoServerContainerImage(cr)
 		if actualImage != desiredImage {
 			existing.Spec.Template.Spec.Containers[0].Image = desiredImage
+			if existing.Spec.Template.ObjectMeta.Labels == nil {
```
wtam2018, line 965: Is this change also made due to a failing test?

iam-veeramalla: Yes, It is made to fix the failing test. I will try to investigate more on what was the reason for unit test failure tomorrow and get back.

iam-veeramalla: I can only suspect it has something to do with the kubernetes client version update to `1.22.2`.

controllers/argocd/configmap.go
iam-veeramalla: This is the failing unit test
`TestReconcileArgoCD_reconcileTLSCerts_withUpdate`

However, I figured out the root cause of the issue. 
`The latest version of controller-runtime overrides the object instead of updating`.

Please refer to the below PR.
https://github.com/kubernetes-sigs/controller-runtime/pull/1651

iam-veeramalla: > The change of logic/behavior is that now it only sets controller owner reference when object does not exist (or when a new cm is created).

That makes sense as existing objects(configmaps) would already have the owner reference set.

controllers/argocd/deployment.go
iam-veeramalla: However, I figured out the root cause of the issue.
The latest version of controller-runtime overrides the object instead of updating.

Please refer to the below PR.
kubernetes-sigs/controller-runtime#1651

controllers/argocd/configmap.go
wtam2018: https://github.com/kubernetes-sigs/controller-runtime/pull/1651 is a fix to the fake client.  The question is that is the current reconciler not doing the right thing and needs to be changed? If the fix to the fake client only affect testing, the failing test should be fixed.

wtam2018 (approved):
LGTM
Thanks @iam-veeramalla

PR #508, opened 2021-12-06
fix: ApplicationSet Controller Does Not Receive Proxy Configuration

iam-veeramalla:
Signed-off-by: iam-veeramalla <abhishek.veeramalla@gmail.com>

**What type of PR is this?**
> /kind bug

**What does this PR do / why we need it**:
When using  Operator on Red Hat Openshift 4.7, the argocd-applicationset-controller does not have the global proxy configuration applied.

This causes us to be unable to use ApplicationSet resources due to the following error in the argocd-applicationset-controller logs:
```
Error during fetching repo: `git fetch origin HEAD --tags --force` failed timeout after 1m30s
```

**Have you updated the necessary documentation?**
NA

**Which issue(s) this PR fixes**:
https://github.com/redhat-developer/gitops-operator/issues/249

**How to test changes / Special notes to the reviewer**:
Please refer
https://github.com/redhat-developer/gitops-operator/issues/249

PR #538, opened 2022-01-20
[cherry-pick]: ApplicationSet Controller Does Not Receive Proxy Configuration #508

iam-veeramalla:
**What type of PR is this?**
> /kind bug
 cherry-pick #508 

When using  Operator on Red Hat Openshift 4.7, the argocd-applicationset-controller does not have the global proxy configuration applied.

This causes us to be unable to use ApplicationSet resources due to the following error in the argocd-applicationset-controller logs:
```
Error during fetching repo: `git fetch origin HEAD --tags --force` failed timeout after 1m30s
```

NA

https://github.com/redhat-developer/gitops-operator/issues/249

Please refer
https://github.com/redhat-developer/gitops-operator/issues/249

### DIFF
--- a/controllers/argocd/applicationset.go
+++ b/controllers/argocd/applicationset.go
@@ -132,16 +132,21 @@ func (r *ReconcileArgoCD) reconcileApplicationSetDeployment(cr *argoprojv1a1.Arg
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


wtam2018 (approved):
LGTM

PR #562, opened 2022-02-15
feat: add support for config management plugins 2.0

jopit:

**What type of PR is this?**

> /kind enhancement

Adds support for Config Management Plugins 2.0, see https://argo-cd.readthedocs.io/en/stable/user-guide/config-management-plugins/#configure-plugin-via-sidecar

Apply the following manifests to a cluster (if OpenShift, remove the `runAsUser` tag)
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: cmp-plugin
  namespace: argocd
data:
  plugin.yaml: |
    kind: ConfigManagementPlugin
    metadata:
      name: cmp-plugin
    spec:
      version: v1.0
      generate:
        command: [sh, -c, 'echo "{\"kind\": \"ConfigMap\", \"apiVersion\": \"v1\", \"metadata\": { \"name\": \"$ARGOCD_APP_NAME\", \"namespace\": \"$ARGOCD_APP_NAMESPACE\", \"annotations\": {\"Foo\": \"$FOO\", \"Bar\": \"baz\"}}}"']
      discover:
        find:
          command: [sh, -c, 'echo "FOUND"; exit 0']
      allowConcurrency: true
      lockRepo: true
---
kind: ArgoCD
metadata:
  name: argocd
  namespace: argocd
spec:
  server:
    route:
      enabled: true
  repo:
    sidecarContainers:
    - name: cmp
      command: [/var/run/argocd/argocd-cmp-server] # Entrypoint should be Argo CD lightweight CMP server i.e. argocd-cmp-server
      image: busybox # This can be off-the-shelf or custom-built image
      securityContext:
        runAsNonRoot: true
        runAsUser: 999
      volumeMounts:
        - mountPath: /var/run/argocd
          name: var-files
        - mountPath: /home/argocd/cmp-server/plugins
          name: plugins
        - mountPath: /tmp
          name: tmp
        # Remove this volumeMount if you've chosen to bake the config file into the sidecar image.
        - mountPath: /home/argocd/cmp-server/config/plugin.yaml
          subPath: plugin.yaml
          name: cmp-plugin
    volumes:
    - configMap:
        name: cmp-plugin
      name: cmp-plugin
```
this registers a plugin that will accept all applications.  Then add the following application
```yaml
kind: Application
metadata:
  name: guestbook
  namespace: argocd
spec:
  project: default
  source:
    repoURL: 'https://github.com/argoproj/argocd-example-apps.git'
    path: guestbook
    targetRevision: HEAD
    plugin:
      env:
        - name: FOO
          value: myfoo
  destination:
    server: 'https://kubernetes.default.svc'
    namespace: argocd
  syncPolicy:
    automated: {}
```
Once the application has synced successfully, check for the existence in the `argocd` namespace of a config map called `guestbook`.  Note that none of the other manifests from the application will be applied.


controllers/argocd/deployment.go
```diff
@@ -831,7 +841,24 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argoprojv1a1.ArgoCD) error
 		repoEnv = argoutil.EnvMerge(repoEnv, []corev1.EnvVar{{Name: "ARGOCD_EXEC_TIMEOUT", Value: fmt.Sprintf("%d", *cr.Spec.Repo.ExecTimeout)}}, true)
 	}
 
-	deploy.Spec.Template.Spec.InitContainers = getArgoRepoInitContainers(cr)
+	deploy.Spec.Template.Spec.InitContainers = []corev1.Container{{
```
wtam2018, line 844: It looks like the user will no longer be able to specfic the initContainer config. It is going to be a upgrade impact. Can we continue to support initContainer without affecting sidecar container?  Is it correct to pass the `repoEnv` to the initContainers?

jopit: The user will still be able to specify the initContainers config.  The code just below this (at line 859) appends the user-specified initContainers to the deploy.Spec.Template.Spec.InitContainers array.

I'll change the code to not pass the `repoEnv` to the copyutil init container.

wtam2018 (approved):
LGTM

PR #563, opened 2022-02-15
chore: add missing bundle versions

jopit:

**What type of PR is this?**

> /kind chore

After releasing 0.2.0 of the argocd-operator, we didn't set up the master branch for the next version.  That means on master, any changes to the CRD have gone into the 0.2.0 bundle folder.  Since we've already split off the branch for 0.3.0, the next version should be 0.4.0.  This PR is to fix this situation


iam-veeramalla:
Had a call/discussion with @jopit on why the operator version is updated as `0.4.0` instead of `0.3.0`. 

Since we already branched out `0.3.0(release-0.3)` and we know some changes that are in master are not going into `release-0.3`, it is better to update the operator version to `0.4.0` in master. This way when we create the next release branch we dont have to update the `operator version` and `manifests` in the release branch.

However, I see an overhead of merging back the `0.3.0` bundle folder in `release-0.3` branch back into master after/during `v0.3.0` release.

@jopit and I have decided that 
1. we shall merge this PR for now, if not the `0.2.0` bundle folder in the master would keep updating.
2. we can decide about this overhead in the next cabal, if there is a change in the decision we can act accordingly with a new PR if required.

iam-veeramalla (approved):
LGTM, Thanks @jopit for catching this.

wtam2018 (commented):
I'd change 0.4.0 to 0.3.0 and cherry pick the missing files to release-0.3 since 0.3 has not been released yet. @jopit

PR #567, opened 2022-02-17
cleanup: remove openshift related code from openshift package

jaideepr97:
**What type of PR is this?**

> /kind bug
> /kind chore
 /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

removes openshift related code from the reconcilerHook 

Fixes #?


wtam2018:
please specific the scope of this PR. There are still openshift api dependencies remain. Please specify how to test. Once they code are removed, please identify where they are moved to. Please update the doc. After openshift specify coded are moved to gitops operator. The community operator probably will not support features have need openshift api.

jaideepr97:
@wtam2018 This PR is only targeting the openshift package, which contains logic specific to (and only invoked when)  argocd-operator being consumed as a dependency within the gitops-operator. This includes the following:
- setting clusterconfig permissions for the `argocd-application-controller` clusterrole
- setting commands/arguments for red hat redis deployment
- setting commands/arguments for red hat redis ha server statefulset 
- configuring cluster secret with empty namespaces to allow managing cluster resources
- setting openshift-specific permissions for `argocd-application-controller` role 

This code is currently not included when packaging the upstream argocd-operator releases. As such, this code is entirely irrelevant to the argocd-operator when used by itself. I don't think any tests are required for the removal of this code within argocd-operator itself. 

This code will be moved into a new package within the gitops-operator controllers folder as it is currentlyte 

The only way to test it that I could think of (and did) is to a new custom gitops-operator image with the added openshift package, and test that the behavior of the operator is unchanged after moving the code there. 

(see: https://github.com/redhat-developer/gitops-operator/pull/295 for corresponding gitops-operator PR to add this package there
see: https://spaces.redhat.com/display/GITOPS/How+to+create+your+custom+Gitops+operator+catalog+for+testing for how to create a custom gitops-operator catalog image for testing)

There should not be any documentation update required for this PR either, since it is only refactoring some code. The behavior for neither operator will change in any way 


Whether argocd-operator should continue supporting other parts of the openshift-api (such as route creation) is a separate discussion from the is PR, the way I see it

chetan-rns:
Please remove the Makefile target `run-openshift` that points to openshift.go

wtam2018:
@jaideepr97 please take a look at the test failure.

controllers/openshift/openshift.go
```diff
@@ -1,347 +0,0 @@
...
-// which requires additional args in comparison to stock redis.
-func getArgsForRedhatHaRedisInitContainer() []string {
-	return []string{
-		"sh",
-		"/readonly-config/init.sh",
-	}
-}
-
-// policyRulesForClusterConfig defines rules for cluster config.
-func policyRulesForClusterConfig() []rbacv1.PolicyRule {
```
jannfis, line 164: So with removing this, would the Operator only ever be able to reconcile a namespace-scoped instance? If the users would want cluster scope, would they need to put in the roles by themselves?

jaideepr97: This code is actually irrelevant to the argocd-operator itself at the moment since we don't include the openshift package when building the argocd-operator. argocd-operator behavior would remain unchanged

The hook (when registered) overwrites the app-controller clusterrole privileges to something that is acceptable to openshift
But the default clusterrole itself [currently has permissions to do everything on all resources](https://github.com/argoproj-labs/argocd-operator/blob/569db9415782c3c27cb13ddb01e9e3fdb2c31e1e/controllers/argocd/policyrule.go#L9) (maybe something we should look into also?)

saumeya:
Also merging this might need re-verification of operator docs that mention openshift specific operations.

jaideepr97:
@saumeya I think those are separate issues
The openshift related documentation we have is mostly around installation (if I'm not wrong). The behavior of the operator on openshift itself would also not be affected by this PR (creating routes etc)
This PR is only removing the part of the code that was used exclusively by GitOps operator, and should have no visible effect to the users of either operator

iam-veeramalla:
@jaideepr97 I think we should also remove the below file and its relevant entries. If the OpenShift package is removed, I don't see a point of maintaining the below hooks. WDYT ?

https://github.com/argoproj-labs/argocd-operator/blob/68e1197b38ab9ce2461d8825765564eece2e1e4d/controllers/argocd/hooks.go

https://github.com/argoproj-labs/argocd-operator/search?q=applyReconcilerHook

iam-veeramalla:
@jaideepr97 probably the next step can be to make the below code neater.
https://github.com/argoproj-labs/argocd-operator/search?q=applyReconcilerHook

iam-veeramalla (approved):

PR #573, opened 2022-02-22
chore: make 0.3.0 the latest bundle version

jopit:

**What type of PR is this?**

> /kind chore

PR https://github.com/argoproj-labs/argocd-operator/pull/563 updated the latest bundle version to be 0.3.0.  After discussion with Abhishek and William, this is the wrong thing to do, the latest bundle version should be 0.3.0 as 0.3 has not yet been released.  This PR fixes that.


bundle/manifests/argocd-operator.clusterserviceversion.yaml
```diff
@@ -1048,4 +1048,4 @@ spec:
   provider:
     name: Argo CD Community
   replaces: argocd-operator.v0.1.0
```
iam-veeramalla, line 1050: it should be replacing `argocd-operator-v0.2.0` ?

jopit: Thanks, I missed that.

config/manifests/bases/argocd-operator.clusterserviceversion.yaml
```diff
@@ -660,5 +660,5 @@ spec:
   maturity: alpha
   provider:
     name: Argo CD Community
-  replaces: argocd-operator.v0.1.0
-  version: 0.2.0
+  replaces: argocd-operator.v0.2.0
+  version: 0.0.0
```
iam-veeramalla, line 664: should be 0.3.0 ?

jopit: It doesn't matter, it gets replaced in the generated CSV by the correct version.

iam-veeramalla (approved):
LGTM, Thanks @jopit

PR #553, opened 2022-02-03
fix: argocd-tls-certs-cm is overwritten on any change

iam-veeramalla:
Signed-off-by: iam-veeramalla <abhishek.veeramalla@gmail.com>

**What type of PR is this?**
> /kind bug

**What does this PR do / why we need it**:
#552 

**Have you updated the necessary documentation?**
* [x] Documentation update is required by this PR.
* [x] Documentation has been updated.

**Which issue(s) this PR fixes**:
Fixes #552 

**How to test changes / Special notes to the reviewer**:
1. Create an Argo CD Instance
2. Deploy the below Argo CD CR.
```
apiVersion: argoproj.io/v1alpha1
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: example-argocd
spec:
  server:
    route:
      enabled: true
```
3. Add a cert to argocd-tls-certs-cm configmap.
```
..........
data:
  test.example.com: '-----BEGIN CERTIFICATE-----  -----END CERTIFICATE-----'
```
4. Argo CD controller reconciles the configmap changes and remove the added cert.

PR #576, opened 2022-02-28
[Release v0.2.1] fix argocd-tls-certs-cm overwritting and CVE 2022-24348

### LINKED ISSUE #552: argocd-tls-certs-cm is overwritten on any change as of v0.2.0
**Describe the bug**
Prior to v0.2.0, certificates for TLS validation could be managed via the ArgoCD GUI or by controlling the argocd-tls-certs-cm ConfigMap directly. The value of argocd-tls-certs-cm was set initially when ArgoCD was deployed but not altered thereafter.

After upgrading tov0.2.0, this ConfigMap is reset to the value of tls.intitialCerts in the ArgoCD CR after any change. This renders the certificate management via the GUI or previous direct ConfigMap control useless and deletes any certificates that were added to the ConfigMap prior to the upgrade.

Workaround:  Add the needed cert to ArgoCD CR spec  tls.initialCerts.  
See https://argocd-operator.readthedocs.io/en/latest/reference/argocd/#tls-options.

Workaround example: https://github.com/iam-veeramalla/openshift-gitops-examples/blob/master/argocd/GITOPS-1725/argocd-initialTLScerts.yaml

**Expected behavior**
Revert back to the original behavior

**Screenshots**

**Additional context**

iam-veeramalla:
> /kind bug

1. cherrypicks #553 
2. fixes CVE 2022-24348 by upgrading Argo CD to v2.2.5

#552 
CVE 2022-24348

Fixes #552 
CVE 2022-24348

#552  is a clean cherry-pick from master branch. There is no additional testing required.

Steps to test CVE 2022-24348:

- Create namespace helm-guestbook
- Create ArgoCD instance somewhere
- Create application with this manifest
```
kind: Application
metadata:
  name: helm-guestbook
  namespace: <argo namespace>
spec:
  destination:
    namespace: helm-guestbook
    server: https://kubernetes.default.svc/
  project: default
  source:
    helm:
      valueFiles:
      - /foo.yaml
    path: helm-guestbook
    repoURL: https://github.com/argoproj/argocd-example-apps
    targetRevision: HEAD
```
- Try to sync application
-  Inspect helm-guestbook Application. If it has the below condition the fix is not working.
```
  conditions:
  - lastTransitionTime: "2022-02-16T19:20:22Z"
    message: 'rpc error: code = Unknown desc = `helm template . --name-template ........
```

### DIFF
--- a/build/util/Dockerfile
+++ b/build/util/Dockerfile
@@ -1,5 +1,5 @@
-# Argo CD v2.2.2
-FROM quay.io/argoproj/argocd@sha256:358c244c96313ca3bf9f588dc870d8123fc22ffa5c231c57da10f77b8d671c66
+# Argo CD v2.2.5
+FROM quay.io/argoproj/argocd@sha256:0cdf5aac80f0a77307cbb288a0c1e5e39d4bac0cafb3c285a6d3c656b9f8e910
 
 USER root
 
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -58,7 +58,7 @@ const (
 	ArgoCDDefaultArgoImage = "quay.io/argoproj/argocd"
 
 	// ArgoCDDefaultArgoVersion is the Argo CD container image digest to use when version not specified.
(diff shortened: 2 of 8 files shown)


jannfis (approved):
LGTM!

PR #577, opened 2022-02-28
feat: add permissions to manage monitoring resources

jaideepr97:
**What type of PR is this?**

> /kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
 /kind enhancement
> /kind documentation
> /kind code-refactoring

Adds permissions for managing `monitoring.coreos.com` resources to the `argocd-application-controller` role in order for argocd-operator to create `prometheusRules` (if there is one included in any application that is being managed by any argo-cd instance controlled by the argocd-operator)

Fixes https://issues.redhat.com/browse/GITOPS-1638

- build operator out of this branch
- create Argo CD instance
- use https://github.com/jaideepr97/gitops-examples/tree/master/bgd-k8s to create an application within argo-cd and sync it
- application should be sync'd and healthy

### DIFF
--- a/controllers/openshift/openshift.go
+++ b/controllers/openshift/openshift.go
@@ -99,6 +99,17 @@ func getPolicyRuleForApplicationController() []rbacv1.PolicyRule {
 				"watch",
 			},
 		},
+		{
+			APIGroups: []string{
+				"monitoring.coreos.com",
+			},
+			Resources: []string{
+				"*",
+			},
+			Verbs: []string{
+				"*",
+			},
+		},
 	}
 }
 


controllers/openshift/openshift.go
```diff
@@ -76,32 +76,14 @@ func reconcilerHook(cr *argoprojv1alpha1.ArgoCD, v interface{}, hint string) err
 				logv.Error(err, "failed to retrieve Cluster Role admin")
 				return err
 			}
-			policyRules := getPolicyRuleForApplicationController()
```
jopit, line 79: This change gives the application controller additional permissions when run with the hooks being active.  Before the change, the verbs were restricted to `get, list, watch`, after the change, the verbs will be `*`.

jaideepr97: @jopit actually, the application-controller already has those permissions today
The application-controller policy rules are actually being defined in 2 different places 
They are first defined [here](https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/policyrule.go#L9) and assigned to the role during creation time. This already gives the application-controller permissions to do everything, and then when the reconcilerHook code is called, this 2nd set of permissions is appended to the existing rules but have no effect since they are a subset of what has already been provided to the role. 

I think whoever wrote this 2nd set of rules might have missed that these permissions had already been defined elsewhere

jaideepr97: according to this set of permissions, the application-controller would not be able to create any of the resources that are required for an application

iam-veeramalla: true

controllers/openshift/openshift.go
```diff
@@ -76,32 +76,14 @@ func reconcilerHook(cr *argoprojv1alpha1.ArgoCD, v interface{}, hint string) err
...
+			policyRules := argocd.PolicyRuleForApplicationController()
 			policyRules = append(policyRules, clusterRole.Rules...)
 			o.Rules = policyRules
 		}
 	}
 	return nil
 }
 
-func getPolicyRuleForApplicationController() []rbacv1.PolicyRule {
```
iam-veeramalla, line 87: ✅

controllers/openshift/openshift.go
chetan-rns: An `admin` in OpenShift doesn't have * permissions for all the resources e.g deleting resource quotas. With these additional privileges, the application controller would be more than an `admin`

chetan-rns: > o.Rules = policyRules

@jaideepr97 I think the hook is overwriting these [permissions](https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/policyrule.go#L9)

jaideepr97: Ah okay, I see what you're saying 
I wonder why those original permissions are even there then because it looks like they'll always get overwritten 
 how does the application-controller create any resources if it doesn't have `create` permissions then ?  :thinking:

jaideepr97: Ah okay, I see what you mean @chetan-rns 
maybe we should be deleting the other set of permissions then, since they'll always get overwritten by this more conservative set of permissions

jaideepr97:
@iam-veeramalla the motivation is mentioned in the description already and there is a JIRA ticket linked
Currently application-controller doesn't have permissions to create prometheusRules on sync (when there is one included within the application) 
The other change regarding `Deep Insight` is just fixing a typo

controllers/openshift/openshift.go
chetan-rns: @jaideepr97 The hooks overwrite the permissions only on OpenShift environments. I think we still need to create the other set of permissions for Kubernetes.

jaideepr97: yeah, @chetan-rns I just followed the code and it looked like  `applyReconciler` was always getting called so I assumed the other set of permissions would always get overwritten
but it just occurred to me that we likely prevent this hook from getting registered at all when we package the argocd-operator

jaideepr97:
hey @iam-veeramalla 
this is still relevant, and not fixed by another other PR afaik

iam-veeramalla (approved):
LGTM, Thanks @jaideepr97

PR #579, opened 2022-03-01
chore: update appset controller

ishitasequeira:
**What type of PR is this?**
kind chore

**What does this PR do / why we need it**: ApplicationSet is merged to ArgoCD repo as a part of v2.3 ([GITOPS-1621](https://issues.redhat.com/browse/GITOPS-1621)). ArgoCD Operator needs to be updated accordingly

**Which issue(s) this PR fixes**: ([GITOPS-1653](https://issues.redhat.com/browse/GITOPS-1653))

Fixes #?

Signed-off-by: ishitasequeira <isequeir@redhat.com>

### DIFF
--- a/controllers/argocd/applicationset.go
+++ b/controllers/argocd/applicationset.go
@@ -76,6 +76,10 @@ func (r *ReconcileArgoCD) reconcileApplicationSetController(cr *argoprojv1a1.Arg
 
 // reconcileApplicationControllerDeployment will ensure the Deployment resource is present for the ArgoCD Application Controller component.
 func (r *ReconcileArgoCD) reconcileApplicationSetDeployment(cr *argoprojv1a1.ArgoCD, sa *corev1.ServiceAccount) error {
+	applicationSetDisabled := isApplicationSetDisabled(cr.Spec.ApplicationSet)
+	if applicationSetDisabled {
+		log.Info("reconciling for application set deployment, but application set is disabled")
+	}
 	deploy := newDeploymentWithSuffix("applicationset-controller", "controller", cr)
 
 	setAppSetLabels(&deploy.ObjectMeta)
@@ -176,6 +180,12 @@ func (r *ReconcileArgoCD) reconcileApplicationSetDeployment(cr *argoprojv1a1.Arg
 	}}
 
 	if existing := newDeploymentWithSuffix("applicationset-controller", "controller", cr); argoutil.IsObjectFound(r.Client, cr.Namespace, existing.Name, existing) {
+		if applicationSetDisabled {
+			if err := r.Client.Delete(context.TODO(), existing); err != nil {
+				return err
+			}
+			return nil
+		}
 
 		existingSpec := existing.Spec.Template.Spec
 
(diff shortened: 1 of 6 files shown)


jaideepr97:
@ishitasequeira could you run `make bundle generate manifests` and push again?

controllers/argocd/applicationset.go
```diff
@@ -176,6 +180,12 @@ func (r *ReconcileArgoCD) reconcileApplicationSetDeployment(cr *argoprojv1a1.Arg
 	}}
 
 	if existing := newDeploymentWithSuffix("applicationset-controller", "controller", cr); argoutil.IsObjectFound(r.Client, cr.Namespace, existing.Name, existing) {
+		if applicationSetDisabled {
+			if err := r.Client.Delete(context.TODO(), existing); err != nil {
```
jaideepr97, line 184: can we have a `log.Info` here that mentions that the applicationSet deployment is being deleted because applicationSet was disabled?

controllers/argocd/applicationset.go
```diff
@@ -224,9 +241,18 @@ func (r *ReconcileArgoCD) reconcileApplicationSetServiceAccount(cr *argoprojv1a1
 	}
 
 	if exists {
+		if applicationSetDisabled {
+			if err := r.Client.Delete(context.TODO(), sa); err != nil {
```
jaideepr97, line 245: can we have a `log.Info` here that mentions that the applicationSet service account is being deleted because applicationSet was disabled?

controllers/argocd/applicationset.go
```diff
@@ -328,6 +356,12 @@ func (r *ReconcileArgoCD) reconcileApplicationSetRole(cr *argoprojv1a1.ArgoCD) (
 		if !errors.IsNotFound(err) {
 			return nil, fmt.Errorf("failed to reconcile the role for the service account associated with %s : %s", role.Name, err)
 		}
+		if applicationSetDisabled {
+			if err := r.Client.Delete(context.TODO(), role); err != nil {
```
jaideepr97, line 360: can we have a `log.Info` here that mentions that the applicationSet role is being deleted because applicationSet was disabled?

controllers/argocd/applicationset.go
```diff
@@ -378,8 +416,17 @@ func (r *ReconcileArgoCD) reconcileApplicationSetRoleBinding(cr *argoprojv1a1.Ar
 	}
 
 	if roleBindingExists {
+		if applicationSetDisabled {
+			if err := r.Client.Delete(context.TODO(), roleBinding); err != nil {
```
jaideepr97, line 420: can we have a `log.Info` here that mentions that the applicationSet roleBinding is being deleted because applicationSet was disabled?

wtam2018:
@ishitasequeira is "disabled" the only thing needed? I don't think we need to add this flag.

wtam2018:
Instead of adding a disabled flag, I think we should pull in a v2.3 (rc) and install/run it to see what exactly need to be changed.  You are sill installing v2.2, right?

controllers/argocd/applicationset.go
```diff
@@ -206,12 +216,19 @@ func (r *ReconcileArgoCD) reconcileApplicationSetDeployment(cr *argoprojv1a1.Arg
 	if err := controllerutil.SetControllerReference(cr, deploy, r.Scheme); err != nil {
 		return err
 	}
```
jannfis, line 218: We probably should do that also only if ApplicationSet is not disabled. It just makes no sense to call `SetControllerReference` if we're not going to reconcile anyway.

jannfis (commented):
Just a nit. Otherwise, looks fine!

wtam2018 (changes requested):
Let's rethink if it makes sense to add "disabled" flag to applicationset. It does not provide really benefit to the users. It only makes the code more complicated and harder to maintain. None of the core servers (e.g. application controller, repo server) has disabled flag. Also, we are talking about tighter integration of applicationset to argocd controller upstream.

PR #582, opened 2022-03-04
fix: Set subject on CA cert used for signing self-signed certs

jannfis:

**What type of PR is this?**

/kind bug

See https://github.com/redhat-developer/gitops-operator/issues/277

Recent versions of Firefox refuse to process TLS certificates that are signed by an unnamed CA. The Operator uses a CA certificate that's created without a X509 `Subject`, and thus, Firefox chokes.

For an existing installation, the `argocd-tls` and `argocd-ca` secrets must be deleted and reconciled back by the Operator.

Fixes https://github.com/redhat-developer/gitops-operator/issues/277

* Install a recent version of Firefox (I used 97.0.1 (64-bit), installed via Flatpak)
* Create simple ArgoCD for the Operator to reconcile and to create a new Argo CD instance. If you use `route`, make sure that `.server.route.tls.termination` is set to `passthrough` (or left blank, because `passthrough` is the default)
* Point Firefox to the Argo CD UI
* Connection should be successful (the unknown certificate message is expected, but should be skippable)

### DIFF
--- a/controllers/argocd/secret.go
+++ b/controllers/argocd/secret.go
@@ -92,7 +92,7 @@ func newCASecret(cr *argoprojv1a1.ArgoCD) (*corev1.Secret, error) {
 		return nil, err
 	}
 
-	cert, err := argoutil.NewSelfSignedCACertificate(key)
+	cert, err := argoutil.NewSelfSignedCACertificate(cr.Name, key)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


wtam2018 (approved):

wtam2018:
LGTM
thanks @jannfis

PR #588, opened 2022-03-17
bug: update workflow to avoid deleting Secrets while reconciling cluster permissions

ishitasequeira:
**What type of PR is this?**
   /kind bug

While reconciling cluster permissions, every time the Secret was deleted if the secret belongs to a cluster config instance.
In this PR, we are updating the logic to not delete the Secret, but to remove the `namespaces` field from the Secret.

Fixes [GITOPS-1777](https://issues.redhat.com/browse/GITOPS-1777)

* build operator out of this branch
* create Argo CD instance
* Enable route for Argo CD instance to be able to access UI
* Log in to Argo CD UI -> Settings -> Clusters, edit the existing cluster to see the Secret getting created
* Update ArgoCD instance to add a label `food: pizza` and sync it
* See that the Secret does not get deleted automatically


jopit (approved):
LGTM

PR #597, opened 2022-03-28
Fix lint errors

chetan-rns:
**What type of PR is this?**
 /kind cleanup

Fix all lint errors that were reported by `golangci-lint run`

`golangci-lint run` should not report any lint errors

### DIFF
--- a/controllers/argocd/hooks.go
+++ b/controllers/argocd/hooks.go
@@ -11,9 +11,8 @@ var (
 	hooks = []Hook{}
 )
 
-// Hook changes resources as they are created or updated by the
-// reconciler.
-type Hook func(*argoprojv1alpha1.ArgoCD, interface{}) error
+// Hook changes resources as they are created or updated by the reconciler.
+type Hook func(*argoprojv1alpha1.ArgoCD, interface{}, string) error
 
 // Register adds a modifier for updating resources during reconciliation.
 func Register(h ...Hook) {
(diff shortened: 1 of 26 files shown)


jannfis:
This does not seem to only remove simple linter errors.

chetan-rns:
> This does not seem to only remove simple linter errors.

@jannfis It includes all the errors reported by `golangci-lint run`

controllers/argocd/hooks.go
```diff
@@ -13,7 +13,7 @@ var (
 
 // Hook changes resources as they are created or updated by the
 // reconciler.
-type Hook func(*argoprojv1alpha1.ArgoCD, interface{}, string) error
```
jopit, line 16: The string parameter shouldn't be removed; the `Hook` is an api that external applications can use and the string is a hint that may be used in the future to disambiguate the context of a hook call.

jannfis: You can use something like `// nolint:<category>` to suppress lint issues for specific targets. I believe this may have been a check in the `unparam` catagory, so adding a `// nolint:unparam` comment just above the type definition should silent the linter.

controllers/argocd/hooks.go
```diff
@@ -22,6 +21,7 @@ func Register(h ...Hook) {
 	hooks = append(hooks, h...)
 }
 
+// nolint:unparam
```
iam-veeramalla, line 24: @chetan-rns I am slightly unsure why this is used. Can you please explain me ?

chetan-rns: Currently, golangci-lint throws an error because the `hint` parameter is not used anywhere. The comment will suppress the lint error since we want to keep the parameter. https://github.com/argoproj-labs/argocd-operator/pull/597#discussion_r837152358

iam-veeramalla (approved):
LGTM, Thanks @chetan-rns

PR #602, opened 2022-04-05
Fix docs to say host instead of URL

### LINKED ISSUE #601: Docs incorrectly use .url instead of .host
**Describe the bug**
This bug was created by me, I'll make a PR shortly to update it. 
In https://github.com/argoproj-labs/argocd-operator/blob/master/docs/reference/api.html.md#argocdstatus, it states that "URL is the url for the hostname to use for Ingress/Route resources" but it was renamed to be `Host` instead of `URL`. Docs need to be updated to reflect this.

reginapizza:
**What type of PR is this?**
 /kind bug

Docs incorrectly use URL instead of host, this will fix it. 

Fixes #601 

N/A

### DIFF
--- a/docs/reference/api.html.md
+++ b/docs/reference/api.html.md
@@ -2216,13 +2216,13 @@ Unknown: For some reason the state of the Argo CD server component could not be 
 </tr>
 <tr>
 <td>
-<code>url</code></br>
+<code>host</code></br>
 <em>
 string
 </em>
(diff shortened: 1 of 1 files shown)


iam-veeramalla:
@reginapizza I have re-triggered the failed jobs.

iam-veeramalla (approved):
LGTM

PR #604, opened 2022-04-08
fix: delete broken pods of statefulSets

saumeya:

**What type of PR is this?**
/kind bug

There is a known issue in Kubernetes (https://github.com/kubernetes/kubernetes/issues/67250) where, if the pods created by StatefulSets fail then they don't get restarted on their own unless manually deleted. This PR adds the deletion in the operator itself, so that it can fetch such a broken pod and restart it.

Fixes #?

1. Run the operator using `make intall` and `make run`
2. In the Application Controller StatefulSet, update the image to any broken url, this will recreate the bug and the `application-controller` pod will show `ImagePullBackOff` or `ErrImagePull`
3. Verify that in a few seconds the pod will be recreated as the StatefulSet reconciler update the image to the correct path

### DIFF
--- a/controllers/argocd/statefulset.go
+++ b/controllers/argocd/statefulset.go
@@ -24,6 +24,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	argoprojv1a1 "github.com/argoproj-labs/argocd-operator/api/v1alpha1"
@@ -495,7 +496,7 @@ func (r *ReconcileArgoCD) reconcileApplicationControllerStatefulSet(cr *argoproj
 		podSpec.Volumes = getArgoImportVolumes(export)
 	}
 
-	invalidImagePod := containsInvalidImage(ss, cr)
+	invalidImagePod := containsInvalidImage(ss, cr, r)
 	if invalidImagePod {
 		if err := r.Client.Delete(context.TODO(), ss); err != nil {
 			return err
@@ -550,11 +551,13 @@ func (r *ReconcileArgoCD) reconcileApplicationControllerStatefulSet(cr *argoproj
 		}
 		return nil // StatefulSet found with nothing to do, move along...
 	}
-
+	// pod := &corev1.PodList{}
+	// fmt.Println("PODSSS ", r.Client.List(context.TODO(), pod))
 	// Delete existing deployment for Application Controller, if any ..
 	deploy := newDeploymentWithSuffix("application-controller", "application-controller", cr)
 	if argoutil.IsObjectFound(r.Client, deploy.Namespace, deploy.Name, deploy) {
 		if err := r.Client.Delete(context.TODO(), deploy); err != nil {
+
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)


controllers/argocd/statefulset.go
```diff
@@ -591,3 +596,24 @@ func updateNodePlacementStateful(existing *appsv1.StatefulSet, ss *appsv1.Statef
...
+		log.Error(err, "failed to initialize k8s client")
+	}
+
+	labelSel := common.ArgoCDKeyName + " = " + fmt.Sprintf("%s-%s", cr.Name, "application-controller")
+	po, err := k8s.CoreV1().Pods(ss.Namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: labelSel})
+	if err != nil {
+		log.Error(err, "failed to get Pod List")
+	}
+
+	if po.Items[0].Status.ContainerStatuses[0].State.Waiting != nil && (po.Items[0].Status.ContainerStatuses[0].State.Waiting.Reason == "ImagePullBackOff" || po.Items[0].Status.ContainerStatuses[0].State.Waiting.Reason == "ErrImagePull") {
```
jannfis, line 615: Please revisit this piece of code under the following premises:

* What happens when there are no pods?
* What happens when the pod has no container status information?
* What if there are multiple replicas, and more than one are in failure state?

Also, please investigate if the error conditions `ImagePullBackOff` and `ErrImagePull` have constants in the Kubernetes API.

controllers/argocd/statefulset.go
```diff
@@ -495,6 +495,11 @@ func (r *ReconcileArgoCD) reconcileApplicationControllerStatefulSet(cr *argoproj
 		podSpec.Volumes = getArgoImportVolumes(export)
 	}
 
+	brokenPod := getBrokenPod(ss, cr)
+	if brokenPod != nil {
+		r.Client.Delete(context.TODO(), brokenPod)
```
jannfis, line 500: Please do check value of err, and if not nil, log out an error.

controllers/argocd/statefulset.go
```diff
@@ -591,3 +596,24 @@ func updateNodePlacementStateful(existing *appsv1.StatefulSet, ss *appsv1.Statef
...
+
+	k8s, err := initK8sClient()
+	if err != nil {
+		log.Error(err, "failed to initialize k8s client")
+	}
+
+	labelSel := common.ArgoCDKeyName + " = " + fmt.Sprintf("%s-%s", cr.Name, "application-controller")
+	po, err := k8s.CoreV1().Pods(ss.Namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: labelSel})
+	if err != nil {
+		log.Error(err, "failed to get Pod List")
```
jannfis, line 612: Should be: "Failed to list pods", and the error reason should be logged as well.

controllers/argocd/statefulset.go
iam-veeramalla: @jannfis agree, We need to handle the cases like sharding where controller can have multiple replicas.

iam-veeramalla: @jannfis @saumeya I would delete the statefulset itself instead of pod. It helps in scenarios where there are multiple pods.

if one of the pods is in `ImagePullBackoff` state the other pod would also be in the same state. So by deleting the stateful set operator would recreate the stateful set with new image.

saumeya: > Please revisit this piece of code under the following premises:
> 
> * What happens when there are no pods?
> * What happens when the pod has no container status information?
> * What if there are multiple replicas, and more than one are in failure state?
> 
> Also, please investigate if the error conditions `ImagePullBackOff` and `ErrImagePull` have constants in the Kubernetes API.

Thanks Jann, I'll search and work on these points.

saumeya: > @jannfis @saumeya I would delete the statefulset itself instead of pod. It helps in scenarios where there are multiple pods.
> 
> if one of the pods is in `ImagePullBackoff` state the other pod would also be in the same state. So by deleting the stateful set operator would recreate the stateful set with new image.

Makes sense, I'll try adding this to the logic

saumeya: I found no constants for the error conditions for the image wait reasons, the type is only **string** in the api. Found this on SO but didn't really point to any constants - https://stackoverflow.com/questions/57821723/list-of-all-reasons-for-container-states-in-kubernetes

controllers/argocd/statefulset.go
saumeya: I am logging out 'err', is there some other error reason that I should add?

controllers/argocd/statefulset.go
```diff
@@ -591,3 +598,31 @@ func updateNodePlacementStateful(existing *appsv1.StatefulSet, ss *appsv1.Statef
 		*changed = true
 	}
 }
+
+//to fetch the broken pod of StatefulSet if any, this function intends to delete the broken pod that cannot
```
iam-veeramalla, line 602: Please correct the comments as per the standards
https://go.dev/blog/godoc

I would suggest something like this.
// Returns true if a StatefulSet has  pods in ErrImagePull or ImagePullBackoff state.

controllers/argocd/statefulset.go
```diff
@@ -591,3 +598,31 @@ func updateNodePlacementStateful(existing *appsv1.StatefulSet, ss *appsv1.Statef
...
+	}
+	brokenPod := false
+	labelSel := common.ArgoCDKeyName + " = " + fmt.Sprintf("%s-%s", cr.Name, "application-controller")
+	po, err := k8s.CoreV1().Pods(ss.Namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: labelSel})
+	if err != nil {
+		log.Error(err, "Failed to list Pods")
+	}
+
+	if len(po.Items) > 0 {
+		for _, p := range po.Items {
```
iam-veeramalla, line 618: Do you think we need to iterate through the list of pods and break on the first pod in ImagePullBackOff state ? Can we instead verify this condition on the first pod as the behavior should be same for all the pods.

controllers/argocd/statefulset_test.go
```diff
@@ -364,3 +364,10 @@ func Test_UpdateNodePlacementStateful(t *testing.T) {
 		t.Fatalf("updateNodePlacement failed, value of changed: %t", actualChange)
 	}
 }
+func Test_ContainsBrokenPod(t *testing.T) {
+	a := makeTestArgoCD()
+	ss := &appsv1.StatefulSet{}
+	if containsBrokenPod(ss, a) {
```
iam-veeramalla, line 370: we are passing an empty statefulset to find brokenPods ? Can you add more tests ?

controllers/argocd/statefulset.go
saumeya: thanks, will update

controllers/argocd/statefulset.go
saumeya: Ok, in that case I'll remove the for loop and break, but we will still need the two if conditions

controllers/argocd/statefulset_test.go
saumeya: Yes, I'm still not sure how to test it properly because, the function fetches the podStatus, and since no pod exists it will return false always. So just added empty values.

iam-veeramalla (approved):
LGTM, However, I notice that operator reconciles in a loop as every time the statefulset pod goes into `ImagePullBackoff` state, operator deletes the pod and this goes on until the image is available or corrected. 

@saumeya please make sure to remove this code as soon as the upstream kubernetes issue is fixed.
https://github.com/kubernetes/kubernetes/issues/67250

saumeya:
> LGTM, However, I notice that operator reconciles in a loop as every time the statefulset pod goes into `ImagePullBackoff` state, operator deletes the pod and this goes on until the image is available or corrected.
> 
> @saumeya please make sure to remove this code as soon as the upstream kubernetes issue is fixed. [kubernetes/kubernetes#67250](https://github.com/kubernetes/kubernetes/issues/67250)

Thanks @iam-veeramalla. Will keep a watch on that issue.

PR #613, opened 2022-04-12
chore:[release-0.3] update argocd to v2.3.3 in Dockerfile and defaults.go

ishitasequeira:
**What type of PR is this?**
/kind chore

chore:[release-0.3] update argocd to v2.3.3 in Dockerfile and defaults.go
Added missing 0.2.1 crds

Fixes #?

### DIFF
--- a/build/util/Dockerfile
+++ b/build/util/Dockerfile
@@ -1,11 +1,11 @@
-# Argo CD v2.2.2
-FROM quay.io/argoproj/argocd@sha256:358c244c96313ca3bf9f588dc870d8123fc22ffa5c231c57da10f77b8d671c66
+# Argo CD v2.3.3
+FROM quay.io/argoproj/argocd@sha256:dd738f234fcdb0aac8631a0fd1aafbbcd86f936480b06e8377b033ef7a764f71
 
 USER root
 
 # Ensure system dependencies are installed
 RUN apt-get update && \
-    apt-get install -y curl && \
+    apt-get install -y curl python3-pip && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Install the Google Cloud SDK (CLI)
--- a/deploy/olm-catalog/argocd-operator/0.3.0/argoproj.io_applicationsets.yaml
+++ b/deploy/olm-catalog/argocd-operator/0.3.0/argoproj.io_applicationsets.yaml
@@ -1,3 +1,5 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -58,7 +58,7 @@ const (
 	ArgoCDDefaultArgoImage = "quay.io/argoproj/argocd"
 
 	// ArgoCDDefaultArgoVersion is the Argo CD container image digest to use when version not specified.
-	ArgoCDDefaultArgoVersion = "sha256:bac1aeee8e78e64d81a633b9f64148274abfa003165544354e2ebf1335b6ee73" // v2.2.5
+	ArgoCDDefaultArgoVersion = "sha256:dd738f234fcdb0aac8631a0fd1aafbbcd86f936480b06e8377b033ef7a764f71" // v2.3.3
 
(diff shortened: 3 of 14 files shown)


ishitasequeira:
@jopit @iam-veeramalla @wtam2018 PTAL.

build/util/Dockerfile
```diff
@@ -1,11 +1,11 @@
...
-FROM quay.io/argoproj/argocd@sha256:358c244c96313ca3bf9f588dc870d8123fc22ffa5c231c57da10f77b8d671c66
+# Argo CD v2.3.3
+FROM quay.io/argoproj/argocd@sha256:dd738f234fcdb0aac8631a0fd1aafbbcd86f936480b06e8377b033ef7a764f71
 
 USER root
 
 # Ensure system dependencies are installed
 RUN apt-get update && \
-    apt-get install -y curl && \
+    apt-get install -y curl python3-pip && \
```
wtam2018, line 8: Please check to see if this is needed.

ishitasequeira: ArgoCD removed python from their image in 2.3, thus we need to add it here.

wtam2018:
LGTM

build/util/Dockerfile
iam-veeramalla, line 8: Python dependency is required to build the util AFAIR. Can you try building the util image remving the python dependency and verify ?

ishitasequeira: `make util-build` fails after removing the python dependency

wtam2018: I remember I raised this question with @iam-veeramalla . He was able to build without the python3 dependency. It would be nice to pull in just what we need. If we indeed need it, then obviously we should have it. What is error that you see after you remove it?

ishitasequeira:
```/home/argocd/google-cloud-sdk/install.sh
Welcome to the Google Cloud CLI!
WARNING: You appear to be running this script as root. This may cause 
the installation to be inaccessible to users other than the root user.
/home/argocd/google-cloud-sdk/install.sh: 210: python: not found
The command '/bin/sh -c curl -sL https://sdk.cloud.google.com/ > /tmp/install.sh &&     bash /tmp/install.sh --disable-prompts --install-dir=/home/argocd &&     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*' returned a non-zero code: 127
make: *** [util-build] Error 127```

**Review by iam-veeramalla — APPROVED** [review]
LGTM , Thanks @ishitasequeira

File: deploy/olm-catalog/argocd-operator/0.3.0/argoproj.io_applications.yaml [hunk 2]
```diff
@@ -1,7 +1,6 @@
 apiVersion: apiextensions.k8s.io/v1
```
jopit, line 1: The changes to this file should be reverted.  They are added by the `make bundle` step (not sure why, but they are in the 0.1 and 0.2 releases as well), which will be run as part of the release process.

deploy/olm-catalog/argocd-operator/0.3.0/argoproj.io_applicationsets.yaml
```diff
@@ -1,3 +1,5 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1
```
jopit, line 3: The changes to this file should be reverted. They are added by the make bundle step (not sure why, but they are in the 0.1 and 0.2 releases as well). I believe these particular changes may cause the CI tests to fail when we try to release on operator hub

deploy/olm-catalog/argocd-operator/0.3.0/argoproj.io_appprojects.yaml
```diff
@@ -1,7 +1,6 @@
 apiVersion: apiextensions.k8s.io/v1
```
jopit, line 1: The changes to this file should be reverted. They are added by the make bundle step (not sure why, but they are in the 0.1 and 0.2 releases as well), which will be run as part of the release process.

jopit (changes requested):

PR #615, opened 2022-04-13
Added probes to Dex container

sabre1041:
Enables probes on dex container which ensures instance is running properly.

It has been observed in some circumstances that dex becomes unhealthy and requires manual deletion of the pod to bring it back into working order

### DIFF
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -103,6 +103,9 @@ const (
 	// ArgoCDDefaultDexHTTPPort is the default HTTP listen port for Dex.
 	ArgoCDDefaultDexHTTPPort = 5556
 
+	// ArgoCDDefaultDexMetricsPort is the default Metrics listen port for Dex.
+	ArgoCDDefaultDexMetricsPort = 5558
+
 	// ArgoCDDefaultDexServiceAccountName is the default Service Account name for the Dex server.
 	ArgoCDDefaultDexServiceAccountName = "argocd-dex-server"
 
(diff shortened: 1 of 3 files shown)


christianh814 (commented):
LGTM!

iam-veeramalla:
Changes looks good to me, Investigating on the failed test - `ReadyReplicas for Dex is not 1`

iam-veeramalla:
I could reproduce the failing E2E for dex on my local as well.
```
status:
  conditions:
  - lastTransitionTime: "2022-04-13T17:56:50Z"
    lastUpdateTime: "2022-04-13T17:56:50Z"
    message: Deployment does not have minimum availability.
    reason: MinimumReplicasUnavailable
    status: "False"
    type: Available
```


Steps to Reproduce:
Checkout the PR
Run the operator using `make install run`
Create a basic Argo CD and let operator to reconcile dex and other pods.
```
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: basic
spec: {}
```

@sabre1041 Were you able to get the dex pod running with these changes ?

sabre1041:
@iam-veeramalla figured out the issue. needed to remove the Readiness Probe and extend the Liveness Probe. Updated PR and tests are now passing

iam-veeramalla:
> 

Looking at the problem description adding only `Liveness Probe` makes sense. But I think we might also need Readiness Probe. 
Reference from Argo CD Helm Repo.
https://github.com/argoproj/argo-helm/blob/58926e3d790f7c4c42ea082269cb921fb9afaeb5/charts/argo-cd/templates/dex/deployment.yaml#L92

iam-veeramalla (approved):
LGTM, Thanks @sabre1041

PR #624, opened 2022-04-22
chore: update the go version and dependencies

ishitasequeira:
**What type of PR is this?**
 /kind chore

Upgrade go version to 1.18 and go.mod dependencies

Fixes #?

Execute below commands in different terminal windows:
* make install run
* make test


api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
ishitasequeira, line 1: `make generate` automatically seems to be adding this now. If we remove this line, the ci fails. Not sure if there is a workaround for this.

jopit: It's actually the `go fmt` command that's doing this as of golang 1.17: https://stackoverflow.com/questions/10646531/golang-conditional-compilation/67937234#67937234

If we're moving to golang 1.18, we shouldn't need to work around this.

sagikazarmark (approved):
Tested the changes to the toolchain locally. Works for me on Go 1.18

iam-veeramalla:
@ishitasequeira there are many manifests file changes introduced by this PR. Are all of them expected ? Can you please re-verify  ?

ishitasequeira:
> @ishitasequeira there are many manifests file changes introduced by this PR. Are all of them expected ? Can you please re-verify ?

@iam-veeramalla, the manifests are auto-updated after executing `make bundle`.

jannfis:
> the manifests are auto-updated after executing

I think this is due to the update of K8s dependencies.

docs/reference/argocd.md
```diff
@@ -44,7 +44,7 @@ Name | Default | Description
 [**StatusBadgeEnabled**](#status-badge-enabled) | `true` | Enable application status badge feature.
 [**TLS**](#tls-options) | [Object] | TLS configuration options.
 [**UsersAnonymousEnabled**](#users-anonymous-enabled) | `true` | Enable anonymous user access.
-[**Version**](#version) | v2.2.2 (SHA) | The tag to use with the container image for all Argo CD components.
```
jannfis, line 47: The version bump of the image is not part of this change. Has this been done before, and this is just docs update?

ishitasequeira: Yes, We changed the container image while releasing 0.3.0.

api/v1alpha1/zz_generated.deepcopy.go
jaideepr97: I find that this happens when I use vscode
whenever I need to get rid of this I edit the file in the native fedora editor and that sticks

jaideepr97:
LGTM, thanks @ishitasequeira !

rishabh625:
It LGTM except that docs could be changed in sperate PR , @jannfis : can u plz take a look ?

jannfis (approved):
LGTM

PR #631, opened 2022-05-02
update deprecated cronjob reference

ishitasequeira:
Signed-off-by: ishitasequeira <ishiseq29@gmail.com>

**What type of PR is this?**
/kind chore

**What does this PR do / why we need it**:
CronJob has been deprecated in v1.21+, and unavailable in v1.25+.
References of CronJob were updated to use "batch/v1" in PR #477 but couple of places were missed out.
Updating missed places to use "batch/v1" for CronJob.

**Have you updated the necessary documentation?**

* [ ] Documentation update is required by this PR.
* [ ] Documentation has been updated.

**Which issue(s) this PR fixes**:

Fixes #?
https://issues.redhat.com/browse/GITOPS-1940

**How to test changes / Special notes to the reviewer**:
- make install run
- make test
All tests should pass successfully.

PR #635, opened 2022-05-04
feat: add support for creating argocd notifications workloads 

jaideepr97:
**What type of PR is this?**

> /kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
/kind enhancement
> /kind documentation
> /kind code-refactoring

This is 1/2 PRs for adding support for spinning up the notifications-controller within argocd-operator. This PR adds the `.spec.Notifications` section within the Argo CD CR, which users can use to enable/disable the creation of argocd-notifications workload. It also adds a field within the Argo CD `status` to reflect the live state of the notifications controller. This PR creates the required deployment, serviceAccount, role and roleBinding required to have the controller brought up on the cluster successfully, and adds the reconciliation logic for these resoureces along with cleaning up the resources when notifications is disabled 

https://github.com/argoproj-labs/argocd-operator/issues/654

1. Start the operator locally and spin up an empty Argo CD instance
2. Notice that out of the box Argo CD instance has `.spec.notifications.enabled` set to false and `.status` doesn't reflect 
anything about the notifications controller since it is currently disabled
3. Set `.spec.notifications.enabled` to `true`
4. Verify that deployment, sa, role and rolebinding for notifications-controller are created
5. Verify that `.status.notificationsController` is `running` after deployment has come up
6. Disable notifications by setting `.spec.notifications.enabled` to false
7. Verify that deployment, sa, role and roleBinding are properly cleaned up, and `.status` once again does not mention notifications controller


controllers/argocd/util.go
```diff
@@ -974,8 +981,31 @@ func setResourceWatches(bldr *builder.Builder, clusterResourceMapper, tlsSecretM
...
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			newCR, ok := e.ObjectNew.(*argoprojv1a1.ArgoCD)
+			if !ok {
+				return false
+			}
+			oldCR, ok := e.ObjectOld.(*argoprojv1a1.ArgoCD)
+			if !ok {
+				return false
+			}
+			if (oldCR.Spec.Notifications.Enabled != newCR.Spec.Notifications.Enabled) && !newCR.Spec.Notifications.Enabled {
```
ishitasequeira, line 996: I guess this condition can be simplified to `oldCR.Spec.Notifications.Enabled && !newCR.Spec.Notifications.Enabled`. WDYT?

openshift.go
```diff
@@ -1,3 +1,4 @@
+//go:build openshift
```
ishitasequeira, line 1: Should we revert this? Not sure if we need it.

controllers/argocdexport/reconcile.go
```diff
@@ -46,7 +45,7 @@ func setResourceWatches(bld *builder.Builder) *builder.Builder {
 	bld.For(&argoproj.ArgoCDExport{})
 
 	// Watch for changes to CronJob sub-resources owned by ArgoCDExport instances.
-	bld.Owns(&batchv1b1.CronJob{})
+	bld.Owns(&batchv1.CronJob{})
```
ishitasequeira, line 48: There is already a PR [#631](https://github.com/argoproj-labs/argocd-operator/pull/631) merged for this. Can we rebase the PR with master?

controllers/argocd/rolebinding_test.go
```diff
@@ -6,12 +6,13 @@ import (
...
 	"testing"
 
-	"github.com/argoproj-labs/argocd-operator/common"
 	"github.com/stretchr/testify/assert"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/argoproj-labs/argocd-operator/common"
```
ishitasequeira, line 15: I can see this is re-ordered at bunch of places. Any specific reason to change this ordering? or was this auto-updated?

controllers/argocd/notifications.go
```diff
@@ -0,0 +1,428 @@
...
+		return nil
+	}
+
+	// roleBinding does not exist but should, so it should be created
+	if err := controllerutil.SetControllerReference(cr, desiredRoleBinding, r.Scheme); err != nil {
+		return err
+	}
+
+	log.Info(fmt.Sprintf("Creating role %s", desiredRoleBinding.Name))
+	return r.Client.Create(context.TODO(), desiredRoleBinding)
```
ishitasequeira, line 246: Same as above comment, move to line 209?

controllers/argocd/notifications.go
```diff
@@ -0,0 +1,428 @@
...
+	// role does not exist but should, so it should be created
+	if err := controllerutil.SetControllerReference(cr, desiredRole, r.Scheme); err != nil {
+		return nil, err
+	}
+
+	log.Info(fmt.Sprintf("Creating role %s", desiredRole.Name))
+	err := r.Client.Create(context.TODO(), desiredRole)
+	if err != nil {
+		return nil, err
+	}
```
ishitasequeira, line 180: Same here too.. Can we move this to line 146?

controllers/argocd/notifications.go
```diff
@@ -0,0 +1,428 @@
...
+		return nil
+	}
+
+	// deployment does not exist but should, so it should be created
+	if err := controllerutil.SetControllerReference(cr, desiredDeployment, r.Scheme); err != nil {
+		return err
+	}
+
+	log.Info(fmt.Sprintf("Creating deployment %s", desiredDeployment.Name))
+	return r.Client.Create(context.TODO(), desiredDeployment)
```
ishitasequeira, line 406: Same here too.. Can we move this to line 321?

controllers/argocd/notifications.go
```diff
@@ -0,0 +1,428 @@
...
+	// SA doesn't exist but should, so it should be created
+	if err := controllerutil.SetControllerReference(cr, sa, r.Scheme); err != nil {
+		return nil, err
+	}
+
+	log.Info(fmt.Sprintf("Creating serviceaccount %s", sa.Name))
+	err := r.Client.Create(context.TODO(), sa)
+	if err != nil {
+		return nil, err
+	}
```
ishitasequeira, line 130: Can we move this inside `FetchObject` call before updating value of `saExists = false` (line 104) and update the workflow accordingly?

ishitasequeira (changes requested):
Thanks @jaideepr97. Went through the code and added some comments on the workflow. Let me know WDYT.

controllers/argocd/util.go
jaideepr97: will do :+1:

controllers/argocd/notifications.go
jaideepr97: I'm not sure I follow
what would the benefit of doing that be?

ishitasequeira: It is more like, if fetching sa object errors out, we know that we do not have an existing sa. So we just create it there itself and return rather than performing other set of checks.

Also, we have followed similar approach everywhere in the repo.

jaideepr97: If you're referring to the workflow that is followed in reconciling the default resources (like server, repo-server etc) maybe in that case a check is not required because we need those resources to exist on the cluster under all circumstances as long as the owner Argo CD instance is around
but notifications resources are optional, so I think we should be more deliberate about when they should be created and make sure they are cleaned up when not required

ishitasequeira:
Just a few nitpicks which could be ignored. It works the way it has been mentioned in the **How to test**. 
LGTM!! 
@jopit PTAL

controllers/argocd/status.go
```diff
@@ -222,6 +226,30 @@ func (r *ReconcileArgoCD) reconcileStatusServer(cr *argoprojv1a1.ArgoCD) error {
 	return nil
 }
 
+// reconcileStatusNotifications will ensure that the Notifications status is updated for the given ArgoCD.
+func (r *ReconcileArgoCD) reconcileStatusNotifications(cr *argoprojv1a1.ArgoCD) error {
+	status := "Unknown"
+
+	deploy := newDeploymentWithSuffix("notifications-controller", "controller", cr)
+	if argoutil.IsObjectFound(r.Client, cr.Namespace, deploy.Name, deploy) {
+		status = "Running"
```
jopit, line 235: When the deployment is found, the other status reconciler methods only set the status to "Running" when the `deploy.Status.ReadyReplicas` is the same as `deploy.Spec.Replicas`, otherwise they set it to "Pending" -- is there a reason not to do the same for the status of the notificationsController?

jopit (changes requested):

controllers/argocd/notifications.go
```diff
@@ -0,0 +1,428 @@
...
+		}
+
+		// roleBinding exists and should. Reconcile roleBinding if changed
+		if !reflect.DeepEqual(existingRoleBinding.RoleRef, desiredRoleBinding.RoleRef) {
+			// if the RoleRef changes, delete the existing role binding and create a new one
+			if err := r.Client.Delete(context.TODO(), existingRoleBinding); err != nil {
+				return err
+			}
+		}
+		if !reflect.DeepEqual(existingRoleBinding.Subjects, desiredRoleBinding.Subjects) {
```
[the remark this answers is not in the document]
jaideepr97: Good catch, thanks @jopit !

controllers/argocd/status.go
jaideepr97: @jopit I wasn't quite sure if there was a use case for us to support multiple replicas of the notifications-controller at the time
However I realize now these are 2 different checks, so I will update the logic here (which should default to 1 replica within the deployment, unless we decide to allow users to be able to express the replica count for the notifications-controller through the CR at some point)
Thanks for catching this!

jannfis (commented):
Just some comments. Great PR @jaideepr97

controllers/argocd/policyrule.go
```diff
@@ -147,6 +147,69 @@ func policyRuleForServer() []v1.PolicyRule {
...
+			},
+			Resources: []string{
+				"applications",
+				"appprojects",
+			},
+			Verbs: []string{
+				"get",
+				"list",
+				"patch",
+				"update",
```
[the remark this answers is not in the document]
jaideepr97: @jannfis not too sure, I used the upstream install manifest as reference to populate the resources
https://github.com/argoproj/argo-cd/blob/master/manifests/install.yaml#L9180

controllers/argocd/notifications.go
```diff
@@ -0,0 +1,428 @@
...
+		VolumeMounts: []corev1.VolumeMount{
+			{
+				Name:      "tls-certs",
+				MountPath: "/app/config/tls",
+			},
+			{
+				Name:      "argocd-repo-server-tls",
+				MountPath: "/app/config/reposerver/tls",
+			},
+		},
```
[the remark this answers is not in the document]
jaideepr97: @jannfis not too sure, I used the upstream install manifest as reference to populate the resources
https://github.com/argoproj/argo-cd/blob/master/manifests/install.yaml#L9848

jannfis (approved):
LGTM.

@jopit @ishitasequeira You have requested changes, thus, PR requires your formal approval too.

ishitasequeira (approved):
LGTM

PR #650, opened 2022-05-13
feat: added env variable to remove managed-by label on argocd instance deletion

ciiay:
**What type of PR is this?**
/kind enhancement

This PR introduces a new environment variable "REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION", and changes the default behave on namespace with "managed-by" label when the target Argo CD instance gets removed by the gitops-operator. By default, when the Argo CD instance with the target namespace gets deleted, the namespace keeps "managed-by" label value. Only when "REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION" is set to "true" in the Subscription config, it will remove the "managed-by" label on Argo CD deletion.

Fixes https://issues.redhat.com/browse/GITOPS-1826

> Without "REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION" set to "true" in argocd-operator Subscription
1. Install argocd-operator with namespace A and create an Argo CD instance
2. Create a new namespace B and set "argocd.argoproj.io/managed-by: <namespace A>" label
3. Delete the Argo CD instance with namespace A
4. Check the new namespace B and see if it still has the "managed-by" label which sets to namespace A

> With "REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION" set to "true" in argocd-operator Subscription
1. Same as 1 from above process
2. Edit Subscription of argocd-operator as below
```
...
spec:
...
  config:
    env:
      - name: REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION
        value: 'true'
...
```
3. Same as 2 -3 from above process
4. Check the new namespace B and see the "managed-by" label no longer exists

### DIFF
--- a/controllers/argocd/argocd_controller_test.go
+++ b/controllers/argocd/argocd_controller_test.go
@@ -103,79 +103,72 @@ func TestReconcileArgoCD_Reconcile(t *testing.T) {
 	}
 }
 
-func TestReconcileArgoCD_Reconcile_withRemoveManagedByLabelOnArgocdDeletionNotSet(t *testing.T) {
+func TestReconcileArgoCD_Reconcile_RemoveManagedByLabelOnArgocdDeletion(t *testing.T) {
 	restoreEnv(t)
 	logf.SetLogger(ZapLogger(true))
-	a := makeTestArgoCD(deletedAt(time.Now()), addFinalizer(common.ArgoCDDeletionFinalizer))
-	r := makeTestReconciler(t, a)
-
-	nsArgocd := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{
-		Name: a.Namespace,
-	}}
-	err := r.Client.Create(context.TODO(), nsArgocd)
-	assert.NoError(t, err)
-
-	ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{
-		Name: "newNamespaceTest",
-		Labels: map[string]string{
-			common.ArgoCDManagedByLabel: a.Namespace,
-		}},
-	}
-	err = r.Client.Create(context.TODO(), ns)
-	assert.NoError(t, err)
-
-	req := reconcile.Request{
-		NamespacedName: types.NamespacedName{
-			Name:      a.Name,
-			Namespace: a.Namespace,
-		},
-	}
-
-	_, err = r.Reconcile(context.TODO(), req)
-	assert.NoError(t, err)
-
-	// Check if the managed-by label still exists in the new namespace
-	assert.NoError(t, r.Client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, ns))
-	assert.Equal(t, ns.Labels[common.ArgoCDManagedByLabel], a.Namespace)
-}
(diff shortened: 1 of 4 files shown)


wtam2018 (commented):
Please update the documentation.

controllers/argocd/deployment.go
```diff
@@ -1293,6 +1293,13 @@ func isDexDisabled() bool {
 	return false
 }
 
+func removeManagedByLabelOnArgoCDDeletion() bool {
```
jaideepr97, line 1296: can we call this `isRemoveManagedByLabelOnArgoCDDeletion()` or something similar instead?
May be better if the purpose of the function is more clear from just reading the name IMO

controllers/argocd/argocd_controller_test.go
```diff
@@ -102,6 +103,82 @@ func TestReconcileArgoCD_Reconcile(t *testing.T) {
...
+	_, err = r.Reconcile(context.TODO(), req)
+	assert.NoError(t, err)
+
+	// Check if the managed-by label gets removed from the new namespace
+	assert.NoError(t, r.Client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, ns))
+	if _, ok := ns.Labels[common.ArgoCDManagedByLabel]; ok {
+		t.Errorf("Expected the label[%v] to be removed from the namespace[%v]", common.ArgoCDManagedByLabel, ns.Name)
+	}
+}
+
```
jaideepr97, line 181: Can we use a table format for these tests instead 
that would help avoid redundancy of testing code 
please refer to this example https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/status_test.go#L55

docs/usage/custom_roles.md
```diff
@@ -38,3 +38,19 @@ spec:
           - name: SERVER_CLUSTER_ROLE
             value: custom-server-role
 ```
+
+
+When an target Argo CD instance gets deleted, namespaces with the `argocd.argoproj.io/managed-by` label of that Argo CD instance's namespace will keep the label. Users can remove the `argocd.argoproj.io/managed-by` label value on Argo CD deletion by setting the environment variable `REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION` to `true` in the Subscription.
```
jaideepr97, line 43: can we rephrase this as: 

"When an Argo CD instance is deleted, namespaces managed by that instance (via the `argocd.argoproj.io/managed-by` label ) will retain the label by default.  Users can change this behavior by setting the environment variable `REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION` to `true` in the Subscription.

ciiay: Thanks for the help. Updated.

jannfis (approved):
LGTM. Thanks @ciiay.

redox72:
In my org I do not have permissions to edit the subscription config but I need this fix as well.
Is it possible to add an option to edit this via the argocd instance object?

PR #666, opened 2022-05-27
fix: upgrade Argo CD to handle CVE-2022-29165

iam-veeramalla:

**What type of PR is this?**
> /kind bug

Fixes **CVE-2022-29165**

```
A critical vulnerability has been discovered in Argo CD which would allow unauthenticated users to impersonate as any Argo CD user or role, including the admin user, by sending a specifically crafted JSON Web Token (JWT) along with the request. In order for this vulnerability to be exploited, [anonymous access](https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/#anonymous-access) to the Argo CD instance must have been enabled.

In a default Argo CD installation, anonymous access is disabled. To find out if anonymous access is enabled in your instance, please see the Workarounds section of this advisory below.

The vulnerability can be exploited to impersonate as any user or role, including the built-in admin account regardless of whether that account is enabled or disabled. Also, the attacker does not need an account on the Argo CD instance in order to exploit this.

If anonymous access to the instance is enabled, an attacker can:

    Escalate their privileges, effectively allowing them to gain the same privileges on the cluster as the Argo CD instance, which is cluster admin in a default installation. This will allow the attacker to create, manipulate and delete any resource on the cluster.

    Exfiltrate data by deploying malicious workloads with elevated privileges, thus bypassing any redaction of sensitive data otherwise enforced by the Argo CD API

We strongly recommend that all users of Argo CD update to a version containing this patch as soon as possible, regardless of whether or not anonymous access is enabled in your instance.
```

NA

**CVE-2022-29165**

Upgrade to Argo CD dependency version to v2.3.4

### DIFF
--- a/build/util/Dockerfile
+++ b/build/util/Dockerfile
@@ -1,5 +1,5 @@
-# Argo CD v2.4.0-rc1
-FROM quay.io/argoproj/argocd@sha256:92ea7377a66fa3e79dd1d7c804307a8b8829049fcf30df3884be764945d8ce7e
+# Argo CD v2.4.0-rc2
+FROM quay.io/argoproj/argocd@sha256:e8f78f3217012b52c482f5d38c162391cdc5dbfb5a1402667454bde8126700bf
 
 USER root
 
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -55,7 +55,7 @@ const (
 	ArgoCDDefaultArgoImage = "quay.io/argoproj/argocd"
 
 	// ArgoCDDefaultArgoVersion is the Argo CD container image digest to use when version not specified.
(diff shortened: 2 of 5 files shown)


jaideepr97:
LGTM, thanks @iam-veeramalla 
Although I don't really think this has high impact on the operator

iam-veeramalla:
> LGTM, thanks @iam-veeramalla 
> Although I don't really think this has high impact on the operator 

@jaideepr97 it does, marked as critical as Dependabot as we consume Argo CD as a dependency.

jaideepr97:
@iam-veeramalla Yeah, but we only use it for the application API right?

PR #672, opened 2022-06-02
fix: mount secret to service account for dex if its not there

rishabh625:
Signed-off-by: rishabh625 <rishabhmishra625@gmail.com>

**What type of PR is this?**

/kind bug

> /kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

since k8s v1.24 secret is not auto mounted to SA , dex creation fails as oauth client used for dex is service account. This PR creates a token and adds the token to service account

Fixes  [326](https://github.com/redhat-developer/gitops-operator/issues/326)

Run this on ocp v4.11 and on lower version to check backward compatibility

### DIFF
--- a/controllers/argocd/util.go
+++ b/controllers/argocd/util.go
@@ -384,7 +384,7 @@ func (r *ReconcileArgoCD) getDexOAuthClientSecret(cr *argoprojv1a1.ArgoCD) (*str
 		sa.Secrets = append(sa.Secrets, *tokenSecret)
 		err = r.Client.Update(context.TODO(), sa)
 		if err != nil {
-			return nil, errors.New("Failed to Add ServiceAccount token for OAuth client secret")
+			return nil, errors.New("failed to add ServiceAccount token for OAuth client secret")
 		}
 	}
 
--- a/controllers/argocd/util.go
+++ b/controllers/argocd/util.go
@@ -377,11 +377,15 @@ func (r *ReconcileArgoCD) getDexOAuthClientSecret(cr *argoprojv1a1.ArgoCD) (*str
 		if err != nil {
 			return nil, errors.New("unable to locate and create ServiceAccount token for OAuth client secret")
 		}
-		tokenSecret := &corev1.ObjectReference{
+		tokenSecret = &corev1.ObjectReference{
 			Name:      secret.Name,
 			Namespace: cr.Namespace,
 		}
 		sa.Secrets = append(sa.Secrets, *tokenSecret)
+		err = r.Client.Update(context.TODO(), sa)
+		if err != nil {
+			return nil, errors.New("Failed to Add ServiceAccount token for OAuth client secret")
+		}
 	}
 
 	// Fetch the secret to obtain the token
--- a/controllers/argocd/util.go
+++ b/controllers/argocd/util.go
@@ -29,6 +29,7 @@ import (
 	"text/template"
 
 	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"gopkg.in/yaml.v2"
 
@@ -376,6 +377,10 @@ func (r *ReconcileArgoCD) getDexOAuthClientSecret(cr *argoprojv1a1.ArgoCD) (*str
 		err := r.Client.Create(context.TODO(), secret)
 		if err != nil {
 			return nil, errors.New("unable to locate and create ServiceAccount token for OAuth client secret")
+		}
+		err = controllerutil.SetControllerReference(cr, secret, r.Scheme)
+		if err != nil {
+			return nil, err
 		}
 		tokenSecret = &corev1.ObjectReference{
 			Name:      secret.Name,
--- a/controllers/argocd/util.go
+++ b/controllers/argocd/util.go
@@ -353,7 +353,6 @@ func (r *ReconcileArgoCD) getDexOAuthClientSecret(cr *argoprojv1a1.ArgoCD) (*str
 	if err := argoutil.FetchObject(r.Client, cr.Namespace, sa.Name, sa); err != nil {
 		return nil, err
 	}
-
 	// Find the token secret
 	var tokenSecret *corev1.ObjectReference
 	for _, saSecret := range sa.Secrets {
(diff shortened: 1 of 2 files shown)


chetan-rns:
Can we include a unit test for this change?

rishabh625:
@iam-veeramalla @chetan-rns : it's difficult to add test for this here , as code is specific to openshift and will fail in CI if added

was planning for below test , but later realize that cannot add
```
+//TestReconcileArgoCD_reconcileDexOAuthClientSecret This test make sures that if dex is enabled a service account is created with token stored in a secret which is used for oauth
+func TestReconcileArgoCD_reconcileDexOAuthClientSecret(t *testing.T) {
+       logf.SetLogger(ZapLogger(true))
+       a := makeTestArgoCD()
+       a.Spec.Dex.OpenShiftOAuth = true
+       r := makeTestReconciler(t, a)
+       assert.NoError(t, createNamespace(r, a.Namespace, ""))
+       assert.NoError(t, r.reconcileDexServiceAccount(a))
+       sa := &corev1.ServiceAccount{}
+       assert.NoError(t, argoutil.FetchObject(r.Client, a.Namespace, common.ArgoCDDefaultDexServiceAccountName, sa))
+       tokenExists := false
+       for _, saSecret := range sa.Secrets {
+               if strings.Contains(saSecret.Name, "dex-server-token") {
+                       tokenExists = true
+               }
+       }
+       assert.True(t, tokenExists, "Dex is enabled but unable to create oauth client secret")
+}
```

rishabh625:
Hi ,
since we cannot call reconcilation on 	testClient := testclient.NewSimpleClientset()
i am finding difficult to add unit test , can anyone guide on this please?

rishabh625:
> Hi , since we cannot call reconcilation on testClient := testclient.NewSimpleClientset() i am finding difficult to add unit test , can anyone guide on this please?

@chetan-rns @iam-veeramalla Added the unit test case

controllers/argocd/util.go
```diff
@@ -364,7 +363,29 @@ func (r *ReconcileArgoCD) getDexOAuthClientSecret(cr *argoprojv1a1.ArgoCD) (*str
...
+			return nil, errors.New("unable to locate and create ServiceAccount token for OAuth client secret")
+		}
+		tokenSecret = &corev1.ObjectReference{
+			Name:      secret.Name,
+			Namespace: cr.Namespace,
+		}
+		sa.Secrets = append(sa.Secrets, *tokenSecret)
+		err = r.Client.Update(context.TODO(), sa)
+		if err != nil {
+			return nil, errors.New("Failed to Add ServiceAccount token for OAuth client secret")
```
chetan-rns, line 387:
```suggestion
			return nil, errors.New("failed to add ServiceAccount token for OAuth client secret")
```

controllers/argocd/util_test.go
```diff
@@ -680,3 +681,16 @@ func generateEncodedPEM(t *testing.T, host string) []byte {
 	encoded := argoutil.EncodeCertificatePEM(cert)
 	return encoded
 }
+
+//sTestReconcileArgoCD_reconcileDexOAuthClientSecret This test make sures that if dex is enabled a service account is created with token stored in a secret which is used for oauth
```
chetan-rns, line 685:
```suggestion
// TestReconcileArgoCD_reconcileDexOAuthClientSecret This test make sures that if dex is enabled a service account is created with token stored in a secret which is used for oauth
```

controllers/argocd/util_test.go
```diff
@@ -680,3 +681,16 @@ func generateEncodedPEM(t *testing.T, host string) []byte {
...
+func TestReconcileArgoCD_reconcileDexOAuthClientSecret(t *testing.T) {
+	logf.SetLogger(ZapLogger(true))
+	a := makeTestArgoCD()
+	a.Spec.Dex.OpenShiftOAuth = true
+	r := makeTestReconciler(t, a)
+	assert.NoError(t, createNamespace(r, a.Namespace, ""))
+	_, err := r.reconcileServiceAccount(common.ArgoCDDefaultDexServiceAccountName, a)
+	assert.NoError(t, err)
+	_, err = r.getDexOAuthClientSecret(a)
+	assert.NoError(t, err)
```
chetan-rns, line 695: Can we test if the SA has a token secret? 

```
	assert.NoError(t, argoutil.FetchObject(r.Client, a.Namespace, sa.Name, sa))
	tokenExists := false
	for _, saSecret := range sa.Secrets {
		if strings.Contains(saSecret.Name, "dex-server-token") {
			tokenExists = true
		}
	}
	assert.True(t, tokenExists, "Dex is enabled but unable to create oauth client secret")
```
like mentioned here https://github.com/argoproj-labs/argocd-operator/pull/672#issuecomment-1147268186

chetan-rns (commented):
The PR looks good to me. Do we need to remove the token secret if the user removes the Argo CD instance?  WIll setting owner reference take care of the clean-up?

controllers/argocd/util_test.go
rishabh625: applied

controllers/argocd/util.go
```diff
@@ -364,7 +364,33 @@ func (r *ReconcileArgoCD) getDexOAuthClientSecret(cr *argoprojv1a1.ArgoCD) (*str
 	}
 
 	if tokenSecret == nil {
-		return nil, errors.New("unable to locate ServiceAccount token for OAuth client secret")
+		secret := &corev1.Secret{
```
jaideepr97, line 367: Optional: could we have a comment here that mentions that this is done for k8s v1.24? just so we can have the context for it when going through the code in the future?

Not a required change by any means

rishabh625: yeah sure , will add comment Thanks @jaideepr97

rishabh625:
Also @jaideepr97 : it doesn't matter that which PR goes first unfying sso one or this one , it would be only merge conflict that either have to solve and it would be a easy one

controllers/argocd/util.go
```diff
@@ -364,7 +364,33 @@ func (r *ReconcileArgoCD) getDexOAuthClientSecret(cr *argoprojv1a1.ArgoCD) (*str
 	}
 
 	if tokenSecret == nil {
```
rishabh625, line 366: `// This change is done, due to change of reduction of secret-based service account tokens in k8s v1.24 so from k8s v1.24 no default secret for service account, but for dex to work we need to provide token of secret used by dex service account  as a oauth token, this helps to achieve that, in log run we should see do dex really requires a secret or it manages to create one using TokenRequest API
`

rishabh625: is this ok @jaideepr97 ?

jaideepr97: looks fine, just as long as we have something! 
thanks

jaideepr97:
@rishabh625 I agree, that's why I think we should merge this first 
This would be pretty easy for me to pull into my branch, so i see no reason to hold it up

jopit (approved):
LGTM

PR #677, opened 2022-06-08
feat: Added routes,ingress,svc code for applicationset

### LINKED ISSUE #639: Applicationset Controller does not deploy service or enable port to reach new webhook feature
**Describe the bug**
As per [v0.3.0 of the Applicationset controller](https://blog.argoproj.io/applicationset-controller-v0-3-0-for-argo-cd-now-available-3f56bd643312), it has support for wehbook notifications, the current installation of the Applicationset controller being reconciled only installs a `Deployment`, `RoleBinding`s, `ServiceAccount` and `Role`, but no `Service`. The deployment is also missing the new port `7000` exposed for webhook events.

**To Reproduce**
1. Install the latest ArgoCD operator release
2. Enable the applicationSet controller option in your CRD installation
3. The installation only creates the above mentioned resources without exposing a port
4. The webhook is unreachable, since the deployment doesn't expose the port and a manual Service creation is needed.

**Expected behavior**
A similar service to the one created by the ApplicationSet controller project yamls should be created to allow the webhook to be reachable.
The deployment should also expose the port 7000 to allow for communication to the webhook.

rishabh625:
Signed-off-by: rishabh625 <rishabhmishra625@gmail.com>

**What type of PR is this?**

/kind bug

> /kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

This PR creates service,routesor ingress and exposes port 7000 for webhook and 8080 for metrics of application controller

* [] Documentation update is required by this PR.

Fixes #639


rishabh625:
@jopit  @iam-veeramalla  plz review this PR!!

controllers/argocd/applicationset.go
```diff
@@ -455,3 +461,52 @@ func setAppSetLabels(obj *metav1.ObjectMeta) {
...
+
+// reconcileApplicationSetService will ensure that the Service is present for the ApplicationSet webhook and metrics component.
+func (r *ReconcileArgoCD) reconcileApplicationSetService(cr *argoprojv1a1.ArgoCD) error {
+	log.Info("reconciling applicationset service")
+
+	svc := newServiceWithSuffix(common.ApplicationSetServiceNameSuffix, common.ApplicationSetServiceNameSuffix, cr)
+	if cr.Spec.ApplicationSet == nil {
+		// if ensureAutoTLSAnnotation(svc, common.ArgoCDServerTLSSecretName, cr.Spec.Server.WantsAutoTLS()) {
+		// 	return r.Client.Update(context.TODO(), svc)
+		// }
```
jannfis, line 473: Can you please remove these lines, instead of commenting them out?

controllers/argocd/applicationset.go
```diff
@@ -455,3 +461,52 @@ func setAppSetLabels(obj *metav1.ObjectMeta) {
...
+			log.Info(fmt.Sprintf("Deleting applicationset controller service %s as applicationset is disabled", svc.Name))
+			err = r.Delete(context.TODO(), svc)
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		if argoutil.IsObjectFound(r.Client, cr.Namespace, svc.Name, svc) {
+			return nil // Service found, do nothing
+		}
```
jannfis, line 488: I'm not sure whether that should be the correct reconciliation behavior. I know it's probably similar to other services we reconcile, but we really should think about it. The behavior here will only ever create the service, but if we want to make changes to the service in the code, it will never adapt unless the service is actually deleted.

jannfis (changes requested):
I have a couple of items that need to be addressed, please see below.

Also, please provide end-to-end tests (kuttl) along with this change.

tests/k8s/1-001_validate_basic/05-applicationset.yaml
```diff
@@ -0,0 +1,21 @@
...
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ArgoCD
+metadata:
+  name: example-argocd
+spec:
+  applicationSet: 
+    webhookSpec:
+          ingress:
+            enabled: true
```
jannfis, line 20: This looks a little odd in indenting.

jannfis (commented):
I have some more comments.

Also, regarding the end-to-end tests, we should make sure to test full reconciliation. For example, can we disable the webhook server again, will other settings result in correct behavior, etc. Please be more diligent with the tests.

api/v1alpha1/argocd_types.go
```diff
@@ -827,6 +829,24 @@ type SSHHostsSpec struct {
...
+type WebhookServerSpec struct {
+
+	// Host is the hostname to use for Ingress/Route resources.
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Host",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:fieldGroup:Server","urn:alm:descriptor:com.tectonic.ui:text"}
+	Host string `json:"host,omitempty"`
+
+	// Ingress defines the desired state for an Ingress for the Application set webhook metricscomponent.
+	Ingress ArgoCDIngressSpec `json:"ingress,omitempty"`
+
+	// Insecure toggles the insecure flag.
```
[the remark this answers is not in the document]
rishabh625: I have to remove this insecure is not used

jannfis (approved):
LGTM!

Red-M:
Removing `deploy/catalog_source.yaml` means the docs for installing via OLM are no longer correct and means that you can't install the operator via this repo.

PR #667, opened 2022-05-27
fix: Operator does not create necessary RoleBindings in all cases

ciiay:
**What type of PR is this?**
/kind bug

**What does this PR do / why we need it**:
Fixes [GITOPS-1963](https://issues.redhat.com/browse/GITOPS-1963) issue.

**Have you updated the necessary documentation?**
* [ ] Documentation update is required by this PR.
* [ ] Documentation has been updated.

**Which issue(s) this PR fixes**:
Fixes #GITOPS-1963 Operator does not create necessary RoleBindings in all cases

**How to test changes / Special notes to the reviewer**:
a. $ oc create namespace central-argocd
b. $ oc create namespace argocd-1
c. $ oc label ns argocd-1 [argocd.argoproj.io/managed-by=central-argocd](http://argocd.argoproj.io/managed-by=central-argocd)
d. $ oc project central-argocd
e. $ oc apply -f argocd-basic.yaml
f. $ oc get roles -n central-argocd 
Now observe that all 4 roles/rolebindings including dex and redis service are created for central-argocd ns

PR #682, opened 2022-06-16
fix: Operator does not create necessary RoleBindings in all cases (#667)

jopit:
**What type of PR is this?**
/kind chore

Cherry-picks PR #667 into release-0.4


ishitasequeira (approved):
LGTM

PR #699, opened 2022-06-28
Fix that new host name isn't applied to ingress

### LINKED ISSUE #558: Changing host does not update ingress
**Describe the bug**
When setting up an `ArgoCd`, and later changing the host the ingress will not be updated with the new host information.

**To Reproduce**
Do a kubectl apply on something like:

```yaml
kind: ArgoCD
metadata:
  name: argocd
spec:
  server:
    ingress:
      enabled: true
```

followed by a kubectl apply on something like:

```yaml
kind: ArgoCD
metadata:
  name: argocd
spec:
  server:
    host: foo.bar

    ingress:
      enabled: true
```

then the host will still be set to the default value when describing the ingress

**Expected behavior**
That the ingress will be updated with the new host

mosuke5:
**What type of PR is this?**
/kind bug

Fixed a problem where changes to `.spec.server.host` in the ArgoCD CR were not reflected in the Ingress resource by changing the order of processing in `reconcileArgoServerIngress()`.

Fixes #558

I have added a test scenario in `ingress_test.go`. You can be verified by a test run.
If you want to check manually, you can do so as follows.

```
$ cat <<EOF | kubectl apply -f - 
kind: ArgoCD
metadata:
  name: argocd-sample
spec:
  server:
    host: before.example.com
    ingress:
      enabled: true
EOF
argocd.argoproj.io/argocd-sample configured

$ kubectl get ingress argocd-sample-server -o yaml
...
spec:
  rules:
  - host: before.example.com
    http:
      paths:
      - backend:
          service:
            name: argocd-sample-server
            port:
              name: http
        path: /
        pathType: ImplementationSpecific
  tls:
  - hosts:
    - before.example.com
    secretName: argocd-secret

$ cat <<EOF | kubectl apply -f - 
kind: ArgoCD
metadata:
  name: argocd-sample
spec:
  server:
    host: after.example.com
    ingress:
      enabled: true
EOF
argocd.argoproj.io/argocd-sample configured

$ kubectl get ingress argocd-sample-server -o yaml
...
spec:
  rules:
  - host: after.example.com
    http:
      paths:
      - backend:
          service:
            name: argocd-sample-server
            port:
              name: http
        path: /
        pathType: ImplementationSpecific
  tls:
  - hosts:
    - after.example.com
    secretName: argocd-secret
```


controllers/argocd/ingress.go
```diff
@@ -162,7 +162,11 @@ func (r *ReconcileArgoCD) reconcileArgoServerIngress(cr *argoprojv1a1.ArgoCD) er
 	if err := controllerutil.SetControllerReference(cr, ingress, r.Scheme); err != nil {
 		return err
 	}
-	return r.Client.Create(context.TODO(), ingress)
+
+	if !found {
+		return r.Client.Create(context.TODO(), ingress)
+	}
+	return r.Client.Update(context.TODO(), ingress)
```
jaideepr97, line 169: Hi @mosuke5 Thanks for your PR! 
Can we add a check here to see if anything has changed before calling Update? Otherwise we might have a lot of unnecessary api calls being made

mosuke5: @jaideepr97 Thank you for your review. I fixed to update only when  annotations and spec of ingress resource are changed.

simonfelding:
bump. this also applies to HTTPRoutes from the Kubernetes Gateway API

olivergondza:
@mosuke5, Thanks for filing this PR. Please resolve the merge conflict or indicate that you are no longer interested to contribute this patch.

PR #700, opened 2022-06-28
Fix: reconcile repo server on mountSAToken and mountSAToken Name change setting

rishabh625:
**What type of PR is this?**

/kind bug

This PR aims to identify all the fields of RepoServer and ensures as CR changes for repo server fields new deployment is triggered with updated values

As a part of PR have seen all fields  of spec.Repo
logLevel, logFormat, image, version are reconciled as a part of init containers 
reconcilation on change for env, volumes, volumeMounts, initContainers is already present
sidecarContainers are reconciled as a part of comtainers

left with two fields  mountSAToken, serviceaccount reconiclation logic is in this PR

Fixes #?

1) Make default deployment of ArgoCD via operator
2) Create a service Account similar to default service account of a namespace
3) Add that Service Account into ArgoCD CR .spec.repo.serviceAccount

Check pod using this new ServiceAccount

### DIFF
--- a/controllers/argocd/util.go
+++ b/controllers/argocd/util.go
@@ -1027,30 +1027,30 @@ func (r *ReconcileArgoCD) setResourceWatches(bldr *builder.Builder, clusterResou
 
 	// Add new predicate to delete Notifications Resources. The predicate watches the Argo CD CR for changes to the `.spec.Notifications.Enabled`
 	// field. When a change is detected that results in notifications being disabled, we trigger deletion of notifications resources
-	repoDeployment := predicate.Funcs{
-		UpdateFunc: func(e event.UpdateEvent) bool {
-			newCR, ok := e.ObjectNew.(*argoprojv1a1.ArgoCD)
-			if !ok {
-				return false
-			}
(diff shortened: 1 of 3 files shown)


controllers/argocd/util.go
```diff
@@ -1025,6 +1025,30 @@ func (r *ReconcileArgoCD) setResourceWatches(bldr *builder.Builder, clusterResou
...
+	// 			err := r.reconcileRepoDeployment(newCR, useTLSForRedis)
+	// 			if err != nil {
+	// 				log.Error(err, fmt.Sprintf("Failed to delete notifications controller resources for ArgoCD %s in namespace %s",
+	// 					newCR.Name, newCR.Namespace))
+	// 			}
+	// 		}
+	// 		return true
+	// 	},
+	// }
+
```
jaideepr97, line 1051: @rishabh625 This looks like an unintended change

rishabh625: Yeah Thanks @jaideepr97  good catch, i almost forgot that I commented it.

rishabh625:
Thanks @jannfis for merging this

PR #710, opened 2022-07-06
fix: argocd login won't crash when .data is repopulated in argocd-secret

reginapizza:
**What type of PR is this?**
 /kind bug

This PR makes the following changes.
1) If argocd-secret does not exist (has been deleted) , the reconciler recreates the secret and regenerates a new server session key.
2) When the admin password changes, the reconciler adds the new password hash to the argocd_secret.

When .data field has been deleted from argocd-secret and get regenerated, this fixes a bug where argocd login no longer works afterwards. 

https://issues.redhat.com/browse/GITOPS-2006

Verifying the error behavior: 
- checkout to recent version of master branch `git checkout master`
- Add the Argo CD instance: 
```
kind: ArgoCD
metadata:
  name: argocd
spec: {}
EOF 
```
- go to OpenShift console and under Administrator View, set namespace as `default` and navigate to Secrets > `argocd-secret`. View the yaml for this secret and delete the entire `.data` section, then save. You should be able to see that the data section gets repopulated but is missing the `server.secretkey` field. Logging into argocd will throw an error stating that this secretKey is missing and login fails. (workaround is restarting the argocd-server pod)

Testing the fix: 
- checkout to this PR
- Go to Makefile and change the version # (just to something else) and change image base tag to `IMAGE_TAG_BASE ?= quay.io/{your quay or docker username}/argocd-operator`
- Run `make docker-build`, `make docker-push` and `make deploy` again
- go back to the OpenShift console and still under Administrator View, set namespace as `default` and go again to Secrets > `argocd-secret`
- add the argocd instance again, and try deleting the secret again the same way as before. This time when you delete it and hit save, you should see that the entire secret regenerates, including the `server.secretkey`.
- to test that login now works, get the route of the argocd instance (no route is created automatically, but you can run `oc edit argocd` and then change `route: enabled: true` in the spec. Now once you run `oc get routes` you should get the route.
- get the admin password from the OpenShift UI under Secrets > `argocd-cluster` (in default namespace). Copy the data for admin.password at the bottom of the page.
- Now login to argocd with `argocd login [route you had from earlier]` and supply the username as `admin` and password as the password from the step above. This should succeed now.

### DIFF
--- a/controllers/argocd/secret.go
+++ b/controllers/argocd/secret.go
@@ -293,16 +293,24 @@ func (r *ReconcileArgoCD) reconcileClusterSecrets(cr *argoprojv1a1.ArgoCD) error
 func (r *ReconcileArgoCD) reconcileExistingArgoSecret(secret *corev1.Secret, clusterSecret *corev1.Secret, tlsSecret *corev1.Secret) error {
 	changed := false
 
+	if secret.Data == nil {
+		secret.Data = make(map[string][]byte)
+	}
+
+	if secret.Data[common.ArgoCDKeyServerSecretKey] == nil {
+		sessionKey, err := generateArgoServerSessionKey()
+		if err != nil {
+			return err
+		}
+		secret.Data[common.ArgoCDKeyServerSecretKey] = sessionKey
+	}
+
 	if hasArgoAdminPasswordChanged(secret, clusterSecret) {
 		pwBytes, ok := clusterSecret.Data[common.ArgoCDKeyAdminPassword]
 		if ok {
 			hashedPassword, err := argopass.HashPassword(strings.TrimRight(string(pwBytes), "\n"))
 			if err != nil {
 				return err
-			}
-
-			if secret.Data == nil {
-				secret.Data = make(map[string][]byte)
 			}
 
 			secret.Data[common.ArgoCDKeyAdminPassword] = []byte(hashedPassword)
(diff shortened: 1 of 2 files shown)


controllers/argocd/secret.go
```diff
@@ -305,8 +305,14 @@ func (r *ReconcileArgoCD) reconcileExistingArgoSecret(secret *corev1.Secret, clu
...
 			}
 
+			sessionKey, err := generateArgoServerSessionKey()
+			if err != nil {
+				return err
+			}
+
 			secret.Data[common.ArgoCDKeyAdminPassword] = []byte(hashedPassword)
 			secret.Data[common.ArgoCDKeyAdminPasswordMTime] = nowBytes()
+			secret.Data[common.ArgoCDKeyServerSecretKey] = sessionKey
```
jannfis, line 315: I believe we should only generate a new server key when the field in the Secret's data is empty. And we should not wrap it into a branch of when `hasArgoAdminPasswordChanged()` has returned true.

A little context: The server key is used for all kind of things, and should be persistent usually. It is used amongst others to sign JWTs. So if this field still exists in the Secret, it must stay the same.

Looking at this method, we should probably:
1. Move the `if secret.Data == nil` to the top of the method, out of the `hasArgoAdminPasswordChanged` condition
2. Do the same with `generateArgoServerSessionKey()`, but do this only if `secret.Data[common.ArgoCDKeyServerSecretKey] is not set to a value

jannfis:
Also, please make sure to have proper Go version installed locally before running `make bundle` - I think we use Go 1.18 meanwhile.

jaideepr97:
@reginapizza I think this could probably use a unit test as well

jaideepr97:
/lgtm
thanks @reginapizza

wtam2018 (approved):

PR #727, opened 2022-07-16
Add control RBAC for Prometheus

zvlb:
**What type of PR is this?**
/kind bug

If create Prometheus-Operator without OLM Service Account "k8s-prometheus" not exist. It's reason why Prometheus not working in Kubernetes.
Add control RBAC for prometheus to Operator

* [!] Documentation update is required by this PR.

Local tests with Kubernetes cluster

### DIFF
--- a/docs/usage/basics.md
+++ b/docs/usage/basics.md
@@ -71,7 +71,7 @@ The operator will create these ConfigMaps for the cluster and set the initial va
 ### Secrets
 
 There is a Secret that is used by Argo CD named `argocd-secret`. The `argocd-server` component reads this secret to
-obtain the admin password for authentication.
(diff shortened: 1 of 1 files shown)


vzemtsov:
> Hey @jopit @iam-veeramalla @jannfis Along the same lines of the conversation we had a while back about whether or not the Argo CD operator should be spinning up new grafana instances - should this apply to prometheus as well? It seems we have an incomplete implementation of prometheus within the operator right now, but I'm wondering if it should be a part of the operator at all
> 
> Let me know your thoughts, thanks

I'll put in my opinion, even though you don't ask.
Now, to deploy Prometheus, you need a working Prometheus-operator in cluster. This makes ArgoCD-Operator imperfect.

Don't deploy Prometheus/Grafana looks like a good idea

snorwin:
@zvlb I think it worth to rebase the PR and try to get it merged.

Issue #262, opened 2021-03-04
Argo CD service names should be referenced by their FQDN


jannfis:
The Argo CD service names should be referenced by their FQDN in order to allow cluster-wide `no_proxy` settings to apply to all instances of Argo CD installed by the operator when a global proxy is set up.

So, instead of e.g. `--repo-server argocd-repo-server:8081`, the reference should be made as `--repo-server argocd-repo-server.<namespace>.svc.cluster.local` - this will also consider an already existing entry of `.cluster.local` in the `no_proxy` env var.

PR #731, opened 2022-07-20
Delete svc.cluster.local from generated svc-routes

zvlb:
**What type of PR is this?**
/kind bug

If Kubernetes/OpenShift Cluster Name not "cluster.local" - routes by svc name like `<SERVICE_NAME>.<NAMESPACE>.svc.cluster.local` - not working.
Delete `.svc.cluster.local` from generated routes name.


vzemtsov:
@jomkz Can you approve this?

wtam2018:
@vzemtsov Can you please provide some details on the issue that you are running into? The servers should be communicating over the cluster local addresses. I am pretty sure these changes will break the deployments when cluster proxy is enabled. When cluster proxy is enabled, these communication will try to go through the proxy with your changes and it will fail.

vzemtsov:
What you mean when you say "cluster proxy"? Kube-proxy?
With my changes all work good. Cluster communication worked correctly.

About problem:
When you install Kubernetes Cluster, you can set ClusterName (https://kubernetes.io/docs/reference/config-api/kubeadm-config.v1beta3/). If you did this cluster domain changed from `*.cluster.local` to `*.<CLUSTER_NAME>`. But in argocd-operator cluster for communication use hardcode cluster-domain - `*.cluster.local`.
I fix it.

(For communication in cluster you can use DNS-record without cluster-domain, like `<SERVICE_NAME>.<NAMESPACE>`)

P.S. Sorry for my english

wtam2018:
What I mean by cluster proxy is the OpenShift cluster wide proxy. https://docs.openshift.com/container-platform/4.11/networking/enable-cluster-wide-proxy.html

wtam2018:
@jannfis please take a look to see if this PR will affect this [fix](https://github.com/argoproj-labs/argocd-operator/issues/262).

vzemtsov:
> @jannfis please take a look to see if this PR will affect this [fix](https://github.com/argoproj-labs/argocd-operator/issues/262).
Oh. I don't see [this](https://github.com/argoproj-labs/argocd-operator/issues/262) PR from @jannfis.

May be I can add new field to CRD ArgoCD (like `ClusterName` or `ClusterDomain`) and set by default to `cluster.local`.
If for someone (like me) needs change this, than can set this in CR
Or, if it's impossible, I can add function for read ClusterName from KubeApi. (I don't do this before, but I can try)

P.S. Sorry for my English

PR #740, opened 2022-07-25
fix: imagePullPolicy in dex deployment

saumeya:
**What type of PR is this?**
 /kind bug

Fixes https://github.com/redhat-developer/gitops-operator/issues/132
Removing the `ImagePullPolicy` in dex deployment
The default policy of `IfNotPresent` property will be used in the deployment.

Fixes #?

### DIFF
--- a/controllers/argocd/dex.go
+++ b/controllers/argocd/dex.go
@@ -229,10 +229,9 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoprojv1a1.ArgoCD) error 
 			"/shared/argocd-dex",
 			"rundex",
 		},
-		Image:           getDexContainerImage(cr),
-		ImagePullPolicy: corev1.PullAlways,
-		Name:            "dex",
-		Env:             proxyEnvVars(),
+		Image: getDexContainerImage(cr),
+		Name:  "dex",
+		Env:   proxyEnvVars(),
 		LivenessProbe: &corev1.Probe{
 			ProbeHandler: corev1.ProbeHandler{
 				HTTPGet: &corev1.HTTPGetAction{
--- a/controllers/argocd/dex_test.go
+++ b/controllers/argocd/dex_test.go
@@ -421,7 +421,6 @@ func TestReconcileArgoCD_reconcileDexDeployment(t *testing.T) {
 				},
 				VolumeMounts: []corev1.VolumeMount{
 					{Name: "static-files", MountPath: "/shared"}},
-				ImagePullPolicy: corev1.PullAlways,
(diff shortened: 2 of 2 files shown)


controllers/argocd/dex.go
```diff
@@ -229,10 +229,9 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoprojv1a1.ArgoCD) error
 			"/shared/argocd-dex",
 			"rundex",
 		},
-		Image:           getDexContainerImage(cr),
-		ImagePullPolicy: corev1.PullAlways,
```
iam-veeramalla, line 233: Removing the "PullAlways" as default lgtm but do you want to provide an option to the users if they want to use "PullAlways" ?

Also please fix the failing e2e.

iam-veeramalla (changes requested):
Removing the "PullAlways" as default lgtm but do you want to provide an option to the users if they want to use "PullAlways" ?

Also please fix the failing e2e.

controllers/argocd/dex.go
saumeya: Thanks, so the behaviour will be if the image tag is :latest or empty - the imagePullPolicy will set to 'Always' but if the tag is specified then it will be 'IfNotPresent'.

saumeya: Default image pull policy
When you (or a controller) submit a new Pod to the API server, your cluster sets the imagePullPolicy field when specific conditions are met:

if you omit the imagePullPolicy field, and the tag for the container image is :latest, imagePullPolicy is automatically set to Always;
if you omit the imagePullPolicy field, and you don't specify the tag for the container image, imagePullPolicy is automatically set to Always;
if you omit the imagePullPolicy field, and you specify the tag for the container image that isn't :latest, the imagePullPolicy is automatically set to IfNotPresent.

iam-veeramalla:
The following test failed.
```
    logger.go:42: 18:36:56 | 1-021_validate_notification_controller/3-disable_notification | test step failed 3-disable_notification
    case.go:254: failed in step 3-disable_notification
    case.go:256: resource matched of kind: apps/v1, Kind=Deployment
```

iam-veeramalla:
@saumeya unfortunately the log captured by the e2e tests are not significant enough to identify the root cause for the failure. 
Can you run the below test on you setup to understand the issue ?
`kubectl kuttl test ./tests/k8s --config ./tests/kuttl-tests.yaml --test 1-021_validate_notification_controller`

iam-veeramalla (approved):
LGTM, Thanks @saumeya

PR #760, opened 2022-08-22
fix: Keycloak/RHSSO login yields in certificate signed by unknown authority 

### LINKED ISSUE #761: Keycloak/RHSSO login yields in certificate signed by unknown authority
**Describe the bug**
Argo CD Operator can make use of RHSSO & KeyCloak through OIDC in addition to Dex. However, with a recent security fix applied, the certificate of RHSSO  or Keycloak cannot be validated in some scenarios. 

Single Sign On with Argo CD Operator can fail with the error message "x509: certificate signed by unknown authority" under the following circumstances:
```
    You are using RHSSO or KeyCloak as SSO provider and
    You are using a self-signed certificate for your route endpoints, or you are using a private CA to issue certificates
```

This behavior is a result of a [security fix](https://github.com/argoproj/argo-cd/security/advisories/GHSA-7943-82jg-wmw5) in Argo CD, which enforces a strict validation of TLS certificates on the configued OIDC endpoints.

iam-veeramalla:
**What type of PR is this?**
> /kind bug

Single Sign On with GitOps Operator can fail with the error message "x509: certificate signed by unknown authority" under the following circumstances:

You are using RHSSO (KeyCloak) as SSO provider and You are using a self-signed certificate for your route endpoints, or you are using a private CA to issue certificates.

This behavior is a result of a [security fix](https://github.com/argoproj/argo-cd/security/advisories/GHSA-7943-82jg-wmw5) in Argo CD, which enforces a strict validation of TLS certificates on the configued OIDC endpoints.

**Workaround & Mitigation**
There are multiple workarounds available:
For Argo CD Operator 0.4.0, you can disable TLS validation for the OIDC (Keycloak/RHSSO) endpoint in the ArgoCD spec:
```
spec:
  extraConfig:
    oidc.tls.insecure.skip.verify: "true"
```

For Argo CD Operator 0.3.0 or less, you need to patch the `argocd-cm` ConfigMap in your instance's namespace as follows
```
oc patch configmap argocd-cm --patch='{"data": {"oidc.tls.insecure.skip.verify": "true"}}'
```

Fixes #761 
Closes #761 

2. Run the e2e test using `kubectl kuttl test ./tests/ocp --config ./tests/kuttl-tests.yaml --test 1-001_validate_rhsso`

(or)

3. You can also test this manually by
4. Create the Argo CD Instance using Keycloak as SSO provider.
```
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: route
spec:
  sso:
    provider: keycloak
    keycloak:
       verifyTLS: false
       rootTLS: "--- Some Dummy Root CA ---"
  server:
    route:
      enabled: true
```
5. Give the operator 3-4 minutes to reconcile.
6. Verify that `rootCA` is added to the oidc.config in `argocd-cm` ConfigMap.
7. Modify the `rootCA` and verify that the operator reconciles the changes.

### DIFF
--- a/controllers/argocd/keycloak.go
+++ b/controllers/argocd/keycloak.go
@@ -1398,60 +1398,65 @@ func (r *ReconcileArgoCD) reconcileKeycloakForOpenShift(cr *argoprojv1a1.ArgoCD)
 		}
 	}
 
-	// If Keycloak deployment exists and a realm is already created for ArgoCD, Do not create a new one.
-	cfg, err := r.prepareKeycloakConfig(cr)
-	if err != nil {
-		return err
-	}
+	if existingDC.Status.AvailableReplicas == expectedReplicas {
 
-	// keycloakRouteURL is used to update the OIDC configuration for ArgoCD.
-	keycloakRouteURL := cfg.KeycloakURL
-	if existingDC.Status.AvailableReplicas == expectedReplicas &&
-		existingDC.Annotations["argocd.argoproj.io/realm-created"] == "false" {
-
-		// Create a keycloak realm and publish.
-		response, err := createRealm(cfg)
+		cfg, err := r.prepareKeycloakConfig(cr)
 		if err != nil {
-			log.Error(err, fmt.Sprintf("Failed posting keycloak realm configuration for ArgoCD %s in namespace %s",
+			return err
+		}
+		// keycloakRouteURL is used to update the OIDC configuration for ArgoCD.
+		keycloakRouteURL := cfg.KeycloakURL
+
+		// If Keycloak deployment exists and a realm is already created for ArgoCD, Do not create a new one.
+		if existingDC.Annotations["argocd.argoproj.io/realm-created"] == "false" {
+
+			// Create a keycloak realm and publish.
+			response, err := createRealm(cfg)
+			if err != nil {
+				log.Error(err, fmt.Sprintf("Failed posting keycloak realm configuration for ArgoCD %s in namespace %s",
+					cr.Name, cr.Namespace))
+				return err
+			}
+
+			if response == successResponse {
+				log.Info(fmt.Sprintf("Successfully created keycloak realm for ArgoCD %s in namespace %s",
+					cr.Name, cr.Namespace))
+
+				// TODO: Remove the deleteOAuthClient invocation once the issue is resolved.
+				// OAuthClient configuration does not get deleted from previous instances occasionally.
+				// It is safe to delete before updating the OIDC config.
+				// https://github.com/openshift/client-go/issues/209
+				err = deleteOAuthClient(cr)
+				if err != nil {
+					return err
+				}
+
+				// Update Realm creation. This will avoid posting of realm configuration on further reconciliations.
+				err = r.Client.Get(context.TODO(), types.NamespacedName{Name: existingDC.Name, Namespace: existingDC.Namespace}, existingDC)
+				if err != nil {
+					return err
+				}
+
+				existingDC.Annotations["argocd.argoproj.io/realm-created"] = "true"
+				err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+					return r.Client.Update(context.TODO(), existingDC)
+				})
+
+				if err != nil {
+					return err
+				}
+
+			}
+		}
+
+		// Updates OIDC Configuration in the argocd-cm when Keycloak is initially configured
+		// or when user requests to update the OIDC configuration through `.spec.sso.keycloak.rootCA`.
+		err = r.updateArgoCDConfiguration(cr, keycloakRouteURL)
+		if err != nil {
+			log.Error(err, fmt.Sprintf("Failed to update OIDC Configuration for ArgoCD %s in namespace %s",
 				cr.Name, cr.Namespace))
 			return err
 		}
-
-		if response == successResponse {
-			log.Info(fmt.Sprintf("Successfully created keycloak realm for ArgoCD %s in namespace %s",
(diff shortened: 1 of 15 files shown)


controllers/argocd/keycloak.go
```diff
@@ -307,7 +308,7 @@ func getKeycloakContainer(cr *argoprojv1a1.ArgoCD) corev1.Container {
 			{ContainerPort: 8888, Name: "ping", Protocol: "TCP"},
 		},
 		ReadinessProbe: &corev1.Probe{
-			FailureThreshold: 10,
+			FailureThreshold: 20,
```
iam-veeramalla, line 311: This change is not so relevant to this PR but I noticed that sometimes keycloak might take more time to be ready or accept traffic.

controllers/argocd/keycloak.go
```diff
@@ -1392,9 +1398,8 @@ func (r *ReconcileArgoCD) reconcileKeycloakForOpenShift(cr *argoprojv1a1.ArgoCD)
 		}
 	}
 
-	// If Keycloak deployment exists and a realm is already created for ArgoCD, Do not create a new one.
-	if existingDC.Status.AvailableReplicas == expectedReplicas &&
-		existingDC.Annotations["argocd.argoproj.io/realm-created"] == "false" {
+	// Proceed with the keycloak configuration only once the keycloak pod is up and running.
+	if existingDC.Status.AvailableReplicas == expectedReplicas {
```
iam-veeramalla, line 1402: There is no major change here .

Previously, we only reconcile Keycloak configuration and update the OIDC configuration once(for the very first time when keycloak instance is created). But now, with the addition of `rootCA` property users should be able to add/remove/update the OIDC configuration at any given time.

controllers/argocd/keycloak.go
```diff
@@ -1489,9 +1501,8 @@ func (r *ReconcileArgoCD) reconcileKeycloak(cr *argoprojv1a1.ArgoCD) error {
 		}
 	}
 
-	// If Keycloak deployment exists and a realm is already created for ArgoCD, Do not create a new one.
-	if existingDeployment.Status.AvailableReplicas == expectedReplicas &&
-		existingDeployment.Annotations["argocd.argoproj.io/realm-created"] == "false" {
+	// Proceed with the keycloak configuration only once the keycloak pod is up and running.
```
iam-veeramalla, line 1504: Same here

controllers/argocd/sso.go
```diff
@@ -85,7 +85,8 @@ func (r *ReconcileArgoCD) reconcileSSO(cr *argoprojv1a1.ArgoCD) error {
 		}
 	}
 
-	if cr.Spec.SSO != nil && (cr.Spec.SSO.Image != "" || cr.Spec.SSO.Version != "" || cr.Spec.SSO.VerifyTLS != nil || cr.Spec.SSO.Resources != nil) {
+	if cr.Spec.SSO != nil && (cr.Spec.SSO.Image != "" || cr.Spec.SSO.Version != "" ||
```
iam-veeramalla, line 88: Improve the readability of code

controllers/argocd/sso.go
```diff
@@ -164,7 +165,8 @@ func (r *ReconcileArgoCD) reconcileSSO(cr *argoprojv1a1.ArgoCD) error {
 		if cr.Spec.SSO.Provider == v1alpha1.SSOProviderTypeKeycloak {
 			// Relevant SSO settings at play are `DISABLE_DEX`, `.spec.dex`, `.spec.sso` fields, `.spec.sso.keycloak`, `.spec.sso.dex`
 
-			if (cr.Spec.SSO.Keycloak != nil) && (cr.Spec.SSO.Image != "" || cr.Spec.SSO.Version != "" || cr.Spec.SSO.Resources != nil || cr.Spec.SSO.VerifyTLS != cr.Spec.SSO.Keycloak.VerifyTLS) {
+			if (cr.Spec.SSO.Keycloak != nil) && (cr.Spec.SSO.Image != "" || cr.Spec.SSO.Version != "" ||
```
iam-veeramalla, line 168: 1. Improve the readability of code.
2. The below condition should be updated as it is breaking on the CR.
```
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: route
spec:
  sso:
   provider: keycloak
   keycloak:
     verifyTLS: false
  server:
    route:
      enabled: true
```

jaideepr97: @iam-veeramalla why is it breaking in this case? if `cr.spec.SSO.verifyTLS` is not expressed I think default should be false, that should mean `cr.spec.SSO.verifyTLS` and `cr.spec.SSO.keycloak.verifyTLS` should match and it should not error out

iam-veeramalla: @jaideep, if `cr.spec.SSO.verifyTLS` is not expressed, it will default to `true` not `false` as it is not a boolean it references a boolean.

jopit: I don't think it will have a boolean value at all, since it is a pointer, it will just be `nil`.  Both `cr.spec.SSO.verifyTLS` and `cr.spec.SSO.verifyTLS` are pointers to boolean, so checking if they have different values is going to be more complex.  Basically they are different if one is set and the other is not, or one points to true and the other points to false.

jopit: My comment above is irrelevant.  The change Abhishek made is correct, since `cr.spec.SSO.verifyTLS` being a pointer means it should be nil if keycloak is configured through `cr.spec.sso.keycloak`.

jopit (approved):
LGTM

PR #778, opened 2022-09-08
fix: remove default help chat url and text

### LINKED ISSUE #762: HelpChat cannot be disabled via Operator-Property
**Describe the bug**
When setting `helpChatURL: ''` within ArgoCD custom resource, the value is not propagated to the argocd-cm ConfigMap

**To Reproduce**
1. Configure ArgoCD custom resource with
```
spec:
  helpChatURL: ''
```
2. Look at ConfigMap argocd-cm. There it says:
```
  help.chatUrl: 'https://mycorp.slack.com/argo-cd'
```

An empty string would hide the "Chat" button within ArgoCD. With the URL set, the button is always shown.

**Workaround**
It is possible to configure the ArgoCD custom resource as follows to get the `''` in argocd-cm

```
spec:
  extraConfig:
    help.chatUrl: ''
```

**Expected behavior**
The empty string should appear in argocd-cm.

**Additional context**

The root cause seems to be the `len(...) > 0` here:
https://github.com/argoproj-labs/argocd-operator/blob/150519d954d16c0f60675e1b22a5f06671d53fad/controllers/argocd/configmap.go#L87-L89

saumeya:

**What type of PR is this?**
 /kind bug

Right now there is a default dummy value for help chat url and text. This button leads to dummy page and is not good user experience. The default should hence be blank and in case the user wants to add the help chat , they can do so using the argocd CR

Fixes #762 

`make install run` and check there is no button for help chat

### DIFF
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -164,10 +164,10 @@ const (
 	ArgoCDDefaultGrafanaVersion = "sha256:afef23a1b4cf159ec3180aac3ad693c10e560657313bfe3ec81f344ace6d2f05" // 6.7.2
 
 	// ArgoCDDefaultHelpChatURL is the default help chat URL.
-	ArgoCDDefaultHelpChatURL = "https://mycorp.slack.com/argo-cd"
+	ArgoCDDefaultHelpChatURL = ""
 
 	// ArgoCDDefaultHelpChatText is the default help chat text.
-	ArgoCDDefaultHelpChatText = "Chat now!"
+	ArgoCDDefaultHelpChatText = ""
 
 	// ArgoCDDefaultIngressPath is the path to use for the Ingress when not specified.
 	ArgoCDDefaultIngressPath = "/"


iam-veeramalla:
@saumeya please add a unit test for this behavior

controllers/argocd/configmap_test.go
```diff
@@ -133,8 +133,8 @@ func TestReconcileArgoCD_reconcileArgoConfigMap(t *testing.T) {
 		"dex.config":                         "",
 		"ga.anonymizeusers":                  "false",
 		"ga.trackingid":                      "",
-		"help.chatText":                      "Chat now!",
-		"help.chatUrl":                       "https://mycorp.slack.com/argo-cd",
+		"help.chatText":                      "",
+		"help.chatUrl":                       "",
```
iam-veeramalla, line 137: ok, you modified the test.

iam-veeramalla (approved):
LGTM

controllers/argocd/configmap_test.go
saumeya: thanks

PR #781, opened 2022-09-19
feat: Remove Argo CD as a base Image from the Export Dockerfile

### LINKED ISSUE #182: Remove Need for Util Image
Currently the operator makes use of a [custom image](https://quay.io/repository/jmckind/argocd-operator-util) that is based on the upstream argo-cd container image for the utilities needed to perform an export. The operator should be refactored to perform the export steps without the need for this image, as it quite large and makes the operator release process as well as offline installs more difficult.

### LINKED ISSUE #279: ArgoCDExport to AWS not working as expected
There are currently two issues with the export job to AWS:

1. **There is no way to specify your default region**, so the script only supports buckets in AWS default region (USEast). A simple fix for this would be to add a key names `aws.bucket.region` to the kubernetes Secret used by ArgoCDExport, let the backup script use this and add a `--region` argument when creating the S3 bucket. Without a default region specified, the script fails with the following error message: `make_bucket failed: s3://<redacted> An error occurred (IllegalLocationConstraintException) when calling the CreateBucket operation: The unspecified location constraint is incompatible for the region specific endpoint this request was sent to.`

2. Even with 1. above fixed, **the script will only work the first time**, because the script attempts to create the bucket on every run. So, if the bucket already exists, the backup will fail. If the script should indeed create the bucket, it needs to first check if it already exists. Alternatively, you could assume that the bucket always exists, thereby removing the need to check and create the bucket. With the current implementation, the backup fails with the following error message: `make_bucket failed: s3://<redacted> An error occurred (BucketAlreadyOwnedByYou) when calling the CreateBucket operation: Your previous request to create the named bucket succeeded and you already own it.`

Juding by the way the [util.sh](https://github.com/argoproj-labs/argocd-operator/blob/master/build/util/util.sh) script is constructed, p.2 above could potentially be relevant also when backing up to Google Cloud or Azure.

### LINKED ISSUE #596: ArgoCD Export to GCP backend fails - gcloud not in PATH
**Describe the bug**
When the export job to a gcp backend is created, the container crashes with the following error:
```
$ kubectl logs -n argocd-production argocd-export-gcp-m7qc8 -f
exporting argo-cd
creating argo-cd backup
encrypting argo-cd backup
pushing argo-cd backup to gcp
/usr/local/bin/argocd-operator-util: line 87: gcloud: command not found
```

**To Reproduce**
Set up an export to GCP:
```
kind: ArgoCDExport
metadata:
  name: argocd-export-gcp
  namespace: argocd-production
  labels:
    test: version-1
spec:
  argocd: argocd
  storage:
    backend: gcp
    secretName: argocd-export-gcp-secret
```

The the export is created, the above error occurs.

Checking the pod, I find the following image:
`quay.io/argoprojlabs/argocd-operator-util@sha256:411388a09cbba0502639e0190a68d82ae2d7716863e35b2af7322384b2c6af48`

Run that image, looking for gcloud command. Gcloud is installed, but not in the `PATH`

```
docker run --rm -ti --name tmp --entrypoint bash quay.io/argoprojlabs/argocd-operator-util@sha256:411388a09cbba0502639e0190a68d82ae2d7716863e35b2af7322384b2c6af48

argocd@d9946953d046:~$ which gcloud
argocd@d9946953d046:~$ cd /
argocd@d9946953d046:/$ find -name gcloud
find: './var/cache/apt/archives/partial': Permission denied
find: './var/cache/ldconfig': Permission denied
find: './proc/tty/driver': Permission denied
./home/argocd/google-cloud-sdk/lib/googlecloudsdk/googledata/devsite/site-cloud/en/sdk/gcloud
./home/argocd/google-cloud-sdk/platform/gsutil_py2/gslib/vendored/oauth2client/tests/data/gcloud
./home/argocd/google-cloud-sdk/bin/gcloud
./home/argocd/google-cloud-sdk/.install/.backup/lib/googlecloudsdk/googledata/devsite/site-cloud/en/sdk/gcloud
./home/argocd/google-cloud-sdk/.install/.backup/bin/gcloud
find: './etc/ssl/private': Permission denied
find: './root': Permission denied

argocd@d9946953d046:/$ echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

Need to include add `./home/argocd/google-cloud-sdk/bin/` to the `PATH`

iam-veeramalla:

**What type of PR is this?**
> /kind bug

Closes #182 
Closes #279 
Closes #596 

NA

Fixes #182 
Fixes #279 
Fixes #596 

1. Pull the PR on to your local and build a export image using `make util-build`
(or)
You can alternatively use the image that I built for testing 
Image - abhishekf5/argocdexportimage
Version - "sha256:c468e2ce7d9c740b6704239fcbc2bfb853c127dbf8b133e322cb2ddc889c84d2"

2. Copy the util image SHA and update to the defaults.go 
- `ArgoCDDefaultExportJobImage`
- `ArgoCDDefaultExportJobVersion`

3. Run the Argo CD Operator using `make install run`
4. Verify the util image for back up to your favorite cloud provider.

`logs which shows s3 bucket is **NOT** created if it already exist`

```
argocd-operator aveerama$ kubectl logs example-argocdexport-xvr2q 
exporting argo-cd
creating argo-cd backup
encrypting argo-cd backup
pushing argo-cd backup to aws
upload: ../../backups/argocd-backup.yaml to s3://rk-testargoexport-todayrrr/argocd-backup.yaml
argo-cd export complete
```

`logs which shows s3 bucket **is** created if it already exist`

```
argocd-operator aveerama$ kubectl logs example-argocdexport-7ph7x 
exporting argo-cd
creating argo-cd backup
encrypting argo-cd backup
pushing argo-cd backup to aws
make_bucket: rk-testargoexport-todayrrr
upload: ../../backups/argocd-backup.yaml to s3://rk-testargoexport-todayrrr/argocd-backup.yaml
argo-cd export complete
```

### DIFF
--- a/build/util/Dockerfile
+++ b/build/util/Dockerfile
@@ -1,24 +1,43 @@
-# Argo CD v2.5.3
-FROM quay.io/argoproj/argocd@sha256:8283a9f06033c2377dc61b03daf4994a3ab961c53d79ed32b9aebadf79bb4858
+FROM quay.io/argoproj/argocd as argocd
+
+# Final Image
+FROM docker.io/library/ubuntu:22.04
 
 USER root
 
-# Ensure system dependencies are installed
-RUN apt-get update && \
-    apt-get install -y curl python3-pip && \
+ENV ARGOCD_USER_ID=999
+
+RUN groupadd -g $ARGOCD_USER_ID argocd && \
+    useradd -r -u $ARGOCD_USER_ID -g argocd argocd && \
+    mkdir -p /home/argocd && \
+    chown argocd:0 /home/argocd && \
+    chmod g=u /home/argocd && \
+    apt-get update && \
+    apt-get dist-upgrade -y && \
+    apt-get install -y \
+    git git-lfs tini curl python3-pip gpg tzdata && \
+    apt-get clean && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Install the AWS CLI
 RUN pip3 install awscli
+
+# Install the Microsoft Azure CLI
+RUN curl -sL https://aka.ms/InstallAzureCLIDeb | bash && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Install the Google Cloud SDK (CLI)
 RUN curl -sL https://sdk.cloud.google.com > /tmp/install.sh && \
     bash /tmp/install.sh --disable-prompts --install-dir=/home/argocd && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-# Install the Microsoft Azure CLI
-RUN curl -sL https://aka.ms/InstallAzureCLIDeb | bash && \
-    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+ENV PATH="/usr/local/google-cloud-sdk/bin:${PATH}" 
+
+COPY entrypoint.sh /usr/local/bin/uid_entrypoint.sh
+RUN chmod +x /usr/local/bin/uid_entrypoint.sh
+
+# Install Argo CD CLI from argocd
+COPY --from=argocd /usr/local/bin/argocd /usr/local/bin/argocd
 
 # Copy util wrapper script
 COPY util.sh /usr/local/bin/argocd-operator-util
--- a/build/util/Dockerfile
+++ b/build/util/Dockerfile
@@ -1,4 +1,5 @@
-FROM quay.io/argoproj/argocd as argocd
+# Argo CD v2.5.4
+FROM quay.io/argoproj/argocd@sha256:d5c4f7542580f275ea92fd07da8569cfe9fb6e49b75bbefef01255ccc4b77d57 as argocd
 
 # Final Image
 FROM docker.io/library/ubuntu:22.04
--- a/build/util/Dockerfile
+++ b/build/util/Dockerfile
@@ -6,6 +6,7 @@ FROM docker.io/library/ubuntu:22.04
 USER root
 
 ENV ARGOCD_USER_ID=999
+ENV DEBIAN_FRONTEND noninteractive
 
 RUN groupadd -g $ARGOCD_USER_ID argocd && \
     useradd -r -u $ARGOCD_USER_ID -g argocd argocd && \
--- a/build/util/entrypoint.sh
+++ b/build/util/entrypoint.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# If we're started as PID 1, we should wrap command execution through tini to
+# prevent leakage of orphaned processes ("zombies").
+if test "$$" = "1"; then
+	exec tini -- $@
+else
+	exec "$@"
+fi
--- a/build/util/util.sh
+++ b/build/util/util.sh
@@ -21,6 +21,7 @@ BACKUP_FILENAME=argocd-backup.yaml
 BACKUP_EXPORT_LOCATION=/tmp/${BACKUP_FILENAME}
 BACKUP_ENCRYPT_LOCATION=/backups/${BACKUP_FILENAME}
 BACKUP_KEY_LOCATION=/secrets/backup.key
+DEFAULT_BACKUP_BUCKET_REGION="us-east-1"
 
 export_argocd () {
     echo "exporting argo-cd"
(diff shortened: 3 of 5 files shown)


build/util/Dockerfile
```diff
@@ -1,24 +1,43 @@
-# Argo CD v2.4.7
-FROM quay.io/argoproj/argocd@sha256:f887f854ab22f7f29f915aae2b765f2948d1555d61e9ce3ca9e659f8df22ab2b
+FROM quay.io/argoproj/argocd as argocd
+
+# Final Image
+FROM docker.io/library/ubuntu:22.04
```
iam-veeramalla, line 4: Fixes #182

build/util/Dockerfile
```diff
@@ -1,24 +1,43 @@
...
+
 # Install the Google Cloud SDK (CLI)
 RUN curl -sL https://sdk.cloud.google.com > /tmp/install.sh && \
     bash /tmp/install.sh --disable-prompts --install-dir=/home/argocd && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-# Install the Microsoft Azure CLI
-RUN curl -sL https://aka.ms/InstallAzureCLIDeb | bash && \
-    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+ENV PATH="/usr/local/google-cloud-sdk/bin:${PATH}"
```
iam-veeramalla, line 34: Fixes #596

build/util/util.sh
```diff
@@ -61,7 +61,7 @@ push_aws () {
     echo "pushing argo-cd backup to aws"
     BACKUP_BUCKET_NAME=`cat /secrets/aws.bucket.name`
     BACKUP_BUCKET_URI="s3://${BACKUP_BUCKET_NAME}"
-    aws s3 mb ${BACKUP_BUCKET_URI}
+    aws s3 mb ${BACKUP_BUCKET_URI} --region us-east-1
```
iam-veeramalla, line 64: Fixes #279

build/util/entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
iam-veeramalla, line 1: This files is a new addition which is a part of the fix #182

iam-veeramalla:
@sagikazarmark This PR adds some major changes to the export Dockerfile like `removing the need for Argo CD base image`. FYI

build/util/util.sh
sagikazarmark: It still attempts to create a bucket though. I can't imagine any single enterprise environment where this isn't going to be a major red flag.

Also, it's an implicit behavior not necessarily easy to find, so I'd strongly recommend to get rid of it completely.

sagikazarmark (commented):
Thanks for the ping @iam-veeramalla 

I had one comment, but it looks good overall.

build/util/util.sh
wtam2018: Hardcoding `--region us-east-1` is probably a bad idea.

iam-veeramalla: If you send your create bucket request to the s3.amazonaws.com endpoint, the request go to the us-east-1 region by default.  

The behavior of Argo CD Export has always been to create bucket in us-east-1, This change is required as the new version of AWS CLI requires a region to be provided. This can be parameterized to allow users to provide region of their choice but I don't see much use.

wtam2018: Default is one thing. There should be a default but hardcoding to always use us-east-1 is the right thing to do?

iam-veeramalla:
> Thanks for the ping @iam-veeramalla
> 
> I had one comment, but it looks good overall.

Totally agree to this but let's take this as a different change and I would like to ask this in the discussions before I remove it :)

sagikazarmark:
@iam-veeramalla do you think this change will make it to the next release?

build/util/util.sh
iam-veeramalla: @wtam2018 
sorry I missed this one, 

so previously, the command to create s3 bucket using the aws CLI was 
`aws s3 mb <foo>` -> which would create the bucket `foo` in `us-east-1` region.

But in the new version of AWS CLI, the command got changed to 
`aws s3 mb <foo> --region <us-east-1>` 

I tried to keep the behavior similar.

Are you saying we should parameterize the region using some command line arguments ? and may be keep the default(if nothing is provided by the user) as `us-east-1` ?

wtam2018: Yes. I am suggesting to parameterize the region.

iam-veeramalla: @wtam2018 made requested changes. PTAL

wtam2018 (approved):

PR #793, opened 2022-10-10
add support for new subkeys in resource customizations

reginapizza:
**What type of PR is this?**
 /kind enhancement

This PR adds support for new resource customization subkeys that were introduced in Argo CD (see[ this article](https://medium.com/argo-project/argo-cd-v2-1-first-release-candidate-is-ready-c1aab7795638) for more information about that)

In gitops-operator: https://github.com/redhat-developer/gitops-operator/issues/248
JIRA issue: https://issues.redhat.com/browse/GITOPS-1561 

1. start your k8s cluster
2. run `make docker-build`, `make docker-push`, and `make deploy`
3. add the argocd instance:
```
kind: ArgoCD
metadata:
  name: argocd
spec: {}
EOF
```
4. run `oc edit argocd argocd`
5. add the following customizations for `resourceHealthChecks`, `resourceIgnoreDifferences`, and `resourceActions` to your argocd spec: 
```
spec:
  resourceIgnoreDifferences:
    all:
      jqPathExpressions:
          - xyz
          - abc
      jsonPointers:
          - xyz
          - abc
      managedFieldsManagers:
          - xyz
          - abc
    resourceIdentifiers:
      - group: apps
        kind: deployments
        customization:
          jqPathExpressions:
            - xyz
            - abc
          jsonPointers:
            - xyz
            - abc
          managedFieldManagers:
            - xyz
            - abc
      - group: batch
        kind: jobs
        customization:
          jqPathExpressions:
            - xyz
            - abc
          jsonPointers:
            - xyz
            - abc
          managedFieldsManagers:
            - xyz
            - abc
  resourceHealthChecks:
    - group: certmanager.k8s.io
      kind: Certificate
      check: |
        hs = {}
        if obj.status ~= nil then
          if obj.status.conditions ~= nil then
            for i, condition in ipairs(obj.status.conditions) do
              if condition.type == "Ready" and condition.status == "False" then
                hs.status = "Degraded"
                hs.message = condition.message
                return hs
              end
              if condition.type == "Ready" and condition.status == "True" then
                hs.status = "Healthy"
                hs.message = condition.message
                return hs
              end
            end
          end
        end
        hs.status = "Progressing"
        hs.message = "Waiting for certificate"
        return hs
  resourceActions:
    - group: apps
      kind: Deployment
      action: |
        discovery.lua: |
        actions = {}
        actions["restart"] = {}
        return actions
        definitions:
        - name: restart
          # Lua Script to modify the obj
          action.lua: |
            local os = require("os")
            if obj.spec.template.metadata == nil then
                obj.spec.template.metadata = {}
            end
            if obj.spec.template.metadata.annotations == nil then
                obj.spec.template.metadata.annotations = {}
            end
            obj.spec.template.metadata.annotations["kubectl.kubernetes.io/restartedAt"] = os.date("!%Y-%m-%dT%XZ")
            return obj
```

6. check your argocd-cm to make sure it was added correctly with `kubectl describe configmap argocd-cm`.


jaideepr97:
heads up @reginapizza 
there have been some changes to the deprecation event emission logic since https://github.com/argoproj-labs/argocd-operator/pull/797 was merged

The problem with having a single toggle variable is that once it has been toggled after the first time the event is emitted, the event will not be emitted for any other namespace/instance that follows and may also need that event 

Instead we may be able to extend the new map that has been put in place to track which instances have emitted the event already, and which ones have not (but need to). This would help ensure that all instances (and by extension the teams that may be controlling those instances) using the old resource customizations format have sufficient warning about it

controllers/argocd/configmap.go
```diff
@@ -331,9 +360,27 @@ func (r *ReconcileArgoCD) reconcileArgoConfigMap(cr *argoprojv1a1.ArgoCD) error
...
+		for k, v := range c {
+			cm.Data[k] = v
+		}
+	}
+
+	// old format of ResourceCustomizations, use should not be encouraged but is supported for now
 	if c := getResourceCustomizations(cr); c != "" {
+		// Emit event providing users with deprecation notice for ResourceCustomization
+		if !ResourceCustomizationDeprecationWarningEmitted {
+			err := argoutil.CreateEvent(r.Client, "Warning", "Deprecated", "ResourceCustomizations is deprecated, please use the new format, ResourceCustomizationsNew, instead.", "DeprecationNotice", cr.ObjectMeta, cr.TypeMeta)
```
jaideepr97, line 375: If we have a timeline for when we will remove support for the old format we should mention it here

wtam2018 (commented):
I don't see any documentation updates.

controllers/argocd/configmap.go
```diff
@@ -331,9 +362,42 @@ func (r *ReconcileArgoCD) reconcileArgoConfigMap(cr *argoprojv1a1.ArgoCD) error
...
+		// Emit event providing users with deprecation notice for ResourceCustomization
+		// if !ResourceCustomizationDeprecationWarningEmitted {
+		// err := argoutil.CreateEvent(r.Client, "Warning", "Deprecated", "ResourceCustomizations is deprecated, please use the new format, ResourceCustomizationsNew, instead.", "DeprecationNotice", cr.ObjectMeta, cr.TypeMeta)
+		// ResourceCustomizationDeprecationWarningEmitted = true
+		// if err != nil {
+		// 	return err
+		// }
+		// }
+
+		// Emit event providing users with deprecation notice for ResourceCustomization if not emitted already
```
jaideepr97, line 384: we can probably get rid of the commented code

controllers/argocd/configmap.go
```diff
@@ -144,15 +144,46 @@ func getRBACScopes(cr *argoprojv1a1.ArgoCD) string {
...
+
+var RCdeprecationEventEmissionTracker = make(map[string]RCdeprecationEventEmissionStatus)
+
+// GetResourceCustomizations loads Resource Customizations from argocd-cm ConfigMap. Please note that ResourceCustomizations are being deprecated in favor of ResourceCustomizationsNew.
 func getResourceCustomizations(cr *argoprojv1a1.ArgoCD) string {
 	rc := common.ArgoCDDefaultResourceCustomizations
 	if cr.Spec.ResourceCustomizations != "" {
 		rc = cr.Spec.ResourceCustomizations
 	}
+	log.Info("You are using an old format for ResourceCustomizations, please use the new format.")
```
iam-veeramalla, line 159: This should ideally be a warning but I understand using the current logging implementation its not possible.

So, can we at least do this ?
	log.Info("[WARNING] You are using an old format for ResourceCustomizations, please use the new format.")
	
	and this in future with the zap logger implementation can be changed to `log.warn`.

controllers/argocd/configmap.go
iam-veeramalla: Yes, Please

controllers/argocd/configmap_test.go
```diff
@@ -31,6 +31,7 @@ import (
 
 	"github.com/argoproj-labs/argocd-operator/api/v1alpha1"
 	argoprojv1alpha1 "github.com/argoproj-labs/argocd-operator/api/v1alpha1"
+	argov1alpha1 "github.com/argoproj-labs/argocd-operator/api/v1alpha1"
```
iam-veeramalla, line 34: This is a duplicate import. 
`argoprojv1alpha1` and `argov1alpha1` are both the same.

controllers/argocd/configmap.go
```diff
@@ -144,15 +144,46 @@ func getRBACScopes(cr *argoprojv1a1.ArgoCD) string {
 	return scopes
 }
 
-// getResourceCustomizations will return the resource customizations for the given ArgoCD.
+type RCdeprecationEventEmissionStatus struct {
+	ResourceCustomizationsDeprecationWarningEmitted bool
+}
+
+var RCdeprecationEventEmissionTracker = make(map[string]RCdeprecationEventEmissionStatus)
```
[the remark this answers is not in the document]
iam-veeramalla: If you would still consider to use it, Please change the variable 
`RCdeprecationEventEmissionTracker` to `RCDeprecationEventEmissionTracker`.

iam-veeramalla (dismissed):

controllers/argocd/configmap_test.go
```diff
@@ -791,6 +792,114 @@ func TestReconcileArgoCD_reconcileArgoConfigMap_withResourceCustomizations(t *te
 	}
 }
 
+func TestReconcileArgoCD_reconcileArgoConfigMap_withNewResourceCustomizations(t *testing.T) {
```
iam-veeramalla, line 795: Thanks for the unit tests :) . Can you also write some e2e tests ?

iam-veeramalla (commented):
I will test the PR locally and update the results.

api/v1alpha1/argocd_types.go
```diff
@@ -339,6 +339,18 @@ type ArgoCDPrometheusSpec struct {
 	Size *int32 `json:"size,omitempty"`
 }
 
+// Configuration to customize resource behavior (optional) can be configured via splitted sub keys.
+type ResourceCustomizationsNew struct {
```
wtam2018, line 343: We probably should find a better name than "New" as it will get old pretty quickly.

jaideepr97:
tested it locally and it works fine 
LGTM, thanks @reginapizza

api/v1alpha1/argocd_types.go
wtam2018: After all, this new struct is not necessary.

controllers/argocd/configmap.go
```diff
@@ -144,15 +144,42 @@ func getRBACScopes(cr *argoprojv1a1.ArgoCD) string {
 	return scopes
 }
 
-// getResourceCustomizations will return the resource customizations for the given ArgoCD.
+// GetResourceCustomizations loads Resource Customizations from argocd-cm ConfigMap. Please note that ResourceCustomizations are being deprecated in favor of ResourceCustomizationsNew.
```
wtam2018, line 147: Don't change it to capitalization of the func name.

wtam2018 (dismissed):

api/v1alpha1/argocd_types.go
```diff
@@ -689,10 +710,22 @@ type ArgoCDSpec struct {
 	// RepositoryCredentials are the Git pull credentials to configure Argo CD with upon creation of the cluster.
 	RepositoryCredentials string `json:"repositoryCredentials,omitempty"`
 
-	// ResourceCustomizations customizes resource behavior. Keys are in the form: group/Kind.
+	// ResourceCustomizations customizes resource behavior. Keys are in the form: group/Kind. Please note that this is being deprecated in favor of in favor of ResourceHealthChecks, ResourceIgnoreDifferences, and ResourceActions.
```
jaideepr97, line 713: @reginapizza small typo here

tests/k8s/1-025_validate_resource_ignore_differences/01-argocd-with-resource-ignore-difference.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: argoproj.io/v1alpha1
```
jaideepr97, line 1: @reginapizza I think we should name this file `02-argocd-with-resource-ignore-difference` instead
kuttl uses the naming of the file as indication for the order in which the steps should be carried out 

01-whatever is followed by 01-assert, followed by 02-whatever followed by 02-assert for eg

right now it seems like there are multiple asserts for the same step

tests/k8s/1-025_validate_resource_ignore_differences/01-assert.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: kuttl.dev/v1beta1
```
jaideepr97, line 1: Likewise this should be `02-assert.yaml`

tests/k8s/1-026_validate_resource_action/01-argocd-with-resource-action.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: argoproj.io/v1alpha1
```
jaideepr97, line 1: `02-argocd-with-resource-action,yaml`

tests/k8s/1-026_validate_resource_action/01-assert.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: kuttl.dev/v1beta1
```
jaideepr97, line 1: `03-assert.yaml`

jaideepr97:
@reginapizza some minor comments that should help the e2e test pass
other than that LGTM :)

jaideepr97:
> some minor comments that should help the e2e test pass
Oops, it seems I was mistaken when I made those comments, feel free to ignore

controllers/argocd/configmap.go
```diff
@@ -144,15 +144,60 @@ func getRBACScopes(cr *argoprojv1a1.ArgoCD) string {
...
-// getResourceCustomizations will return the resource customizations for the given ArgoCD.
+// getResourceCustomizations loads Resource Customizations from argocd-cm ConfigMap
 func getResourceCustomizations(cr *argoprojv1a1.ArgoCD) string {
 	rc := common.ArgoCDDefaultResourceCustomizations
 	if cr.Spec.ResourceCustomizations != "" {
 		rc = cr.Spec.ResourceCustomizations
 	}
+
+	// TODO: change this to log.warn once it's available with zap logger
+	log.Info("[WARNING] You are using an old format for ResourceCustomizations, please use the new format.")
```
wtam2018, line 155: I'd get rid of it in that case. We already log events, right?

controllers/argocd/configmap.go
```diff
@@ -144,15 +144,60 @@ func getRBACScopes(cr *argoprojv1a1.ArgoCD) string {
...
 	return rc
 }
 
+// preferred way to add health customizations with `resource.customizations.health`
```
wtam2018, line 159: I'd get rid of  "preferred way". Once the old way is removed, this comment becomes very puzzling as it suggests that there are other ways to do it.

api/v1alpha1/argocd_types.go
```diff
@@ -541,6 +541,27 @@ type ArgoCDServerServiceSpec struct {
...
+	Group string `json:"group"`
+	Kind  string `json:"kind"`
+	Check string `json:"check"`
+}
+
+// Resource Customization for ignore difference
+type ResourceIgnoreDifference struct {
+	Group             string `json:"group"`
+	Kind              string `json:"kind"`
+	JqPathExpressions string `json:"JqPathExpressions"`
```
wtam2018, line 555: Should `JqPathExpressions` be a `[]string`?

docs/reference/argocd.md
```diff
@@ -957,10 +957,133 @@ spec:
 
 ## Resource Customizations
 
-The configuration to customize resource behavior. This property maps directly to the `resource.customizations` field in the `argocd-cm` ConfigMap.
+There are two ways to customize resource behavior: with subkeys (`resourceHealthChecks`, `resourceIgnoreDifferences`, and `resourceActions`) and without subkeys (`resourceCustomizations`). `resourceCustomizations` maps directly to the `resource.customizations` field in the `argocd-cm` ConfigMap, while each of the subkeys maps directly to their own field in the `argocd-cm`. `resourceHealthChecks` will map to `resource.customizations.health`, `resourceIgnoreDifferences` to `resource.customizations.ignoreDifferences`, and `resourceActions` to `resource.customizations.actions`
```
wtam2018, line 960: The documentation is versioned, right? The user will need to know in which version the subkey config is available.

docs/reference/argocd.md
```diff
@@ -957,10 +957,133 @@ spec:
...
+            if obj.spec.template.metadata.annotations == nil then
+                obj.spec.template.metadata.annotations = {}
+            end
+            obj.spec.template.metadata.annotations["kubectl.kubernetes.io/restartedAt"] = os.date("!%Y-%m-%dT%XZ")
+            return obj
+  resourceIgnoreDifferences:
+  - group: admissionregistration.k8s.io
+    kind: MutatingWebhookConfiguration
+    jqPathExpressions: |
+      jsonPointers: |
```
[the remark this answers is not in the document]
wtam2018: It does look like there are other ignore differences customization we can do, namely `managedFieldsManagers` and `jsonPointers` See https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/#system-level-configuration.  Also, It is possible to configure ignoreDifferences to be applied to all resources by setting this key `resource.customizations.ignoreDifferences.all`.  Is it supported/tested by the operator?

api/v1alpha1/argocd_types.go
jaideepr97: yeah probably should be

jaideepr97: Actually, I think we just take whatever is specified by the user as a single string and put it in the config map because that is how it is documented upstream as well since it is all behind a `|`
```
data:
  resource.customizations.ignoreDifferences.admissionregistration.k8s.io_MutatingWebhookConfiguration: |
    jqPathExpressions:
    - '.webhooks[]?.clientConfig.caBundle'
```

so this probably doesn't need to be an array, user is responsible for specifying how many ever expressions they want to

docs/reference/argocd.md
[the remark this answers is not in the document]
jaideepr97: > It does look like there are other ignore differences customization we can do, namely managedFieldsManagers and jsonPointers See https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/#system-level-configuration

I think those are all behind a `|` so users should be able to specify them if they want and we would just take them as is and paste them here 
That was the reason the 3rd fields were all originally just called `Customization` instead of `check` and `action` and `jqexpressionpath`, maybe we should go back to that 

 >resource.customizations.ignoreDifferences.all. Is it supported/tested by the operator?
 
 not at present to my knowledge, this would need additional changes but should be straight forward

PR #794, opened 2022-10-12
changes to support the feature apps in any namespaces

ishitasequeira:
**What type of PR is this?**
/kind enhancement

The PR adds support for the feature apps in any namespace added upstream. To do so, the PR adds new command args `--additional-namespace` to ArgoCD ApplicationController and ArgoCD Server components. Also, the PR adds new roles in the respective namespaces for ArgoCD Server to perform actions in the namespace.

https://issues.redhat.com/browse/GITOPS-2341

* Run `make install run`
* Create namespaces `argocd-test` and `argocd-1`
* Deploy an ArgoCD instance using below yaml file
```
kind: ArgoCD
metadata:
  name: example-argocd
spec:
  controller:
    sourceNamespaces:
    - argocd
    - argocd-1
  notifications:
    enabled: True
    env:
      - name: foo
        value: bar
  server:
    ingress:
      enabled: true
    sourceNamespaces:
    - argocd-test
    - argocd-1
```
* Check if the ArgoCD pods came up successfully using command `kubectl get pods | grep example-argocd`
* Check whether the new roles were created in namespaces `argocd-test` and `argocd-1`
* Check if the namespaces `argocd-test` and `argocd-1` have a new label `argocd.argoproj.io/managed-by-cluster-argocd`

* e2e test for the PR.
`kubectl kuttl test ./tests/k8s --config ./tests/kuttl-tests.yaml --test 1-024_validate_apps_in_any_namespace`


controllers/argocd/rolebinding.go
```diff
@@ -208,6 +208,11 @@ func getCustomRoleName(name string) string {
 	return ""
 }
 
+// Returns the name of the role for the supported namespaces for ArgoCDServer in the format of "SupportedNameSpace_SERVER_CLUSTER_ROLE"
+func getCustomRoleNameForSupportedNamespaces(name string, namespace string) string {
+	return fmt.Sprintf("%s_%s", namespace, common.ArgoCDServerClusterRoleEnvName)
```
jaideepr97, line 213: @ishitasequeira 
why are we appending SERVER_CLUSTER_ROLE to the name of this role? Sounds like it would be regular role and not actually a cluster role

controllers/argocd/policyrule.go
```diff
@@ -209,6 +209,27 @@ func policyRuleForNotificationsController() []v1.PolicyRule {
 	}
 }
 
+func policyRuleForServerSupportedNamespaces() []v1.PolicyRule {
```
jaideepr97, line 212: @ishitasequeira 
if a namespace is already being managed by argo-cd (has the managed-by label) then that namespace will already have a server role with permissions to manage applications (https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/policyrule.go#L117)
this might lead to redundancy if we create a separate role with a subset of the same permissions

do we want to first check if this is the case?

jaideepr97:
@ishitasequeira 

what happens if the same namespace features in the "sourceNamespaces" list of 2 different argo-cd instances?
would both instances try to manage the same application?

I think it would also be good to have some e2e tests written out that can cover scenarios like this

ishitasequeira:
> @ishitasequeira
> 
> what happens if the same namespace features in the "sourceNamespaces" list of 2 different argo-cd instances? would both instances try to manage the same application?
> 

Yes, we would reconcile the roles for both the argo-cd instances.

controllers/argocd/rolebinding.go
ishitasequeira: Updated the role name to be of the format `sourceNamespace_targetNamespace_argocd-server`

jaideepr97:
@ishitasequeira that might be a problem, no?
having multiple argo-cd instances managing the same application doesn't sound like a good idea to me

ishitasequeira:
> @ishitasequeira that might be a problem, no? having multiple argo-cd instances managing the same application doesn't sound like a good idea to me
> 
> should we maybe consider some kind of mechanism to check if an app is already being managed by some other instance?

As mentioned in @jannfis's upstream [PR](https://github.com/argoproj/argo-cd/pull/9755) description,

```
If there are multiple instances of Argo CD installed to the same cluster, the behavior when more than one instance specifies same allowed namespace patterns (and have access to those namespaces) is undefined. This scenario should be avoided (needs to be documented).
```

I will update our documentation with this comment.

jaideepr97:
@ishitasequeira 
I see
I suppose we could leave it to the users to be careful...however, I think since we are an operator (as opposed to upstream) maybe there is some level of expectation that we would go further to make the users' experience more convenient? 
I'm fine with leaving it up to the user, but I think it might be worth discussing if we should try and put some guardrails in place all the same

sbose78:
@jaideepr97 Agreed. A little over-engineered solution could be having an admission webhook disallow specification of a namespace in `sourceNamespaces` if it was already specified for a different ArgoCD instance.

jaideepr97:
@sbose78 that could work 
maybe a slightly more straightforward way could be to just maintain a map to track which namespaces are watched by which instances, and not create roles when we encounter a namespace that's already being watched by a different instance 
(Also log it the operator and emit a warning event if necessary mentioning that xyz namespace is already being watched?)

ishitasequeira:
> @sbose78 that could work maybe a slightly more straightforward way could be to just maintain a map to track which namespaces are watched by which instances, and not create roles when we encounter a namespace that's already being watched by a different instance (Also log it the operator and emit a warning event if necessary mentioning that xyz namespace is already being watched?)

I think managing a map might be difficult as we would have to take care of cleaning it up too if the ArgoCD instance is updated later and a namespace is removed from it.. I was wondering if we could use a label like `[argocd.argoproj.io/managed-by](http://argocd.argoproj.io/managed-by)` and we check if the namespace has this label or not.

jaideepr97:
> I think managing a map might be difficult as we would have to take care of cleaning it up too if the ArgoCD instance is updated later and a namespace is removed from it.

@ishitasequeira FWIW I don't think cleanup would be too bad, we would just need to watch for changes in the CR when a namespace is removed from the watch list, and when a namespace is deleted - in both cases we check if that namespace has any roles that need to be cleaned up & then remove that entry from the map

(We already have similar watches in place to react to different events like removal of SSO in CR or deletion of managed namespace)

ishitasequeira:
As agreed in the [slack conversation](https://coreos.slack.com/archives/C01RQH8KQ87/p1666799197470809), we are attaching a label `argocd.argoproj.io/managed-by-cluster-argocd` to every namespace added as a `SourceNamespace`.

ishitasequeira:
@iam-veeramalla @jopit @jaideepr97 could you review the PR?

jaideepr97:
@ishitasequeira should we maybe include the name/namespace of the managing cluster argo-cd in the label? in case we have multiple cluster scoped instances on the same cluster it could get difficult to tell who is managing that namespace (even though there's no risk of multiple instances managing that namespace)

controllers/argocd/role.go
```diff
@@ -156,6 +176,86 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
+		}
+		if managedNamespace {
+			continue
+		}
+		log.Info(fmt.Sprintf("Reconciling role for %s", namespace.Name))
+
+		// Update namespace with managed-by-cluster-argocd label
+		namespace.Labels[common.ArgoCDManagedByClusterArgoCDLabel] = cr.Name
+		if err := r.Client.Update(context.TODO(), namespace); err != nil {
+			log.Error(err, fmt.Sprintf("failed to remove label from namespace [%s]", namespace.Name))
```
jaideepr97, line 225: I think this should say "failed to add label to namesapce" instead

jaideepr97 (commented):
@ishitasequeira it seems like we may not be handling clean up in this PR. When someone removes a namespace from a source list, I think the label should be removed from that namespace and the roles and rolebindings will need to be cleaned up as well.

Another point that occurred to me was that if one instance tries to watch a namespace that is already being watched by a different instance we could consider alerting them through a warning event or something else

controllers/argocd/role.go
```diff
@@ -156,6 +176,86 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
+				break
+			}
+		}
+		if managedNamespace {
+			continue
+		}
+		log.Info(fmt.Sprintf("Reconciling role for %s", namespace.Name))
+
+		// Update namespace with managed-by-cluster-argocd label
+		namespace.Labels[common.ArgoCDManagedByClusterArgoCDLabel] = cr.Name
```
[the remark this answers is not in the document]
ishitasequeira: We are setting `cr.Name` as the value for the label. Added a new check for checking if another `cr.Name` is assigned to the label for avoiding overwrite issues.

controllers/argocd/role.go
```diff
@@ -156,6 +176,86 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
+
+		managedNamespace := false
+		// do not reconcile roles for namespaces already containing managed-by label
+		// as it already contains roles reconciled during reconcilation of ManagedNamespaces
+		for _, ns := range r.ManagedNamespaces.Items {
+			if reflect.DeepEqual(ns, namespace) {
+				managedNamespace = true
+				break
+			}
+		}
```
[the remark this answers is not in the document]
ishitasequeira: Added a map of ArgoCD instance to NamespaceList under ReconcileArgocd for optimization and cleanup

api/v1alpha1/argocd_types.go
```diff
@@ -105,6 +105,9 @@ type ArgoCDApplicationControllerSpec struct {
 
 	// Env lets you specify environment for application controller pods
 	Env []corev1.EnvVar `json:"env,omitempty"`
+
+	// SourceNamespaces defines the namespaces application resources are allowed to be created in
+	SourceNamespaces []string `json:"sourceNamespaces,omitempty"`
```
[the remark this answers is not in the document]
ishitasequeira: Updated this to include SourceNamespaces globally and only once.

bundle/manifests/argoproj.io_applications.yaml
```diff
@@ -1806,6 +1806,10 @@ spec:
                   reconciled using the latest git version
                 format: date-time
                 type: string
+              resourceHealthSource:
+                description: 'ResourceHealthSource indicates where the resource health
+                  status is stored: inline if not set or appTree'
+                type: string
```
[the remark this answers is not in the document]
ishitasequeira: Updated ArgoCD version to include the feature. Updated manifests along with it.

controllers/argocd/role.go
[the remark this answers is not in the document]
jaideepr97: should we use the namespace instead of name? might be easier since namespace names are unique but you can have multiple instances with the same name in different namespaces

It would also be consistent with how we use the instance namespace for managed-by label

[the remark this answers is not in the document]
ishitasequeira: I kept it the instance name as we could have multiple argocd instances in the same namespace as well. That way it would be difficult to understand which argocd instance is managing the namespace.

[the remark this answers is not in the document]
jaideepr97: I don't think you can have multiple argocd instances in one namespace 
that sounds like it would cause a lot of problems

controllers/argocd/role.go
```diff
@@ -156,6 +181,114 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
+		if managedNamespace {
+			continue
+		}
+
+		// reconcile roles only if another ArgoCD instance is not already set as value for managed-by-cluster-argocd label
+		if value, ok := namespace.Labels[common.ArgoCDManagedByClusterArgoCDLabel]; !ok || value != "" {
+			if value != "" && value != cr.Name {
+				log.Info(fmt.Sprintf("Namespace already has label set to argocd instance %s. Thus, skipping namespace %s", value, namespace.Name))
+				continue
+			}
```
jaideepr97, line 249: I think this should be 
```
if value, ok := namespace.Labels[common.ArgoCDManagedByClusterArgoCDLabel]; ok && value != "" {
				log.Info(fmt.Sprintf("Namespace already has label set to argocd instance %s. Thus, skipping namespace %s", value, namespace.Name))
				continue
			}

```

controllers/argocd/role.go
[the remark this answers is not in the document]
jaideepr97: @ishitasequeira this is not exactly what i had in mind when I suggested the change, and as far as I can tell isn't really optimizing much 

(https://github.com/argoproj-labs/argocd-operator/pull/794/files#diff-2fe0a2bb5baaf6170a23a7c08adc70117adfa25b7d481923aa8626a38cefd447R45) here I think we can update `ManagedSourceNamespaces` and `ManagedNamespaces` to be of type `map[string]string` where the key is the namespace name and value can be empty. We can see [here](https://github.com/argoproj-labs/argocd-operator/blob/90eadf8356c81bf66bd1928d2354fc6e34aed3ef/controllers/argocd/argocd_controller.go#L127) that the operator repopulates the list of namespaces for a specific instance each time that instance is being reconciled. This means that it is as good as saying each instance has its own copy of `ReconcileArgoCD` (and `ManagedSourceNamespaces` and `ManagedNamespaces` by extension). So we don't need to store a list of watched source namespaces for ALL argo-cd instances in the map, just for the current instance in question.

By using a map instead of a list, instead of lines 202-207 we can do something like 
```
if _, ok := r.ManagedNamespaces[sourceNamespace]; ok {
             managedNamespace = true
    }
``` 
since map lookup is O(1) I think considering how many times the operator reconciles, overtime the savings could be huge especially if one instance is managing multiple namespaces 

The same would apply for lines 234-239 but with `r.ManagedSourceNamespaces`

controllers/argocd/role.go
```diff
@@ -156,6 +181,114 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
+					if n.Name == namespace.Name {
+						alreadyManagedSourceNS = true
+						break
+					}
+				}
+				if !alreadyManagedSourceNS {
+					value.Items = append(value.Items, *namespace)
+				}
+			}
+			r.ManagedSourceNamespaces[cr.Name] = value
```
jaideepr97, line 286: If we were to follow what I suggested in the earlier comment, this part of the code would not be needed 
`r.ManagedSourceNamespaces` could be populated the same way as is done [here](https://github.com/argoproj-labs/argocd-operator/blob/90eadf8356c81bf66bd1928d2354fc6e34aed3ef/controllers/argocd/util.go#L1347) for `r.ManagedNamespaces` i.e, populate the map/list each time there is a reconciliation done for a specific argocd instance in question

controllers/argocd/role.go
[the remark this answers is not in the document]
ishitasequeira: Using it the way you have described does make sense. I have addressed this in the following commit.

Also, updating `r.ManagedNamespaces` to map includes bunch of changes.. I would create a separate task for it as part of code optimization.

controllers/argocd/role.go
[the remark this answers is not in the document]
ishitasequeira: Makes sense, have updated it.

controllers/argocd/role.go
ishitasequeira: This was added the it is to make sure that argocd-operator is able to update the map even if the operator restarts and the instance and namespaces preexist with `managed-by-cluster-argocd` label. 

I have updated the logic around Role creation thus removing the above concern.

controllers/argocd/util.go
```diff
@@ -1360,6 +1368,88 @@ func (r *ReconcileArgoCD) setManagedNamespaces(cr *argoproj.ArgoCD) error {
...
+	for _, ns := range r.ManagedSourceNamespaces {
+		managedNamespace := false
+		if cr.DeletionTimestamp == nil {
+			for _, namespace := range cr.Spec.SourceNamespaces {
+				if namespace == ns {
+					managedNamespace = true
+					break
+				}
+			}
+		}
```
jaideepr97, line 1406: Do we need this check here? `r.ManagedSourceNamespaces` is populated by directly checking the cluster for which namespaces have the label assigned to the CR currently being reconciled 
This check seems unnecessary to me

jaideepr97: Ah sorry  I think i see what you're doing here - trying to synchronize `r.ManagedSourceNamespaces` with `cr.spec.SourceNamespaces`
but I think there might be a better way to handle that 


Here we will be running this whole loop each time the operator reconciles, even when no changes to `cr.spec.SourceNamespaces` has been made
instead maybe we could add a predicate here that watches for changes to `cr.spec.SourceNamespaces`. We can compare the lists in the `oldCR` and `newCR` by checking if any namespace is present in `oldCR` but not in `newCR`
for those namespaces we can invole `removeUnmanagedSourceNamespaceResources`
that way we only invoke this function when some actual change has been detected to the list of managed source namespaces

let me know what you think

controllers/argocd/role.go
```diff
@@ -156,6 +181,91 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
+				return nil, fmt.Errorf("failed to reconcile the role for the service account associated with %s : %s", name, err)
+			}
+			log.Info(fmt.Sprintf("Fetching Role Error: %s", err.Error()))
+		}
+
+		// do not reconcile roles for namespaces already containing managed-by-cluster-argocd label
+		// as it already contains roles reconciled during reconcilation of ManagedNamespaces
+		if _, ok := r.ManagedSourceNamespaces[sourceNamespace]; ok {
+			// If sourceNamespace includes the name but role is missing in the namespace, create the role
+			if reflect.DeepEqual(existingRole, v1.Role{}) {
```
jaideepr97, line 228: This should  be 
```
if !reflect.DeepEqual(existingRole, v1.Role{}) { ... }
```

controllers/argocd/role.go
```diff
@@ -156,6 +181,91 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
+	for _, sourceNamespace := range cr.Spec.SourceNamespaces {
+
+		namespace := &corev1.Namespace{}
+		if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: sourceNamespace}, namespace); err != nil {
+			return nil, err
+		}
+
+		managedNamespace := false
+		// do not reconcile roles for namespaces already containing managed-by label
+		// as it already contains roles reconciled during reconcilation of ManagedNamespaces
```
jaideepr97, line 197: could we specify in this comment that the roles created during reconciliation of Managednamespaces includes permissions to manipulate application resources specifically and that is why we don't need to create those roles again here? would be more clear to whoever is reading the code that way

controllers/argocd/util.go
```diff
@@ -1360,6 +1368,88 @@ func (r *ReconcileArgoCD) setManagedNamespaces(cr *argoproj.ArgoCD) error {
...
+
+	return nil
+}
+
+// removeUnmanagedSourceNamespaceResources cleansup resources from SourceNamespaces if namespace is not managed by argocd instance.
+// It also removes the managed-by-cluster-argocd label from the namespace
+func (r *ReconcileArgoCD) removeUnmanagedSourceNamespaceResources(cr *argoproj.ArgoCD) error {
+	if _, ok := r.ManagedSourceNamespaces[cr.Namespace]; !ok {
+		return nil
+	}
```
jaideepr97, line 1395: Please remove this as discussed

controllers/argocd/util.go
```diff
@@ -1360,6 +1368,88 @@ func (r *ReconcileArgoCD) setManagedNamespaces(cr *argoproj.ArgoCD) error {
...
+	for _, namespace := range namespaces.Items {
+		r.ManagedSourceNamespaces[namespace.Name] = ""
+	}
+
+	return nil
+}
+
+// removeUnmanagedSourceNamespaceResources cleansup resources from SourceNamespaces if namespace is not managed by argocd instance.
+// It also removes the managed-by-cluster-argocd label from the namespace
+func (r *ReconcileArgoCD) removeUnmanagedSourceNamespaceResources(cr *argoproj.ArgoCD) error {
```
jaideepr97, line 1392: I think we can simplify this function a little by providing the list of sourceNamespaces along with the cr as inputs 

Then the function just needs to delete the resources and remove the labels from those namespaces 
that might make this function a little more focused in scope and flexible in terms of where and how it is called

controllers/argocd/role.go
ishitasequeira: Will add it.

controllers/argocd/util.go
ishitasequeira: Instead of taking the SourceNamespaces as an input, we are taking it directly from cr.spec.SourceNamespaces. So, the function will not be simplified as such.

controllers/argocd/role.go
ishitasequeira: If we were not able to find the existingRole in line 216, it errors out and continues the operations to create the necessary role by checking that the value of existingRole matches the value `v1.Role{}` (value set at line 215). Thus I believe this part of code does not need to change.

controllers/argocd/util.go
ishitasequeira: I think if we add a predicate to the check for changes in SourceNamespaces, we might not be able to prioritize ManagedBy label over MangedByClusterArgocd label. 

For example, consider the scenario where the namespace was added as SourceNamespace and then the ManagedBy label was added to the namespace. In this case, we would reconcile resources for the namespace only for ManagedBy label and not for MangedByClusterArgocd label.

I think then the question would be, do we need to prioritize ManagedBy label over ManagedByClusterArgocd label as we are reconciling ManagedBy label resources first?

Lemme know WDYT..

controllers/argocd/role.go
jaideepr97: oops yeah you're right 
sorry ignore that

controllers/argocd/util.go
jaideepr97: @ishitasequeira as another small optimization maybe we can rewrite this code to use only one loop instead of a nested one

```
for _, ns := rnage cr.Spec.SourceNamespaces {
   managedNamespace := false 
   if cr.DeletionTimestamp == nil {
     if _, ok := r.ManagedSoruceNamespaces[ns]; ok {
       managedNamesapce = true
       break
       }
     }
   }
```

let me know what you think

ishitasequeira: In the double for loop, what we are trying to do is, we are trying to find out which of the `r.ManagedSoruceNamespaces` is not part of `cr.Spec.SourceNamespaces` so as to perform the cleanup in that namespace. The for loop you have suggested would not help in achieving the same.

jaideepr97: yeah sorry you're right 
maybe `sourceNamespaces` could also be a map to be honest

ishitasequeira: `sourceNamespaces` is part of the  ArgoCD CR. IMO, it would not make sense to ask users to send a map when they just need to send a list of namespaces.

jaideepr97: I meant we can make a map for it before accessing it in the backend logic

controllers/argocd/role.go
```diff
@@ -156,6 +181,92 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
...
+
+		namespace := &corev1.Namespace{}
+		if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: sourceNamespace}, namespace); err != nil {
+			return nil, err
+		}
+
+		// do not reconcile roles for namespaces already containing managed-by label
+		// as it already contains roles with permissions to manipulate application resources
+		// reconciled during reconcilation of ManagedNamespaces
+		if value, ok := namespace.Labels[common.ArgoCDManagedByLabel]; ok {
```
jaideepr97, line 198: here I think we should add
```if value, ok := namespace.Labels[common.ArgoCDManagedByLabel]; ok && value != ""```
if you set managed-by = "" then operator cleans up roles effectively meaning ns is no longer managed, but still we dont create the new role if needed because the logic only checks if the label is present, not if it is actually managed

jaideepr97:
I have tested this PR locally and it works as expected for the most part 
Couple things I have discussed with @ishitasequeira that we should document :

1. No namespace can be managed by multiple argo-cd instances (cluster scoped or namespace scoped) i.e, only one of either `managed-by` or `managed-by-cluster-argocd` labels can be applied to a given namespace. We will be prioritizing `managed-by` label in case of a conflict as this feature is currently in beta, so the new roles/rolebindings will not be created if namespace is already labelled with `managed-by` label, and they will be deleted if a namespace is first added to the `sourceNamespacs` list and is later also labelled with `managed-by` label 

2. This also means users cannot currently create/manage apps and create app resources in the same non-control plane namespace (as they both require their own labels) out of the box 
as a workaround users will have to create custom roles

@ishitasequeira please update the e2e test to ensure proper cleanup of roles/rolebindings as discussed

jaideepr97:
other than that this LGTM, thanks a lot @ishitasequeira !!

ishitasequeira:
@jaideepr97 I have updated the documentation and the e2e test.

@iam-veeramalla @wtam2018 PTAL.

jaideepr97 (changes requested):
@ishitasequeira Please add a step in the e2e test that removes a namespace from a source list, and then checks that the role/rolebinding is removed from the namespace that was removed from the list

jaideepr97 (approved):

jaideepr97:
Merged, thanks again @ishitasequeira  !

PR #795, opened 2022-10-12
fix: Emit sso deprecation event only once for each instance that needs it

jaideepr97:
**What type of PR is this?**

 /kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

This PR fixes a flaw in an earlier fix meant to stop the spamming of SSO related deprecation notice events. 

**Original issue:** operator would spam the cluster with deprecation warning events when it identifies deprecated fields in use
**Original fix**: stopped the spamming but made it so that only the first problematic instance would receive deprecation notice events, operator would stop emitting them completely after that point and subsequent instances would not be notified at all.
**Current fix:** ensures that every problematic instance sees exactly one event generated per deprecated field - such that all instance owners are sufficiently notified and no one is spammed with events. 

Fixes #?

1. Run operator
2. Create multiple argo-cd instances in different namespaces, and ensure they contain either `.spec.dex` or `.spec.sso.verifyTLS` or both (for e.g)
3. Ensure that there is only 1 deprecation notice event created for each deprecated field, in each namespace 
4. Delete argo-cd instance in one of the namespaces and re-create it with the same faults, another set of events should be generated
5. Delete one of the namespaces containing argo-cd instance and recreate it with the same name and new argo-cd instance with the same faults. Confirm that another set of events is generated.

### DIFF
--- a/controllers/argocd/sso.go
+++ b/controllers/argocd/sso.go
@@ -36,10 +36,23 @@ const (
 	multipleSSOConfiguration string = "multiple SSO configuration: "
 )
 
+// DeprecationEventEmissionStatus is meant to track which deprecation events have been emitted already. This is temporary and can be removed in v0.0.6 once we have provided enough
+// deprecation notice
+type DeprecationEventEmissionStatus struct {
+	SSOSpecDeprecationWarningEmitted    bool
+	DexSpecDeprecationWarningEmitted    bool
+	DisableDexDeprecationWarningEmitted bool
+}
+
 var (
 	templateAPIFound     = false
 	ssoConfigLegalStatus string
 )
+
+// DeprecationEventEmissionTracker map stores the namespace containing ArgoCD instance as key and DeprecationEventEmissionStatus as value,
+// where DeprecationEventEmissionStatus tracks the events that have been emitted for the instance in the particular namespace.
+// This is temporary and can be removed in v0.0.6 when we remove the deprecated fields.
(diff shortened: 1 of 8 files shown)


controllers/argocd/sso.go
```diff
@@ -35,15 +36,23 @@ const (
...
+	DexSpecDeprecationWarningEmitted    bool
+	DisableDexDeprecationWarningEmitted bool
+}
 
+var (
 	templateAPIFound     = false
 	ssoConfigLegalStatus string
 )
 
+// map that stores the namespace that contains an Argo CD instance as key, and a DeprecationEventEmissionStatus struct as value, which is responsible for tracking which
```
ishitasequeira, line 52: nit: rephrase

```
DeprecationEventEmissionTracker map stores the namespace containing ArgoCD instance as key and DeprecationEventEmissionStatus as value, 
where DeprecationEventEmissionStatus tracks the events that have been emitted for the instance in the particular namespace.

This is temporary and can be removed in v0.0.6 when we remove the deprecated fields.
```

ishitasequeira (commented):
1 small nit to make it more understandable..

controllers/argocd/sso.go
jaideepr97: done, thanks

ishitasequeira:
@jaideepr97 , the PR needs rebase!!

PR #806, opened 2022-11-18
Update rbac for source namespaces

ishitasequeira:
**What type of PR is this?**
/kind bug

**What does this PR do / why we need it**:
Add missing permissions and update rolebinding reconciliation logic for sourceNamespaces feature

**Have you updated the necessary documentation?**

* [ ] Documentation update is required by this PR.
* [ ] Documentation has been updated.

**Which issue(s) this PR fixes**:

Fixes #?

**How to test changes / Special notes to the reviewer**:

PR #807, opened 2022-11-21
Revert "Update rbac for source namespaces"

wtam2018:
Reverts argoproj-labs/argocd-operator#806


wtam2018:
Per @jaideepr97 https://github.com/argoproj-labs/argocd-operator/pull/806 is still needed although there are more problems.  See below from @jaideepr97 .
The issue we were seeing in testing on Friday is that apps weren't showing up in the UI due to permission issues
after that was fixed the app itself shows up but the dependent resources either weren't showing up or weren't getting created even when argo-cd said the app is healthy and synced
So the PR was open to keep tweaking it and seeing what other permission issues could be fixed through there

Issue #816, opened 2022-11-25
Argo CD .status.host is updated with wrong value when using Ingress on OpenShift clusters


iam-veeramalla:
**Describe the bug**

Argo CD .status.host is updated with wrong value when using Ingress on OpenShift clusters

`Openshift 4.11`
==================

Create test-namespace 'test-ns'

**Create Argo CD instance with the manifest** 
```
    apiVersion: argoproj.io/v1alpha1    
    kind: ArgoCD    
    metadata:      
      name: example-argocd      
      namespace: test-ns
      labels:        
        example: ingress    
    spec:      
      server:        
        host: example-argocd.<apps_domain_name>
        grpc:          
          ingress:            
            enabled: true        
        ingress:          
          enabled: true          
          tls:          
          - hosts:              
            - example-argocd        
        insecure: true
```

**Actual results:**

The ingress is created and it's getting Argo CD's .spec.server.host as HOSTS (.spec.rules[0].host):

```
NAME                    CLASS    HOSTS                                                    ADDRESS                                             
example-argocd-server   <none>   example-argocd.apps.rjeczkowaws-411z.ocp-gitops-qe.com   router-default.apps.rjeczkowaws-411z.ocp-gitops-qe.com 
```

**$ oc get ingress/example-argocd-server -o jsonpath="{.spec.rules[0].host}" -n test-ns**
`example-argocd.apps.rjeczkowaws-411z.ocp-gitops-qe.com`

_Argo CD on .status.host field is updated with `router-default.apps.xxxxxxx-411.ocp-xxxx-xx.com` which is the same as Ingress ADDRESS (`.status.loadBalancer.ingress[0].hostname`)._

**$ oc get argocd -o jsonpath="{.items[*]['status.host']}" -n test-ns**
`router-default.apps.rjeczkowaws-411z.ocp-gitops-qe.com`

$ oc get ingress/example-argocd-server -o jsonpath="{.status.loadBalancer.ingress[0].hostname}" -n test-ns
router-default.apps.rjeczkowaws-411z.ocp-gitops-qe.com

**Expected behavior**

Argo CD should copy the value of Ingress .spec.rules[0].host (instead of .status.loadBalancer.ingress[0].hostname) to .status.host because that's the URL of Argo CD.

**Screenshots**
NA

**Additional context**
NA

PR #815, opened 2022-11-25
fix: secure connection with dex, use https (#814)

iam-veeramalla:
Signed-off-by: iam-veeramalla <abhishek.veeramalla@gmail.com>

**What type of PR is this?**
> /kind bug

**What does this PR do / why we need it**:
Fixes #814 
Closes #814 

**Have you updated the necessary documentation?**
NA

**Which issue(s) this PR fixes**:
Fixes #814 

**How to test changes / Special notes to the reviewer**:
Please refer #814

PR #818, opened 2022-11-28
[cherry-pick] dex secure connection and ingress on openshift fixes

### LINKED ISSUE #814: Argo CD Server fails to connect to Dex
**Describe the bug**
One of the ways to enable OIDC/OAuth with Argo CD is to use Dex. Argo CD Server connects with Dex to verify the authentication of a user. 

However, this seems to be broken on the master. 

Screenshot 2022-11-25 at 5 42 04 PM

This can also be reproduced by logging into the Argo CD Server Pod

Screenshot 2022-11-25 at 4 52 34 PM

If we look at the above screenshot and the error message, it is evident that HTTPS connection is accepted. 

**To Reproduce**
1. Login into the Argo CD Server Pod.
2. Try to connect to the Dex Service using HTTP.
`curl <dex-service-name>.<namespace>.svc.cluster.local:5556`

**Expected behavior**
The above should not throw 400 (`client sent a HTTP request to an HTTPS server`)

**Screenshots**
Provided above.

**Additional context**
NA

iam-veeramalla:
**What type of PR is this?**
> /kind bug

cherry-picks #815 #816 

Fixes #814 #816 

Please refer #814 #816


controllers/argocd/status.go
```diff
@@ -250,16 +252,33 @@ func (r *ReconcileArgoCD) reconcileStatusNotifications(cr *argoprojv1a1.ArgoCD)
 func (r *ReconcileArgoCD) reconcileStatusHost(cr *argoprojv1a1.ArgoCD) error {
 	cr.Status.Host = ""
 	cr.Status.Phase = "Available"
-	if cr.Spec.Server.Route.Enabled {
-		if !IsRouteAPIAvailable() {
-			log.Info("Routes not available in non-OpenShift environments, please use Ingresses instead")
-			return nil
-		}
+
+	if (cr.Spec.Server.Route.Enabled || cr.Spec.Server.Ingress.Enabled) && IsRouteAPIAvailable() {
```
wtam2018, line 256: I think this logic to adjust the status based on OpenShift route host should go to gitops operator's ReconcilerHook.

iam-veeramalla: @wtam2018 All the OpenShift Routes logic should move out of this operator. 

This will be a major change which we should target in the near future. I think it is out of the scope of this PR.

controllers/argocd/status.go
```diff
@@ -250,16 +252,33 @@ func (r *ReconcileArgoCD) reconcileStatusNotifications(cr *argoprojv1a1.ArgoCD)
 func (r *ReconcileArgoCD) reconcileStatusHost(cr *argoprojv1a1.ArgoCD) error {
 	cr.Status.Host = ""
 	cr.Status.Phase = "Available"
-	if cr.Spec.Server.Route.Enabled {
-		if !IsRouteAPIAvailable() {
-			log.Info("Routes not available in non-OpenShift environments, please use Ingresses instead")
-			return nil
-		}
```
wtam2018, line 257: We should preserve the behavior that when the operator is deployed on a non-openshift cluster and enabling route, the above message is logged.

iam-veeramalla: I initially added it but during the code review, @jaideepr97 and I realized that it is a redundant piece of code.

https://github.com/argoproj-labs/argocd-operator/blob/2090abd831d1600ad31f81d71dfc7cff4652f9c0/controllers/argocd/util.go#L787

iam-veeramalla: Also, it looks out of the place to log this error message during statusReconciliation after all the processing is done.

wtam2018: If it is in the original code, we should log preserve message. The fix has the side effect of deleting the log message.

controllers/argocd/status.go
wtam2018: It is not necessary a major change. Just moving the logic here to openshift.go in gitops operator. However, the problem is the argocd operator is broken or not fixed in this issue.

controllers/argocd/status.go
jaideepr97: I think it should be fine to remove this, since it should be common knowledge that routes are only available on openshift and not kubernetes, at least amongst the users of this operator

controllers/argocd/status.go
jaideepr97: I think we would need to think a little bit about how to properly handle this through the hooks
I also think it may be best handled in 0.6/1.8

iam-veeramalla:
@jaideepr97 or @wtam2018 can one of you please approve this change ? If you feel this PR requires any changes, add your comments so that we can discuss.

For now, I don't see any comments on the PR.

wtam2018 (approved):

jaideepr97:
@iam-veeramalla strange
@wtam2018 and I both left comments here which for some reason only show up when you look at the file changes, not on the main PR page

PR #821, opened 2022-12-06
Fix error in resource customizations documentation

reginapizza:
**What type of PR is this?**
 /kind bug
 /kind documentation

Fixes [GITOPS-2451](https://issues.redhat.com/browse/GITOPS-2451) to correct documentation about resource customizations

Change is already present in release-0.5 branch.

### DIFF
--- a/docs/reference/argocd.md
+++ b/docs/reference/argocd.md
@@ -964,16 +964,33 @@ There are two ways to customize resource behavior- the first way, only available
 
 ### Resource Customizations (with subkeys) Example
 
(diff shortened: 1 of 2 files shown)


docs/reference/argocd.md
```diff
@@ -964,16 +964,50 @@ There are two ways to customize resource behavior- the first way, only available
...
 kind: ArgoCD
 metadata:
-  name: example-argocd
-  labels:
-    example: resource-customizations
+  name: argocd
 spec:
+  resourceIgnoreDifferences:
+    all:
+      jqPathExpressions:
```
wtam2018, line 977: I wasn't aware that these fields have been changed to slices. That is good. Thanks. Please realistic examples instead of xyz and abc.

PR #845, opened 2023-01-24
Progressive delivery of applications using ApplicationSets

iam-veeramalla:
**What type of PR is this?**
> /kind enhancement

This is an experimental, alpha-quality feature that allows you to control the order in which the ApplicationSet controller will create or update the Applications owned by an ApplicationSet resource. It may be removed in future releases or modified in backwards-incompatible ways.

The feature only interacts with the health of managed Applications. It is **NOT** intended to support direct integrations with other Rollout controllers (such as the native ReplicaSet controller or Argo Rollouts).
Progressive Rollouts watch for the managed Application resources to become "Healthy" before proceeding to the next stage.

**Enabling Progressive Rollouts**
As an experimental feature, progressive rollouts must be explicitly enabled, in one of the below 3 ways.
1. Pass `--enable-progressive-rollouts` to the ApplicationSet controller args.
    To achieve this using the operator, Users can add the above argument in the Argo CD CR
```
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: basic
spec:
  applicationSet:
    extraCommandArgs:
     - --enable-progressive-rollouts
```
2. Set `ARGOCD_APPLICATIONSET_CONTROLLER_ENABLE_PROGRESSIVE_ROLLOUTS=true` in the ApplicationSet controller environment variables in the below mentioned way.
```
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: basic
spec:
  applicationSet:
    env:
     - name: ARGOCD_APPLICATIONSET_CONTROLLER_ENABLE_PROGRESSIVE_ROLLOUTS
       value: "true"
```
3. Set `applicationsetcontroller.enable.progressive.rollouts: true` in the ArgoCD `ConfigMap`.
```
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: basic
spec:
   extraConfig:
     applicationsetcontroller.enable.progressive.rollouts: true
```

1. Run the operator manually using `make install run`
2. Once the operator is running, submit any one of the 3 Argo CD CRs mentioned above.
3. The expected behavior for each of the example is provided above.


iam-veeramalla:
> ting that fiel

Hi @jaideepr97 , the implementation is same as upstream.

Upstream also supports all the three ways and three are completely unique

iam-veeramalla:
@jaideepr97 fyi
https://argo-cd.readthedocs.io/en/latest/operator-manual/applicationset/Progressive-Rollouts/#enabling-progressive-rollouts

jaideepr97:
@iam-veeramalla I understand that 
I'm saying as an operator I think we can try and make things more user friendly 
what if we just had a toggle under `.spec.applicationSets.enableProgressiveRollout` that set all these fields to true under the hood?

iam-veeramalla:
> @iam-veeramalla I understand that I'm saying as an operator I think we can try and make things more user friendly what if we just had a toggle under `.spec.applicationSets.enableProgressiveRollout` that set all these fields to true under the hood?

@jaideepr97 Thanks for the suggestion.
The idea here is to implement the `extraCommandArgs` and `env` support so that in the future, we don't need to add more fields to the Argo CD CR w.r.t command arguments or env variable for ApplicationSets.
It might look like we are adding too many place holders for this PR but we are actually not. The applicationset controller lacks these options which are already supported by `controller`, `server`, `repo-server` and other components.

examples/argocd-route.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   labels:
     example: route
 spec:
+  applicationSet: {}
```
jaideepr97, line 8: @iam-veeramalla is this intentionally added here to the example route?

iam-veeramalla: ah no, good catch. I will get rid of it.

jaideepr97:
@iam-veeramalla could we add unit tests as well?

iam-veeramalla:
@jaideepr97 added new tests and removed an unintentional change. PTAL, Thanks :)

docs/reference/argocd.md
```diff
@@ -97,6 +99,23 @@ spec:
...
+metadata:
+  name: example-argocd
+  labels:
+    example: applicationset
+spec:
+  applicationSet:
+    extraCommandArgs:
+      - --foo
+      - bar
+```
```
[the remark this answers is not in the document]
iam-veeramalla: I think we should keep it very generic. In the release notes text when we describe the feature, we can add a point for this.

jaideepr97 (approved):

PR #847, opened 2023-02-02
feat: read dex clientSecret from argocd-secret

iam-veeramalla:

**What type of PR is this?**
> /kind enhancement

This PR will introduce a behavioral change in the way operator configures Dex for OpenShift connector.

Currently, Operator configures dex openshift connector to the Argo CD Instance when `.spec.dex.OpenShiftOAuth: true` && `.spec.provider: dex`. While this configuration works as expected, the oidc configuration which includes the `clientSecret` is placed in the configmap as plain text. This can be a security risk.

Once this PR is merged, operator will provide the reference of the `clientSecret` in the configmap and not the actual secret. Argo CD Server will read this `clientSecret` reference from `argocd-secret`.

The new configuration will look something like below
`kubectl get cm argocd-cm -oyaml`

```
data:
  .....
  dex.config: |
    connectors:
    - config:
        ....
        clientSecret: $oidc.dex.clientSecret
       .....
       .....
```

NA

2. Use the below Argo CD CR to create Argo CD Instance with Dex configuration.
```
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: route
spec:
  sso:
   provider: dex
   dex:
    openShiftOAuth: true
  server:
    route:
      enabled: true
```
3. Login into the Argo CD UI using the route url.
4. Verify that the login is working fine with the new implementation.
5. Verify in the Argo CD Configmap that `clientSecret` is not comprised but is just a reference to the `oidc.dex.clientSecret` in `argocd-secret`.

### DIFF
--- a/common/keys.go
+++ b/common/keys.go
@@ -228,4 +228,7 @@ const (
 
 	// ArgoCDServerClusterRoleEnvName is an environment variable to specify a custom cluster role for Argo CD server
 	ArgoCDServerClusterRoleEnvName = "SERVER_CLUSTER_ROLE"
+
+	// ArgoCDDexSecretKey is used to reference Dex secret from Argo CD secret into Argo CD configmap
+	ArgoCDDexSecretKey = "oidc.dex.clientSecret"
 )
--- a/controllers/argocd/dex.go
+++ b/controllers/argocd/dex.go
@@ -143,10 +143,6 @@ func (r *ReconcileArgoCD) reconcileDexConfiguration(cm *corev1.ConfigMap, cr *ar
 
 // getOpenShiftDexConfig will return the configuration for the Dex server running on OpenShift.
 func (r *ReconcileArgoCD) getOpenShiftDexConfig(cr *argoprojv1a1.ArgoCD) (string, error) {
-	clientSecret, err := r.getDexOAuthClientSecret(cr)
-	if err != nil {
-		return "", err
-	}
 
 	groups := []string{}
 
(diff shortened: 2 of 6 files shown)


controllers/argocd/dex.go
```diff
@@ -164,7 +160,7 @@ func (r *ReconcileArgoCD) getOpenShiftDexConfig(cr *argoprojv1a1.ArgoCD) (string
 		Config: map[string]interface{}{
 			"issuer":       "https://kubernetes.default.svc", // TODO: Should this be hard-coded?
 			"clientID":     getDexOAuthClientID(cr),
-			"clientSecret": *clientSecret,
+			"clientSecret": "$oidc.dex.clientSecret",
```
jannfis, line 163: You have defined the `ArgoCDDexSecretKey` constant already, would it make sense to use it here?

jannfis (approved):
LGTM, with a minor comment.

reginapizza (approved):
Looks good and works for me! The original request was that the user wished that the token would not be exposed as that is a security concern, but an additional issue we were running into was that even writing the clientSecret in a variable was getting overwritten during reconciliation. With this PR even if the user wishes to change their secret, they can do so in argocd-secret and it won't get overwritten. Two birds with one stone- thanks Abhishek! :D

common/keys.go
```diff
@@ -228,4 +228,7 @@ const (
 
 	// ArgoCDServerClusterRoleEnvName is an environment variable to specify a custom cluster role for Argo CD server
 	ArgoCDServerClusterRoleEnvName = "SERVER_CLUSTER_ROLE"
+
+	// ArgoCDDexSecretKey is used to reference Dex secret from Argo CD secret into Argo CD configmap
+	ArgoCDDexSecretKey = "oidc.dex.clientSecret"
```
jaideepr97, line 233: Should this be `$oidc.dex.clientSecret`?

PR #854, opened 2023-02-09
fix: remove managed-by label from terminating namespaces

### LINKED ISSUE #853: Managed namespace in terminating state blocks app resource deployment to other managed namespaces
**Describe the bug**
When an Argo CD instance is created in namespace-scoped mode and manages multiple other namespaces - if any of those managed namespaces enters terminating state for extended amounts of time (due to a finalizer on a resource for eg) the argo-cd instance is blocked and unable to deploy app resources to any of the other managed namespaces 

**To Reproduce**
1) Install namespace-scoped Argo CD:

```kubectl create ns gitops-service-argocd```

```kubectl apply -f https://raw.githubusercontent.com/redhat-appstudio/managed-gitops/main/manifests/base/gitops-service-argocd/base/argo-cd.yaml```
 
2) Create a Namespace managed by Argo CD.
```
apiVersion: v1
kind: Namespace
metadata:
  name: jane
  labels:
    argocd.argoproj.io/managed-by: gitops-service-argocd
```
3) Create 'ConfigMap' with a finalizer, in Namespace.
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-config-map-2
  namespace: jane
  
  finalizers:
  - some.random/finalizer
 
data: {}
```
4) Delete the 'Namespace', putting it into 'Terminating' state. The Namespace will never finish deleting, due to the finalizer on the 'ConfigMap'.

```kubectl delete namespace jane```

5) create a new managed 'Namespace'
```
apiVersion: v1
kind: Namespace
metadata:
  name: john
  labels:
    argocd.argoproj.io/managed-by: gitops-service-argocd
```
6) 'kubectl apply' the following Argo CD Application
```
kind: Application
metadata:
  name: my-app
  namespace: gitops-service-argocd
spec:
  destination:
    namespace: john
    server: https://kubernetes.default.svc
  project: default
  source:
    path: kustomize-guestbook
    repoURL: https://github.com/argoproj/argocd-example-apps
    targetRevision: master
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
```

7) You will see that the Argo CD Application fails to deploy, and has the following status:

```
status:
    conditions:
    - lastTransitionTime: "2023-02-02T17:58:41Z"
      message: 'failed to sync cluster https://172.30.0.1:443: failed to load initial
        state of resource Service: services is forbidden: User "system:serviceaccount:gitops-service-argocd:gitops-service-argocd-argocd-application-controller"
        cannot list resource "services" in API group "" in the namespace "jane"'
      type: ComparisonError
    - lastTransitionTime: "2023-02-02T17:58:41Z"
      message: 'failed to sync cluster https://172.30.0.1:443: failed to load initial
        state of resource Service: services is forbidden: User "system:serviceaccount:gitops-service-argocd:gitops-service-argocd-argocd-application-controller"
        cannot list resource "services" in API group "" in the namespace "jane"'
      type: ComparisonError
    health:
      status: Healthy
    reconciledAt: "2023-02-02T17:58:41Z"
    sync:
      comparedTo:
        destination:
          namespace: john
          server: https://kubernetes.default.svc
        source:
          path: kustomize-guestbook
          repoURL: https://github.com/argoproj/argocd-example-apps
          targetRevision: master
      status: Unknown
```

Also in the Argo CD UI you may notice none of the application resources are deployed in the "John" namespace

**Expected behavior**
Argo CD should not be blocked from deploying resources to non-terminating managed namespaces

### PR DESCRIPTION
**Pull request by jaideepr97**
Signed-off-by: Jaideep Rao <jaideep.r97@gmail.com>

**What type of PR is this?**

/kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

This PR removes the "managed-by" label from a previously managed now terminating namespace, which triggers the removal of this namespace from the argo-cd instance's cluster secret thus unblocking it from deploying resources to other managed namespaces 
This seems like a cleaner approach than directly deleting the namespace from the cluster secret as we already have existing code paths to achieve this that could be triggered by removing the label from the terminating namespace - and while it is generally not a good idea to delete labels from namespaces (if not added by the operator) however, it maybe okay in this case since the namespace is scheduled for deletion anyway  

Fixes #853?

repeat steps documented in linked issue and confirm that app resources are in fact deployed to "John" namespace

### DIFF
--- a/controllers/argocd/role.go
+++ b/controllers/argocd/role.go
@@ -103,8 +103,13 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
 
 	// create policy rules for each namespace
 	for _, namespace := range r.ManagedNamespaces.Items {
-		// Skip terminating namespaces.
+		// If encountering a terminating namespace remove managed-by label from it and skip reconciliation - This should trigger
+		// clean-up of roles/rolebindings and removal of namespace from cluster secret
 		if namespace.DeletionTimestamp != nil {
+			if _, ok := namespace.Labels[common.ArgoCDManagedByLabel]; ok {
+				delete(namespace.Labels, common.ArgoCDManagedByLabel)
+				_ = r.Client.Update(context.TODO(), &namespace)
+			}
 			continue
 		}
 
--- a/controllers/argocd/rolebinding.go
+++ b/controllers/argocd/rolebinding.go
@@ -110,8 +110,13 @@ func (r *ReconcileArgoCD) reconcileRoleBinding(name string, rules []v1.PolicyRul
 	}
 
 	for _, namespace := range r.ManagedNamespaces.Items {
-		// Skip terminating namespaces.
+		// If encountering a terminating namespace remove managed-by label from it and skip reconciliation - This should trigger
+		// clean-up of roles/rolebindings and removal of namespace from cluster secret
 		if namespace.DeletionTimestamp != nil {
+			if _, ok := namespace.Labels[common.ArgoCDManagedByLabel]; ok {
+				delete(namespace.Labels, common.ArgoCDManagedByLabel)
+				_ = r.Client.Update(context.TODO(), &namespace)
+			}
 			continue
 		}
 
--- a/tests/k8s/1-031_validate_handle_terminating_namespaces/01-install.yaml
+++ b/tests/k8s/1-031_validate_handle_terminating_namespaces/01-install.yaml
@@ -1,0 +1,207 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: gitops-service-argocd
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ArgoCD
+metadata:
+  finalizers:
+  - argoproj.io/finalizer
+  name: gitops-service-argocd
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: controllers/argocd/role.go [hunk 1]
```diff
@@ -103,8 +103,13 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
 
 	// create policy rules for each namespace
 	for _, namespace := range r.ManagedNamespaces.Items {
-		// Skip terminating namespaces.
+		// If encountering a terminating namespace remove managed-by label from it and skip reconciliation - This should trigger
```
jannfis, line 106: It _should_ trigger or it _will_ trigger?

Can we back this up by a test?

jaideepr97: @jannfis 
it will trigger
Added a test case, PTAL 
(Only checking for synced state of app because the app likely runs privileged containers which will fail to come up on OCP and will stay in progressing/degraded state. Sync state anyway should confirm that the resources were at least deployed in the target namespace, which should be enough to address the original bug)

jannfis (approved):
LGTM!

PR #860, opened 2023-02-18
fix: operator overwrites server replicas when autoscale is enabled

jaideepr97:
**What type of PR is this?**

 /kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

Currently, the operator does not respect server autoscale configuration expressed in CR. The HPA controller tries to edit replica setting in server deployment, but operator overwrites it. This PR changes the logic to only reconcile on replica drift if autoscale is disabled. 

Fixes #?

1. Deploy operator
2. Create Argo CD instance
3. set `.spec.server.autoscale.enabled` to true
4. set required no. of replicas in `.spec.server.autoscale.hpa`
5. confirm that server replicas are spun up by hpa and replicas are not subsequently killed


controllers/argocd/deployment.go
```diff
@@ -35,26 +35,30 @@ import (
...
 )
 
+var (
+	defaultRepoServerReplicas int32 = 1
+	defaultServerReplicas     int32 = 1
+)
+
 // getArgoCDRepoServerReplicas will return the size value for the argocd-repo-server replica count if it
-// has been set in argocd CR. Otherwise, nil is returned if the replicas is not set in the argocd CR or
+// has been set in argocd CR. Otherwise, default value is returned if the replicas is not set in the argocd CR or
```
wtam2018, line 44: The reason we want to return nil as opposed to our default is to make a distinction that the user has explicit set the number of replicas. If the user does not set the value explicit, the code will honor the default value in the Depployment.

controllers/argocd/deployment.go
```diff
@@ -53,7 +53,6 @@ func getArgoCDServerReplicas(cr *argoprojv1a1.ArgoCD) *int32 {
 	if !cr.Spec.Server.Autoscale.Enabled && cr.Spec.Server.Replicas != nil && *cr.Spec.Server.Replicas >= 0 {
 		return cr.Spec.Server.Replicas
 	}
-
```
wtam2018, line 56: You can probably this file change.

wtam2018 (approved):

PR #872, opened 2023-03-09
fix: dex client secret is not updated in the argocd-secret intermittently

iam-veeramalla:
**What type of PR is this?**
> /kind bug

This PR will fix the issue with `dex client secret is not updated in the argocd-secret intermittently.`
The actual enhancement of securing the dex client secret (OIDC) is part of #847 

This looks like a race condition or sometimes when operator tries to get the dex client secret from Argo CD dex service account (from the secret named token) the token secret is not created by then. This can be fixed in the reconciliation of existing Argo CD secret which is called much later multiple times.
This is an intermittent behavior and can be reproduced more on the bundle installation rather than the local installation.

NA

https://issues.redhat.com/browse/GITOPS-2714

I have added a new automation test which will validate the behavior
**Run the below test**
`kubectl kuttl test ./tests/ocp --config ./tests/kuttl-tests.yaml --test 1-004_validate_dex_clientsecret`

### DIFF
--- a/controllers/argocd/secret.go
+++ b/controllers/argocd/secret.go
@@ -335,11 +335,13 @@ func (r *ReconcileArgoCD) reconcileExistingArgoSecret(cr *argoprojv1a1.ArgoCD, s
 	}
 
 	if cr.Spec.SSO != nil && cr.Spec.SSO.Provider == v1alpha1.SSOProviderTypeDex {
-		if secret.Data[common.ArgoCDDexSecretKey] == nil {
-			dexOIDCClientSecret, err := r.getDexOAuthClientSecret(cr)
-			if err != nil {
-				return nil
-			}
+		dexOIDCClientSecret, err := r.getDexOAuthClientSecret(cr)
+		if err != nil {
+			return nil
+		}
+		actual := string(secret.Data[common.ArgoCDDexSecretKey])
+		expected := *dexOIDCClientSecret
+		if actual != expected {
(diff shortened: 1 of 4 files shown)


controllers/argocd/secret.go
```diff
@@ -335,11 +335,13 @@ func (r *ReconcileArgoCD) reconcileExistingArgoSecret(cr *argoprojv1a1.ArgoCD, s
...
 	if cr.Spec.SSO != nil && cr.Spec.SSO.Provider == v1alpha1.SSOProviderTypeDex {
-		if secret.Data[common.ArgoCDDexSecretKey] == nil {
-			dexOIDCClientSecret, err := r.getDexOAuthClientSecret(cr)
-			if err != nil {
-				return nil
-			}
+		dexOIDCClientSecret, err := r.getDexOAuthClientSecret(cr)
+		if err != nil {
+			return nil
+		}
```
jannfis, line 341: Is it intended to return `nil` instead of `err` here?

controllers/argocd/secret.go
```diff
@@ -335,11 +335,13 @@ func (r *ReconcileArgoCD) reconcileExistingArgoSecret(cr *argoprojv1a1.ArgoCD, s
...
+		actual := string(secret.Data[common.ArgoCDDexSecretKey])
+		expected := *dexOIDCClientSecret
```
jannfis, line 343: Before dereferencing a pointer, please _always_ make sure it's not `nil`

PR #896, opened 2023-04-25
feat: Unify sso status & remove deprecated .spec.sso & .spec.dex fields 

svghadi:
**What type of PR is this?**
/kind enhancement

This PR removes `.status.dex` and `.status.ssoConfig` & unifies them under `.status.sso` to reflect the workload status of whichever SSO provider is configured (dex/keycloak). It also removes the deprecated `.spec.dex`, `.spec.sso.image`, `.spec.sso.resources`, `.spec.sso.verifyTLS`, `.spec.sso.version` fields & back backward compatibility code added in #646. 

NOTE: This PR introduces **breaking changes** to ArgoCD CRD

2. Create empty argo-cd instance
```yaml
kind: ArgoCD
metadata:
  name: argocd
spec: {}
```
3. Observe argocd `.status.sso: Unknown` 
4. Set `.spec.sso.provider=dex`
5. Observe argocd `.status.sso: Failed` due to dex configuration failure
6. Set `.spec.sso.dex.Config="test"`
7.  Observe argocd `.status.sso: Running`(`Pending` if deployment creation is in progress), dex configured successfully 
8. Set `.spec.sso.provider=keycloak`
9. Observe argocd `.status.sso: Failed` due to invalid sso configurations
10. Remove `.spec.sso.dex`
11. Observe argocd `.status.sso: Running`(`Pending` if deployment creation is in progress), keycloak configured successfully 
12. Verify failure scenarios due to old fields in ArgoCD CR
     - Creation of argocd fails with `strict decoding error: unknown field "spec.dex"`
     ```yaml
     kind: ArgoCD
     metadata:
       name: argocd-old
     spec:
       dex:
         config: test-config
     ```
     - Creation of argocd fails with `strict decoding error: unknown field "spec.sso.verifyTLS"`
     ```yaml
     kind: ArgoCD
     metadata:
       name: argocd-old
     spec:
       sso:
         provider: keycloak
         verifyTLS: false
     ```
13. e2e test for the PR.
`kubectl kuttl test ./tests/k8s --config ./tests/kuttl-tests.yaml --test 1-015_validate_sso_status`

### DIFF
--- a/deploy/olm-catalog/argocd-operator/0.7.0/argocd-operator.v0.7.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/argocd-operator/0.7.0/argocd-operator.v0.7.0.clusterserviceversion.yaml
@@ -349,38 +349,6 @@ spec:
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:fieldGroup:Controller
         - urn:alm:descriptor:com.tectonic.ui:resourceRequirements
-      - description: Config is the dex connector configuration.
-        displayName: Configuration
-        path: dex.config
-        x-descriptors:
-        - urn:alm:descriptor:com.tectonic.ui:fieldGroup:Dex
-        - urn:alm:descriptor:com.tectonic.ui:text
-      - description: Image is the Dex container image.
-        displayName: Image
-        path: dex.image
-        x-descriptors:
-        - urn:alm:descriptor:com.tectonic.ui:fieldGroup:Dex
-        - urn:alm:descriptor:com.tectonic.ui:text
-      - description: OpenShiftOAuth enables OpenShift OAuth authentication for the
-          Dex server.
-        displayName: OpenShift OAuth Enabled'
-        path: dex.openShiftOAuth
-        x-descriptors:
-        - urn:alm:descriptor:com.tectonic.ui:fieldGroup:Dex
-        - urn:alm:descriptor:com.tectonic.ui:booleanSwitch
-      - description: Resources defines the Compute Resources required by the container
-          for Dex.
-        displayName: Resource Requirements'
-        path: dex.resources
-        x-descriptors:
-        - urn:alm:descriptor:com.tectonic.ui:fieldGroup:Dex
-        - urn:alm:descriptor:com.tectonic.ui:resourceRequirements
-      - description: Version is the Dex container image tag.
-        displayName: Version
-        path: dex.version
-        x-descriptors:
-        - urn:alm:descriptor:com.tectonic.ui:fieldGroup:Dex
-        - urn:alm:descriptor:com.tectonic.ui:text
       - description: GAAnonymizeUsers toggles user IDs being hashed before sending
           to google analytics.
         displayName: Google Analytics Anonymize Users'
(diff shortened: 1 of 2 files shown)


jaideepr97:
Thanks for the PR @svghadi 
Looks pretty good so far, nice work!

Just a couple things that occurred to me that we missed last time but can probably address now through this PR:
1. Can we convert strings to lower case during direct string comparisons? There are a lot of places we do this, like comparing `.spec.sso.provider` to `dex` or `keycloak`, lets make sure we do `strings.ToLower(cr.spec.sso.provider)` wherever that happens
2. I think we should stop returning an error [here](https://github.com/argoproj-labs/argocd-operator/pull/896/files#diff-30c42e628eebe205919571b78686864b3d5aaab1d82bb09bdb56a1c53bc3d81fL719) after `reconcileSSO` since SSO is ultimately a non-critical component and we shouldn't block reconciliation just because of that. I think logging the error should be good enough


Will the docs updates be made in a separate PR?

controllers/argocd/secret.go
```diff
@@ -184,7 +184,7 @@ func (r *ReconcileArgoCD) reconcileArgoSecret(cr *argoprojv1a1.ArgoCD) error {
 		common.ArgoCDKeyTLSPrivateKey:      tlsSecret.Data[common.ArgoCDKeyTLSPrivateKey],
 	}
 
-	if cr.Spec.Dex != nil || (cr.Spec.SSO != nil && cr.Spec.SSO.Provider == v1alpha1.SSOProviderTypeDex) {
+	if cr.Spec.SSO != nil && cr.Spec.SSO.Provider.ToLower() == v1alpha1.SSOProviderTypeDex.ToLower() {
```
jaideepr97, line 187: @svghadi the case conversion is only required on `spec.sso.provider` since that is coming from the user. `v1alpha1.SSOPRoivderTypeDex` is already defined in lower case so this is not required
Let's try to simplify where we can

jaideepr97:
Thanks @svghadi 
There are a lot of unit tests that have a `setEnvVarFunc` field that was required for DISABLE_DEX testing. This field is not required in these tests anymore, I think we should get rid of it as well

controllers/argocd/secret.go
svghadi: Here, `.ToLower()` on `v1alpha1.SSOPRoivderTypeDex` is used to resolve type mismatch error
> invalid operation: cr.Spec.SSO.Provider.ToLower() == v1alpha1.SSOProviderTypeDex (mismatched types string and "github.com/argoproj-labs/argocd-operator/api/v1alpha1".SSOProviderType)

But I agree, we should remove unwanted calls. I will try to simplify by updating `ToLower()` function to return `SSOProviderType` instead of `string` so that we can remove extra call on already defined values.

jaideepr97 (approved):
LGTM, thanks @svghadi :) 

Approving this for now but not merging it yet since we need to have a discussion about bumping the API version because of breaking changes introduced in this PR

wtam2018:
/hold 
Do not merge until we are clear on upgrading Argo CD CRD to v1beta1

svghadi:
The PR is ready for merge. We will upgrade Argo CD CRD to v1beta1 in upcoming v0.8 release.

cc: @iam-veeramalla

PR #910, opened 2023-05-11
feat: Add dynamic sharding for argocd controller

ishitasequeira:
**What type of PR is this?**
/kind enhancement

This PR enables dynamic scaling of application controller based on the configured MinShards, MaxShards and ClustersOnEachShard flag. The feature is enabled for users if they set the flag `Spec.Controller.Sharding.EnableDynamicSharding`.

https://issues.redhat.com/browse/GITOPSRVCE-409

* Install the operator
* Create ArgoCD instance with below fields set:
```
spec:
    controller:
        sharding:
            enableDynamicScaling: true
            minShards: 2
            maxShards: 4
            clustersOnEachShard: 1
```
* Add atleast 3 clusters to argocd instance
* See that the number of replicas was increased to 3 and the controller pods increased to 3


api/v1alpha1/argocd_types.go
```diff
@@ -115,6 +115,18 @@ type ArgoCDApplicationControllerShardSpec struct {
 
 	// Replicas defines the number of replicas to run in the Application controller shard.
 	Replicas int32 `json:"replicas,omitempty"`
+
+	// EnableDynamicScaling defines whether dynamic scaling should be enabled for Application Controller component
+	EnableDynamicScaling bool `json:"enableDynamicScaling,omitempty"`
```
anandf, line 120: Should this be renamed as `DynamicScalingEnabled` ? I see an existing property `StatusBadgeEnabled` and thought we can follow a similar naming convention.

api/v1alpha1/argocd_types.go
```diff
@@ -115,6 +115,18 @@ type ArgoCDApplicationControllerShardSpec struct {
...
+
+	// MinShards defines the minimum number of shards at any given point
+	MinShards int32 `json:"minShards,omitempty"`
+
+	// MaxShards defines the maximum number of shards at any given point
+	MaxShards int32 `json:"maxShards,omitempty"`
+
+	// clustersOnEachShard defines the maximum number of clusters managed by each argocd shard
+	ClustersOnEachShard int32 `json:"clustersOnEachShard,omitempty"`
```
anandf, line 129: suggestion : `ClustersPerShard`  ?

api/v1alpha1/groupversion_info.go
```diff
@@ -15,8 +15,8 @@ limitations under the License.
 */
 
 // Package v1alpha1 contains API Schema definitions for the argoproj.io v1alpha1 API group
-//+kubebuilder:object:generate=true
-//+groupName=argoproj.io
+// +kubebuilder:object:generate=true
```
anandf, line 18: Check formatting

api/v1alpha1/argocdexport_types.go
```diff
@@ -30,22 +30,22 @@ import (
...
-//+operator-sdk:csv:customresourcedefinitions:resources={{PersistentVolumeClaim,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{Pod,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{Prometheus,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{ReplicaSet,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{Route,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{Secret,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{Service,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{ServiceMonitor,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{StatefulSet,v1,""}}
+// +operator-sdk:csv:customresourcedefinitions:resources={{ArgoCD,v1alpha1,""}}
```
anandf, line 33: check formatting

controllers/argocd/statefulset.go
```diff
@@ -441,13 +441,56 @@ func getArgoControllerContainerEnv(cr *argoprojv1a1.ArgoCD) []corev1.EnvVar {
...
 	var replicas int32 = common.ArgocdApplicationControllerDefaultReplicas
 
-	if cr.Spec.Controller.Sharding.Replicas != 0 && cr.Spec.Controller.Sharding.Enabled {
+	if cr.Spec.Controller.Sharding.EnableDynamicScaling {
+		if cr.Spec.Controller.Sharding.MinShards < 1 {
+			log.Info("Minimum number of shards cannot be less than 1. Setting default value to 1")
+			cr.Spec.Controller.Sharding.MinShards = 1
+		}
+
+		if cr.Spec.Controller.Sharding.MaxShards < cr.Spec.Controller.Sharding.MinShards {
```
anandf, line 453: Should this check be done as part of validation (admission controller) when creating the CR ? Changing customer provided `spec` should be avoided.

controllers/argocd/statefulset.go
```diff
@@ -441,13 +441,56 @@ func getArgoControllerContainerEnv(cr *argoprojv1a1.ArgoCD) []corev1.EnvVar {
...
+		}
+
+		clusterSecrets, err := r.getClusterSecrets(cr)
+		if err != nil {
+			log.Info("Error retreiving cluster secrets for ArgoCD instance %s", cr.Name)
+		}
+
+		clustersOnEachShard := cr.Spec.Controller.Sharding.ClustersOnEachShard
+		if clustersOnEachShard < 1 {
+			log.Info("clustersOnEachShard cannot be less than 1. Defaulting to 1.")
```
anandf, line 464: Similar comment. Add validation logic.

jaideepr97: Specifying both, max & min no. of shards at the same time as specifying the no. of clusters per shard can lead to a situation where one has to override the other

in this case it looks like we would prioritize staying within the max-min window over honoring no. of clusters per shard 

Is this intentional behavior, and if so should we document it somewhere?

controllers/argocd/statefulset.go
jaideepr97: I would agree, we should not be writing back to the `spec`

we can always work with local variables rather than changing spec fields though

jaideepr97:
I wonder if we could use an HPA for this? 
@ishitasequeira @anandf

api/v1alpha1/argocd_types.go
```diff
@@ -39,22 +39,22 @@ func init() {
...
-//+operator-sdk:csv:customresourcedefinitions:resources={{PersistentVolumeClaim,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{Pod,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{Prometheus,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{ReplicaSet,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{Route,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{Secret,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{Service,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{ServiceMonitor,v1,""}}
-//+operator-sdk:csv:customresourcedefinitions:resources={{StatefulSet,v1,""}}
+// +operator-sdk:csv:customresourcedefinitions:resources={{ArgoCD,v1alpha1,""}}
```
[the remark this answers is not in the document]
ishitasequeira: The spacing is added automatically everytime I run `make build`. Will remove it manually.

controllers/argocd/statefulset.go
ishitasequeira: @jaideepr97 Yes, this is the intentional behavior. We would want to keep the number of replicas of application controller between the MinShards and MaxShards window. The clustersPerShard helps in deciding whether we should increase the number of replicas or not.

ishitasequeira:
> I wonder if we could use an HPA for this? @ishitasequeira @anandf

@jaideepr97 could you expand more on this?

docs/reference/argocd.md
```diff
@@ -155,6 +155,10 @@ AppSync | 3m | AppSync is used to control the sync frequency of ArgoCD Applicati
 Sharding.enabled | false | Whether to enable sharding on the ArgoCD Application Controller component. Useful when managing a large number of clusters to relieve memory pressure on the controller component.
 Sharding.replicas | 1 | The number of replicas that will be used to support sharding of the ArgoCD Application Controller.
 Env | [Empty] | Environment to set for the application controller workloads
+Sharding.dynamicScalingEnabled | true | Whether to enable dynamic scaling of the ArgoCD Application Controller component. This will ignore the configuration of `Sharding.enabled` and `Sharding.replicas`
+Sharding.minShards | 1 | The minimum number of replicas of the ArgoCD Application Controller component.
```
anandf, line 159: Can we mention the validation critierias in a separate column ?
- ClustersPerShard, MinShards and MaxShards must be 1 or > 1
- MaxShards must always be greater than MinShards

anandf:
Please add the code for validation webhook, otherwise looks good to me.

api/v1alpha1/argocd_validation_webhook.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (cr *ArgoCD) ValidateArgocdCR() error {
+
+	var minShards int32 = cr.Spec.Controller.Sharding.MinShards
+	var maxShards int32 = cr.Spec.Controller.Sharding.MaxShards
+
+	if cr.Spec.Controller.Sharding.DynamicScalingEnabled {
+		if minShards < 1 {
+			return fmt.Errorf("Minimum number of shards cannot be less than 1. Setting default value to 1")
```
anandf, line 87: Error message should be sufficient. `Setting default value to 1` is not required. The CR creation itself would fail and the user as to fix the CR and re-create it.
Was wondering if we can specify the full path of the variable so that it would be easy for the user to figure out what needs to be changed. Something like `spec.controller.sharding.minShards cannot be less than 1`

api/v1alpha1/argocd_validation_webhook.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+
+		if maxShards < minShards {
+			return fmt.Errorf("Maximum number of shards cannot be less than minimum number of shards. Setting maximum shards same as minimum shards")
```
anandf, line 91: Similar comment for error message.

api/v1alpha1/argocd_validation_webhook.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+
+		clustersPerShard := cr.Spec.Controller.Sharding.ClustersPerShard
+		if clustersPerShard < 1 {
+			return fmt.Errorf("ClustersPerShard cannot be less than 1. Defaulting to 1")
```
anandf, line 96: Similar comment for error message.

controllers/argocd/statefulset.go
```diff
@@ -441,13 +442,61 @@ func getArgoControllerContainerEnv(cr *argoprojv1a1.ArgoCD) []corev1.EnvVar {
...
 }
 
-func (r *ReconcileArgoCD) reconcileApplicationControllerStatefulSet(cr *argoprojv1a1.ArgoCD, useTLSForRedis bool) error {
+func (r *ReconcileArgoCD) getArgoControllerReplicaCount(cr *argoprojv1a1.ArgoCD) int32 {
 	var replicas int32 = common.ArgocdApplicationControllerDefaultReplicas
+	var minShards int32 = cr.Spec.Controller.Sharding.MinShards
+	var maxShards int32 = cr.Spec.Controller.Sharding.MaxShards
+
+	if cr.Spec.Controller.Sharding.DynamicScalingEnabled {
+		if minShards < 1 {
```
anandf, line 451: Is this check really needed ? As the `ValidatingWebHook` should restrict such invalid configuration.

controllers/argocd/statefulset.go
```diff
@@ -441,13 +442,61 @@ func getArgoControllerContainerEnv(cr *argoprojv1a1.ArgoCD) []corev1.EnvVar {
...
+			err := errors.New("incorrect configuration")
+			log.Error(err, "Minimum number of shards cannot be less than 1. Setting default value to 1")
+			minShards = 1
+		}
+
+		if maxShards < minShards {
```
anandf, line 457: Is this check really needed ? As the `ValidatingWebHook` should restrict such invalid configuration.

controllers/argocd/statefulset.go
```diff
@@ -441,13 +442,61 @@ func getArgoControllerContainerEnv(cr *argoprojv1a1.ArgoCD) []corev1.EnvVar {
...
+			err := errors.New("incorrect configuration")
+			log.Error(err, "Maximum number of shards cannot be less than minimum number of shards. Setting maximum shards same as minimum shards")
+			maxShards = minShards
+		}
+
+		clustersPerShard := cr.Spec.Controller.Sharding.ClustersPerShard
+		if clustersPerShard < 1 {
```
anandf, line 464: Is this check really needed ? As the `ValidatingWebHook` should restrict such invalid configuration.

anandf (approved):

controllers/argocd/statefulset.go
jaideepr97: @ishitasequeira how does this interface with the actual distribution algorithms that we are developing for cluster reallocation? For example, `clustersPerShard` is being used in a pretty simple way right now, but if in the future we have load based balacing in place and argo-cd wants to place 3 small clusters on one shard and 1 big cluster on another shard, that would go against whatever is defined in `clustersPerShard` by the user

Maybe we should at least make it clear to the user that whatever they configure in `clustersPerShard` won't always be honored

controllers/argocd/statefulset.go
jaideepr97: @ishitasequeira @anandf 
hmmm is a validation webhook for this overkill? I think we can achieve this kind of min/max validation just through CRD openapi schema validation


```
// +kubebuilder:validation:Maximum
specifies the maximum numeric value that this field can have.

// +kubebuilder:validation:Minimum
specifies the minimum numeric value that this field can have. Negative numbers are supported.
```
https://book.kubebuilder.io/reference/markers/crd-validation.html

controllers/argocd/statefulset.go
ishitasequeira: @jaideepr97, this interface takes care of. distributing the clusters across multiple shards evenly for now. This is more of an intermediate step till ArgoCD starts considering load on a cluster as a factor to distribute shards. Considering load on each cluster is a bit of a complex thing in current scenario for ArgoCD, thus we decided to go step by step to achieve dynamic scaling of shards. 

As far as not honoring `clustersPerShard` goes, we would not honor the value in `clustersPerShard` even if the number of shards goes beyond the `minShards - maxShards` range. I can add this as a note in the docs.

controllers/argocd/statefulset.go
ishitasequeira: @jaideepr97 kubebuilder could surely be used, but we also want to have validation where we check that the `maxReplicas > minReplicas`. This check is available as a beta feature in k8s v1.25 but not in earlier versions. Thus the use of validation webhook.

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules

anandf: By using kubebuilder annotations, we might be able to support these validations in `OCP 4.10` and `OCP 4.11` which use `1.23` and `1.24`versions of kubernetes respectively.

The problem with having validation logic in controller is that the user is not notified about it, until he finds a problem and sets on to debug a problem by going through the controller logs. If some values in the CR is invalid, user should be notified about it during the CR creation time, so that the user gets a chance to fix the CR and re-create it. It would also avoid silent misconfigurations that can become a pain to debug.

@jaideepr97 Can we have this validation webhook code (even if its a trivial check) and slowly move all validations done in controller to this webhook in subsequent PRs (if we decide to go for it) ?

ishitasequeira: @jaideepr97 following our internal conversations over [slack](https://redhat-internal.slack.com/archives/C01RQH8KQ87/p1684877099000859) around how we could prevent customers from having mis-configurations, we could utilize the validation webhook for addressing this issues. 

Also, I agree to the point by @anandf about notifying customers during CR creation time rather than just adding logs around it and having customers go through logs to understand the problem.

config/webhook/manifests.yaml
```diff
@@ -0,0 +1,56 @@
+
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
```
jaideepr97, line 4: @ishitasequeira @anandf  why do we need to introduce a mutating webhook in addition to the validation webhook?

config/webhook/manifests.yaml
```diff
@@ -0,0 +1,56 @@
...
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /validate-managed-gitops-redhat-com-v1alpha1-gitopsdeployment
+  failurePolicy: Fail
+  name: vgitopsdeployment.kb.io
+  rules:
+  - apiGroups:
+    - managed-gitops.redhat.com
```
jaideepr97, line 48: @ishitasequeira @anandf I don't think it's a good idea to introduce Red Hat managed services specific configuration here in argocd-operator 

I think if this is required we should think of a way to do this in gitops-operator instead

config/webhook/manifests.yaml
```diff
@@ -0,0 +1,56 @@
...
+    apiVersions:
+    - v1alpha1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - gitopsdeployments
```
jaideepr97, line 55: @ishitasequeira @anandf I don't think it's a good idea to introduce Red Hat managed services specific configuration here in argocd-operator

I think if this is required we should think of a way to do this in gitops-operator instead

controllers/argocd/statefulset.go
jaideepr97: On further thought, I think I like the idea of moving all validation into the validating webhook
it does seem like a cleaner approach and would help clean up the operator code as well, so I am in favor of it

I would like @iam-veeramalla 's  thoughts on this as well though

config/webhook/manifests.yaml
ishitasequeira: You are right.. I will remove the Mutating Webhook.

config/webhook/manifests.yaml
ishitasequeira: Ahh.. that's my bad.. I will update the configuration for argocd-operator instead.

jaideepr97:
@ishitasequeira would you mind contributing an e2e test as well? I don't think the unit test is enough to ensure that the validating webhook is in fact catching misconfigurations at run time

ishitasequeira:
> @ishitasequeira would you mind contributing an e2e test as well? I don't think the unit test is enough to ensure that the validating webhook is in fact catching misconfigurations at run time

@jaideepr97 I am working on a separate PR for e2e tests for this epic. The e2e tests require creating multiple clusters and adding them to the argocd instance to see the scaling in effect. Currently we don't have the setup available for adding multiple clusters.

Issue to track the e2e tests: https://issues.redhat.com/browse/GITOPSRVCE-535

ishitasequeira:
> Looking at the implementation of the validating webhook itself, I don't think this is sufficient
> 
> My understanding is that it is quite a bit more involved than what is included in this PR. The validating webhook needs to run its own server (maybe in a separate pod?) and be captured within a request handler, which receives admission review objects from the k8s api server. None of this is being handled in this PR
> 
> I think the idea of moving all CR validation to a validating webhook is good and I am in favor of it, but I'm of the opinion that it should be done in a separate PR, as it has very little do with autoscaling, which is what this PR is about. I would suggest that the validation required for autoscaling be moved back into the operator, until a separate effort to implement the webhook is made, and all CR valdation can be moved there. I'd rather not block this PR because of the webhook stuff
> 
> (edit: here's a sample implmentation of a validating webhook that can be used for reference https://developers.redhat.com/articles/2021/09/17/kubernetes-admission-control-validating-webhooks#a_prometheus_rule_validator)

@jaideepr97 I agree with you that more configuration is required. But we don't need the Handle function as you mentioned. There is another way to configure webhooks where `ValidateCreate`, `ValidateUpdate`, etc. functions take care of handling the resources. For reference: https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation.html 

I have created a separate epic in Service Core backlog [GITOPSRVCE-606](https://issues.redhat.com/browse/GITOPSRVCE-606) for adding ValidationWebhook and will create a separate PR for the same. For this PR, I have added a `TODO` for the same.

api/v1alpha1/argocd_types.go
```diff
@@ -115,6 +115,20 @@ type ArgoCDApplicationControllerShardSpec struct {
 
 	// Replicas defines the number of replicas to run in the Application controller shard.
 	Replicas int32 `json:"replicas,omitempty"`
+
+	// DynamicScalingEnabled defines whether dynamic scaling should be enabled for Application Controller component
+	DynamicScalingEnabled bool `json:"dynamicScalingEnabled,omitempty"`
```
jannfis, line 120: Consider making this a pointer to a bool, instead of a bool value. With a pointer, the value can be `nil` in addition to `true` or `false`, which means the option is not set. This allows us to change the default value for this parameter later on.

controllers/argocd/statefulset.go
jannfis: I'd not agree with this conclusion. The validation webhook should never replace sanity checks within the code. Imagine a validation webhook analogue to JavaScript code validating user input in a web UI. You still want to make sure that data is valid in the backend, and not rely on the frontend validation.

jannfis (commented):
Good work, only have some comments! 

Have you considered some end to end tests for this scenario, too?

controllers/argocd/statefulset.go
jannfis: > moving all validation into the validating webhook

I would disagree that this is something we should do. The validation should happen at both places, similar to how you'd have validation in a frontend/backend scenario. You validate in the frontend (here: webhook) to let the user know early-on that something is wrong, but you validate in the backend (here: controller) as well to ensure you never process invalid data.

We could make functions for the validation logic, to avoid redundancy in code and to prevent validation logic potentially drifting apart.

api/v1alpha1/argocd_types.go
ishitasequeira: That's a good point!! I will make it a pointer to a bool value.

ishitasequeira:
> Good work, only have some comments!
> 
> Have you considered some end to end tests for this scenario, too?

I am currently working on adding the e2e tests. I should have them added by today EOD / Monday.

ishitasequeira:
@jannfis @jaideepr97 @anandf I have added the e2e test cases. PTAL.

anandf:
> I have added the e2e test cases. PTAL.
@ishitasequeira 
Current E2E Tests validates the scale down use case. Can you add tests for the below use cases as well.
- Scale Up scenario. Keep increasing the number of cluster secrets and check if the number of instances are not going beyond the `maxShards`
- Scenario where the `maxShards` is sufficiently low and `clustersPerShard` is not given weightage. Lets say `maxShard=1`, `clustersPerShard=1`, but the acutal number of clusters is 10. In that case, `clustersPerShard` would not be honoured and only 1 instance of application controller runs.

tests/k8s/1-032_validate_dynamic_scaling/02-assert.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ArgoCD
+metadata:
+  name: argocd
+status:
+  phase: Available
+---
+apiVersion: v1
+kind: Pod
```
anandf, line 9: Rather than the pod, can we assert on the `StatefulSet` object's `.status.availableReplicas` value ? Current style of assertion does not guarentee that there are exactly 3 pods. Even if there are 5 pods, this test would still succeed.

tests/k8s/1-032_validate_dynamic_scaling/02-add-cluster-secrets.yaml
```diff
@@ -0,0 +1,41 @@
...
+  name: argocd
+  labels:
+    foo: bar
+spec:
+  controller:
+    sharding:
+      dynamicShardingEnabled: true
+      minShards: 1
+      maxShards: 4
+      clustersPerShard: 1
```
jaideepr97, line 41: @ishitasequeira is this step necessary? ideally we should not have to do anything to trigger reconciliation, the operator should automatically notice that these secrets are created and that autoscaling must be done 

do we need to add a watch for secrets with this specific cluster label maybe?

ishitasequeira: The watcher is already added on Secrets. Removing the re-triggering of reconcilation

tests/k8s/1-032_validate_dynamic_scaling/06-assert.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: argoproj.io/v1alpha1
```
anandf, line 1: Any specific reason why the file is name prefixed with `06` instead of `04` which is the next available integer in the sequence of steps ?

ishitasequeira: Missed renaming the files. Thanks for catching that. Updated the file names.

deploy/olm-catalog/argocd-operator/0.6.0/argoproj.io_argocds.yaml
```diff
@@ -537,10 +537,29 @@ spec:
                     description: Sharding contains the options for the Application
                       Controller sharding configuration.
                     properties:
+                      clustersPerShard:
```
anandf, line 540: @ishitasequeira Are we supporting these new properties in older version (0.6.0) as well ?

ishitasequeira: The file was automatically generated after executing `make bundle manifests`. I updated the file to not include these new properties.

jannfis (approved):
LGTM!

controllers/argocd/custommapper.go
```diff
@@ -153,3 +153,32 @@ func (r *ReconcileArgoCD) namespaceResourceMapper(o client.Object) []reconcile.R
 
 	return result
 }
+
+// namespaceResourceMapper maps a watch event on a namespace, back to the
```
jaideepr97, line 157: nit: comment should say clusterSecretResourceMapper

controllers/argocd/statefulset.go
```diff
@@ -451,13 +451,60 @@ func getArgoControllerContainerEnv(cr *argoprojv1a1.ArgoCD) []corev1.EnvVar {
 	return env
 }
 
-func (r *ReconcileArgoCD) reconcileApplicationControllerStatefulSet(cr *argoprojv1a1.ArgoCD, useTLSForRedis bool) error {
+func (r *ReconcileArgoCD) getArgoControllerReplicaCount(cr *argoprojv1a1.ArgoCD) int32 {
```
jaideepr97, line 454: nit: can we rename this to `getApplicationControllerReplicaCount`?

jaideepr97 (approved):
just a couple nits, i wouldn't block merging for them though
LGTM otherwise, thanks @ishitasequeira !

PR #918, opened 2023-05-26
fix: don't set phase to available during host reconciliation 

### LINKED ISSUE #917: issue in phase reconciliation logic
**Describe the bug**
Currently, Argo CD phase reconciliation seems to be broken. Phase appears as available even if one or more core components (app-controller, server, repo-server, redis) are still pending. This is happening because phase is being manipulated in multiple places (including while setting the host status for the instance). 

**To Reproduce**
1. deploy the operator
2. create an argo-cd instance 
```
cat <<EOF | oc apply -f -
kind: ArgoCD
metadata:
  name: argocd
spec:
EOF

```
3. verify that phase is initially set to `Available` and core components are `running`
4. edit argo-cd to set the image version to an invalid version:
```
spec:
  version: dummy-version
```
5. See that core component pods are restarted with invalid image and stuck in `pending`
6. Confirm that argo-cd instance phase is still set to `available`  

**Expected behavior**
Phase should ONLY be set to available if all core components are running AND there is a valid host set 

**Additional context**

**Screenshots**
phase

**Additional context**

jaideepr97:
**What type of PR is this?**

 /kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

This PR fixees phase reconciliation logic in the operator. Phase is manipulated both in `reconcileStatusPhase`  as well as `reconcileStatusHost`. While it is enough to know that the route/ingress is not up to set the instance phase to `pending`, it is not enough to know that these resources are up to set it to `available`, as there are other requirements that need to be met before phase can be set to `available`. Therefore. `reconcileStatusHost` should not try to set the phase to available under any circumstances. 
After this PR absence of successful route/ingress will set the phase to pending, while presence of these resources will not keep the phase as is (i.e dependent on the core workloads) 

Fixes #917 

same as those listed out in #917 - except actual behavior should now match existing behavior


svghadi:
Line 351 is setting `Status.Phase = "Available"` in `reconcileStatusHost`. Shouldn't it be removed or it was intentional?
https://github.com/argoproj-labs/argocd-operator/blob/92eccbbbb7a54e6344f3b067510db95cbd7995d6/controllers/argocd/status.go#L331-L354

Also, we don't set `Status.Phase` to `Pending` anywhere in `else` section of main `if` loop which runs on k8s. I am not sure if we need to address this, just an observation.

svghadi:
I also noticed we mention `Argocd.Status.Phase` has 5 values in [docs](https://argocd-operator.readthedocs.io/en/latest/reference/api.html/#argoproj.io/v1alpha1.ArgoCDStatus), however we never set it to any value other than `Pending` & `Available` in the code. We can address this in some other PR.

jaideepr97:
> Line 351 is setting `Status.Phase = "Available"` in `reconcileStatusHost`. Shouldn't it be removed or it was intentional?
> 
> https://github.com/argoproj-labs/argocd-operator/blob/92eccbbbb7a54e6344f3b067510db95cbd7995d6/controllers/argocd/status.go#L331-L354
> 
> Also, we don't set `Status.Phase` to `Pending` anywhere in `else` section of main `if` loop which runs on k8s. I am not sure if we need to address this, just an observation.

Thanks! updated 
yes for your 2nd point we can address that in a different PR

iam-veeramalla (approved):
/lgtm

PR #919, opened 2023-05-31
fix: replace kube-rbac-proxy image tag v0.8.0 with its digest equivalent (#510)

### LINKED ISSUE #510: kube-rbac-proxy Image with tag hinders disconnected deployments
Hi community,

after upgrading to ArgoCD operator 0.1.0 from 0.0.15, we experienced issues with the argocd-operator-controller-manager Pod.

**Describe the bug**
Pod argocd-operator-controller-manager is not able to start, due to failing to access the image gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0 https://github.com/argoproj-labs/argocd-operator/blob/755639898c156155085e1fcbe2c3c5e74fec332d/config/default/manager_auth_proxy_patch.yaml#L13

We are running OpenShift 4.7 in a **disconnected environment**, which is why the image from gcr.io is unpullable.
Other ArgoCD operator images are mirrored to an internal registry and access is realized with ImageContentSourcePolicies (by this process: https://docs.openshift.com/container-platform/4.7/operators/admin/olm-restricted-networks.html). ImageContentSourcePolicies, however, require images to be referenced by a SHA-digest (see https://access.redhat.com/solutions/4817401 / Red Hat subscription required for access)
For this image, this does not seem possible, as it is referenced by a tag rather than a SHA-digest and can therefore not be rewritten with ImageContentSourcePolicies.

**To reproduce**
1. In an OpenShift 4.7. environment without internet access, mirror the OLM catalog and its images to a local registry (as described in https://docs.openshift.com/container-platform/4.7/operators/admin/olm-restricted-networks.html)
2. Apply the created ImageContentSourcePolicies
3. Install the ArgoCD operator from Operatorhub
4. ImageContentSourcePolicies enable access to quay.io/argoprojlabs/argocd-operator@sha256:36a65debc06ec68237e5cb6ce9dd2e5076407b00b05d511654e152bd03f2496a but not to gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0

**Expected behavior**
Image kube-rbac-proxy should be referenced with SHA digest rather than a tag to enable usage of ImageContentSourcePolicies.

mortya:
**What type of PR is this?**

/kind bug

This PR fixes installation in environments that require images to be mirrored by digest rather than tag, such as openshift in a "restricted" network install.

Note that this PR includes modifying older CSV versions.  I'm not sure that's kosher.  If not, please free to leave out that part, or kick it back to me for fixing.

N/A

Fixes #510 

Issue #510 includes a description of how to test that the actual problem is fixed.

That said, it's a lot of work. It should also be possible to upgrade an existing installation, and just look at the output of 
```
ns=argocd; # or whatever
kubectl -n $ns get po -l control-plane=controller-manager;
kubectl -n $ns get po -l control-plane=controller-manager -o json | jq '.items[].spec.containers[0].image' -r
```
If the pod is running and has a kube-rbac-proxy image specified by digest instead of tag, you know it's working.

### DIFF
--- a/bundle/manifests/argocd-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/argocd-operator.clusterserviceversion.yaml
@@ -1085,7 +1085,7 @@ spec:
                 - --upstream=http://127.0.0.1:8080/
                 - --logtostderr=true
                 - --v=10
-                image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
+                image: gcr.io/kubebuilder/kube-rbac-proxy@sha256:db06cc4c084dd0253134f156dddaaf53ef1c3fb3cc809e5d81711baa4029ea4c
                 name: kube-rbac-proxy
                 ports:
                 - containerPort: 8443
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -10,7 +10,7 @@ spec:
     spec:
       containers:
       - name: kube-rbac-proxy
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
+        image: gcr.io/kubebuilder/kube-rbac-proxy@sha256:db06cc4c084dd0253134f156dddaaf53ef1c3fb3cc809e5d81711baa4029ea4c
         args:
         - "--secure-listen-address=0.0.0.0:8443"
         - "--upstream=http://127.0.0.1:8080/"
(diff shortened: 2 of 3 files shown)


jaideepr97:
Hi @mortya thanks for your PR! 

I think it would be best to leave this change out of the released version bundles 
would you mind restricting this to master and the 0.7.0 bundle?

mortya:
Hi, @jaideepr97 !  I do not at all mind modifying the PR to only apply to new cluster service versions.  Done.  Not sure what the fashion is here for correcting PRs; did it as an amend and a force push rather than a new commit.  Hope that's OK!

PR #921, opened 2023-05-31
fix: add reference doc for Extra Config (#858)

### LINKED ISSUE #858: .spec.extraConfig Missing
The documentation for `.spec.extraConfig` is missing from the document https://argocd-operator.readthedocs.io/en/latest/reference/argocd/

mortya:
**What type of PR is this?**

/kind documentation

The reference for argocd does not include extraConfig.  I spent a while looking for something like that.  Issue #858 implies I'm not the only one.

Fixes #858

### DIFF
--- a/docs/reference/argocd.md
+++ b/docs/reference/argocd.md
@@ -17,6 +17,7 @@ Name | Default | Description
 [**Controller**](#controller-options) | [Object] | Argo CD Application Controller options.
 [**Dex**](#dex-options) | [Object] | Dex configuration options.
 [**DisableAdmin**](#disable-admin) | `false` | Disable the admin user.
+[**ExtraConfig**](#extra-config) | [Empty] | A catch-all mechanism to populate the argocd-cm configmap.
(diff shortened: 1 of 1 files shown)


jaideepr97:
hi @mortya 

This is already documented here: https://github.com/argoproj-labs/argocd-operator/blob/master/docs/usage/extra-config.md

mortya:
Thanks, I did eventually find that other doc.  But it wasn't in the reference for the actual argocd CRD, which is where I initially looked.  Normally that sort of type reference lists all features of the type even if they are also documented elsewhere.  So it should document extraConfig.

mortya:
Thanks, @jaideepr97.  I did eventually find that other doc. But it wasn't in the reference for the actual argocd CRD, which is where I initially looked. Normally that sort of type reference lists all features of the type at a basic level, even if they are also (better) documented elsewhere. So it should document extraConfig.

jannfis (approved):
LGTM! Thank you!

PR #925, opened 2023-06-06
ArgoCD Operator Fails to Start: Update Operator Role for Monitoring Flag

ArthurVardevanyan:
Operator version 0.6.0 does not start with ```.spec.monitoring.enabled``` set to ```true```

/kind bug

```bash
W0606 02:35:07.896600       1 reflector.go:324] pkg/mod/k8s.io/client-go@v0.23.1/tools/cache/reflector.go:167: failed to list *v1.PrometheusRule: prometheusrules.monitoring.coreos.com is forbidden: User "system:serviceaccount:argocd:argocd-operator-controller-manager" cannot list resource "prometheusrules" in API group "monitoring.coreos.com" at the cluster scope
```

N/A: https://github.com/argoproj-labs/argocd-operator/blob/release-0.6/docs/usage/monitoring.md


jannfis:
@jaideepr97 can you PTAL at the CSV changes here? I know you had some concerns about previous version CSVs in other PRs.

jaideepr97:
Thanks @ArthurVardevanyan! 

Yeah, I don't think it makes sense to change older version bundles unless we are planning to do a follow up release for that version, maybe, as it interferes with the history should we need to reference it 

Would you mind keeping the change to master and `v0.7.0`?

ArthurVardevanyan:
> Thanks @ArthurVardevanyan!
> 
> Yeah, I don't think it makes sense to change older version bundles unless we are planning to do a follow up release for that version, maybe, as it interferes with the history should we need to reference it
> 
> Would you mind keeping the change to master and `v0.7.0`?

Updated

jaideepr97 (approved):

dan1el-k:
It seems that this fixed was missed for the 0.7.0 release.
The issue still exists and the Permission are still missing in the Role.

PR #935, opened 2023-06-13
fix: Revert appSync type implemented in #926

ItsKev:
**What type of PR is this?**
/kind bug

I've implemented a breaking API change in #926 . This PR reverts the appSync type back to `metav1.Duration` and leaves the API as it was before.

Fixes -

Set the `controller.appSync` value to 1 minute and verify that the controller performs reconciliation every minute instead of the default 3 minutes.


jannfis (approved):
Thanks a lot for fixing this, @ItsKev !

PR #944, opened 2023-07-07
updated CRDs based on argocd v2.7.6

reginapizza:
**What type of PR is this?**
> /kind chore

updates our CRDs based on [argocd 2.7.6 CRDs](https://github.com/argoproj/argo-cd/tree/v2.7.6/manifests/crds)


iam-veeramalla (approved):
Thanks for catching this @reginapizza

PR #954, opened 2023-07-26
Move to only adding two roles for managed namespaces

selrahal:
**What type of PR is this?**

/kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

**What does this PR do / why we need it**: Right now grafana and redis (non-ha) roles are created in managed namespaces. 

**Which issue(s) this PR fixes**: https://issues.redhat.com/browse/GITOPS-3156

Fixes #?


jaideepr97 (approved):
/lgtm 
Thanks @selrahal ! 
It may be worth noting that we are in the process of redesigning and rewriting this operator, so this behavior will change very soon anyway

PR #961, opened 2023-08-03
Add labelSelector option to filter the ArgoCD instances for reconciliation

raghavi101:
**What type of PR is this?**

> /kind enhancement

This PR introduces an env variable and a flag :  `ARGOCD_LABEL_SELECTOR`  and `--label-selector` which can be used to assign distinct labels to the ArgoCD instances, enabling each controller instance to be tailored to oversee only the corresponding ArgoCD instances identified by the label selector. By implementing this approach, the desired outcome can be effectively attained through targeted management and configuration of the instances.

Fixes #?

1. Start by exporting a desired label-selector : `export ARGOCD_LABEL_SELECTOR=foo=bar`
2.  On the terminal run the following to start the operator locally : `make install run` 
3. Apply a basic argocd instance manifest. eg. 

```
kind: ArgoCD
metadata:
  name: test1
  namespace: test-argocd
  labels:
    example: basic
spec: {}
```

4. The above argocd instance will not be reconciled as it is not matched with the label selector option of the operator. To reconcile the argocd instance add the same label to it : `kubectl label argocd test1 foo=bar -n test-argocd`
5. The same applies to changes in the argocd instance which has been labelled and reconciled initially but if it gets unlabelled later `kubectl label argocd test1 foo- -n test-argocd` then it will not be reconciled by the operator.


bundle/manifests/argocd-operator.clusterserviceversion.yaml
```diff
@@ -1043,13 +1043,16 @@ spec:
...
                 - --leader-elect
+                - --label-selector
                 command:
                 - /manager
                 env:
                 - name: WATCH_NAMESPACE
                   valueFrom:
                     fieldRef:
                       fieldPath: metadata.annotations['olm.targetNamespaces']
+                - name: LABEL_SELECTOR
```
ishitasequeira, line 1054: nit: maybe rename this as `ARGOCD_LABEL_SELECTOR`

controllers/argocd/argocd_controller.go
```diff
@@ -104,90 +106,109 @@ func (r *ReconcileArgoCD) Reconcile(ctx context.Context, request ctrl.Request) (
...
-	ActiveInstanceReconciliationCount.WithLabelValues(argocd.Namespace).Inc()
+	// Check if the labelSelectorValue matches the r.LabelSelector (command-line option)
+	if r.LabelSelector != labelSelectorValue {
+		fmt.Println("Skipping reconciliation as labelSelector does not match.")
+		return reconcile.Result{}, fmt.Errorf("Skipping reconciliation as labelSelector does not match for the give ArgoCD instance : %v", argocd.Name)
+	}
 
-	if argocd.GetDeletionTimestamp() != nil {
+	// Check if the $LABEL_SELECTOR environment variable is set in the argocd instance and begin reconcilliation for the same.
+	if r.LabelSelector == labelSelectorValue || labelSelectorValue == common.ArgoCDDefaultLabelSelector {
```
ishitasequeira, line 125: This check is not required as the above `if` block will satisfy the required check.

controllers/argocd/argocd_controller.go
```diff
@@ -104,90 +106,109 @@ func (r *ReconcileArgoCD) Reconcile(ctx context.Context, request ctrl.Request) (
...
-		// increment instance count with new phase and decrement instance count with old phase
-		// update the phase in corresponding map entry
-		// total instance count remains the same
-		if oldPhase := ActiveInstanceMap[argocd.Namespace]; oldPhase != newPhase {
-			ActiveInstanceMap[argocd.Namespace] = newPhase
-			ActiveInstancesByPhase.WithLabelValues(newPhase).Inc()
-			ActiveInstancesByPhase.WithLabelValues(oldPhase).Dec()
+	// Fetch the value of the $LABEL_SELECTOR environment variable from the argocd instance
+	var labelSelectorValue string
+	for _, envVar := range argocd.Spec.Server.Env {
```
ishitasequeira, line 111: The LABEL_SELECTOR value will not be in `argocd.Spec.Server.Env` instead we would need to find the label in `argocd.metadata.labels` and search for the specified key in label selectors.

ishitasequeira (changes requested):

raghavi101:
Hi @ishitasequeira thanks for your review! I've updated the code with recent changes PTAL.

config/manager/manager.yaml
```diff
@@ -55,5 +56,9 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.annotations['olm.targetNamespaces']
+        - name : ARGOCD_LABEL_SELECTOR
+        valueFrom:
+            fieldRef:
+              fieldPath: metadata.labels[ARGOCD_LABEL_SELECTOR]
```
ishitasequeira, line 62: indentation is not correct

controllers/argocd/argocd_controller.go
```diff
@@ -45,6 +46,7 @@ type ReconcileArgoCD struct {
 	ManagedNamespaces *corev1.NamespaceList
 	// Stores a list of SourceNamespaces as values
 	ManagedSourceNamespaces map[string]string
+	LabelSelector           string
```
ishitasequeira, line 49: nit: Add a comment describing the field

main.go
```diff
@@ -78,7 +79,10 @@ func main() {
 	var enableLeaderElection bool
 	var probeAddr string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", fmt.Sprintf(":%d", common.OperatorMetricsPort), "The address the metric endpoint binds to.")
+	var labelSelectorFlag string
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
```
ishitasequeira, line 83: This seems to have got copied by mistake. This line is already present in the code

ishitasequeira (changes requested):
Left a couple of comments!!

controllers/argocd/argocd_controller.go
```diff
@@ -105,6 +108,19 @@ func (r *ReconcileArgoCD) Reconcile(ctx context.Context, request ctrl.Request) (
...
+		return reconcile.Result{}, err
+	}
+
+	// Match the value of labelSelector from ReconcileArgoCD to labels from the argocd instance
+	if !labelSelector.Matches(labels.Set(argocd.Labels)) {
+		reqLogger.Info("the ArgoCD instance does not match the label selector %s and skipping for reconcillation", r.LabelSelector)
+		return reconcile.Result{}, nil
+	}
+
 	newPhase := argocd.Status.Phase
```
anandf, line 124: Can you add a log, saying the argocd instance is selected by the label selector for reconcillation and for default value, add a log message saying that all argocd instances are selected for reconcillation. This might help in debugging issues.

saumeya:
@raghavi101 Can you add steps/instructions to review the changes

bundle/manifests/argocd-operator.clusterserviceversion.yaml
```diff
@@ -1733,6 +1733,7 @@ spec:
               containers:
               - args:
                 - --leader-elect
+                - --label-selector
```
anandf, line 1736: This command line arg can be removed as the Environment variable is set. Also what would be the behaviour if both command line argument `--label-selector` and the environment variable `ARGOCD_LABEL_SELECTOR` is used but with different values ? Which one would take more precedence ?

common/defaults.go
```diff
@@ -178,6 +178,9 @@ const (
 	// ArgoCDKeycloakImage is the default Keycloak Image used for the non-openshift platforms when not specified.
 	ArgoCDKeycloakImage = "quay.io/keycloak/keycloak"
 
+	// Default Label Selector is used when the Label Selector is not mentioned/empty.
```
anandf, line 181: Probably you can mention that, `by default all ArgoCD instances would be reconciled`.

common/keys.go
```diff
@@ -228,4 +228,7 @@ const (
 
 	// ArgoCDDexSecretKey is used to reference Dex secret from Argo CD secret into Argo CD configmap
 	ArgoCDDexSecretKey = "oidc.dex.clientSecret"
+
+	// Label Selector is an env variable for ArgoCD instance reconcilliation.
+	ArgoCDLabelSelector = "ARGOCD_LABEL_SELECTOR"
```
anandf, line 233: `ArgoCDLabelSelector` can be renamed a `ArgoCDLabelSelectorKey`, like the previous ones and comments should start with the name of the variable.

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -24,6 +24,7 @@ spec:
         - "--health-probe-bind-address=:8081"
         - "--metrics-bind-address=127.0.0.1:8080"
         - "--leader-elect"
+        - "--label-selector"
```
anandf, line 27: This argument may not be required.

controllers/argocd/argocd_controller.go
```diff
@@ -104,6 +107,20 @@ func (r *ReconcileArgoCD) Reconcile(ctx context.Context, request ctrl.Request) (
...
+
+	// Fetch labelSelector from r.LabelSelector (command-line option)
+	labelSelector, err := labels.Parse(r.LabelSelector)
+	if err != nil {
+		reqLogger.Info(fmt.Sprintf("error parsing the labelSelector '%s'.", labelSelector))
+		return reconcile.Result{}, err
+	}
+	// Match the value of labelSelector from ReconcileArgoCD to labels from the argocd instance
+	if !labelSelector.Matches(labels.Set(argocd.Labels)) {
+		reqLogger.Info(fmt.Sprintf("the ArgoCD instance does not match the label selector '%s' and skipping for reconcillation", r.LabelSelector))
```
anandf, line 121: probably the log message can include the name of the Argo CD instance and its namespace to give more clarity.

go.mod
```diff
@@ -63,6 +64,7 @@ require (
 	github.com/prometheus/client_model v0.4.0 // indirect
 	github.com/prometheus/common v0.43.0 // indirect
 	github.com/prometheus/procfs v0.10.1 // indirect
+	github.com/sirupsen/logrus v1.9.3 // indirect
```
anandf, line 67: Why this new dependency ?

main.go
anandf: `8080` port is hardcoded. Probably it got introduced when resolving merge conflicts.

common/defaults.go
anandf: Also the go doc comment should start with the variable name i.e `ArgoCDDefaultLabelSelector`

controllers/argocd/argocd_controller_test.go
```diff
@@ -102,6 +102,37 @@ func TestReconcileArgoCD_Reconcile(t *testing.T) {
 	}
 }
 
+func TestReconcileArgoCD_ReconcileLabel(t *testing.T) {
```
anandf, line 105: Can you add few more test cases, where multiple argocd instances and label selector selects only few of them. Another test could be to test the default instance when label selector is not used, then all argocd instances will be reconciled.

anandf (changes requested):
Thanks @raghavi101 for these changes. Added few comments and suggestions. Please address them.

controllers/argocd/argocd_controller_test.go
```diff
@@ -102,6 +102,65 @@ func TestReconcileArgoCD_Reconcile(t *testing.T) {
 	}
 }
 
+func TestReconcileArgoCD_LabelSelector(t *testing.T) {
```
ishitasequeira, line 105: Can you add more unit tests for scenarios below?
* No label selector provided and all argocd instances are reconciled
* Multiple ArgoCD instances present with matching label present on some and absent on some. Only instances matching the label are reconciled.

docs/usage/environment_variables.md
```diff
@@ -7,6 +7,7 @@ The following environment variables are available in `argocd-operator`:
...
 | `SERVER_CLUSTER_ROLE` | none | Administrators can configure a common cluster role for all the managed namespaces in role bindings for the Argo CD server with this environment variable. Note: If this environment variable contains custom roles, the Operator doesn’t create the default admin role. Instead, it uses the existing custom role for all managed namespaces. |
 | `REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION` | false | When an Argo CD instance is deleted, namespaces managed by that instance (via the `argocd.argoproj.io/managed-by` label ) will retain the label by default. Users can change this behavior by setting the environment variable `REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION` to `true` in the Subscription. |
+| `ARGOCD_LABEL_SELECTOR` | none | The argocd-operator can be labelled (eg: `export ARGOCD_LABEL_SELECTOR=foo=bar`). Follwoing this, the argocd instances can be matched with the same label like so `kubectl label argocd test1 foo=bar -n test-argocd` which will facilitate the reconciliattion of one or more argocd instances. This will enable each controller instance to be tailored to oversee only the corresponding ArgoCD instances identified by the label selector. |
```
ishitasequeira, line 10: As we do not label argocd-operator, let's rephrase this? Something like below maybe?

```
The label selector can be set on argocd-opertor by exporting `ARGOCD_LABEL_SELECTOR` (eg: `export ARGOCD_LABEL_SELECTOR=foo=bar`). The labels can be added to the argocd instances using the command `kubectl label argocd test1 foo=bar -n test-argocd`. This will enable the operator instance to be tailored to oversee only the corresponding ArgoCD instances having the matching label selector.
```

main.go
```diff
@@ -79,8 +81,11 @@ func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
-	flag.StringVar(&metricsAddr, "metrics-bind-address", fmt.Sprintf(":%d", common.OperatorMetricsPort), "The address the metric endpoint binds to.")
+	var labelSelectorFlag string
+
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
```
ishitasequeira, line 86: Any reason to change this line?

main.go
```diff
@@ -94,6 +99,13 @@ func main() {
 
 	printVersion()
 
+	// Check the label selector format eg. "foo=bar"
+	if _, err := labels.Parse(labelSelectorFlag); err != nil {
+		setupLog.Info("error parsing the labelSelector '%s'.", labelSelectorFlag)
```
ishitasequeira, line 104: use `setupLog.Error` instead

ishitasequeira (changes requested):
@raghavi101 , left some comments. PTAL.

controllers/argocd/argocd_controller_test.go
raghavi101: Both the scenarios are covered in the unit tests but they are directed mainly towards single instance because there is a duplicity issue while using `makeTestArgoCD()`

main.go
raghavi101: thanks for spotting this

controllers/argocd/argocd_controller_test.go
ishitasequeira: We can create multiple instances and update the name/labels of the instance. Doesn't that work? Also, you could try something like:

```
a := makeTestArgoCD(func(a *argoprojv1alpha1.ArgoCD) {
		a.Name = "test1"
                a.labels = ""
	})
```

raghavi101: Made the changes

ishitasequeira (changes requested):
Just a last comment.

controllers/argocd/argocd_controller_test.go
```diff
@@ -102,6 +102,157 @@ func TestReconcileArgoCD_Reconcile(t *testing.T) {
...
+	//ctx := context.Background()
+	a := makeTestArgoCD()
+	a.Name = "argo-test-1"
+	b := makeTestArgoCD()
+	b.Name = "argo-test-2"
+	c := makeTestArgoCD()
+	c.Name = "argo-test-3"
+	rt1 := makeTestReconciler(t, a)
+	rt2 := makeTestReconciler(t, b)
+	rt3 := makeTestReconciler(t, c)
```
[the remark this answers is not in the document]
raghavi101: Yes corect, this worked. I assumed each ArgoCD instance needs a compartmentalized reconcilers to request reconciliation. But I made the above changes, it works. Thanks!

controllers/argocd/argocd_controller_test.go
```diff
@@ -102,6 +102,153 @@ func TestReconcileArgoCD_Reconcile(t *testing.T) {
...
+		},
+	}
+	resTest, err := rt.Reconcile(context.TODO(), reqTest)
+	assert.Error(t, err)
+	if resTest.Requeue {
+		t.Fatal("reconcile requeued request")
+	}
+
+	//not reconciled should return error
+	rt.LabelSelector = "foo=bar"
```
ishitasequeira, line 174: If we are already setting the LabelSelector in line 160, we should not have a need to define it here.

controllers/argocd/argocd_controller_test.go
```diff
@@ -102,6 +102,153 @@ func TestReconcileArgoCD_Reconcile(t *testing.T) {
...
+		},
+	}
+	resTest2, err := rt.Reconcile(context.TODO(), reqTest2)
+	assert.Error(t, err)
+	if resTest2.Requeue {
+		t.Fatal("reconcile requeued request")
+	}
+
+	//not reconciled should return error
+	rt.LabelSelector = "foo=bar"
```
ishitasequeira, line 188: Same here

ishitasequeira (approved):
Last few changes for repeated label assignment. But overall looks good.

ishitasequeira (approved):
Thanks @raghavi101. The PR LGTM!

anandf (approved):

config/manager/manager.yaml
```diff
@@ -55,5 +56,7 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.annotations['olm.targetNamespaces']
+        - name: ARGOCD_LABEL_SELECTOR
+          value: ""
```
svghadi, line 60: Should we remove `ARGOCD_LABEL_SELECTOR` if it's value is set to empty string here? @raghavi101 

`env.StringFromEnv` in main.go handles empty or missing env value by setting it to default.

bundle/manifests/argocd-operator.clusterserviceversion.yaml
jaideepr97: @anandf @raghavi101 was there a resolution to this comment?

controllers/argocd/argocd_controller_test.go
```diff
@@ -102,6 +102,151 @@ func TestReconcileArgoCD_Reconcile(t *testing.T) {
...
+	}
+	resTest3, err := r2.Reconcile(context.TODO(), reqTest3)
+	//Instance 'b' is not reconciled and an error is expeced
+	assert.Error(t, err)
+	if resTest3.Requeue {
+		t.Fatal("reconcile requeued request")
+	}
+
+}
+
```
jaideepr97, line 249: @raghavi101 @anandf do we need 2 separate tests for this?
I feel like this could be handled in a single combined  test case:
- create multiple instances - some with no label, others with 2 or more different labels and no label filter set
- ensure all instances reconcile
- add a labelselector that matches the labels on one of the existing instances we just made, re-reconcile all of them and now only  the one with the matching label should not error out

jaideepr97: also its a little cleaner to do 
```
a := makeTestArgoCD( func( a *argoproj.ArgoCD) {
          a. Name = "argo-test-1"
          a.Labels["some-label"] = "some val"
          })
```
instead of
```
a := makeTestArgoCD()
a.Name = "argo-test-1"
```

raghavi101: Combined both the test cases, thanks @jaideepr97

config/manager/manager.yaml
raghavi101: Cleaned it, thanks @svghadi

bundle/metadata/annotations.yaml
```diff
@@ -5,7 +5,7 @@ annotations:
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: argocd-operator
   operators.operatorframework.io.bundle.channels.v1: alpha
-  operators.operatorframework.io.metrics.builder: operator-sdk-v1.10.0+git
+  operators.operatorframework.io.metrics.builder: operator-sdk-v1.32.0
```
jaideepr97, line 8: @raghavi101 I don't think we want to make this change here

deploy/olm-catalog/argocd-operator/0.8.0/argocd-operator.v0.8.0.clusterserviceversion.yaml
```diff
@@ -218,8 +218,9 @@ metadata:
     capabilities: Deep Insights
     categories: Integration & Delivery
     certified: "false"
+    createdAt: "2023-10-18T11:54:09Z"
     description: Argo CD is a declarative, GitOps continuous delivery tool for Kubernetes.
-    operators.operatorframework.io/builder: operator-sdk-v1.10.0+git
+    operators.operatorframework.io/builder: operator-sdk-v1.32.0
```
jaideepr97, line 223: @raghavi101 here as well

bundle/metadata/annotations.yaml
raghavi101: Yeah I ran `make bundle` locally to fix the failing pipeline. Trying to figure out how to fix that CI fail. Will revert these changes back

bundle/manifests/argocd-operator.clusterserviceversion.yaml
raghavi101: @anandf correct me if I am wrong, for applying the label-selector the operator needs to be patched that can be done via setting the env var which means the flag option might become null and void. Referring the gitops-operator's [kuttl test](https://github.com/redhat-developer/gitops-operator/pull/607/files#diff-3ec4329326c83654616bbf537b95eb50b8aa36e4ae684bd6f1dff3cdfc0f1bc9R11)

anandf (approved):

PR #965, opened 2023-08-08
update control plane

reginapizza:
**What type of PR is this?**
 /kind chore

Forgot to update the control plane from argocd-controller to controller-manager in 0.7.0 release


iam-veeramalla (approved):
LGTM

PR #967, opened 2023-08-16
add build.os config for readthedocs

reginapizza:
**What type of PR is this?**
 /kind enhancement
 /kind documentation

The build.image config key on .readthedocs.yaml has been deprecated, and will be removed on October 16, 2023. Even though we weren't using it, we still have to have this key otherwise our docs will break. 

I used the same configuration that argocd upstream has: https://github.com/argoproj/argo-cd/blob/master/.readthedocs.yml

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,3 +8,7 @@ mkdocs:
 python:
   install:
     - requirements: docs/requirements.txt
+build:
+  os: "ubuntu-22.04"
+  tools:
+    python: "3.7"


.readthedocs.yml
```diff
@@ -8,3 +8,7 @@ mkdocs:
 python:
   install:
     - requirements: docs/requirements.txt
+build:
+  os: "ubuntu-22.04"
+  tools:
+    python: "3.7"
```
jaideepr97, line 14: how do we decide what goes here?

ishitasequeira: I think to avoid build failures.. ref: https://github.com/argoproj/argo-cd/pull/12956

ishitasequeira (approved):
LGTM!

.readthedocs.yml
jaideepr97: ya i meant did we pick those versions for a reason or does it not matter?

PR #974, opened 2023-08-23
code-refactoring: organize & rename constants to follow uniform format

jaideepr97:
**What type of PR is this?**

> /kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
 /kind code-refactoring

Renames and organizes all constants being used by the operator to follow a specific, uniform system
also sets up for movement of component specific constants into their own folders easily in a future PR

Defaults are now divided into sections based on components vs general argo-cd defaults
Keys are split into domain specific keys and general argo-cd keys 
env vars are extracted into a separate file
Values are split into names, suffixes and values 

All future constants should follow the same system 

Fixes #?


common/defaults.go
```diff
@@ -14,66 +14,16 @@
 
 package common
 
+// app-controller
 const (
 	// ArgoCDApplicationControllerComponent is the name of the application controller control plane component
 	ArgoCDApplicationControllerComponent = "argocd-application-controller"
 
 	// ArgoCDApplicationControllerDefaultShardReplicas is the default number of replicas that the ArgoCD Application Controller Should Use
 	ArgocdApplicationControllerDefaultReplicas = 1
```
ciiay, line 23: Should this key be like `ArgoCD...` instead of `Argocd...` as other keys?

common/defaults.go
```diff
@@ -307,30 +339,31 @@ vs-ssh.visualstudio.com ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC7Hr1oTWqNqOlzGJOf
 )
 
 // DefaultLabels returns the default set of labels for controllers.
-func DefaultLabels(name, component string) map[string]string {
+func DefaultLabels(resourceName, instanceName, component string) map[string]string {
```
ciiay, line 342: Can we put one of `resourceName` and `instanceName` as optional arg? I see a lot of `Labels:    common.DefaultLabels(cr.Name, cr.Name, ""),` which is quite confusing.

ciiay (changes requested):
Thanks for the great effort @jaideepr97 . I like the reorganization of common folder and renamed keys which follow a good naming rule. I left a few small comments. And I see you updated the keys for new `permissions` and `workloads` package, while after merging of `monitoring` and `networking` package now they are still in old keys which don't compile right. Also you may want to replace `ArgoCDManagedByLabel` with `ArgoCDArgoprojKeyManagedBy` in comments as well, only two places. Awesome work 👏

common/defaults.go
jaideepr97: we should pass in both `resourceName` and `instanceName` because they go into different labels, serving different purposes

> Labels:    common.DefaultLabels(cr.Name, cr.Name, ""),
Yeah, we are going to change this behavior as a part of the rewrite 
resource name and instance name shouldn't be the same

common/defaults.go
jaideepr97: nice catch

jaideepr97: Although, this will be changing soon in my app-controller PR when all these constants will be moved and renamed in the app-controller package

ciiay (approved):
LGTM
Thanks @jaideepr97

PR #985, opened 2023-08-31
Add tls cert mount appset controller

ishitasequeira:
**What type of PR is this?**
/kind enhancement

ArgoCD recently added support for custom TLS certificates for ApplicationSet Gitlab SCM Providers. ApplicationSetController added a new option `--scm-root-ca-path` and expects the certificate to be mounted on the path specified. As part of this PR, we added a new parameter to `spec.applicationSet.scmRootCaPath` to ArgoCD spec which takes the data from a hard-coded configMap named `argocd-appset-gitlab-scm-tls-certs-cm` containing the certificate and volume mounts it on the applicationset-controller pods on the path specified in `spec.applicationSet.scmRootCaPath`. The PR also adds a watcher on the the configMap to reconcile ArgoCD for any changes to the ConfigMap.

Fixes [GITOPS-3107](https://issues.redhat.com/browse/GITOPS-3107)

1. Start the argocd-operator using `make install run`
2. Create a configMap named `argocd-appset-gitlab-scm-tls-certs-cm` with TLS cert details.
3. Create the ArgoCD instance with below configuration in the same namespace as the configMap.
```
kind: ArgoCD
metadata:
  name: argocd
  namespace: test-1-32-appsets-scm-tls-mount
spec:
  applicationSet:
    scmRootCaPath: <path_of_the_cert_for_volume_mount>
```
4. Describe the applicationset controller pod to see if the volume and volume mount is present or not.

### DIFF
--- a/api/v1beta1/argocd_types.go
+++ b/api/v1beta1/argocd_types.go
@@ -160,7 +160,8 @@ type ArgoCDApplicationSet struct {
 
 	WebhookServer WebhookServerSpec `json:"webhookServer,omitempty"`
 
-	SCMRootCaPath string `json:"scmRootCaPath,omitempty"`
+	// SCMRootCAPath is the path where the Gitlab SCM Provider's TLS certificate is mounted on the ApplicationSet Controller.
+	SCMRootCAPath string `json:"scmRootCaPath,omitempty"`
 }
 
 // ArgoCDCASpec defines the CA options for ArgCD.
--- a/api/v1beta1/argocd_types.go
+++ b/api/v1beta1/argocd_types.go
@@ -159,6 +159,8 @@ type ArgoCDApplicationSet struct {
 	LogLevel string `json:"logLevel,omitempty"`
 
 	WebhookServer WebhookServerSpec `json:"webhookServer,omitempty"`
+
+	ScmRootCaPath string `json:"scmRootCaPath,omitempty"`
 }
 
 // ArgoCDCASpec defines the CA options for ArgCD.
--- a/api/v1beta1/argocd_types.go
+++ b/api/v1beta1/argocd_types.go
@@ -160,7 +160,7 @@ type ArgoCDApplicationSet struct {
 
 	WebhookServer WebhookServerSpec `json:"webhookServer,omitempty"`
 
-	ScmRootCaPath string `json:"scmRootCaPath,omitempty"`
+	SCMRootCaPath string `json:"scmRootCaPath,omitempty"`
 }
 
 // ArgoCDCASpec defines the CA options for ArgCD.
--- a/api/v1beta1/argocd_types.go
+++ b/api/v1beta1/argocd_types.go
@@ -160,8 +160,8 @@ type ArgoCDApplicationSet struct {
 
 	WebhookServer WebhookServerSpec `json:"webhookServer,omitempty"`
 
-	// SCMRootCAPath is the path where the Gitlab SCM Provider's TLS certificate is mounted on the ApplicationSet Controller.
-	SCMRootCAPath string `json:"scmRootCaPath,omitempty"`
+	// SCMRootCAConfigMap is the name of the config map that stores the Gitlab SCM Provider's TLS certificate which will be mounted on the ApplicationSet Controller (optional).
+	SCMRootCAConfigMap string `json:"scmRootCAConfigMap,omitempty"`
 }
 
 // ArgoCDCASpec defines the CA options for ArgCD.
--- a/common/values.go
+++ b/common/values.go
@@ -77,6 +77,9 @@ const (
 	// ArgoCDTLSCertsConfigMapName is the upstream hard-coded TLS certificate data ConfigMap name.
 	ArgoCDTLSCertsConfigMapName = "argocd-tls-certs-cm"
 
+	// ArgoCDAppSetGitlabSCMTLSCertsConfigMapName is the hard-coded ApplicationSet Gitlab SCM TLS certificate data ConfigMap name.
+	ArgoCDAppSetGitlabSCMTLSCertsConfigMapName = "argocd-appset-gitlab-scm-tls-certs-cm"
+
 	// ArgoCDRedisServerTLSSecretName is the name of the TLS secret for the redis-server
 	ArgoCDRedisServerTLSSecretName = "argocd-operator-redis-tls"
 
--- a/controllers/argocd/applicationset.go
+++ b/controllers/argocd/applicationset.go
@@ -46,9 +46,9 @@ func getArgoApplicationSetCommand(cr *argoproj.ArgoCD) []string {
 	cmd = append(cmd, "--loglevel")
 	cmd = append(cmd, getLogLevel(cr.Spec.ApplicationSet.LogLevel))
 
-	if cr.Spec.ApplicationSet.ScmRootCaPath != "" {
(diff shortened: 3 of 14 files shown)


api/v1beta1/argocd_types.go
```diff
@@ -159,6 +159,8 @@ type ArgoCDApplicationSet struct {
 	LogLevel string `json:"logLevel,omitempty"`
 
 	WebhookServer WebhookServerSpec `json:"webhookServer,omitempty"`
+
+	ScmRootCaPath string `json:"scmRootCaPath,omitempty"`
```
anandf, line 163: would it be better to name as `SCMRootCAPath`?

api/v1beta1/argocd_types.go
```diff
@@ -159,6 +159,8 @@ type ArgoCDApplicationSet struct {
 	LogLevel string `json:"logLevel,omitempty"`
 
 	WebhookServer WebhookServerSpec `json:"webhookServer,omitempty"`
+
+	SCMRootCaPath string `json:"scmRootCaPath,omitempty"`
```
anandf, line 163: Can we name it as`SCMRootCAPath` ?

common/values.go
```diff
@@ -77,6 +77,9 @@ const (
 	// ArgoCDTLSCertsConfigMapName is the upstream hard-coded TLS certificate data ConfigMap name.
 	ArgoCDTLSCertsConfigMapName = "argocd-tls-certs-cm"
 
+	// ArgoCDAppSetGitlabSCMTLSCertsConfigMapName is the hard-coded ApplicationSet Gitlab SCM TLS certificate data ConfigMap name.
+	ArgoCDAppSetGitlabSCMTLSCertsConfigMapName = "argocd-appset-gitlab-scm-tls-certs-cm"
```
anandf, line 81: Any specific reason for going with config map and not a secret ?

ishitasequeira: Followed what we have used for TLS certs.

api/v1beta1/argocd_types.go
ishitasequeira: Updated

docs/reference/argocd.md
```diff
@@ -85,6 +85,7 @@ Resources | [Empty] | The container compute resources.
 LogLevel | info | The log level to be used by the ArgoCD Application Controller component. Valid options are debug, info, error, and warn.
 LogFormat | text | The log format to be used by the ArgoCD Application Controller component. Valid options are text or json.
 ParallelismLimit | 10 | The kubectl parallelism limit to set for the controller (`--kubectl-parallelism-limit` flag)
+SCMRootCaPath (#add-tls-certificate-for-gitlab-scm-provider-to-applicationsets-controller) | [Empty] | The path where the Gitlab SCM Provider's TLS certificate is mounted on the ApplicationSet Controller. The TLS certificate is picked from a configMap `"argocd-appset-gitlab-scm-tls-certs-cm"` and mounted on the applicationset-controller as a volume mount.
```
anandf, line 88: Correct  name to `SCMRootCAPath`

tests/k8s/1-033_validate_applicationset_tls_scm_volume_mount/01-install.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+---
+apiVersion: argoproj.io/v1beta1
+kind: ArgoCD
+metadata:
+  name: argocd
+  namespace: test-1-32-appsets-scm-tls-mount
+spec:
+  applicationSet:
+    scmRootCaPath: /app/tls/cert
```
anandf, line 56: Can user provide any path to mount the CM volumen ? Are there any restrictions on what this path can be ? Like should be under `/app/tls` ?

controllers/argocd/applicationset.go
```diff
@@ -144,9 +149,26 @@ func (r *ReconcileArgoCD) reconcileApplicationSetDeployment(cr *argoproj.ArgoCD,
 			},
 		},
 	}
+	addSCMGitlabVolumeMount := false
+	if cr.Spec.ApplicationSet.SCMRootCAPath != "" {
```
anandf, line 153: Why should we get the CA path from the user, if the operator is going to mount it from a well defined CM. Can the operator not create an empty config map `argocd-appset-gitlab-scm-tls-certs-cm` and mount it in a well defined path like `/app/tls/cert` and if at all user wants to use this feature, all he has to do is to fill Certificate date in config map `argocd-appset-gitlab-scm-tls-certs-cm`. It would be easy for the operator to manage the `Volume` and `VolumeMounts` as its no longer conditonal elements. Since the mount path is well defined, there won't be any path restrictions that user has to be aware of.

anandf: The alternate approach, would be similar to how TLS certs are handled in ArgoCD Server.

```
oc get deploy -n openshift-gitops openshift-gitops-server -o yaml
....
- mountPath: /app/config/tls
          name: tls-certs
....
volumes:
      - configMap:
          defaultMode: 420
          name: argocd-tls-certs-cm
        name: tls-certs
```

```
oc get cm -n openshift-gitops argocd-tls-certs-cm -o yaml

apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-tls-certs-cm
  namespace: openshift-gitops
```

tests/k8s/1-033_validate_applicationset_tls_scm_volume_mount/01-install.yaml
ishitasequeira: User can provide any path to mount the CM volume as long as controller has access to write to that path.

controllers/argocd/applicationset.go
ishitasequeira: We would not want to create an empty CM as the requests for Gitlab client would use these certs only if the parameter `--scm-root-ca-path` is set on the ApplicationController pod. And, we would want to set this path only if user has provided the ConfigMap.

I do like the thought of hard-coding the path of the cert, but I am wondering if there would be any use case where user would want to specify the path. I am open to hard-coding the path for volume mount.

ishitasequeira: Yes, we can do that. I will update the code.

anandf: The one you had suggested in the example, `/app/tls/cert` looks good to me. If we want to give some kind of user configurable option, to explicitly enable this feature, then we can consider taking in the config map name in the ArgoCD spec and keep the mount path hard coded to `/app/tls/cert`

tests/k8s/1-033_validate_applicationset_tls_scm_volume_mount/01-install.yaml
anandf: If a user, who might not the directory structure of the AppSet controller container, provide an invalid directory, then there would be errors. If say using a restricted path like `/cert` or `/etc/tls/cert` needing root access are provided, then it would result in unexpected errors.

ishitasequeira: I updated to use `/app/tls/scm/cert` as the path. I also took ConfigMap Name as the input from users, not restricting them to a specific ConfigMap.

controllers/argocd/applicationset.go
anandf: /lgtm

anandf (approved):

anandf (approved):

PR #986, opened 2023-08-31
code-refactoring: Gitops 3126 implement notifications controller

ciiay:
**What type of PR is this?**
/kind code-refactoring

Implemented notifications controller.

NOTE: this PR just adds the package files and tests, this code is not invoked at any point yet

Fixes #?
GITOPS-3126: https://issues.redhat.com/browse/GITOPS-3126

Unit tests are yet included. Coming soon.


controllers/argocd/notifications/configmap.go
```diff
@@ -0,0 +1,81 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func (nr *NotificationsReconciler) reconcileConfigMap() error {
+
+	nr.Logger.Info("reconciling configMaps")
+
+	name := argoutil.GenerateUniqueResourceName(nr.Instance.Name, nr.Instance.Namespace, ArgoCDNotificationsControllerComponent)
+	labels := common.DefaultLabels(name, nr.Instance.Name, ArgoCDNotificationsControllerComponent)
```
jaideepr97, line 19: @ciiay please move these to the main `notifications_controller.go` file before the reconciliations are triggered 
create global variables called `resourceName` and `resourceLabels` which can be accessed throughout the package

controllers/argocd/notifications/configmap.go
```diff
@@ -0,0 +1,81 @@
...
+	labels[common.AppK8sKeyName] = name
```
jaideepr97, line 20: This is already handled by `DefaultLabels` so this line is not required

controllers/argocd/notifications/configmap.go
```diff
@@ -0,0 +1,81 @@
...
+
+	configMapRequest := workloads.ConfigMapRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        name,
+			Namespace:   nr.Instance.Namespace,
+			Labels:      labels,
+			Annotations: nr.Instance.Annotations,
+		},
+	}
```
jaideepr97, line 29:
```suggestion
	configMapRequest := workloads.ConfigMapRequest{
		ObjectMeta: metav1.ObjectMeta{
			Name:        name,
			Namespace:   nr.Instance.Namespace,
			Labels:      labels,
			Annotations: nr.Instance.Annotations,
		},
		Data: make(map[string]string), 
	}
```

controllers/argocd/notifications/configmap.go
```diff
@@ -0,0 +1,81 @@
...
+		if err = workloads.CreateConfigMap(desiredConfigMap, nr.Client); err != nil {
+			nr.Logger.Error(err, "reconcileConfigMap: failed to create configMap", "name", desiredConfigMap.Name, "namespace", desiredConfigMap.Namespace)
+			return err
+		}
+		nr.Logger.V(0).Info("reconcileConfigMap: configMap created", "name", desiredConfigMap.Name, "namespace", desiredConfigMap.Namespace)
+		return nil
+	}
+
+	nr.Logger.V(0).Info("reconcileConfigMap: configMap updated", "name", existingConfigMap.Name, "namespace", existingConfigMap.Namespace)
+	return nil
```
jaideepr97, line 71: currently the operator does not overwrite updates made to this config map manually 

@ciiay we should not be logging that the operator updated the config map if it did not perform any updates

controllers/argocd/notifications/constants.go
```diff
@@ -0,0 +1,16 @@
...
+	ArgoCDNotificationsControllerComponent = "notifications-controller"
+	RoleKind                               = "Role"
+	TLSCerts                               = "tls-certs"
+	ArgoCDRepoServerTLS                    = "argocd-repo-server-tls"
+	CapabilityDropAll                      = "ALL"
+	VolumeMountPathTLS                     = "/app/config/tls"
+	VolumeMountPathRepoServerTLS           = "/app/config/reposerver/tls"
+	WorkingDirApp                          = "/app"
+	ImageUpgradedLabel                     = "image.upgraded"
+	TimeFormatMST                          = "01022006-150406-MST"
```
jaideepr97, line 15: most of these constants are not notifications related, and are more general purpose. They should not be defined in the notifications package

I will raise a separate PR to put them in a common location, which can be referenced

controllers/argocd/notifications/deployment.go
```diff
@@ -0,0 +1,212 @@
...
+				},
+			},
+			WorkingDir: WorkingDirApp,
+		}},
+		SecurityContext: &corev1.PodSecurityContext{
+			RunAsNonRoot: argoutil.BoolPtr(true),
+		},
+	}
+
+	AddSeccompProfileForOpenShift(nr.Logger, nr.Client, &podSpec)
```
jaideepr97, line 92: I think this function should be a mutation, and registered with the other mutations separately, so it need not be called here 

I will include that in a separate PR. Please remove this line

controllers/argocd/notifications/deployment.go
```diff
@@ -0,0 +1,212 @@
...
+			},
+		},
+
+		Client:    nr.Client,
+		Mutations: []mutation.MutateFunc{mutation.ApplyReconcilerMutation},
+	}
+
+	if replicas := GetArgoCDNotificationsControllerReplicas(nr); replicas != nil {
+		deploymentRequest.Spec.Replicas = replicas
+	}
```
jaideepr97, line 127: @ciiay lets try and avoid making changes to the deployment received from the workloads package, as this runs the risk of overwriting any mutations that have been executed already (mutations are meant to be run at the very last minute to ensure they are not overwritten)

I would suggest setting the replicas while putting together the `deploymentRequest`

controllers/argocd/notifications/notifications.go
```diff
@@ -2,20 +2,91 @@ package notifications
...
+		nr.Logger.Info("reconciling notifications deployment")
+		return err
+	}
+
 	return nil
 }
+
+func (nr *NotificationsReconciler) DeleteResources() error {
+
+	name := argoutil.GenerateUniqueResourceName(nr.Instance.Name, nr.Instance.Namespace, ArgoCDNotificationsControllerComponent)
```
jaideepr97, line 58: don't need this once the resourceName is set for the whole package to access

controllers/argocd/notifications/role.go
```diff
@@ -0,0 +1,151 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func (nr *NotificationsReconciler) reconcileRole() error {
+
+	nr.Logger.Info("reconciling roles")
+
+	roleRequest := permissions.RoleRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        argoutil.GenerateUniqueResourceName(nr.Instance.Name, nr.Instance.Namespace, ArgoCDNotificationsControllerComponent),
```
jaideepr97, line 23: use name defined globally in package

controllers/argocd/notifications/rolebinding.go
```diff
@@ -0,0 +1,104 @@
...
+
+	if !reflect.DeepEqual(existingRoleBinding.RoleRef, desiredRoleBinding.RoleRef) ||
+		!reflect.DeepEqual(existingRoleBinding.Subjects, desiredRoleBinding.Subjects) {
+		existingRoleBinding.RoleRef = desiredRoleBinding.RoleRef
+		existingRoleBinding.Subjects = desiredRoleBinding.Subjects
+		if err = permissions.UpdateRoleBinding(existingRoleBinding, nr.Client); err != nil {
+			nr.Logger.Error(err, "reconcileRoleBinding: failed to update roleBinding", "name", existingRoleBinding.Name, "namespace", existingRoleBinding.Namespace)
+			return err
+		}
+	}
```
jaideepr97, line 90: I think we should try to stay consistent in our coding style across the package 
I would suggest using the same table style logic to conduct resource update used in `deployment.go`

controllers/argocd/notifications/role.go
```diff
@@ -0,0 +1,151 @@
...
+			Namespace:   nr.Instance.Namespace,
+			Labels:      nr.Instance.Labels,
```
jaideepr97, line 25: we should call util.DefaultLabels here to get all the labels we need

controllers/argocd/notifications/rolebinding.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		nr.Logger.Error(err, "reconsileRoleBinding: failed to get serviceaccount", "name", name, "namespace", nr.Instance.Namespace)
+		return err
+	}
+
+	roleBindingRequest := permissions.RoleBindingRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        name,
+			Namespace:   nr.Instance.Namespace,
+			Labels:      nr.Instance.Labels,
```
jaideepr97, line 32: we should call util.DefaultLabels here to get all the labels we need

controllers/argocd/notifications/secret.go
```diff
@@ -0,0 +1,83 @@
...
+
+	nr.Logger.Info("reconciling secrets")
+
+	name := argoutil.GenerateUniqueResourceName(nr.Instance.Name, nr.Instance.Namespace, ArgoCDNotificationsControllerComponent)
+
+	secretRequest := workloads.SecretRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: nr.Instance.Namespace,
+			Labels: map[string]string{
```
jaideepr97, line 25: we should call util.DefaultLabels here to get all the labels we need

pkg/workloads/deployment.go
```diff
@@ -101,3 +102,14 @@ func RequestDeployment(request DeploymentRequest) (*appsv1.Deployment, error) {
...
+func UpdateNodePlacement(existing *appsv1.Deployment, deploy *appsv1.Deployment, changed *bool) {
+	if !reflect.DeepEqual(existing.Spec.Template.Spec.NodeSelector, deploy.Spec.Template.Spec.NodeSelector) {
+		existing.Spec.Template.Spec.NodeSelector = deploy.Spec.Template.Spec.NodeSelector
+		*changed = true
+	}
+	if !reflect.DeepEqual(existing.Spec.Template.Spec.Tolerations, deploy.Spec.Template.Spec.Tolerations) {
+		existing.Spec.Template.Spec.Tolerations = deploy.Spec.Template.Spec.Tolerations
+		*changed = true
+	}
+}
```
jaideepr97, line 115: I don't think this function belongs here 
seems like this is something that could just be handled by `UpdateIfChanged` and we may not need this function at all

controllers/argocd/notifications/util.go
```diff
@@ -0,0 +1,664 @@
...
+	if tag == "" {
+		tag = common.ArgoCDDefaultArgoVersion
+		defaultTag = true
+	}
+	if e := os.Getenv(common.ArgoCDImageEnvVar); e != "" && (defaultTag && defaultImg) {
+		return e
+	}
+
+	return argoutil.CombineImageTag(img, tag)
+}
```
jaideepr97, line 620: This function belongs in a common package accessible to all controller packages 

I will move it in a separate PR

controllers/argocd/notifications/util.go
```diff
@@ -0,0 +1,664 @@
...
+func caseInsensitiveGetenv(s string) (string, string) {
+	if v := os.Getenv(s); v != "" {
+		return s, v
+	}
+	ls := strings.ToLower(s)
+	if v := os.Getenv(ls); v != "" {
+		return ls, v
+	}
+	return "", ""
+}
```
jaideepr97, line 643: These have been moved to the util package in a separate PR

controllers/argocd/notifications/util.go
```diff
@@ -0,0 +1,664 @@
...
+
+func UpdateIfChanged(existingVal, desiredVal interface{}, extraAction func(), changed *bool) {
+	if !reflect.DeepEqual(existingVal, desiredVal) {
+		reflect.ValueOf(existingVal).Elem().Set(reflect.ValueOf(desiredVal).Elem())
+		if extraAction != nil {
+			extraAction()
+		}
+		*changed = true
+	}
+}
```
jaideepr97, line 599: This function belongs in a common package accessible to all controller packages

I will move it in a separate PR

controllers/argocd/notifications/util.go
```diff
@@ -0,0 +1,664 @@
...
+			podspec.SecurityContext = &corev1.PodSecurityContext{}
+		}
+		if podspec.SecurityContext.SeccompProfile == nil {
+			podspec.SecurityContext.SeccompProfile = &corev1.SeccompProfile{}
+		}
+		if len(podspec.SecurityContext.SeccompProfile.Type) == 0 {
+			podspec.SecurityContext.SeccompProfile.Type = corev1.SeccompProfileTypeRuntimeDefault
+		}
+	}
+}
```
jaideepr97, line 664: This function belongs in the mutations package 

I will move it in a separate PR

controllers/argocd/notifications/testing.go
jaideepr97, line 1: This looks like it needs to be removed into a separate, commonly accessible package

controllers/argocd/notifications/secret.go
```diff
@@ -0,0 +1,83 @@
...
+				common.AppK8sKeyName: name,
+			},
+		},
```
jaideepr97, line 28:
```suggestion
		ObjectMeta: metav1.ObjectMeta{
				Name:        resourceName,
				Namespace:   nr.Instance.Namespace,
				Labels:      resourceLabels,
				Annotations: nr.Instance.Annotations,
			},
```

controllers/argocd/notifications/role.go
```diff
@@ -0,0 +1,151 @@
...
+			Annotations: nr.Instance.Annotations,
+		},
```
jaideepr97, line 27:
```suggestion
		ObjectMeta: metav1.ObjectMeta{
			Name:        resourceName,
			Namespace:   nr.Instance.Namespace,
			Labels:      resourceLabels,
			Annotations: nr.Instance.Annotations,
		},
```

controllers/argocd/notifications/configmap.go
ciiay: @jaideepr97 Shall I replace it with `reconcileConfigMap: configMap reconciled`?

controllers/argocd/notifications/role.go
ciiay: This is set to a global value already in the `reconcile()` function in `notifications.go`.

ciiay: Once your PR gets merged, I will update it to `util.DefaultLabels()` instead.

controllers/argocd/notifications/rolebinding.go
ciiay: Global value and only call `DefaultLabels` once in the reconcile() function in `notifications.go`.

controllers/argocd/notifications/testing.go
ciiay: Something like `controller/argocd/controllers_testing.go`?

pkg/workloads/deployment.go
ciiay: Good catch.

controllers/argocd/notifications/notifications.go
```diff
@@ -2,20 +2,91 @@ package notifications
...
 	Scheme   *runtime.Scheme
 	Instance *v1alpha1.ArgoCD
 	Logger   logr.Logger
 }
 
 func (nr *NotificationsReconciler) Reconcile() error {
 
-	// controller logic goes here
+	nr.Logger = ctrl.Log.WithName(ArgoCDNotificationsControllerComponent).WithValues("instance", nr.Instance.Name, "instance-namespace", nr.Instance.Namespace)
+
```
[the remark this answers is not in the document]
ciiay: @jaideepr97 Shouldn't we check this before notifications.go gets called? I think this logic should be located in a higher level function call. Like in the `reconcileControllers()` functions in the current `argocd_controller.go` file.

controllers/argocd/notifications/configmap.go
ciiay: Removing it as discussed.

controllers/argocd/notifications/notifications.go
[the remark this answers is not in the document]
ciiay: As discussed, moving forward with the above solution for now.

controllers/argocd/notifications/testing.go
ciiay: Moving this `testing.go` to `controller/argocd/controllerCommon/` package as discussed. Leaving notifications controller specific testing functions as it is now.

jaideepr97, line 1: why do we need a separate file for this? I think we can move this stuff into `notifcations_test.go`

controllers/argocd/notifications/util.go
jaideepr97: @ciiay please rebase, since these functions have been moved to general locations we can get rid of them from here

controllers/argocd/notifications/secret.go
```diff
@@ -0,0 +1,76 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func (nr *NotificationsReconciler) reconcileSecret() error {
+
+	nr.Logger.Info("reconciling secrets")
+
+	secretRequest := workloads.SecretRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      resourceName,
```
jaideepr97, line 19: this secret needs to be called `argocd-notifications-secret`, and can't follow the generic resource naming system
Please create a new constant in `constants.go` for `NotificationsSecretName` and use that here instead

controllers/argocd/notifications/configmap.go
```diff
@@ -0,0 +1,75 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func (nr *NotificationsReconciler) reconcileConfigMap() error {
+
+	nr.Logger.Info("reconciling configMaps")
+
+	configMapRequest := workloads.ConfigMapRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        resourceName,
```
jaideepr97, line 18: this secret needs to be called `argocd-notifications-cm`, and can't follow the generic resource naming system
Please create a new constant in `constants.go` for `NotificationsConfigMapName` and use that here instead

controllers/argocd/notifications/deployment.go
jaideepr97: @ciiay please remove this line from here

controllers/argocd/notifications/deployment.go
```diff
@@ -0,0 +1,202 @@
...
+			},
+			Selector: &metav1.LabelSelector{
+				MatchLabels: resourceLabels,
+			},
+			Replicas: GetArgoCDNotificationsControllerReplicas(nr),
+		},
+
+		Client:    nr.Client,
+		Mutations: []mutation.MutateFunc{mutation.ApplyReconcilerMutation},
+	}
```
jaideepr97, line 116: I think we should extract the creation of desired deployment and deploymentRequest into 2 separate functions of their own and keep only the reconciliation of the deployment in `reconcileDeployment`. That will allow us to reuse the desired deployment spec in unit tests as well and reduce the size of the `reconcileDeployment` function

jaideepr97:
some general comments about the unit tests

1) We should really avoid using the test name as a variable for conditional checks, I don' think that is good coding practice. If needed lets split the test up into multiple tests

2) there is no need to try and preserve/restore `resourceName` and `resourceLabels`. These variables are re-initialized each cycle, and setting them to something in a test is not going to break the operator 

3) If the operator is not handling a case in the reconciliation cycle, there is no need to add a test for it. Let us simplify the test cases to create/update/delete scenarios

controllers/argocd/notifications/constants.go
```diff
@@ -0,0 +1,27 @@
...
+	APIVersionV1                           = "v1"
+	APIVersionAppsV1                       = "apps/v1"
+	APIVersionRbacV1                       = "rbac.authorization.k8s.io/v1"
+	APIGroupRbacV1                         = "rbac.authorization.k8s.io"
```
jaideepr97, line 26: Some of these have been extracted to the `common` folder
other non-notifications specific constants should also be treated the same 

I think we should create an API and resources section in `common.values.go` to add these strings there and move them out of this package

jaideepr97 (approved):

jaideepr97:
LGTM, thanks a lot @ciiay !!

PR #990, opened 2023-09-07
code-refactoring: generalize some functions to be more accessible to all controller packages

jaideepr97:
**What type of PR is this?**

> /kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
 /kind code-refactoring

- enables CI on operator-redesign branch
- moves `AddSCCForOpenShfit` function into the mutation package
- introduce `controllerCommon` package for code common to all controllers

Fixes #?


controllers/argocd/controllerCommon/helper.go
ciiay, line 1: Are we going to need this in places other than argocd controllers? If no why we have to make two separate helper files?

pkg/cluster/namespace.go
```diff
@@ -19,7 +19,7 @@ type NamespaceRequest struct {
 
 	// array of functions to mutate role before returning to requester
 	Mutations []mutation.MutateFunc
-	Client    interface{}
+	Client    ctrlClient.Client
```
ciiay, line 22: Can we name controller a better way? Probably `cntlr`? `ctrl` makes me feel it's `control`. I know it's from the existing name but it doesn't follow a clear abbreviation rule.

ref: https://www.abbreviations.com/abbreviation/controller

controllers/argocd/util.go
ciiay, line 1: Just a question, will this util.go go away eventually when we finish redesign completely? Because I want to align up import package names across the project.

For example we have different names for same `"github.com/argoproj-labs/argocd-operator/api/v1alpha1"` package everywhere.

controllers/argocd/controllerCommon/argohelper.go
```diff
@@ -0,0 +1,29 @@
+package controllerCommon
```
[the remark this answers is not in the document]
ciiay: Maybe just `cntlrcmn`?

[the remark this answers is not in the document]
jaideepr97: changed to argocdcommon

controllers/argocd/util.go
jaideepr97: as decided, we will do this in a separate PR

pkg/cluster/namespace.go
jaideepr97: updated

ciiay (approved):
LGTM
Thanks!

PR #1015, opened 2023-10-06
code-refactoring: Fixed nil pointer runtime panic error

ciiay:
**What type of PR is this?**
/kind bug
/kind code-refactoring

Fixed nil pointer runtime panic error. When there is a get resource error, the `existingResource` would be nil, and there would be a nil point error on `existingResource.Name` and `existingResourceNamespace`. So it has to be `desiredResource.Name` and `desiredResource.Namespace` instead.

NOTE: this PR just adds the package files and tests, this code is not invoked at any point yet

Fixes #?
Notifications controller and ApplicationSet controller.

Peer review.


jaideepr97 (approved):
nice catch @ciiay 
LGTM, thanks

PR #1016, opened 2023-10-06
code-refactoring: implement repo-server 

ciiay:
What type of PR is this?
/kind code-refactoring

What does this PR do / why we need it:
This PR implements the repo-server controller and adds unit tests where possible.
This PR does not include end-to-end tests since there is a dependency on other controllers for that 

Have you updated the necessary documentation?

 Documentation update is required by this PR.
 Documentation has been updated.
Which issue(s) this PR fixes:

Fixes #?
GITOPS-3125: https://issues.redhat.com/browse/GITOPS-3125

How to test changes / Special notes to the reviewer:
- run Argo CD operator
- apply the following CR:
```
kind: ArgoCD
metadata:
  name: project-gitops
spec:
  repo:
    enabled: true

```
- confirm that repo-server resources come up, and repo-server deployment is running
- disable repo-server as follows
```
kind: ArgoCD
metadata:
  name: project-gitops
spec:
  repo:
    enabled: false
```
- confirm that repo-server resources are deleted


controllers/argocd/notifications/configmap.go
```diff
@@ -27,7 +27,6 @@ func (nr *NotificationsReconciler) reconcileConfigMap() error {
 
 	if err != nil {
 		nr.Logger.Error(err, "reconcileConfigMap: failed to request configMap", "name", desiredConfigMap.Name, "namespace", desiredConfigMap.Namespace)
-		nr.Logger.V(1).Info("reconcileConfigMap: one or more mutations could not be applied")
```
jaideepr97, line 30: why did we get rid of this?

controllers/argocd/reposerver/tlssecret.go
jaideepr97, line 1: we can just call this `secret.go`

pkg/workloads/deployment.go
```diff
@@ -101,3 +105,41 @@ func RequestDeployment(request DeploymentRequest) (*appsv1.Deployment, error) {
...
+	}
+
+	currentDeployment.Spec.Template.ObjectMeta.Labels[key] = NowNano()
+	return UpdateDeployment(currentDeployment, client)
+}
+
+// nowNano returns a string with the current UTC time as epoch in nanoseconds
+func NowNano() string {
+	return fmt.Sprintf("%d", time.Now().UTC().UnixNano())
+}
```
jaideepr97, line 145: this belongs in `util/time.go` if it is not there already

pkg/workloads/deployment.go
```diff
@@ -101,3 +105,41 @@ func RequestDeployment(request DeploymentRequest) (*appsv1.Deployment, error) {
...
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+		return nil
+	}
+
+	currentDeployment.Spec.Template.ObjectMeta.Labels[key] = NowNano()
+	return UpdateDeployment(currentDeployment, client)
+}
```
jaideepr97, line 140: @ciiay 

I'm thinking a better approach for this might be to define this method that only takes in a deployment name, and some modification function, runs that modification and then updates the deployment. Something like
```
func TriggerDeploymentRollout( deployment *appsv1.Deployment, o option) error {
	currentDeployment, err := GetDeployment(deployment.Name, deployment.Namespace, client)
	if err != nil {
		if !errors.IsNotFound(err) {
			return err
		}
		return nil
	}
        // run options fn
	o()
	updateDeployment()
	return
}
```
and each controller that needs this should define this for themselves, so that the final call on these deployments is made by the controller they belong to

pkg/workloads/deployment.go
```diff
@@ -101,3 +105,41 @@ func RequestDeployment(request DeploymentRequest) (*appsv1.Deployment, error) {
...
+		},
+	}
+
+	deployment, err := RequestDeployment(deploymentRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	return deployment, nil
+}
```
jaideepr97, line 126: we should not be needing this function here

controllers/argocd/notifications/configmap.go
ciiay: @jaideepr97 I didn't get why request configMap is "one or more mutation could not be applied". Can you explain when do we need to log this and why?

PR #1018, opened 2023-10-10
code-refactoring: rebase against master branch 

jaideepr97:
**What type of PR is this?**

> /kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
 /kind code-refactoring

This PR pulls in latest changes from master branch into the operator-redesign branch in order to make sure this branch keeps up with the non-refactoring related changes in the master branch 
Not doing this often enough can lead to large and complex merge conflicts further down the line 

Some major changes introduced that we should observe going forwards:
- Concurrent development of new operator while leaving existing code and packages untouched for the most part. We want to avoid introducing any changes to existing code as far as possible, to keep things from breaking and make sure the branch always compiles. The only exception is parts of existing code that will carry into the new code base as well (such as changes to `argocd-_types.go`). Changes to these files are acceptable since they will continue to remain and need to be updated. Avoid changes to files that are definitely going away (such as `deployments.go` for ex)
- Addition of `TOBEREMOVED.go` in every package
Instead of deleting a piece of code that has been refactored/moved/renamed (to be used by the new code), duplicate that code in the new location and move the existing function/set of functions into that package's `TOBEREMOVED.go`. This ensures existing code does not break (references remain in tact) and we have a way to track which code has not yet been replicated in the new codebase
- If renaming/moving constants, do the same
move existing constant to `TOBEREMOVED` and create the renamed constant in the correct location
- If moving/replacing unit tests around, do the same
Move existing unit test to `TOBEREMOVED_test.go` so we don't lose the existing unit test
- Moving component constants to `common` package
In order to avoid potential import cycles it might be best to keep all component constants in the `common` package as oppposed to their own individual packages as was decided before. However, each component gets a dedicated file for it self (`appcontroller.go`, `redis.go` etc) so as to still maintain some isolation and organization for those components
In the future maybe we can keep only the publicly needed constants in `common` for each component but for now let's keep all constants together 
- Separation of `pkg/argoutil` and `pkg/util` packages
Going forwards, `pkg/util` is reserved for pure utility functions that are not kubernetes/argocd-operator specific. This includes things like string/bit manipulation, data structure operations etc. Each unique function should be in a dedicated file following existing patterns. For any "utility" functions that are kubernetes/argo-cd/operator specific place them in `pkg/argoutil` package. This includes things like k8s client/api manipulation/communications,  argo-cd admin password/security related stuff etc
- maintaining both `ReconcileArgoCD` and `ArgoCDReconciler`
`ArgoCDReconciler` is the updated reconciler struct, while `ReconcileArgoCD` is the old reconciler struct. We should maintain both temporarily, since all of the existing code base requires the continued existence of `ReconcileArgoCD`, and keeping it around reduces the no. of changes to old code during rebase against master

Fixes #?

### DIFF
--- a/controllers/argocd/applicationset/deployment.go
+++ b/controllers/argocd/applicationset/deployment.go
@@ -47,7 +47,7 @@ func (asr *ApplicationSetReconciler) reconcileDeployment() error {
 	existingDeployment, err := workloads.GetDeployment(desiredDeployment.Name, desiredDeployment.Namespace, asr.Client)
 	if err != nil {
 		if !errors.IsNotFound(err) {
-			asr.Logger.Error(err, "reconcileDeployment: failed to retrieve deployment", "name", existingDeployment.Name, "namespace", existingDeployment.Namespace)
+			asr.Logger.Error(err, "reconcileDeployment: failed to retrieve deployment", "name", desiredDeployment.Name, "namespace", desiredDeployment.Namespace)
 			return err
 		}
 
--- a/controllers/argocd/applicationset/role.go
+++ b/controllers/argocd/applicationset/role.go
@@ -51,7 +51,7 @@ func (asr *ApplicationSetReconciler) reconcileRole() error {
 	existingRole, err := permissions.GetRole(desiredRole.Name, desiredRole.Namespace, asr.Client)
 	if err != nil {
 		if !errors.IsNotFound(err) {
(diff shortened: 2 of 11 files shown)


ciiay:
Hi @jaideepr97 , is this PR still valid?

PR #1021, opened 2023-10-19
allow enabling ArgoCD workloads independently

ishitasequeira:
**What type of PR is this?**
/kind enhancement

Currently all ArgoCD core components are deployed by default. This PR allows to enable/disable the ArgoCD core components individually.

Fixes #?
* https://issues.redhat.com/browse/GITOPS-3388


controllers/argocd/util.go
```diff
@@ -1602,3 +1602,7 @@ func getApplicationSetHTTPServerHost(cr *argoproj.ArgoCD) string {
 	}
 	return host
 }
+
+func IsComponentEnabled(enabled *bool) bool {
```
anandf, line 1606: Can we have separate functions for each component ? Like `IsRedisEnabled` `IsRepoServerEnabled` `IsApplicationSetEnabled` which can take in a `Spec` object and return `true` or `false` based on the `enabled` flag value for the respective component.  The current function adds little value, as the caller of the function can directly check if the boolean value (`enabled`) is `true` or `false`

ishitasequeira: Sure. I kept the function generic so that it can be used generically for every component as the check remains same.

ishitasequeira:
Do we need to check for depending components to be available as well? For example, `argocd-application-controller` depends on `argocd-repo-server` and `argocd-redis`. So, for starting `argocd-application-controller`, should we check if components `argocd-repo-server` and `argocd-redis` are available?

ishitasequeira:
@anandf , as discussed offline, I have updated the code to skip installing the component if depending components are disabled.

controllers/argocd/deployment.go
```diff
@@ -624,6 +619,11 @@ func (r *ReconcileArgoCD) reconcileRedisDeployment(cr *argoproj.ArgoCD, useTLS b
 
 	existing := newDeploymentWithSuffix("redis", "redis", cr)
 	if argoutil.IsObjectFound(r.Client, cr.Namespace, existing.Name, existing) {
+		if !cr.Spec.Redis.IsEnabled() {
+			// Deployment exists but component enabled flag has been set to false, delete the Deployment
+			log.Info("Redis exists but should be disabled. Deleting existing redis.")
```
anandf, line 624: Error message could be reworded. `Redis deployment must not exist as it is disabled in the spec. Removing the redis deployment`

controllers/argocd/deployment.go
```diff
@@ -1103,6 +1108,13 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argoproj.ArgoCD, useTLSFor
 
 	existing := newDeploymentWithSuffix("repo-server", "repo-server", cr)
 	if argoutil.IsObjectFound(r.Client, cr.Namespace, existing.Name, existing) {
+
+		if !cr.Spec.Repo.IsEnabled() {
+			log.Info("Existing ArgoCD Repo Server found but should be disabled. Deleting Repo Server")
+			// Delete existing deployment for ArgoCD Repo Server, if any ..
+			return nil
```
anandf, line 1115: delete call is missing. Is that something you missed ?

ishitasequeira: Updated.

controllers/argocd/deployment.go
```diff
@@ -1171,6 +1183,18 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argoproj.ArgoCD, useTLSFor
 		return nil // Deployment found with nothing to do, move along...
 	}
 
+	// Skipping creation of repo server if Redis is disabled
+	if !cr.Spec.Redis.IsEnabled() {
+		err := fmt.Errorf("redis configuration is disabled and redis location not configured")
+		log.Error(err, "redis configuration is disabled and redis location not configured")
```
anandf, line 1189: Check error message. redis location not supported yet.

controllers/argocd/deployment.go
```diff
@@ -1292,6 +1316,11 @@ func (r *ReconcileArgoCD) reconcileServerDeployment(cr *argoproj.ArgoCD, useTLSF
 
 	existing := newDeploymentWithSuffix("server", "server", cr)
 	if argoutil.IsObjectFound(r.Client, cr.Namespace, existing.Name, existing) {
+		if !cr.Spec.Server.IsEnabled() {
+			log.Info("Existing ArgoCD Server found but should be disabled. Deleting ArgoCD Server")
+			// Delete existing deployment for ArgoCD Server, if any ..
+			return nil
```
anandf, line 1322: Did you miss the call to delete the server deployment.

controllers/argocd/deployment.go
```diff
@@ -1337,6 +1366,17 @@ func (r *ReconcileArgoCD) reconcileServerDeployment(cr *argoproj.ArgoCD, useTLSF
 		return nil // Deployment found with nothing to do, move along...
 	}
 
+	// Skipping to start Argo CD Server if Repo Server is disabled. Repo Server internally checks for Redis being available. Thus skipping check on Redis here.
+	if !cr.Spec.Repo.IsEnabled() {
+		log.Error(fmt.Errorf("repo server configuration is disabled and repo server location not configured"), "repo server configuration is disabled and repo server location not configured")
```
anandf, line 1371: Check error message

controllers/argocd/statefulset.go
```diff
@@ -696,6 +706,18 @@ func (r *ReconcileArgoCD) reconcileApplicationControllerStatefulSet(cr *argoproj
 		return nil // StatefulSet found with nothing to do, move along...
 	}
 
+	// Skipping to start application controller if Repo Server is disabled. Repo Server internally checks for Redis being available. Thus skipping check on Redis here.
+	if !cr.Spec.Repo.IsEnabled() {
+		err := fmt.Errorf("repo server configuration is disabled and repo server location not configured")
+		log.Error(err, "repo server configuration is disabled and repo server location not configured")
```
anandf, line 712: Check error message

controllers/argocd/applicationset.go
```diff
@@ -204,6 +201,13 @@ func (r *ReconcileArgoCD) reconcileApplicationSetDeployment(cr *argoproj.ArgoCD,
 		return nil // Deployment found with nothing to do, move along...
 	}
 
+	// Skipping to start applicationset controller if Repo Server is disabled. Repo Server internally checks for Redis being available. Thus skipping check on Redis here.
+	if !cr.Spec.Repo.IsEnabled() {
+		err := fmt.Errorf("repo server configuration is disabled and repo server location not configured")
+		log.Error(err, "repo server configuration is disabled and repo server location not configured")
```
anandf, line 207: check error message

anandf (approved):

jaideepr97:
> @anandf , as discussed offline, I have updated the code to skip installing the component if depending components are disabled.

maybe i'm missing something, but doesn't this defeat the point of this feature? from my understanding if we disable app-controller because redis and repo-server are disabled for the given instance, don't we end up with a all-or-nothing configuration instead of being able to control each component independently?

ishitasequeira:
> > @anandf , as discussed offline, I have updated the code to skip installing the component if depending components are disabled.
> 
> maybe i'm missing something, but doesn't this defeat the point of this feature? from my understanding if we disable app-controller because redis and repo-server are disabled for the given instance, don't we end up with a all-or-nothing configuration instead of being able to control each component independently?
> 
> I also see that in this PR currently, if a component like redis or repo-server is disabled, we are not checking to make sure that alternate arrangements are being made (for eg an FQDN for the disabled component running elsewhere being provided, as is described in the linked issue) Is that also by design?

I had included the option of taking urls for dependent components like redis. But that would need extensive testing and would be better off as an epic of it's own. This PR would be a start for that feature.

jaideepr97:
> I had included the option of taking urls for dependent components like redis. But that would need extensive testing and would be better off as an epic of it's own. This PR would be a start for that feature.

I see, thanks

>> @anandf , as discussed offline, I have updated the code to skip installing the component if depending components  are disabled.
>>
> maybe i'm missing something, but doesn't this defeat the point of this feature? from my understanding if we disable app-controller because redis and repo-server are disabled for the given instance, don't we end up with a all-or-nothing configuration instead of being able to control each component independently?

Could you shed some light on this aspect?

bundle/manifests/argoproj.io_argocds.yaml
```diff
@@ -1494,9 +1494,17 @@ spec:
...
+                  enabled:
+                    description: Enabled is the flag to enable Redis during ArgoCD
+                      installation. (optional, default `true`)
+                    type: boolean
                   image:
                     description: Image is the Redis container image.
                     type: string
+                  remote:
+                    description: Remote specifies the remote URL of the Redis container.
+                      (optionnal, default '')
```
anandf, line 1506: Correct typo `optionnal`. Probably you can mention that by default, a local instance managed by the operator is used.

bundle/manifests/argoproj.io_argocds.yaml
```diff
@@ -3018,6 +3030,10 @@ spec:
                     description: MountSAToken describes whether you would like to
                       have the Repo server mount the service account token
                     type: boolean
+                  remote:
+                    description: Remote specifies the remote URL of the Repo Server
+                      container. (optionnal, default '')
```
anandf, line 3035: Please correct typo

api/v1alpha1/argocd_types.go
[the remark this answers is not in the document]
ishitasequeira: The conversion webhook shows up errors if I do not update `v1alpha1` apis along with `v1beta1`.

controllers/argocd/applicationset.go
```diff
@@ -44,8 +44,11 @@ func getArgoApplicationSetCommand(cr *argoproj.ArgoCD) []string {
 	cmd = append(cmd, "entrypoint.sh")
 	cmd = append(cmd, "argocd-applicationset-controller")
 
-	cmd = append(cmd, "--argocd-repo-server")
-	cmd = append(cmd, getRepoServerAddress(cr))
+	if cr.Spec.Repo.IsEnabled() {
+		cmd = append(cmd, "--argocd-repo-server", getRepoServerAddress(cr))
+	} else {
+		log.Info("Repo Server is disabled. This would affect the functioning of ApplicationSet Controller.")
```
jaideepr97, line 50: should this be a warning instead of Info?

ishitasequeira: I didn't see a `log.Warn` option in the logger being used in the operator, thus ended up using `log.Info`.

api/v1beta1/argocd_types.go
```diff
@@ -162,6 +169,13 @@ type ArgoCDApplicationSet struct {
 
 	// SCMRootCAConfigMap is the name of the config map that stores the Gitlab SCM Provider's TLS certificate which will be mounted on the ApplicationSet Controller (optional).
 	SCMRootCAConfigMap string `json:"scmRootCAConfigMap,omitempty"`
+
+	// Enabled is the flag to enable the Application Set Controller during ArgoCD installation. (optional, default `true`)
+	Enabled *bool `json:"enabled,omitempty"`
```
jaideepr97, line 174: This seems unnecessary for appset, since we can already enable/disable appset by leaving it as nil or not nil
is there some way this field will become useful in the future?

controllers/argocd/deployment.go
```diff
@@ -1171,6 +1199,11 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argoproj.ArgoCD, useTLSFor
 		return nil // Deployment found with nothing to do, move along...
 	}
 
+	if !cr.Spec.Repo.IsEnabled() {
+		log.Info("ArgoCD Repo disabled. Skipping starting ArgoCD Repo.")
```
jaideepr97, line 1203: can we call this repo-server instead of "repo"?

api/v1beta1/argocd_types.go
ishitasequeira: Tried keeping it consistent with the format being used for other core components of ArgoCD. It would be better to update that behavior eventually as ApplicationSet is considered a core component for ArgoCD now.

controllers/argocd/util.go
```diff
@@ -131,7 +136,12 @@ func getArgoApplicationControllerCommand(cr *argoproj.ArgoCD, useTLSForRedis boo
 		}
 	}
 
-	cmd = append(cmd, "--repo-server", getRepoServerAddress(cr))
+	if cr.Spec.Repo.IsEnabled() {
+		cmd = append(cmd, "--repo-server", getRepoServerAddress(cr))
+	} else {
+		log.Info("Repo Server is disabled. This would affect the functioning of Application Controller.")
```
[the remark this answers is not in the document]
ishitasequeira: As mentioned in the previous comment, we do not have a `Warn` option available in the logger used in the operator, thus the Info.

controllers/argocd/deployment.go
ishitasequeira: Will update. Thanks.

jaideepr97 (approved):
LGTM thanks @ishitasequeira

PR #1052, opened 2023-11-02
update control plane

reginapizza:
**What type of PR is this?**
 /kind bug

Forgot to update control plane during last PR


iam-veeramalla (approved):
/lgtm, thanks @reginapizza

PR #1063, opened 2023-11-09
code-refactoring: Add argocd server controller

svghadi:
**What type of PR is this?**
/kind code-refactoring

Adds server controller 


controllers/argocd/server/deployment.go
```diff
@@ -0,0 +1,338 @@
...
+		cmd = append(cmd, "--insecure")
+	}
+
+	cmd = append(cmd, "--staticassets")
+	cmd = append(cmd, "/shared/app")
+
+	cmd = append(cmd, "--dex-server")
+	cmd = append(cmd, dex.GetDexServerAddress(sr.Instance.Name, sr.Instance.Namespace))
+
+	// reposever flags
```
chengfang, line 292: typo in // reposerver

controllers/argocd/server/route_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	routev1 "github.com/openshift/api/route/v1"
+)
+
+func TestServerReconciler_createUpdateAndDeleteRoute(t *testing.T) {
+	ns := argocdcommon.MakeTestNamespace()
+	sr := makeTestServerReconciler(t, ns)
+	routev1.AddToScheme(sr.Scheme)
```
chengfang, line 20: `AddToScheme` is deprecated. Can we use `Install` instead?

controllers/argocd/server/route_test.go
```diff
@@ -0,0 +1,114 @@
...
+	route = &routev1.Route{}
+	err = sr.Client.Get(context.TODO(), types.NamespacedName{Name: "argocd-server", Namespace: "argocd",}, route)
+	assert.Error(t, err)
+	assert.True(t, errors.IsNotFound(err))
+}
+
+func TestServerReconciler_routeTLS(t *testing.T) {
+	ns := argocdcommon.MakeTestNamespace()
+	sr := makeTestServerReconciler(t, ns)
+	routev1.AddToScheme(sr.Scheme)
```
chengfang, line 66: `AddToScheme` is deprecated. Can we use `Install` instead?

controllers/argocd/sso/dex/dex.go
```diff
@@ -0,0 +1 @@
+package dex
```
chengfang, line 1: this file seems emtpy. Should it be removed?

controllers/argocd/sso/dex/constants.go
```diff
@@ -0,0 +1,5 @@
+package dex
+
+const (
+	ArgoCDDexControllerComponent = "dex-server"
+)
```
chengfang, line 5: does it make sense to combine `constants` into `util` so to have fewer file to manage, as the above constant is only used in util.

controllers/argocd/server/constants.go
```diff
@@ -0,0 +1,36 @@
...
+	ServerSuffix = "server"
+)
+
+// ingress
+const (
+	// ArgoCDKeyIngressBackendProtocol is the backend-protocol key for labels.
+	ArgoCDKeyIngressBackendProtocol = "nginx.ingress.kubernetes.io/backend-protocol"
+
+	// ArgoCDKeyIngressClass is the ingress class key for labels.
+	ArgoCDKeyIngressClass = "kubernetes.io/ingress.class"
```
chengfang, line 29: `ArgoCDKeyIngressClass` seems not used.

controllers/argocd/server/constants.go
```diff
@@ -0,0 +1,36 @@
...
+
+	// ArgoCDKeyIngressSSLRedirect is the ssl force-redirect key for labels.
+	ArgoCDKeyIngressSSLRedirect = "nginx.ingress.kubernetes.io/force-ssl-redirect"
+
+	// ArgoCDKeyIngressSSLPassthrough is the ssl passthrough key for labels.
+	ArgoCDKeyIngressSSLPassthrough = "nginx.ingress.kubernetes.io/ssl-passthrough"
```
chengfang, line 35: `ArgoCDKeyIngressSSLPassthrough` seems not used.

common/keys.go
```diff
@@ -205,3 +205,4 @@ const (
 	// ArgoCDArgoprojKeyManagedByClusterArgoCD is needed to identify namespace mentioned as sourceNamespace on ArgoCD
 	ArgoCDArgoprojKeyManagedByClusterArgoCD = "argocd.argoproj.io/managed-by-cluster-argocd"
 )
+
```
chengfang, line 208: no change in this file except EOF. Can we revert this file just to simplify the history?

common/values.go
```diff
@@ -87,9 +89,18 @@ const (
...
 // Commnds
 const (
 	LogLevel = "--loglevel"
 )
+
+// annotations
+const (
+	// AnnotationOpenShiftServiceCA is the annotation on services used to
+	// request a TLS certificate from OpenShift's Service CA for AutoTLS
+	AnnotationOpenShiftServiceCA = "service.beta.openshift.io/serving-cert-secret-name"
```
jaideepr97, line 105: since this is a map key i think we should have this in `keys.go`

controllers/argocd/applicationset/role.go
```diff
@@ -80,6 +80,9 @@ func (asr *ApplicationSetReconciler) reconcileRole() error {
 
 func (asr *ApplicationSetReconciler) deleteRole(name, namespace string) error {
 	if err := permissions.DeleteRole(name, namespace, asr.Client); err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
```
jaideepr97, line 85: is this change necessary here? seems unrelated to server package
also not found errors are already handled within `permissions.DeleteRole`

controllers/argocd/server/constants.go
jaideepr97, line 35: These constants are already defined in `common/keys.go` 
I don't think we need them in the server package

controllers/argocd/applicationset/role.go
svghadi: I removed not found error handling in `permissions.DeleteRole` and in some other functions because it was getting difficult to handle logging in controller code where the delete function is called repeatedly on each reconciliation.  For example in case of server role delete, the first delete call would log role deleted message correctly but subsequent calls to delete were logging the same delete message even though it was just exiting without performing a actual delete of the object. 
I couldn't think of a better way to handle it. Please let me know if there's a different approach we could consider. Also if we are okay with having false positive in logs, we can revert this change.

svghadi: An alternative approach could be to avoid relying on `permissions.DeleteRole` and similar delete functions where repeated calls are made. Instead, we could handle the delete explicitly using `client.Delete`.

common/keys.go
svghadi: Reverted.

common/values.go
svghadi: There is already a `ServiceBetaOpenshiftKeyCertSecret` key in `common/keys.go` with same value. Using it instead of adding a new one.

controllers/argocd/applicationset/role.go
jaideepr97: >the first delete call would log role deleted message correctly but subsequent calls to delete were logging the same delete message even though it was just exiting without performing a actual delete of the object.

makes sense

controllers/argocd/server/role.go
```diff
@@ -0,0 +1,411 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:        roleName,
+				Labels:      roleLabels,
+				Annotations: sr.Instance.Annotations,
+			},
+			Client:    sr.Client,
+			Mutations: []mutation.MutateFunc{mutation.ApplyReconcilerMutation},
+		}
+
+		roleRequest.ObjectMeta.Namespace = nsName
```
jaideepr97, line 65: can we move this to where the rest of the fields are being initialized in the roleRequest just above? seems out of place to set this separately here

controllers/argocd/server/role.go
```diff
@@ -0,0 +1,411 @@
...
+			// role found, no changes detected
+			continue
+		} else {
+			// custom role in use, cleanup default role
+			err := sr.deleteRole(desiredRole.Name, desiredRole.Namespace)
+			if err != nil {
+				reconciliationErrors = append(reconciliationErrors, err)
+			}
+			continue
+		}
```
jaideepr97, line 129: I think I would prefer to invert this if else so that we check if `getCustomRole() != nil` then we delete the existing role and continue. That way the majority of the logic does not have to be at a nested level, which reduces the probability of bugs slightly and makes it more maintainable IMO

WDYT?

controllers/argocd/server/role.go
```diff
@@ -0,0 +1,411 @@
...
+			Client:    sr.Client,
+			Mutations: []mutation.MutateFunc{mutation.ApplyReconcilerMutation},
+		}
+
+		roleRequest.ObjectMeta.Namespace = nsName
+
+		if nsName == sr.Instance.Namespace {
+			roleRequest.Rules = getPolicyRulesForArgoCDNamespace()
+		} else {
+			roleRequest.Rules = getPolicyRulesForManagedNamespace()
```
jaideepr97, line 70: we should also add a custom label to differentiate the role for managed namespace vs source namespace 
like here for reference: https://github.com/jaideepr97/argocd-operator-rewrite/blob/main/controllers/argocd/server/role.go#L75C4-L75C92

controllers/argocd/server/role.go
svghadi: Yes. Makes sense.

controllers/argocd/server/role.go
svghadi: done

controllers/argocd/server/role.go
```diff
@@ -0,0 +1,411 @@
...
+			if !errors.IsNotFound(err){
+				sr.Logger.Error(err, "reconcileSourceRoles: failed to retrieve namespace", "name", nsName)
+				reconciliationErrors = append(reconciliationErrors, err)
+			}
+			continue
+		}
+
+		// do not reconcile roles for namespaces already containing managed-by label
+		// as it already contain roles with permissions to manipulate application resources
+		// reconciled during reconcilation of ManagedNamespaces
```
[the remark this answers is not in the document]
svghadi: Makes sense. I have incorporated the suggestion. PTAL. 

Need a bit clarification on naming convention for role & rolebinding in source namespaces. In our current code base, we are naming the resources in source namespaces as `ArgoCDName_TargetNamespace` and in comment it says 
 as `sourceNamespace_targetNamespace_argocd-server` and in your POC code it is `ArgoCDNamespace_TargetNamespace`. Which one should we prefer?
https://github.com/argoproj-labs/argocd-operator/blob/d424ebd71f4d1e67ade00a8b329e3a6e8688950d/controllers/argocd/rolebinding.go#L321-L324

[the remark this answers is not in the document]
jaideepr97: I think it should ideally be `argocdName-argocdNamespace-server`
no need for  targetnamespace in the name 
@svghadi

controllers/argocd/server/server.go
```diff
@@ -20,6 +24,112 @@ type ServerReconciler struct {
...
+	ns, err := cluster.GetNamespace(sr.Instance.Namespace, sr.Client)
+	if err != nil {
+		sr.Logger.Error(err, "Reconcile: failed to retrieve namespace", "name", sr.Instance.Namespace)
+		return err
+	}
+
+	// namespace is deleting, delete the resources
+	if ns.DeletionTimestamp != nil {
+		return sr.DeleteResources()
+	}
```
jaideepr97, line 39: I think this is no longer necessary

We will add some logic to the namespace watcher's event handler to watch for terminating namespaces and trigger deletion of all controllers' resources from there instead of running this check each time in every controller unnecessarily

controllers/argocd/server/clusterrole.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/argoproj-labs/argocd-operator/pkg/permissions"
+	v1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (sr *ServerReconciler) reconcileClusterRole() error {
+
+	crName := getClusterRoleName(sr.Instance.Name, sr.Instance.Namespace)
+	crLables := common.DefaultResourceLabels(crName, sr.Instance.Name, ServerControllerComponent)
```
jaideepr97, line 17: We should generate all the resource names we need in `server.go` before the reconciliation begins, and just use those values wherever applicable, so that we only define them once and it stays consistent across all resources

labels will be generated by `argoutil.GetObjMeta` so we don't need to worry about those here

controllers/argocd/server/clusterrole.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	sr.Logger.Info("reconciling clusterRole")
+
+	crRequest := permissions.ClusterRoleRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        crName,
+			Labels:      crLables,
+			Annotations: sr.Instance.Annotations,
+		},
```
jaideepr97, line 31: call `argoutil.GetObjMeta` 
I think if we pass an empty string in place of namespace it should make it cluster scoped

controllers/argocd/server/clusterrole.go
```diff
@@ -0,0 +1,127 @@
...
+		Rules:     getPolicyRuleForClusterRule(),
+		Client:    sr.Client,
+		Mutations: []mutation.MutateFunc{mutation.ApplyReconcilerMutation},
+	}
+
+	desiredCR, err := permissions.RequestClusterRole(crRequest)
+	if err != nil {
+		sr.Logger.Error(err, "reconcileClusterRole: failed to request cluster role", "name", desiredCR.Name)
+		sr.Logger.V(1).Info("reconcileClusterRole: one or more mutations could not be applied")
+		return err
```
jaideepr97, line 41: Please update to only returning annotated errors here instead of logging them, as described in the coding practices guide https://github.com/argoproj-labs/argocd-operator/blob/operator-redesign/coding-standards-and-best-practices.md#error-handling
Best to follow that everywhere we are currently logging errors (except at the end of the call stack) 
```suggestion
		sr.Logger.V(1).Info("reconcileClusterRole: one or more mutations could not be applied")
		return errors.Wrapf(err, "reconcileClusterRole: failed to request cluster role %s", desiredCR.Name)
```

controllers/argocd/server/clusterrole.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// difference in existing & desired role, reset it
+	if !reflect.DeepEqual(existingCR.Rules, desiredCR.Rules) {
+		existingCR.Rules = desiredCR.Rules
+		if err = permissions.UpdateClusterRole(existingCR, sr.Client); err != nil {
+			sr.Logger.Error(err, "reconcileClusterRole: failed to update cluster role", "name", existingCR.Name)
+			return err
+		}
+		sr.Logger.V(0).Info("reconcileClusterRole: role updated", "name", existingCR.Name)
+	}
```
jaideepr97, line 68: please update to follow this template for all resources 
(see redis PR for reference)

```
roleChanged := false

	fieldsToCompare := []struct {
		existing, desired interface{}
		extraAction       func()
	}{
		{&existingRole.Rules, &desiredRole.Rules, nil},
	}

	for _, field := range fieldsToCompare {
		argocdcommon.UpdateIfChanged(field.existing, field.desired, field.extraAction, &roleChanged)
	}

	if !roleChanged {
		return nil
	}

	if err = permissions.UpdateRole(existingRole, rr.Client); err != nil {
		return errors.Wrapf(err, "reconcileRole: failed to update role %s", existingRole.Name)
	}

	rr.Logger.V(0).Info("role updated", "name", existingRole.Name, "namespace", existingRole.Namespace)
	return nil
```

jaideepr97:
@svghadi added some generic comments that should be applied across all resources managed by the server 
Please rebase after all these PRs are merged and I will do another round of reivew 

https://github.com/argoproj-labs/argocd-operator/pull/1161
https://github.com/argoproj-labs/argocd-operator/pull/1167
https://github.com/argoproj-labs/argocd-operator/pull/1168
https://github.com/argoproj-labs/argocd-operator/pull/1170
https://github.com/argoproj-labs/argocd-operator/pull/1171


Additionally, please make sure to move all the server related  functions that you have already implemented from the old code into the correct `TOBEREMOVED.go` file so we can be sure that we didn't miss refactoring any existing logic

jaideepr97 (approved):
thanks @svghadi !

PR #1084, opened 2023-11-27
test: Add test for notification service with self-signed certs

svghadi:
**What type of PR is this?**
/kind test

This PR adds a e2e test for https://github.com/argoproj/argo-cd/issues/15392. The bug has been fixed in upstream argocd with https://github.com/argoproj/argo-cd/pull/15394. 

The source for `webhook-server` image used in test is added into `tests/auxiliary/webhook-server`. The self-signed certs are generated on linux and are valid for `webhook` endpoint till `2103` year.

Run automated kuttl test

```bash
kubectl kuttl test tests/k8s --config tests/kuttl-tests.yaml --test 1-034_validate_webhook_notifications 
```


tests/k8s/1-034_validate_webhook_notifications/00-setup-webhook-server.yaml
```diff
@@ -0,0 +1,168 @@
...
+    matchLabels:
+      app: webhook
+  template:
+    metadata:
+      labels:
+        app: webhook
+    spec:
+      containers:
+        - name: server
+          image: quay.io/svghadi/webhook-server:latest
```
svghadi, line 31: Should we move this image to `quay.io/argoprojlabs/` similar to `quay.io/argoprojlabs/argocd-notifications-e2e-smtplistener:multiarch` which is used in `1-022_validate_notifications`?

PR #1089, opened 2023-11-29
fix the logic for applicationset resources reconcilation when spec.applicationset.enabled is false

ishitasequeira:
**What type of PR is this?**
/kind bug

Currently when the ApplicationSet resources do not exist and we set `spec.applicationset.enabled = false`, the reconciliation fails giving us below error:
```
2023-11-29T07:15:04Z	ERROR	Reconciler error	{"controller": "argocd", "controllerGroup": "argoproj.io", "controllerKind": "ArgoCD", "ArgoCD": {"name":"argocd","namespace":"test"}, "namespace": "test", "name": "argocd", "reconcileID": "c83d8dba-9944-47c3-b8d1-9f75a60346e9", "error": "roles.rbac.authorization.k8s.io \"argocd-applicationset-controller\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/remote-source/deps/gomod/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:329
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/remote-source/deps/gomod/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/remote-source/deps/gomod/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:227
```

Fixes https://issues.redhat.com/browse/GITOPS-3667 

* Start the operator using `make install run`
* apply the below argocd configmap
```
kind: ArgoCD
metadata:
  name: argocd
spec:
  controller:
    enabled: false
  redis:
    enabled: false
  repo:
    enabled: false
  server:
    enabled: false
  applicationSet:
    enabled: false
```
* No resources should be created for the argocd instance.


controllers/argocd/deployment.go
```diff
@@ -1375,7 +1375,7 @@ func (r *ReconcileArgoCD) reconcileServerDeployment(cr *argoproj.ArgoCD, useTLSF
 		return nil // Deployment found with nothing to do, move along...
 	}
 
-	if !cr.Spec.Controller.IsEnabled() {
+	if !cr.Spec.Server.IsEnabled() {
 		log.Info("ArgoCD Repo Server disabled. Skipping starting repo server.")
```
anandf, line 1379: Log message mentions as `Repo Server`. It should be `Server` right ?

tests/k8s/1-034_validate_applicationset_reconcile_enabled_set_false/01-errors.yaml
```diff
@@ -0,0 +1,63 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ArgoCD
+metadata:
+  name: argocd-test
+  namespace: test
+status:
+  phase: Available
```
anandf, line 7: The ArgoCD should exist and should be in `Available` phase when all the components are disabled.

controllers/argocd/status.go
```diff
@@ -207,7 +207,10 @@ func (r *ReconcileArgoCD) reconcileStatusSSO(cr *argoproj.ArgoCD) error {
 func (r *ReconcileArgoCD) reconcileStatusPhase(cr *argoproj.ArgoCD) error {
 	var phase string
 
-	if cr.Status.ApplicationController == "Running" && cr.Status.Redis == "Running" && cr.Status.Repo == "Running" && cr.Status.Server == "Running" {
+	if ((!cr.Spec.Controller.IsEnabled() && cr.Status.ApplicationController == "Unknown") || cr.Status.ApplicationController == "Running") &&
```
anandf, line 210: Instead of having a status of `Unknown` can we have a state called `Disabled`. It would be more intutive for the user to understand that the component is explicitly disabled.

anandf (changes requested):
- Could we set the component status to `Disabled` instead of `Unknown`.
- Could we add one more E2E scenario where we start from all components enabled, and disable each of them and another scenario for validating the HA scenario. 
- 1-034_validate_applicationset_reconcile_enabled_set_false newly added test is failing. It is probably due to the ArgoCD instance added in the errors file.

controllers/argocd/status.go
ishitasequeira: We do not maintain a `Disabled` state today and add `Unknown` as the state for missing components like `ApplicationSet` if `spec.applicationset = nil`. It would be like adding a new behavior and a new Status as part of bugfix.

tests/kuttl-tests.yaml
```diff
@@ -6,6 +6,6 @@ testDirs:
 suppress:
   - events
 parallel: 1
-timeout: 300
+timeout: 60
```
anandf, line 9: Do we need this change ?

ishitasequeira: reverted

anandf (approved):

controllers/argocd/status.go
anandf: If not as part of this PR, we can create a separate PR for adding the `Disabled` state. That being a string and not part of the API spec, it can be introduced with minimal impact to end users.

PR #1129, opened 2024-01-02
Fix webhook route fail in argocd due to long route name

Rizwana777:
Fix the failing of webhook route in argocd due to long route name

JIRA link:
https://issues.redhat.com/browse/GITOPSRVCE-794

### DIFF
--- a/controllers/argocd/route.go
+++ b/controllers/argocd/route.go
@@ -418,11 +418,16 @@ func (r *ReconcileArgoCD) reconcileApplicationSetControllerWebhookRoute(cr *argo
 // - If the FIRST label ("console-openshift-console" in the above case) is longer than 63 characters, shorten (truncate the end) it to 63.
 // - If any other label is longer than 63 characters, return an error
 // - After all the labels are 63 characters or less, check the length of the overall hostname:
-//   - If the overall hostname is > 255, then shorten the FIRST label until the host name is < 255
+//   - If the overall hostname is > 253, then shorten the FIRST label until the host name is < 253
 //   - After the FIRST label has been shortened, if it is < 20, then return an error (this is a sanity test to ensure the label is likely to be unique)
 func shortenHostname(hostname string) (string, error) {
 	if hostname == "" {
 		return "", nil
+	}
+
+	// Return the hostname as it is if hostname is already within the size limit
+	if len(hostname) <= maxHostnameLength {
+		return hostname, nil
 	}
 
 	// Split the hostname into labels
--- a/controllers/argocd/route.go
+++ b/controllers/argocd/route.go
@@ -29,6 +29,12 @@ import (
 	"github.com/argoproj-labs/argocd-operator/controllers/argoutil"
 )
 
+const (
+	maxLabelLength    = 63
+	maxHostnameLength = 253
+	minFirstLabelSize = 20
+)
+
 var routeAPIFound = false
 
 // IsRouteAPIAvailable returns true if the Route API is present.
@@ -340,10 +346,19 @@ func (r *ReconcileArgoCD) reconcileApplicationSetControllerWebhookRoute(cr *argo
 
 	route.Spec.Host = hostname
 
-	if cr.Spec.Server.Insecure {
-		// Disable TLS and rely on the cluster certificate.
-		route.Spec.Port = &routev1.RoutePort{
-			TargetPort: intstr.FromString("webhook"),
+	route.Spec.Port = &routev1.RoutePort{
+		TargetPort: intstr.FromString("webhook"),
+	}
+
+	// Allow override of TLS options for the Route
+	if cr.Spec.ApplicationSet.WebhookServer.Route.TLS != nil {
+		tls := &routev1.TLSConfig{}
+
+		// Set Termination
+		if cr.Spec.ApplicationSet.WebhookServer.Route.TLS.Termination != "" {
+			tls.Termination = cr.Spec.ApplicationSet.WebhookServer.Route.TLS.Termination
+		} else {
+			tls.Termination = routev1.TLSTerminationEdge
 		}
 
 		// Set Certificate
@@ -399,18 +414,28 @@ func (r *ReconcileArgoCD) reconcileApplicationSetControllerWebhookRoute(cr *argo
 	return r.Client.Update(context.TODO(), route)
 }
 
+// The algorithm used by this function is:
+// - If the FIRST label ("console-openshift-console" in the above case) is longer than 63 characters, shorten (truncate the end) it to 63.
+// - If any other label is longer than 63 characters, return an error
+// - After all the labels are 63 characters or less, check the length of the overall hostname:
+//   - If the overall hostname is > 255, then shorten the FIRST label until the host name is < 255
+//   - After the FIRST label has been shortened, if it is < 20, then return an error (this is a sanity test to ensure the label is likely to be unique)
 func shortenHostname(hostname string) (string, error) {
+	if hostname == "" {
+		return "", nil
+	}
+
 	// Split the hostname into labels
 	labels := strings.Split(hostname, ".")
 
 	// Check and truncate the FIRST label if longer than 63 characters
-	if len(labels[0]) > 63 {
-		labels[0] = labels[0][:63]
+	if len(labels[0]) > maxLabelLength {
+		labels[0] = labels[0][:maxLabelLength]
 	}
 
 	// Check other labels and return an error if any is longer than 63 characters
 	for _, label := range labels[1:] {
-		if len(label) > 63 {
+		if len(label) > maxLabelLength {
 			return "", fmt.Errorf("label length exceeds 63 characters")
 		}
 	}
@@ -418,15 +443,18 @@ func shortenHostname(hostname string) (string, error) {
 	// Join the labels back into a hostname
 	resultHostname := strings.Join(labels, ".")
 
-	// Check and shorten the overall hostname if > 253 characters
-	if len(resultHostname) > 255 {
-		resultHostname = resultHostname[:253]
-	}
+	// Check and shorten the overall hostname
+	if len(resultHostname) > maxHostnameLength {
+		// Shorten the first label until the length is less than 253
+		for len(resultHostname) > maxHostnameLength && len(labels[0]) > 20 {
+			labels[0] = labels[0][:len(labels[0])-1]
+			resultHostname = strings.Join(labels, ".")
+		}
 
-	// Check if the FIRST label is < 20 and return an error if true
-	if len(labels[0]) < 20 {
-		return "", fmt.Errorf("first label length is less than 20 characters")
+		// Check if the first label is still less than 20 characters
+		if len(labels[0]) < minFirstLabelSize {
+			return "", fmt.Errorf("shortened first label is less than 20 characters")
+		}
 	}
-
 	return resultHostname, nil
 }
--- a/controllers/argocd/route.go
+++ b/controllers/argocd/route.go
@@ -17,6 +17,7 @@ package argocd
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	routev1 "github.com/openshift/api/route/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -248,6 +249,13 @@ func (r *ReconcileArgoCD) reconcileServerRoute(cr *argoproj.ArgoCD) error {
 		route.Spec.Host = cr.Spec.Server.Host // TODO: What additional role needed for this?
 	}
 
+	hostname, err := shortenHostname(route.Spec.Host)
+	if err != nil {
+		return err
+	}
+
+	route.Spec.Host = hostname
+
 	if cr.Spec.Server.Insecure {
 		// Disable TLS and rely on the cluster certificate.
 		route.Spec.Port = &routev1.RoutePort{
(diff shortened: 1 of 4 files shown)


controllers/argocd/route.go
```diff
@@ -302,6 +302,10 @@ func (r *ReconcileArgoCD) reconcileApplicationSetControllerWebhookRoute(cr *argo
 		}
 	}
 
+	if len(route.Spec.Host) > 63 {
+		route.Spec.Host = route.Spec.Host[:63]
+	}
+
```
jgwest, line 308: Based on what the code is doing earlier in the function, it seems like `route.Spec.Host` will always be empty at this point (correct me if I am wrong).

jgwest: Also it seems like line 328 will override this:
```
	// Allow override of the Host for the Route.
	if len(cr.Spec.Server.Host) > 0 {
		route.Spec.Host = cr.Spec.ApplicationSet.WebhookServer.Host
	}
```

jgwest (changes requested):

controllers/argocd/route.go
Rizwana777: No, When I added print statement here to print `route.Spec.Host`, I am able to see the host name there(or I may be missing something, please correct me), and as it  is overriding in line number 328, I will add the logic after 328 and adding this logic `reconcileServerRoute` function as well as it is failing with same error there as well

jgwest (changes requested):
So the rule appears to be that a hostname cannot be longer than 255 characters, and an individual label (constitute part of a hostname) cannot be longer than 63 characters.
([source](https://devblogs.microsoft.com/oldnewthing/20120412-00/?p=7873))

So I think our shortening algorithm should be:
- If the FIRST label ("console-openshift-console" in the above case)  is longer than 63 characters, shorten (truncate the end) it to 63.
- If any other label is longer than 63 characters, return an error
- After all the labels are 63 characters or less, check the length of the overall hostname:
	- If the overall hostname is > 255, then shorten the FIRST label until the host name is < 255
	- After the FIRST label has been shortened, if it is < 20, then return an error (this is a sanity test to ensure the label is likely to be unique)

For example:

For example, this:
`myhostnameaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.redhat.com`
would become:
`myhostnameaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.redhat.com`

For example, this:
`myhostnametwentysixletteraaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.redhat.com`
would become: `myhostnametwentysixletteraaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.redhat.com`

WDYT?

controllers/argocd/route.go
```diff
@@ -366,3 +381,35 @@ func (r *ReconcileArgoCD) reconcileApplicationSetControllerWebhookRoute(cr *argo
...
+func shortenHostname(hostname string) (string, error) {
+	// Split the hostname into labels
+	labels := strings.Split(hostname, ".")
+
+	// Check and truncate the FIRST label if longer than 63 characters
+	if len(labels[0]) > 63 {
+		labels[0] = labels[0][:63]
+
+		// Check if the FIRST label is < 20 after truncating and return an error if true
+		if len(labels[0]) < 20 {
```
jgwest, line 394: This is logically impossible, here?

controllers/argocd/route.go
```diff
@@ -366,3 +381,35 @@ func (r *ReconcileArgoCD) reconcileApplicationSetControllerWebhookRoute(cr *argo
 	}
 	return r.Client.Update(context.TODO(), route)
 }
+
+func shortenHostname(hostname string) (string, error) {
```
jgwest, line 385: This appears not to be implementing the full algorithm prescribed, specifically bullet point 3:
After all the labels are 63 characters or less, check the length of the overall hostname:
- If the overall hostname is > 255, then shorten the FIRST label until the host name is < 255
- After the FIRST label has been shortened, if it is < 20, then return an error (this is a sanity test to ensure the label is likely to be unique)

jgwest, line 385: Should make sure this function handles the case where hostname is empty, and add a test for this case.

controllers/argocd/route_test.go
```diff
@@ -213,6 +213,57 @@ func TestReconcileRouteUnsetsInsecure(t *testing.T) {
...
+
+	_, err := r.Reconcile(context.TODO(), req)
+	assert.NoError(t, err)
+
+	loaded := &routev1.Route{}
+	err = r.Client.Get(ctx, types.NamespacedName{Name: testArgoCDName + "-server", Namespace: testNamespace}, loaded)
+	fatalIfError(t, err, "failed to load route %q: %s", testArgoCDName+"-server", err)
+
+	wantRoute := &routev1.RouteSpec{
+		Host: "myhostnameaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
```
jgwest, line 258: Your algorithm is stripping the .com from the end, which should never  happen.

jgwest (changes requested):
Looks like the algorithm as described still needs to be implemented, and the E2E tests are failing.

controllers/argocd/route.go
Rizwana777: I have added this in the function https://github.com/argoproj-labs/argocd-operator/pull/1129/files#diff-34c9d2f4dd86a5067f505bc888a3169f2ff71cca7c59bf83e9f083c91d57b1d6R386, please let me know whether I am checking empty hostname in right way and please correct me for unit test as well for this

controllers/argocd/route.go
Rizwana777: yes, misunderstood this , fixed it now

controllers/argocd/route_test.go
```diff
@@ -213,6 +216,69 @@ func TestReconcileRouteUnsetsInsecure(t *testing.T) {
 	}
 }
 
+func TestReconcileRouteForShorteningHostname(t *testing.T) {
```
jgwest, line 219: It would be good to include the other example I mentioned in the test, as well:
`myhostnametwentysixletteraaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.redhat.com`

Rather than cloning the existing test case, you should be able to [parameterize it, for example](https://github.com/argoproj-labs/argocd-operator/blob/e4d4791fd16468a82824cb734cd44fe18f4d790c/api/v1beta1/argocd_types_test.go#L29C4-L29C5).

We should also have a check to make sure that the first label is not squeezed to be <20 characters.

jgwest (changes requested):

controllers/argocd/route.go
```diff
@@ -366,3 +381,42 @@ func (r *ReconcileArgoCD) reconcileApplicationSetControllerWebhookRoute(cr *argo
...
+	if hostname == "" {
+		return "", fmt.Errorf("hostname is empty")
+	}
```
jgwest, line 388: In this case, we should just `return "", nil`, e.g. correct me if I'm wrong, but the original code (which is calling this function) does not return an error if the hostname is empty.

Rizwana777: here at this [line](https://github.com/argoproj-labs/argocd-operator/blob/41356fb260e9158ae5a628d267eeab97e3ac6939/controllers/argocd/route.go#L229) we are  calling shortenHostname function when `cr.Spec.Server.Route.Enabled` is true , it means route is enable and having a hostname and hostname should not empty and we are overriding the hostname [here](https://github.com/argoproj-labs/argocd-operator/blob/41356fb260e9158ae5a628d267eeab97e3ac6939/controllers/argocd/route.go#L249) so I have returned the error instead of nil? I may be wrong in understanding this please do correct me

controllers/argocd/route_test.go
Rizwana777: Added the test

controllers/argocd/route.go
jgwest: For this case, I think we should preserve the behaviour of the original code: AFAICT, in the original code, if the host name was not specified, then the code would not return an error and would still create a Route with an empty host. Can you confirm? 

You could ask on #gitops-dev if this is desirable behaviour or not, but I'm not familiar enough with the expected behaviour in this case, so my gut is to not modify that property of the existing code.

Rizwana777: Yes @jgwest I tried this without hostname and it is not returning an error, I will keep the behaviour of existing code here, thank you

controllers/argocd/route.go
```diff
@@ -366,3 +381,42 @@ func (r *ReconcileArgoCD) reconcileApplicationSetControllerWebhookRoute(cr *argo
 	}
 	return r.Client.Update(context.TODO(), route)
 }
+
```
jgwest, line 384: One final, simple request, can you add this as a comment to the top of this function:
```
The algorithm used by this function is:
- If the FIRST label ("console-openshift-console" in the above case) is longer than 63 characters, shorten (truncate the end) it to 63.
- If any other label is longer than 63 characters, return an error
- After all the labels are 63 characters or less, check the length of the overall hostname:
    - If the overall hostname is > 255, then shorten the FIRST label until the host name is < 255
    - After the FIRST label has been shortened, if it is < 20, then return an error (this is a sanity test to ensure the label is likely to be unique)
```

This will help anyone else working on this function against quickly understand what it is supposed to do.

jgwest (changes requested):
One more tiny request then good to go.

controllers/argocd/route.go
Rizwana777: Added the comment @jgwest

controllers/argocd/route.go
```diff
@@ -366,3 +381,48 @@ func (r *ReconcileArgoCD) reconcileApplicationSetControllerWebhookRoute(cr *argo
...
+	labels := strings.Split(hostname, ".")
+
+	// Check and truncate the FIRST label if longer than 63 characters
+	if len(labels[0]) > 63 {
+		labels[0] = labels[0][:63]
+	}
+
+	// Check other labels and return an error if any is longer than 63 characters
+	for _, label := range labels[1:] {
+		if len(label) > 63 {
```
anandf, line 406: @Rizwana777 Can you add constants for the magic numbers `63` 253` and `20`

Rizwana777: Done @anandf PTAL

controllers/argocd/route.go
```diff
@@ -392,3 +413,48 @@ func (r *ReconcileArgoCD) reconcileApplicationSetControllerWebhookRoute(cr *argo
...
+// - If the FIRST label ("console-openshift-console" in the above case) is longer than 63 characters, shorten (truncate the end) it to 63.
+// - If any other label is longer than 63 characters, return an error
+// - After all the labels are 63 characters or less, check the length of the overall hostname:
+//   - If the overall hostname is > 255, then shorten the FIRST label until the host name is < 255
+//   - After the FIRST label has been shortened, if it is < 20, then return an error (this is a sanity test to ensure the label is likely to be unique)
+func shortenHostname(hostname string) (string, error) {
+	if hostname == "" {
+		return "", nil
+	}
+
```
jaideepr97, line 427: @Rizwana777 @jgwest should we add a check here to see if shortening is even required before getting into it? 
Just a check to return the hostname as is if it is already within the size limit

Rizwana777: yes agreed @jaideepr97 added a check https://github.com/argoproj-labs/argocd-operator/pull/1129/files#diff-34c9d2f4dd86a5067f505bc888a3169f2ff71cca7c59bf83e9f083c91d57b1d6R432 , PTAL

jaideepr97: @Rizwana777 can this not be a simple 

```
if len(hostname) < maxHostnameLength {
  return hostname, nil
}
```

jaideepr97: Thanks @Rizwana777 one last little thing for improved readability, I would move this line above the previous line, so we do this check and return before splitting the hostname into labels

Rizwana777: @jaideepr97  Done, PTAL

jaideepr97 (approved):
LGTM, thanks @Rizwana777

PR #1132, opened 2024-01-05
Add documentation for enabling/disabling argocd core components

ishitasequeira:
**What type of PR is this?**
/kind documentation

Documentation for the feature of enabling/disabling the individual ArgoCD core components.

### DIFF
--- a/docs/usage/enabling-disabling-argocd-core-components.md
+++ b/docs/usage/enabling-disabling-argocd-core-components.md
@@ -1,0 +1,48 @@
+# Enabling/Disabling Core Components of ArgoCD
+The Operator oversees all Argo CD workloads, including the API server, repository server, application controller, and more.
+
+Currently, the following workloads are managed:
+
+* argocd-server (API server and UI)
+* argocd-repo-server (Repository server)
+* argocd-application-controller (Main reconciliation controller)
+* argocd-applicationset-controller (ApplicationSet reconciliation controller)
+* argocd-redis (volatile cache)
+
+To support installations with minimal resource requirements and to facilitate distribution across clusters or namespaces, the ability to selectively install specific Argo CD components has been introduced.
+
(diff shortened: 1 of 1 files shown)


docs/usage/enabling-disabling-argocd-core-components.md
```diff
@@ -0,0 +1,48 @@
...
+
+* argocd-server (API server and UI)
+* argocd-repo-server (Repository server)
+* argocd-application-controller (Main reconciliation controller)
+* argocd-applicationset-controller (ApplicationSet reconciliation controller)
+* argocd-redis (volatile cache)
+
+To support installations with minimal resource requirements and to facilitate distribution across clusters or namespaces, the ability to selectively install specific Argo CD components has been introduced.
+
+To enable/disable a particular Argo CD workload, a new flag, `spec.<component>.enabled`, has been implemented. The default value of the flag is `true`, implying that if the flag is unspecified, the Argo CD workload is enabled by default.
```
anandf, line 14: Instead of  `Argo CD workload` we can have it as `Argo CD component`

docs/usage/enabling-disabling-argocd-core-components.md
```diff
@@ -0,0 +1,48 @@
...
+  name: example
+spec:
+  controller:
+    enabled: false
+```
+
+In this example, only the controller component is disabled, while all other components continue to run normally.
+
+# Specifying External URLs for Redis and RepoServer Components
+When disabling core components like Redis or Repo Server, you may wish to provide an external URL for components running in external clusters. The remote URL can be set using the `spec.<component>.remote` flag (where the component can only be `redis` or `repo`).
```
anandf, line 33: Consider rewording `where the component can only be `  to `component can either be `

anandf (changes requested):
Left few minor comments. Otherwise looks good to me.

anandf (approved):

PR #1146, opened 2024-01-09
fix logic to allow processor operation to be less than default value

anandrkskd:
**What type of PR is this?**
/kind bug

Currently, when we set application-controller 'status/operation processor' values below 10, it set it to default value of 10.
This PR allows the value to be less than the default value, but is the value is less or equals to 0 it sets it back to the default value.

Fixes [GITOPS-3770](https://issues.redhat.com/browse/GITOPS-3770)

### DIFF
--- a/controllers/argocd/util.go
+++ b/controllers/argocd/util.go
@@ -309,10 +309,11 @@ func (r *ReconcileArgoCD) getArgoServerURI(cr *argoproj.ArgoCD) string {
 // getArgoServerOperationProcessors will return the numeric Operation Processors value for the ArgoCD Server.
 func getArgoServerOperationProcessors(cr *argoproj.ArgoCD) int32 {
 	op := common.ArgoCDDefaultServerOperationProcessors
-	if cr.Spec.Controller.Processors.Operation > op {
-		op = cr.Spec.Controller.Processors.Operation
+	if cr.Spec.Controller.Processors.Operation > 0 {
+		return cr.Spec.Controller.Processors.Operation
+	} else {
+		return op
 	}
-	return op
 }
 
 // getArgoServerStatusProcessors will return the numeric Status Processors value for the ArgoCD Server.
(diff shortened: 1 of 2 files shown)


controllers/argocd/util.go
```diff
@@ -309,10 +309,11 @@ func (r *ReconcileArgoCD) getArgoServerURI(cr *argoproj.ArgoCD) string {
 // getArgoServerOperationProcessors will return the numeric Operation Processors value for the ArgoCD Server.
 func getArgoServerOperationProcessors(cr *argoproj.ArgoCD) int32 {
 	op := common.ArgoCDDefaultServerOperationProcessors
-	if cr.Spec.Controller.Processors.Operation > op {
-		op = cr.Spec.Controller.Processors.Operation
+	if cr.Spec.Controller.Processors.Operation > 0 {
+		return cr.Spec.Controller.Processors.Operation
+	} else {
+		return op
 	}
```
jaideepr97, line 316: This change can be simplified to a single line

```suggestion
	if cr.Spec.Controller.Processors.Operation > 0 {
		op = cr.Spec.Controller.Processors.Operation
	} 
	return op 
```

controllers/argocd/util_test.go
```diff
@@ -389,6 +389,27 @@ func TestGetArgoApplicationControllerCommand(t *testing.T) {
...
+				"--status-processors",
+				"20",
+				"--kubectl-parallelism-limit",
+				"10",
+				"--loglevel",
+				"info",
+				"--logformat",
+				"text",
+			},
+		},
```
jaideepr97, line 412: could we also add a test here that checks for `0 < operation processors < 10` ? 
To ensure that the fix addresses the original use case

jaideepr97 (approved):
LGTM, thanks @anandrkskd! 
Congrats on your first contribution :)

PR #1162, opened 2024-01-15
fix: operator shouldn't overwrite dex config 

chetan-rns:
**What type of PR is this?**

/kind bug

The operator overrides dex connector configuration in argocd-cm upon setting .spec.sso.dex.config in ArgoCD CR. This PR includes the user provided dex config along with the dex connectors. 

Fixes #?

1. Create an Argo CD CR with the dex provider and OpenShiftAuth enabled. Or use the default Argo CD instance in the openshift-gitops namespace.
2. Update the dex config in the CR.
```yaml
    provider: dex
    dex:
      config: |
        expiry:
          idTokens: "1h"
          signingKeys: "12h"
```
3. Ensure that both the connector and the config are present in the argocd-cm.

### DIFF
--- a/controllers/argocd/configmap_test.go
+++ b/controllers/argocd/configmap_test.go
@@ -359,6 +359,28 @@ func TestReconcileArgoCD_reconcileArgoConfigMap_withDexConnector(t *testing.T) {
 		},
 	}
 
+	getDexConfig := func(t *testing.T, k8sClient client.Client) map[string]interface{} {
+		t.Helper()
+
+		cm := &corev1.ConfigMap{}
+		err := k8sClient.Get(context.TODO(), types.NamespacedName{
+			Name:      common.ArgoCDConfigMapName,
+			Namespace: testNamespace,
+		}, cm)
+		assert.NoError(t, err)
+
+		dex, ok := cm.Data["dex.config"]
+		if !ok {
+			t.Fatal("reconcileArgoConfigMap with dex failed")
+		}
+
+		m := make(map[string]interface{})
+		err = yaml.Unmarshal([]byte(dex), &m)
+		assert.NoError(t, err, fmt.Sprintf("failed to unmarshal %s", dex))
+
+		return m
+	}
+
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			sa := &corev1.ServiceAccount{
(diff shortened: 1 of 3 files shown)


controllers/argocd/dex.go
```diff
@@ -160,10 +159,32 @@ func (r *ReconcileArgoCD) getOpenShiftDexConfig(cr *argoproj.ArgoCD) (string, er
 	dex := make(map[string]interface{})
 	dex["connectors"] = connectors
 
+	if err := updateDexConfig(cr, dex); err != nil {
```
jgwest, line 162: Couldn't hurt to add a quick comment here describing what `updateDexConfig` is doing (or update the function name to better indicate its purpose, if that makes sense)

controllers/argocd/configmap_test.go
```diff
@@ -393,29 +393,65 @@ func TestReconcileArgoCD_reconcileArgoConfigMap_withDexConnector(t *testing.T) {
...
+				config := dexConnector["config"]
+				assert.Equal(t, config.(map[interface{}]interface{})["clientID"], "system:serviceaccount:argocd:argocd-argocd-dex-server")
+			}
+			verifyDexConnector(m)
 
-			connectors, ok := m["connectors"]
-			if !ok {
-				t.Fatal("no connectors found in dex.config")
+			// update .dex.config and verify that the dex connector is not overwritten
+			type expiry struct {
```
jgwest, line 430: :+1: , I learn something new about Go every day :grin:

controllers/argocd/configmap_test.go
```diff
@@ -393,29 +393,65 @@ func TestReconcileArgoCD_reconcileArgoConfigMap_withDexConnector(t *testing.T) {
...
+			a.Spec.SSO.Dex.Config = string(dexCfgBytes)
+
+			err = r.reconcileArgoConfigMap(a)
+			assert.NoError(t, err)
+
+			m = getDexConfig(t, r.Client)
+			_, ok := m["expiry"]
+			assert.True(t, ok)
+
+			verifyDexConnector(m)
```
jgwest, line 454: For your consideration, just a suggestion as you are more familiar with the test behaviour from working on it here:
- Would this part of the test work/make sense as it's own entry in the test struct (line ~427), rather than as an addition to the end of the test?
- e.g. in here:
```go
tests := []struct {
		name             string
		updateCrSpecFunc func(cr *argoproj.ArgoCD)
	}{
		{
```

jgwest (changes requested):
A few quick thoughts from working through the PR, then good to go.

controllers/argocd/dex.go
chetan-rns: updated the name .

controllers/argocd/configmap_test.go
chetan-rns: added as a separate test entry

jgwest (approved):
LGTM, thanks @chetan-rns!

iam-veeramalla (approved):

PR #1192, opened 2024-01-25
Deprecate grafana field and add log info warning

Rizwana777:
**What type of PR is this?**

> /kind bug

JIRA Link - https://issues.redhat.com/browse/GITOPS-2049
- removed duplicate imports from some files


jgwest (changes requested):
- See proposed tweak to the deprecation warning, in all the places where it is used.
    - Actually, come to think of it, it couldn't hurt to put this message in a `string` constant.
- `grafana.go` in `controllers/argocd` can also be deleted, AFAICT.
- You can also search the code for 'grafana', as I think there are some unused constants that can be removed here, too.

jgwest:
@Rizwana777: to fix the failing PR checks:
- Run `make bundle` and check in the files under `bundle/manifests/` (these are the files that are listed in the failure message under the failing GitHub action)
- DCO check requires all commits to be signed. This means that when you `git commit` your changes, you need to use `git commit -s`. This will automatically sign the commit. (This check is enforced across all of argocd-operator)
    - You may need to squash your exist commits into a single commit, first, and then re-commit them as signed.

jgwest (changes requested):
We should probably remove the remaining Grafana references/artifacts. I was able to find these, for example, by searching for 'grafana':
- Dockerfile entries in multiple Dockerfiles
- `.md` files under `/docs`
- `grafana/` directory
- `GRAFANA_CONFIG_PATH`

controllers/argocd/ingress.go
```diff
@@ -246,75 +246,19 @@ func (r *ReconcileArgoCD) reconcileGrafanaIngress(cr *argoproj.ArgoCD) error {
 	if argoutil.IsObjectFound(r.Client, cr.Namespace, ingress.Name, ingress) {
 		if !cr.Spec.Grafana.Enabled || !cr.Spec.Grafana.Ingress.Enabled {
 			// Ingress exists but enabled flag has been set to false, delete the Ingress
-			return r.Client.Delete(context.TODO(), ingress)
+			return nil
```
jgwest, line 249: Correct me if I'm wrong, but I believe this `r.client.Delete` call still exist: even though we have deprecated Grafana, we should still clean up any old Ingresses that people might have had in the past.

controllers/argocd/route.go
```diff
@@ -100,60 +100,19 @@ func (r *ReconcileArgoCD) reconcileGrafanaRoute(cr *argoproj.ArgoCD) error {
 	if argoutil.IsObjectFound(r.Client, cr.Namespace, route.Name, route) {
 		if !cr.Spec.Grafana.Enabled || !cr.Spec.Grafana.Route.Enabled {
 			// Route exists but enabled flag has been set to false, delete the Route
-			return r.Client.Delete(context.TODO(), route)
+			return nil
 		}
```
jgwest, line 104: Same as above, I think this `Delete` call should still be here, so that it deletes an old Grafana Route that still exists, even though we have deprecated the functionality.

controllers/argocd/service.go
```diff
@@ -71,34 +71,18 @@ func (r *ReconcileArgoCD) reconcileGrafanaService(cr *argoproj.ArgoCD) error {
 	if argoutil.IsObjectFound(r.Client, cr.Namespace, svc.Name, svc) {
 		if !cr.Spec.Grafana.Enabled {
 			// Service exists but enabled flag has been set to false, delete the Service
-			return r.Client.Delete(context.TODO(), svc)
+			return nil
```
jgwest, line 74: Same as above, think this Delete should still exist.

jgwest (changes requested):
A few more minor items I found going over the changes, then good to go

controllers/argocd/ingress.go
Rizwana777: Yes @jgwest as e2e test for failing, I removed it (as e2e tests were flaky before, I didn't know the reason), added them back

jgwest (approved):
LGTM, thanks @Rizwana777!

controllers/argocd/configmap.go
```diff
@@ -309,12 +306,6 @@ func newConfigMapWithName(name string, cr *argoproj.ArgoCD) *corev1.ConfigMap {
 	return cm
 }
 
-// newConfigMapWithName creates a new ConfigMap with the given suffix appended to the name.
-// The name for the CongifMap is based on the name of the given ArgCD.
-func newConfigMapWithSuffix(suffix string, cr *argoproj.ArgoCD) *corev1.ConfigMap {
-	return newConfigMapWithName(fmt.Sprintf("%s-%s", cr.ObjectMeta.Name, suffix), cr)
-}
-
```
jaideepr97, line 317: @Rizwana777 did you mean to get rid of this function? it is used for other configmaps as well

controllers/argocd/util.go
```diff
@@ -669,7 +641,7 @@ func (r *ReconcileArgoCD) redisShouldUseTLS(cr *argoproj.ArgoCD) bool {
 	tlsSecretName := types.NamespacedName{Namespace: cr.Namespace, Name: common.ArgoCDRedisServerTLSSecretName}
 	err := r.Client.Get(context.TODO(), tlsSecretName, &tlsSecretObj)
 	if err != nil {
-		if !apierrors.IsNotFound(err) {
```
jaideepr97, line 672: I would prefer we keep this as `apierrors` as it is more descriptive and allows us to separately import the go errors pkg as well, if needed

jaideepr97:
Thanks for raising this @Rizwana777 , I am personally happy to see grafana getting deprecated from the operator :) 
Left a few minor comments

controllers/argocd/configmap.go
Rizwana777: @jaideepr97 `newConfigMapWithSuffix` is used only used by grafana right now and I didn't see any place in the code where it is used, so I have removed it , I can add it back if we are using it in future somewhere?

controllers/argocd/util.go
Rizwana777: updated

controllers/argocd/configmap.go
jaideepr97: oops, you're right
my bad

jaideepr97 (approved):
LGTM, thanks @Rizwana777

PR #1198, opened 2024-01-26
Fix status operation value to allow less than 10

anandrkskd:
**What type of PR is this?**
 /kind bug

Currently, when we set application-controller 'status/operation processor' values below 10, it set it to default value of 10.
This PR allows the value to be less than the default value, but is the value is less or equals to 0 it sets it back to the default value.

Fixes [GITOPS-3770](https://issues.redhat.com/browse/GITOPS-3770)


jaideepr97 (approved):
LGTM thanks @anandrkskd

jaideepr97 (changes requested):
please add a test case between 0 and the default

svghadi (approved):
LGTM. Thanks

jaideepr97 (approved):

PR #1202, opened 2024-01-28
code-refactoring: add/modify helper functions

jaideepr97:
**What type of PR is this?**

> /kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
/kind code-refactoring

This PR does:

- Add helper fn for performing partial matches on resources. This is helpful for unit tests when we want to compare objects for a specific set of fields, and ignore fields like `RevisionVersion` etc
- Modify `UpdateIfChanged` to accept batch input to make consumer side logic cleaner 
- Add helper to convert resources from `client.Object` to `runtime.Object` type
- Add nil check for labels before triggering rollout
- Add helper fn to decide if TLS should be used 

Fixes #?


controllers/argocd/applicationset/rolebinding.go
```diff
@@ -78,32 +80,32 @@ func (asr *ApplicationSetReconciler) reconcileRoleBinding() error {
...
-			&desiredRoleBinding.Subjects,
-		},
+	// if roleRef differs, we must delete the rolebinding as kubernetes does not allow updation of roleRef
+	if !reflect.DeepEqual(existingRoleBinding.RoleRef, desiredRoleBinding.RoleRef) {
+		asr.Logger.Info("detected drift in roleRef for rolebinding", "name", existingRoleBinding.Name, "namespace", existingRoleBinding.Namespace)
+		if err := asr.deleteRoleBinding(resourceName, asr.Instance.Namespace); err != nil {
+			return errors.Wrapf(err, "reconcileRoleBinding: unable to delete obsolete rolebinding %s", existingRoleBinding.Name)
+		}
+		return nil
+	}
```
svghadi, line 90: How will be the recreation of rolebinding handled? On next reconciliation? and does it make sense to log it as error instead of info as it is something that shouldn't have happened.

controllers/argocd/notifications/rolebinding.go
```diff
@@ -78,32 +80,32 @@ func (nr *NotificationsReconciler) reconcileRoleBinding() error {
...
-			&desiredRoleBinding.Subjects,
-		},
+	// if roleRef differs, we must delete the rolebinding as kubernetes does not allow updation of roleRef
+	if !reflect.DeepEqual(existingRoleBinding.RoleRef, desiredRoleBinding.RoleRef) {
+		nr.Logger.Info("detected drift in roleRef for rolebinding", "name", existingRoleBinding.Name, "namespace", existingRoleBinding.Namespace)
+		if err := nr.deleteRoleBinding(resourceName, nr.Instance.Namespace); err != nil {
+			return errors.Wrapf(err, "reconcileRoleBinding: unable to delete obsolete rolebinding %s", existingRoleBinding.Name)
+		}
+		return nil
+	}
```
svghadi, line 90: Same as previous comment.

controllers/argocd/argocdcommon/tls.go
svghadi, line 1: Is this file accidentally added in this PR? PR description doesn't have a mention about this.

controllers/argocd/applicationset/rolebinding.go
jaideepr97: yes @svghadi

controllers/argocd/notifications/rolebinding.go
jaideepr97: yes

controllers/argocd/argocdcommon/tls.go
jaideepr97: it was a last minute addition
description is updated

controllers/argocd/applicationset/rolebinding.go
svghadi: Should we return error in this case so that we can trigger the main `Reconcile()` loop again? My concern is that if the main `Reconcile()` loop is completed without error then the rolebinding will not be created until next reconciliation i.e by default 10hr if nothing is changed. 

Or maybe, lets merge it and figure out later when everything is in place, that it is actually  going to affect?

iam-veeramalla (approved):

PR #1205, opened 2024-01-29
fix: ApplicationSet WebhookServer configurations don't reflect in created Route

jparsai:
**What type of PR is this?**
> /kind bug

This PR has fix for the issue where operator doesn't reflect `ArgoCD.Spec.ApplicationSet.WebhookServer` configurations from CR into the Route that is created by operator.

Fixes #?

1. Create ArgoCD instance having below configurations in `Spec.ApplicationSet.WebhookServer` field.
```
applicationSet:
    webhookServer:
    host: webhook-host.<apps.URL>
    route:
        enabled: true
        tls:
          termination: edge 
```
**Actual results:**
- Webhook route created with default generated name, For example: `argocd-applicationset-controller-webhook-argo.apps.URL`
- Few other values that are given in `Spec.ApplicationSet.WebhookServer` are also not reflecting in created Route, like `Route.Annotations`, `Route.Labels`, `Route.Spec.TLS.Termination`, `Route.WildcardPolicy`.

**Expected results:**
- Webhook route created with given name in `Spec.ApplicationSet.WebhookServer.Host` field. As per above example: `webhook-host.<apps.URL>`
- All values given in fields of `Spec.ApplicationSet.WebhookServer` should reflect in created Route.

2. Other configurations of `Spec.ApplicationSet.WebhookServer` should also reflect in Route, like

### DIFF
--- a/controllers/argocd/route.go
+++ b/controllers/argocd/route.go
@@ -308,54 +308,80 @@ func (r *ReconcileArgoCD) reconcileApplicationSetControllerWebhookRoute(cr *argo
 
 	// Allow override of the Annotations for the Route.
 	if len(cr.Spec.ApplicationSet.WebhookServer.Route.Annotations) > 0 {
-		route.Annotations = cr.Spec.Server.Route.Annotations
+		route.Annotations = cr.Spec.ApplicationSet.WebhookServer.Route.Annotations
 	}
 
 	// Allow override of the Labels for the Route.
 	if len(cr.Spec.ApplicationSet.WebhookServer.Route.Labels) > 0 {
 		labels := route.Labels
-		for key, val := range cr.Spec.Server.Route.Labels {
+		for key, val := range cr.Spec.ApplicationSet.WebhookServer.Route.Labels {
 			labels[key] = val
 		}
 		route.Labels = labels
 	}
 
 	// Allow override of the Host for the Route.
-	if len(cr.Spec.Server.Host) > 0 {
+	if len(cr.Spec.ApplicationSet.WebhookServer.Host) > 0 {
 		route.Spec.Host = cr.Spec.ApplicationSet.WebhookServer.Host
 	}
 
-	if cr.Spec.Server.Insecure {
-		// Disable TLS and rely on the cluster certificate.
-		route.Spec.Port = &routev1.RoutePort{
-			TargetPort: intstr.FromString("webhook"),
+	route.Spec.Port = &routev1.RoutePort{
+		TargetPort: intstr.FromString("webhook"),
+	}
+
+	// Allow override of TLS options for the Route
+	if cr.Spec.ApplicationSet.WebhookServer.Route.TLS != nil {
+		tls := &routev1.TLSConfig{}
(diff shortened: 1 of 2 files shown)


controllers/argocd/route.go
```diff
@@ -308,54 +308,75 @@ func (r *ReconcileArgoCD) reconcileApplicationSetControllerWebhookRoute(cr *argo
...
+
+	// Allow override of TLS options for the Route
+	if cr.Spec.ApplicationSet.WebhookServer.Route.TLS != nil {
+		tls := &routev1.TLSConfig{}
+
+		// Set Termination
+		if cr.Spec.ApplicationSet.WebhookServer.Route.TLS.Termination != "" {
+			tls.Termination = cr.Spec.ApplicationSet.WebhookServer.Route.TLS.Termination
+		} else {
+			tls.Termination = routev1.TLSTerminationPassthrough
```
jgwest, line 340: I think this should default to TLSTerminationEdge, because the ApplicationSet webhook does not support [HTTPS AFAICT](https://github.com/argoproj/argo-cd/blob/3c2124235619d8451e2d24c7873e5a6da17354af/cmd/argocd-applicationset-controller/commands/applicationset_controller.go#L271).

If we instead use `TLSTerminationPassthrough`, then OpenShift will pass the HTTPS connection data directly to the applicationset controller webhook, which will throw it away because it's not valid HTTP.

controllers/argocd/route_test.go
```diff
@@ -213,6 +213,123 @@ func TestReconcileRouteUnsetsInsecure(t *testing.T) {
...
+	argoCD := makeArgoCD(func(a *argoproj.ArgoCD) {
+
+		a.Spec.ApplicationSet = &argoproj.ArgoCDApplicationSet{
+			WebhookServer: argoproj.WebhookServerSpec{
+				Host: "webhook-test.org",
+				Route: argoproj.ArgoCDRouteSpec{
+					Enabled: true,
+					TLS: &routev1.TLSConfig{
+						Termination: "passthrough",
+					},
```
jgwest, line 286: Do we want to test the other values in TLSConfig as well (key, ca cert, etc), since these were reported as part of the original bug?

jgwest (changes requested):
LGTM, just some minor requests found while reviewing re: default TLS value, and unit tests.

Following up from standup, I agree with your proposed solution here: I don't think we need to add a new `Insecure` field, because applicationset controller already does not support TLS.

controllers/argocd/route.go
jparsai: This is done.

jgwest (approved):
LGTM :+1: , thanks @jparsai!

controllers/argocd/route_test.go
jgwest: :+1:

controllers/argocd/route.go
jgwest: :+1:

iam-veeramalla (approved):

PR #1224, opened 2024-02-06
Add environment variable to change manager (operator) logging level

### LINKED ISSUE #1223: Support changing log level of manager

While the operator supports changing the log level of the components it manages, it does not provide a mechanism to change the log level of the operator itself. For users who are leveraging logging services like Splunk, where cost is based on the amount of logging, being to able to tune the amount of logging is highly desirable.

The current implementation picks up the log level from the command line using the zap command line options however this isn't useful in a container with a fixed entrypoint. The log level should come from a LOG_LEVEL environment variable that can be set either on the deployment or in an environment variable in the subscription when using OLM.

The environment variable should be set first but could still be overriden by the command line option which might be useful in debugging situations.

I considered modifying the entrypoint to use the `-zap-log-level` command line option and just set it to an environment variable, unfortunately if no environment variable is set the binary refuses to run with an empty setting. Additionally this command line switch isn't user variable since it only takes a limited number of string options (i.e. no warn flag for example), otherwise the user has to use a number based scheme which is obtuse.

**Additional context**
This is coming from Red Hat internal Red Hat issue GITOPS-3946

gnunn1:
**What type of PR is this?**
/kind bug

This enables changing logging level of the manager (i.e. operator) based on an environment variable.

* [*] Documentation update is required by this PR.
* [*] Documentation has been updated.

Fixes #1223 

1. Set the log level on the command line prior to running locally, i.e

```
export LOG_LEVEL=warn
make install run
```

2. In Kubernetes/OpenShift set the LOG_LEVEL environment on the Deployment for a Helm deployment or via the Subscription in OLM.


gnunn1:
The tests that are failing don't seem to be related to what I did in the PR?

svghadi:
It's a flaky one. Failed in my PR aswell few times.

svghadi (approved):
Tested out the changes locally. Looks good. 
Thanks @gnunn1 !!

PR #1258, opened 2024-02-20
Adding sharding unittest

Julik24:
**What type of PR is this?**

 /kind code-refactoring

We want to improve coverage of unit-tests.  This does a bit of refactoring by moving helper operationProcessors and controllerProcessors functions into util-test.go file and adds some test cases for .spec.controller.sharding

### DIFF
--- a/controllers/argocd/util_test.go
+++ b/controllers/argocd/util_test.go
@@ -90,12 +90,6 @@ func maxShardsSharding(n int32) argoCDOpt {
 	}
 }
 
-func clustersPerShardSharding(n int32) argoCDOpt {
-	return func(a *argoproj.ArgoCD) {
-		a.Spec.Controller.Sharding.ClustersPerShard = n
-	}
-}
-
 var imageTests = []struct {
 	name      string
 	pre       func(t *testing.T)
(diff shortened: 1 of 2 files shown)


controllers/argocd/util_test.go
```diff
@@ -544,6 +580,71 @@ func TestGetArgoApplicationControllerCommand(t *testing.T) {
 	}
 }
 
+func TestGetArgoApplicationControllerSharding(t *testing.T) {
+
+	enabled := []v1.EnvVar{
```
AdamSaleh, line 585: Because `enabled` looks to be the same as `replicas0` and just differes from replicas3 in the value, I'd suggest factoring it out into one more simple utility function? i.e `func replicasEnvVar(value string) []v1.EnvVar { ` e.t.c.

AdamSaleh (changes requested):
Looks good, only code improvement I would like to see is to simplify those env-var definitions.

And don't forget to use signoff when commiting, to appease the DCO check. The suggested fix `git rebase HEAD~2 --signoff` and `git push --force-with-lease origin sharding_unittest` should work and you just need to remember add `--signoff` when you will create future commits :)

controllers/argocd/util_test.go
```diff
@@ -544,6 +580,71 @@ func TestGetArgoApplicationControllerCommand(t *testing.T) {
 	}
 }
 
+func TestGetArgoApplicationControllerSharding(t *testing.T) {
```
mehabhalodiya, line 583: I see `Spec.Controller.Sharding` test already exists here: https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/statefulset_test.go#L347 

@AdamSaleh I believe we both looked at this existing test for sharding in one of our calls. Can you please confirm? I see that the existing test and this test do the same thing just that one resides in `statefulset_test.go` and Julia has written inside `util_test.go`.

mehabhalodiya (commented):
I am in favour of closing this PR. :)

controllers/argocd/util_test.go
AdamSaleh: These are very simmilar, but I still think having the additional one is useful, because it focuses on a smaller function - just `getArgoControllerContainerEnv(cr) ` and takes min and max shards into account as well.

mehabhalodiya: I believe everything related to sharding exists, if you want to see dynamic sharding testing, it is written here: https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/statefulset_test.go#L592

Julik24: Agree. So that's why this new test didn't increase test coverage :) Can I close the PR?

PR #373, opened 2021-07-16
fix: Use default image for repository server workload

jannfis:
Signed-off-by: jannfis <jann@mistrust.net>

**What type of PR is this?**

> /kind bug

**What does this PR do / why we need it**:

Fixes a bug from PR #358 where ARGOCD_IMAGE environment is not respected for argocd-repo-server workload anymore. We should not have a separate default value for that.

**Have you updated the necessary documentation?**

* [ ] Documentation update is required by this PR.
* [ ] Documentation has been updated.

**Which issue(s) this PR fixes**:

Fixes #?

**How to test changes / Special notes to the reviewer**:

* Set `ARGOCD_IMAGE` pointing to a custom image
* Run Operator locally
* Create `ArgoCD` CR
* Verify that `argocd-repo-server` deployment uses the custom image instead of the default one

PR #1262, opened 2024-02-21
 fix: Use image configured in spec as fallback before default image for repository server workload

### LINKED ISSUE #623: Repo Server Image should use spec.version if spec.repo.version is not set
**Describe the bug**
When using a different ArgoCD version than the default version set by the operator, we set spec.version in the CR.
Documentation states for spec.version: "**The tag to use with the container image for all Argo CD components**." (https://argocd-operator.readthedocs.io/en/latest/reference/argocd/#version)

The repo server, however, does not respect this value but only looks at spec.repo.version or falls back to defaults 
https://github.com/argoproj-labs/argocd-operator/blob/68e1197b38ab9ce2461d8825765564eece2e1e4d/controllers/argocd/util.go#L149-L159

https://github.com/argoproj-labs/argocd-operator/blob/68e1197b38ab9ce2461d8825765564eece2e1e4d/controllers/argocd/util.go#L168-L172

**To Reproduce**
1. Install ArgoCD with operator version 0.3.0 
2. Create an ArgoCD instance and set spec.version to 2.2.5
3. Observe repo server deployment - it has the image SHA digest of ArgoCD image 2.3.3

**Expected behavior**
Repo Server falls back to spec.version if spec.repo.version ist not set

tomjo:
**What type of PR is this?**
/kind bug

* The repo server container image version will now fallback to spec.version before falling back to the default version if spec.repo.version is not set. Otherwise users get in the confusing situation of all the argo pods other than repo running a newer version than the repo server.
* It also fixes some documentation and code comments which were inaccurate after a previous change (#373). Also cleans up the unused field ArgoCDRepoImageEnvName after that change.

Fixes #623 

Added tests for the repo server image and version

### DIFF
--- a/controllers/argocd/util.go
+++ b/controllers/argocd/util.go
@@ -205,10 +205,10 @@ func getArgoContainerImage(cr *argoproj.ArgoCD) string {
 //
 // 1. from the Spec, the spec.repo field has an image and version to use for
 // generating an image reference.
-// 2. from the Environment, this looks for the `ARGOCD_IMAGE` field and uses
-// that if the spec is not configured.
-// 3. from the Spec, the spec.version field has an image and version to use for
+// 2. from the Spec, the spec.version field has an image and version to use for
 // generating an image reference
+// 3. from the Environment, this looks for the `ARGOCD_IMAGE` field and uses
+// that if the spec is not configured.
 // 4. the default is configured in common.ArgoCDDefaultArgoVersion and
 // common.ArgoCDDefaultArgoImage.
 func getRepoServerContainerImage(cr *argoproj.ArgoCD) string {
--- a/controllers/argocd/util.go
+++ b/controllers/argocd/util.go
@@ -200,27 +200,35 @@ func getArgoContainerImage(cr *argoproj.ArgoCD) string {
 
 // getRepoServerContainerImage will return the container image for the Repo server.
 //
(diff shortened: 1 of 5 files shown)


controllers/argocd/util.go
```diff
@@ -186,27 +186,35 @@ func getArgoContainerImage(cr *argoproj.ArgoCD) string {
...
 //
 // 1. from the Spec, the spec.repo field has an image and version to use for
 // generating an image reference.
-// 2. from the Environment, this looks for the `ARGOCD_REPOSERVER_IMAGE` field and uses
+// 2. from the Environment, this looks for the `ARGOCD_IMAGE` field and uses
 // that if the spec is not configured.
-// 3. the default is configured in common.ArgoCDDefaultRepoServerVersion and
-// common.ArgoCDDefaultRepoServerImage.
+// 3. from the Spec, the spec.version field has an image and version to use for
+// generating an image reference
```
svghadi, line 197: Based on current logic image value from `.spec.version` will be used over `ARGOCD_IMAGE` env var, if both are set, which is fine as it is the same for other components. Can you switch these statements ?

svghadi (changes requested):
Thanks @tomjo for the PR. I have left a comment, rest all looks good.

svghadi (approved):
Thanks alot @tomjo for your contribution.

PR #1299, opened 2024-04-07
Adds podman compatibility in makefile

### LINKED ISSUE #1298: Add podman compatibility to Makefile
I would like to use podman for building the argocd operator instead of docker, without breaking it for people who still use docker.

I'm using Fedora and the Podman Buildah stack, so i have no Docker installed, so per default i can't build the operator

Makefile could detect if it can use docker or podman

Elyytscha:
**What type of PR is this?**

/kind enhancement

Adds Podman compatibility to Makefile

I think no doc update is needed for this

Fixes #1298

```bash
make docker-build docker-push
make bundle-build bundle-push
make catalog-build catalog-push
```

with podman or docker installed, both will work

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ CONTAINER_RUNTIME := $(shell command -v docker 2> /dev/null || command -v podman
 
 # If neither Docker nor Podman is found, print an error message and exit
 ifeq ($(CONTAINER_RUNTIME),)
-$(error "No container runtime (Docker or Podman) found in PATH. Please install one of them.")
+$(warning "No container runtime (Docker or Podman) found in PATH. Please install one of them.")
 endif
 
 # CHANNELS define the bundle channels used in the bundle.
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,14 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= 0.9.0
 
+# Try to detect Docker or Podman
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -5,6 +5,14 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= 0.9.0
 
+# Try to detect Docker or Podman
+CONTAINER_RUNTIME := $(shell command -v docker 2> /dev/null || command -v podman 2> /dev/null)
+
+# If neither Docker nor Podman is found, print an error message and exit
+ifeq ($(CONTAINER_RUNTIME),)
+$(error "No container runtime (Docker or Podman) found in PATH. Please install one of them.")
+endif
```
svghadi, line 14: Can we remove this check or just log the message and continue without exiting? My concern with this is that if I just want to run some non image build command like `make run` or `make install` or some unit test on system without docker/podman, it won't allow me to do so.

svghadi (changes requested):
Thank You @Elyytscha for the PR. I have a minor suggestion, other than that look good.

Makefile
Elyytscha: changed from error to warning, so now it should only print a warning instead of exiting with an error

svghadi (approved):

Elyytscha:
feel free to squash it, if it gets merged, does not need to be 2 commits imo.

svghadi:
Thanks @Elyytscha for the contribution.

PR #1304, opened 2024-04-15
Operator redesign/Add tests for util and argoutil pkg

Julik24:
**What type of PR is this?**

/kind code-refactoring

It's not entirely clear to me how VerifyPassword function works. In its definition it takes parameters (password, hashedPassword), but in function HasArgoAdminPasswordChanged it is called with parameters(expectedPwd, actualPwd).

### DIFF
--- a/pkg/argoutil/auth_test.go
+++ b/pkg/argoutil/auth_test.go
@@ -1,0 +1,67 @@
+package argoutil
+
+import (
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	"testing"
+)
+
+func TestGenerateArgoAdminPassword(t *testing.T) {
+	password, err := GenerateArgoAdminPassword()
+	assert.NoError(t, err)
+	assert.NotNil(t, password)
+}
+
+func TestGenerateArgoServerSessionKey(t *testing.T) {
+	password, err := GenerateArgoServerSessionKey()
+	assert.NoError(t, err)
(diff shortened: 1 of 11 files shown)


pkg/argoutil/auth_test.go
```diff
@@ -0,0 +1,67 @@
...
+			Data: map[string][]byte{
+				"admin-password": old_admin_password,
+			},
+		}
+
+		got := HasArgoAdminPasswordChanged(old_password, old_password)
+		if got != false {
+			t.Errorf("HasAdminPasswordChanged() = %v, want false", got)
+		}
+	})*/
```
jaideepr97, line 66: @Julik24 did you mean to leave this in?

jaideepr97: LGTM other than this

Julik24: @jaideepr97 , that's the test of a function I have a question about: is it ok that HasArgoAdminPasswordChanged(expectedPwd, actualPwd) takes two passwords as parameters and calls with this parameters argopass.VerifyPassword(password, hashedPassword), which must take hash instead of the second password? If it doesn't interfere with anything, I'll just remove the commented code

jaideepr97: good catch @Julik24, that definitely seems like a bug in how the function is written 
I would suggest uncommenting this test so someone can contribute a fix

PR #1317, opened 2024-04-25
Fix redirect uri

anandf:
**What type of PR is this?**

> /kind bug

In the `reconcileArgoConfigMap` method, if dex is enabled, ConfigMap is updated twice - once in the `reconcileDexConfiguration` method and then outside. There is a chance that the redirect URI is different in both these updates, causing wrong values to be persisted in the configMap. To fix that, setting the same value of dex config in both the updates.

Reproduced the problem in a demo cluster and ensured that the problem is not reproducible with this fix.

Fixes https://issues.redhat.com/browse/GITOPS-4358 and https://issues.redhat.com/browse/GITOPS-3736


jannfis (approved):
LGTM

svghadi (approved):
lgtm. Thanks!!

PR #1323, opened 2024-04-26
fix(Redis): Custom Endpoint without Deploying Redis

ArthurVardevanyan:
**What type of PR is this?**

/kind bug

When Redis is set to false, the remote field is not being set, and the argocd controller/repo server fall back to the default value of  ```argocd-redis```

```yaml
  redis:
    enabled: false
    remote: 'argocd-dragonfly:6379'
```

When Redis is set to true, the custom remote field is being set, but the redis deployment is also being deployed
```yaml
  redis:
    enabled: true
    remote: 'argocd-dragonfly:6379'
 ```
 
 Looking at the code,  it seemed simpler to update the condition to be ```enabled + remote``` for a custom endpoint instead of rewriting for ```disabled + remote```.
    
N/A?

Should I open a Separate Issue?

### DIFF
--- a/controllers/argocd/deployment.go
+++ b/controllers/argocd/deployment.go
@@ -553,6 +553,11 @@ func (r *ReconcileArgoCD) reconcileRedisDeployment(cr *argoproj.ArgoCD, useTLS b
 		return nil // Deployment found with nothing to do, move along...
 	}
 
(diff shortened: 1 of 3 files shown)


iam-veeramalla (approved):
Thanks for the contribution @ArthurVardevanyan

PR #1329, opened 2024-05-02
fix: verify DeploymentConfig API when installing keycloak using Template

chetan-rns:
**What type of PR is this?**

/kind bug

Users have the option to disable DeploymentConfig API on their OCP clusters. In such cases, the operator shouldn't watch the DeploymentConfig APIs and shouldn't install Keycloak using Template.

https://docs.openshift.com/container-platform/4.14/installing/cluster-capabilities.html#deployment-config-capability_cluster-capabilities

Fixes https://issues.redhat.com/browse/GITOPS-4489

### DIFF
--- a/controllers/argocd/sso.go
+++ b/controllers/argocd/sso.go
@@ -46,22 +46,17 @@ func CanUseKeycloakWithTemplate() bool {
 }
 
 func verifyKeycloakTemplateAPIs() error {
-	found, err := argoutil.VerifyAPI(deploymentConfig.GroupVersion.Group, deploymentConfig.GroupVersion.Version)
+	var err error
+	deploymentConfigAPIFound, err = argoutil.VerifyAPI(deploymentConfig.GroupVersion.Group, deploymentConfig.GroupVersion.Version)
 	if err != nil {
 		return err
 	}
-	deploymentConfigAPIFound = found
 
-	return verifyTemplateAPI()
-}
-
-// verifyTemplateAPI will verify that the template API is present.
-func verifyTemplateAPI() error {
-	found, err := argoutil.VerifyAPI(template.GroupVersion.Group, template.GroupVersion.Version)
+	templateAPIFound, err = argoutil.VerifyAPI(template.GroupVersion.Group, template.GroupVersion.Version)
 	if err != nil {
 		return err
 	}
-	templateAPIFound = found
+
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


controllers/argocd/sso.go
```diff
@@ -33,13 +34,25 @@ const (
...
+}
+
+func verifyKeycloakTemplateAPIs() error {
+	found, err := argoutil.VerifyAPI(deploymentConfig.GroupVersion.Group, deploymentConfig.GroupVersion.Version)
+	if err != nil {
+		return err
+	}
+	deploymentConfigAPIFound = found
+
+	return verifyTemplateAPI()
```
jgwest, line 55: nit: I recommend moving the contents of verifyTemplateAPI into this function.  'verifyKeycloakTemplateAPIs' is the only function calling it, and we probably want to make sure that both variables are always updated at the same time.

main.go
```diff
@@ -227,7 +227,7 @@ func main() {
 	}
 
 	// Setup Schemes for SSO if template instance is available.
-	if argocd.IsTemplateAPIAvailable() {
+	if argocd.CanUseKeycloakWithTemplate() {
```
jgwest, line 230: WDYT about adding a log statement which indicates whether the functionality will be enabled/disabled, based on the results of the call to `InspectCluster` above?

jgwest (changes requested):
One question I have for this PR:

My concern with this PR is that it is sending new OpenShift 4.15 clusters down an untested path: currently, (I suspect) the vast vast majority of OpenShift users would have DeploymentConfig configured, and thus when keyclock was enabled, it would use the template-based approach. 

Very very few users would be using the non-Template, Ingress-based approach:  basically only argocd-operator users on non-OpenShift.

However, with this PR, we are now sending all those new OpenShift 4.15 users (with cluster that defaults to not having this API) down this mostly untested code path. 

All those users would switch from using the Template API-based codepath that is commonly used, to the ingress-based codepath which is rarely used. (And anything that is rarely used is a lot less likely to work)

For example, if you don't have the required APIs installed, we will no longer install keycloak from the official openshift image on registry.redhat.com: `registry.redhat.io/rh-sso-7/sso76-openshift-rhel8` 

Instead, we install keycloak from `quay.io/keycloak/keycloak`:

```
	// ArgoCDKeycloakImage is the default Keycloak Image used for the non-openshift platforms when not specified.
	ArgoCDKeycloakImage = "quay.io/keycloak/keycloak"
```

As an example of the kind of issues we might encounter, the quay.io/keycloak version hasn't been updated in 2 years, whereas the openshift tag was updated 7 months ago.

An alternative to the approach taken in this PR would be just to disable the keycloak functionality entirely if the user doesn't have the required APIs, and output a message telling the user to install those PRs.

chetan-rns:
@jgwest That's a valid concern. I agree that the Operator should prefer the official image for Keycloak on OCP clusters. I've updated the PR to log and return an error if the user selects Keycloak but the DeploymentConfig API is absent. I also had to check if the Template API is present to ensure it is an OCP cluster. This should prevent new users from installing the unofficial image. 

As a long-term solution, we should update the Template to use Deployment instead.

main.go
```diff
@@ -227,7 +227,8 @@ func main() {
...
+		setupLog.Info("Keycloak instance can be managed using OpenShift Template")
```
jgwest, line 231: LGTM, one final, tiny request then good to go: can we also add an Info statement (same as above) that outputs for the case where OpenShift Template is not available? Basically the opposite case to what is in the if statement above. 

Logging something like "Keycloak instance cannot be managed, as DeploymentConfig API is not installed on the cluster."

jgwest (changes requested):

chetan-rns:
@jgwest @iam-veeramalla This PR might be redundant if we plan to move away from Templates. Until then we can accept this as a temporary solution so users don't see errors when the DeploymentConfig is disabled. WDYT?

jgwest (approved):

jgwest:
Merging as no objections were raised.

Issue #1275, opened 2024-03-03
Support defining custom cluster roles for cluster scoped instances


gnunn1:
**Is your feature request related to a problem? Please describe.**

When using a cluster scoped instance of Argo CD the operator will create a separate and hard coded set of clusterroles and clusterrolebindings. Users that want to lock down the permissions of their instances need the ability to use an alternate set of clusterroles. 

As an example, the new feature in Argo CD 2.10 to auto-respect RBAC (https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#auto-respect-rbac-for-controller) enables users to remove the cluster-wide view all permissions required by a cluster scoped instance and provide a much more tailored set of permissions. However the Argo CD Operator hard codes this cluster role to define view all permissions.

**Describe the solution you'd like**
Enable users to specify their own roles using the environment variables `CONTROLLER_CLUSTER_SCOPE_ROLE` and `SERVER_CLUSTER_SCOPE_ROLE`

**Describe alternatives you've considered**
Updates to the cluster role created by the operator get overridden by the operator so it's not a suitable approach.

gnunn1:
PR is in progress for this.

jparsai:
Reopening since PR was reverted and there is a new PR raised to address this.

PR #1357, opened 2024-05-16
feat: Add support for custom ClusterRoles (controller,server) for cluster scoped instances

jparsai:
**What type of PR is this?**

/kind enhancement

This PR enables user to define custom roles for cluster scoped instances of Argo CD. This is becoming increasingly important with the addition of the auto-respect RBAC feature added in Argo CD 2.10.

- [x]  Documentation update is required by this PR.
- [x]  Documentation has been updated.

**Which issue(s) this PR fixes:**

Fixes https://github.com/argoproj-labs/argocd-operator/issues/1275
Note this also tracked in Red Hat's issue database here: https://issues.redhat.com/browse/GITOPS-2614

**How to test changes / Special notes to the reviewer:**
- Create a namespace and to make it cluster scoped namespace update `ARGOCD_CLUSTER_CONFIG_NAMESPACES` environment variable of Subscription resource. 
- Create ClusterRoles and ClusterRoleBindings for Application controller and server.
- Create Argo CD CR in namespace and set  `Spec.DefaultClusterScopedRoleCreationDisabled` to `true`. 
- All Argo CD components should be healthy.

### DIFF
--- a/controllers/argocd/role.go
+++ b/controllers/argocd/role.go
@@ -280,10 +280,29 @@ func (r *ReconcileArgoCD) reconcileRoleForApplicationSourceNamespaces(name strin
 }
 
 func (r *ReconcileArgoCD) reconcileClusterRole(name string, policyRules []v1.PolicyRule, cr *argoproj.ArgoCD) (*v1.ClusterRole, error) {
+
 	allowed := false
 	if allowedNamespace(cr.Namespace, os.Getenv("ARGOCD_CLUSTER_CONFIG_NAMESPACES")) {
 		allowed = true
 	}
+
+	// Check if it is cluster-scoped instance namespace and user doesn't want to use default ClusterRole
+	if allowed && cr.Spec.DefaultClusterScopedRoleDisabled {
+
+		// In case DefaultClusterScopedRoleDisabled was false earlier and default ClusterRole was created, then delete it.
+		existingClusterRole := &v1.ClusterRole{}
+		if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: GenerateUniqueResourceName(name, cr)}, existingClusterRole); err == nil {
+
+			// Default ClusterRole exists, now delete it
+			if err := r.Client.Delete(context.TODO(), existingClusterRole); err != nil {
+				return nil, fmt.Errorf("failed to delete existing cluster role for the service account associated with %s : %s", name, err)
+			}
+		}
+
+		// Don't create a default ClusterRole
+		return nil, nil
+	}
+
 	clusterRole := newClusterRole(name, policyRules, cr)
 	if err := applyReconcilerHook(cr, clusterRole, ""); err != nil {
 		return nil, err
--- a/controllers/argocd/role_test.go
+++ b/controllers/argocd/role_test.go
@@ -151,6 +151,60 @@ func TestReconcileArgoCD_reconcileClusterRole(t *testing.T) {
 	assert.Contains(t, r.Client.Get(context.TODO(), types.NamespacedName{Name: clusterRoleName}, reconciledClusterRole).Error(), "not found")
 }
 
+func TestReconcileArgoCD_reconcileClusterRole_disabled(t *testing.T) {
+	logf.SetLogger(ZapLogger(true))
+	a := makeTestArgoCD()
+
+	resObjs := []client.Object{a}
+	subresObjs := []client.Object{a}
+	runtimeObjs := []runtime.Object{}
+	sch := makeTestReconcilerScheme(argoproj.AddToScheme)
+	cl := makeTestReconcilerClient(sch, resObjs, subresObjs, runtimeObjs)
+	r := makeTestReconciler(cl, sch)
+
+	workloadIdentifier := common.ArgoCDApplicationControllerComponent
+	clusterRoleName := GenerateUniqueResourceName(workloadIdentifier, a)
+	expectedRules := policyRuleForApplicationController()
+
+	// Set the namespace to be cluster-scoped
+	t.Setenv("ARGOCD_CLUSTER_CONFIG_NAMESPACES", a.Namespace)
+
+	// Disable creation of default ClusterRole
+	a.Spec.DefaultClusterScopedRoleDisabled = true
+
+	// Reconcile ClusterRole
+	_, err := r.reconcileClusterRole(workloadIdentifier, expectedRules, a)
+	assert.NoError(t, err)
+
+	// Ensure default ClusterRole is not created
+	reconciledClusterRole := &v1.ClusterRole{}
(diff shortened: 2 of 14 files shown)


controllers/argocd/rolebinding.go
```diff
@@ -312,13 +311,51 @@ func (r *ReconcileArgoCD) reconcileRoleBinding(name string, rules []v1.PolicyRul
...
+// getCustomRoleName returns value of env variables defined for custom cluster role for namespace scoped Argo CD
+func getCustomRoleName(componentName string, cr *argoproj.ArgoCD) string {
+
+	// return env variable value for application-controller
+	if componentName == common.ArgoCDApplicationControllerComponent && cr.Spec.Controller.Env != nil {
+		for _, env := range cr.Spec.Controller.Env {
+			if env.Name == common.ArgoCDControllerClusterRoleEnvName {
+				return env.Value
+			}
+		}
```
svghadi, line 323: What if we prioritize `cr.Spec.Controller.Env` over `os.Getenv(common.ArgoCDControllerClusterRoleEnvName)` instead of completely removing it? This way, we can ensure that existing users are not affected by the change.  WDYT?

svghadi (commented):
Thanks for the PR @jparsai. I have some concerns regarding backward compatibility. Are we removing the ability to specify cluster roles via subscriptions?

controllers/argocd/rolebinding.go
jparsai: Ahh, it slipped from my mind, I will fix it.

jparsai: It is done, PTAL.

controllers/argocd/rolebinding.go
```diff
@@ -312,13 +312,61 @@ func (r *ReconcileArgoCD) reconcileRoleBinding(name string, rules []v1.PolicyRul
...
+		if cr.Spec.Controller.Env != nil {
+			for _, env := range cr.Spec.Controller.Env {
+				if env.Name == common.ArgoCDControllerClusterRoleEnvName {
+					return env.Value
+				}
+			}
+		}
+
+		// DEPRECATED: Check if env variable is set in Subscription resource
 		return os.Getenv(common.ArgoCDControllerClusterRoleEnvName)
```
svghadi, line 329: Do we want to deprecate subscription way of setting roles? I see value in having both via CR and subscription.

jparsai: Yes, it is changed now, there is also change in implementation, PTAL.

controllers/argocd/role_test.go
```diff
@@ -151,6 +151,34 @@ func TestReconcileArgoCD_reconcileClusterRole(t *testing.T) {
...
+
+	t.Setenv("ARGOCD_CLUSTER_CONFIG_NAMESPACES", a.Namespace)
+	_, err := r.reconcileClusterRole(workloadIdentifier, expectedRules, a)
+	assert.NoError(t, err)
+
+	reconciledClusterRole := &v1.ClusterRole{}
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: clusterRoleName}, reconciledClusterRole)
+	assert.Error(t, err)
+	assert.ErrorContains(t, err, "not found")
+}
```
jgwest, line 180: Can we also add some quick logic at the end of the test that sets `a.Spec.DefaultClusterScopedRoleCreationDisabled = false` and verifies that the roles are created again?

controllers/argocd/role_test.go
```diff
@@ -151,6 +151,34 @@ func TestReconcileArgoCD_reconcileClusterRole(t *testing.T) {
 	assert.Contains(t, r.Client.Get(context.TODO(), types.NamespacedName{Name: clusterRoleName}, reconciledClusterRole).Error(), "not found")
 }
 
+func TestReconcileArgoCD_reconcileClusterRole_disabled(t *testing.T) {
```
jgwest, line 154: It would also be good to test the case where:
- an `ArgoCD` is initially created with `a.Spec.DefaultClusterScopedRoleCreationDisabled = false`
- reconcile is called
- verify cluster roles are created
- then that `ArgoCD` is set `a.Spec.DefaultClusterScopedRoleCreationDisabled = true`
- verify cluster roles are deleted

controllers/argocd/rolebinding_test.go
```diff
@@ -184,6 +184,32 @@ func TestReconcileArgoCD_reconcileClusterRoleBinding(t *testing.T) {
 	assert.NoError(t, r.Client.Get(context.TODO(), types.NamespacedName{Name: expectedName}, clusterRoleBinding))
 }
 
+func TestReconcileArgoCD_reconcileClusterRoleBinding_disabled(t *testing.T) {
+	logf.SetLogger(ZapLogger(true))
```
jgwest, line 188: Same as my above in `role_test.go`, good to check the case where:
- `DefaultClusterScopedRoleCreationDisabled = true` -> `DefaultClusterScopedRoleCreationDisabled = false: ensure ClusterRoleBinding is created`
- `DefaultClusterScopedRoleCreationDisabled = false` -> `DefaultClusterScopedRoleCreationDisabled = true: ensure ClusterRoleBinding is deleted`

controllers/argocd/role.go
```diff
@@ -284,6 +284,12 @@ func (r *ReconcileArgoCD) reconcileClusterRole(name string, policyRules []v1.Pol
 	if allowedNamespace(cr.Namespace, os.Getenv("ARGOCD_CLUSTER_CONFIG_NAMESPACES")) {
 		allowed = true
 	}
+
+	if allowed && cr.Spec.DefaultClusterScopedRoleCreationDisabled {
+		// Don't create a default ClusterRole, because user wants to create a custom ClusterRole for cluster-scoped instance.
+		return nil, nil
+	}
+
```
jgwest, line 292: If `DefaultClusterScopedRoleCreationDisabled` is set to  true, should we delete the cluster roles if they already exist?

controllers/argocd/rolebinding.go
```diff
@@ -340,6 +340,11 @@ func newRoleBindingWithNameForApplicationSourceNamespaces(namespace string, cr *
 
 func (r *ReconcileArgoCD) reconcileClusterRoleBinding(name string, role *v1.ClusterRole, cr *argoproj.ArgoCD) error {
 
+	if cr.Spec.DefaultClusterScopedRoleCreationDisabled {
```
jgwest, line 343: If `DefaultClusterScopedRoleCreationDisabled` is set to  true, should we delete the cluster roles bindings if they already exist?

jgwest (changes requested):
Looks great, just some minor stuff I could think of re: expected user behaviour.

One more request: can you add a simple kuttl test that sets `DefaultClusterScopedRoleCreationDisabled` to true, and ensures that ClusterRole/ClusterRoleBinding are not created?

jgwest (approved):
LGTM, thanks @jparsai!

PR #1363, opened 2024-05-20
feat: Update the default TLS termination policy to reencrypt 

chetan-rns:
**What type of PR is this?**

> /kind enhancement

- This PR updates the default TLS termination policy of OpenShift Route from passthrough to reencrypt. With this change, OpenShift users could use the default ingress certificates for the argocd server without any explicit modification.
- The user/browser/CLI and the Router will communicate using the ingress certificate. The operator relies on OpenShift Service's CA to generate a self-signed certificate in the argocd-server-tls secret which will be used between the Router and the Argo CD server.

Fixes #?

1. Create an Argo CD CR without specifying the server's TLS termination policy.
2. Verify if the server Route is using the Reencrypt policy.
3. Open the server route in a browser and ensure that it is using the ingress operator certificate.

### DIFF
--- a/api/v1alpha1/argocd_types.go
+++ b/api/v1alpha1/argocd_types.go
@@ -960,10 +960,12 @@ func (argocd *ArgoCD) IsDeletionFinalizerPresent() bool {
 	return false
 }
 
-// WantsAutoTLS returns true if user configured a route with reencryption
-// termination policy.
+// WantsAutoTLS returns true if:
+// 1. user has configured a route with reencrypt.
+// 2. user has not configured TLS and we default to reencrypt.
 func (s *ArgoCDServerSpec) WantsAutoTLS() bool {
-	return s.Route.TLS != nil && s.Route.TLS.Termination == routev1.TLSTerminationReencrypt
+	return s.Route.TLS == nil ||
+		(s.Route.TLS != nil && s.Route.TLS.Termination == routev1.TLSTerminationReencrypt)
 }
 
 // WantsAutoTLS returns true if the repository server configuration has set
--- a/api/v1alpha1/argocd_types.go
+++ b/api/v1alpha1/argocd_types.go
@@ -964,8 +964,7 @@ func (argocd *ArgoCD) IsDeletionFinalizerPresent() bool {
 // 1. user has configured a route with reencrypt.
 // 2. user has not configured TLS and we default to reencrypt.
 func (s *ArgoCDServerSpec) WantsAutoTLS() bool {
-	return s.Route.TLS == nil ||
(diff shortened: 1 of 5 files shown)


chetan-rns:
Jira: https://issues.redhat.com/browse/GITOPS-4674

api/v1alpha1/argocd_types.go
```diff
@@ -960,10 +960,12 @@ func (argocd *ArgoCD) IsDeletionFinalizerPresent() bool {
...
 
-// WantsAutoTLS returns true if user configured a route with reencryption
-// termination policy.
+// WantsAutoTLS returns true if:
+// 1. user has configured a route with reencrypt.
+// 2. user has not configured TLS and we default to reencrypt.
 func (s *ArgoCDServerSpec) WantsAutoTLS() bool {
-	return s.Route.TLS != nil && s.Route.TLS.Termination == routev1.TLSTerminationReencrypt
+	return s.Route.TLS == nil ||
+		(s.Route.TLS != nil && s.Route.TLS.Termination == routev1.TLSTerminationReencrypt)
```
anandf, line 968: Do we need the extra not nil check `s.Route.TLS != nil` ? Since `nil` check is already done, it is implied that the other condition would be evaluated only if its not nil. Is my understanding correct or Did I misunderstood the logic?

api/v1beta1/argocd_types.go
```diff
@@ -991,10 +991,12 @@ func (argocd *ArgoCD) IsDeletionFinalizerPresent() bool {
...
 
-// WantsAutoTLS returns true if user configured a route with reencryption
-// termination policy.
+// WantsAutoTLS returns true if:
+// 1. user has configured a route with reencrypt.
+// 2. user has not configured TLS and we default to reencrypt.
 func (s *ArgoCDServerSpec) WantsAutoTLS() bool {
-	return s.Route.TLS != nil && s.Route.TLS.Termination == routev1.TLSTerminationReencrypt
+	return s.Route.TLS == nil ||
+		(s.Route.TLS != nil && s.Route.TLS.Termination == routev1.TLSTerminationReencrypt)
```
anandf, line 999: Same do we need the extra not nil check ?

anandf (commented):
A minor comment, otherwise looks good to me.

svghadi (commented):
Tested out the change, looks good. Thanks @chetan-rns.

api/v1alpha1/argocd_types.go
chetan-rns: You are right, Anand. Updated the logic.

api/v1beta1/argocd_types.go
chetan-rns: Updated.

anandf (approved):
LGTM

jgwest (approved):

PR #1364, opened 2024-05-20
feat: secure redis authentication [CVE-2024-31989]

iam-veeramalla:
**What type of PR is this?**
> /kind bug

Fix CVE-2024-31989

It has been discovered that an unprivileged pod in a different namespace on the same cluster could connect to the Redis server on port 6379. Despite having installed the latest version of the VPC CNI plugin on the EKS cluster, it requires manual enablement through configuration to enforce network policies. This raises concerns that many clients might unknowingly have open access to their Redis servers. This vulnerability could lead to Privilege Escalation to the level of cluster controller, or to information leakage, affecting anyone who does not have strict access controls on their Redis instance. This issue has been patched in version(s) 2.8.19, 2.9.15 and 2.10.10.

NA

Fixes CVE-2024-31989

### DIFF
--- a/controllers/argocd/policyrule.go
+++ b/controllers/argocd/policyrule.go
@@ -41,7 +41,6 @@ func policyRuleForRedis(client client.Client) []v1.PolicyRule {
 				"get",
 				"list",
 				"watch",
-				"create",
 			},
 		},
 	}
@@ -78,7 +77,6 @@ func policyRuleForRedisHa(client client.Client) []v1.PolicyRule {
 				"get",
 				"list",
 				"watch",
-				"create",
 			},
 		},
 	}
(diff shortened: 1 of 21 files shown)


controllers/argocd/policyrule.go
```diff
@@ -29,7 +29,22 @@ func policyRuleForApplicationController() []v1.PolicyRule {
...
+				"",
+			},
+			Resources: []string{
+				"secrets",
+			},
+			Verbs: []string{
+				"get",
+				"list",
+				"watch",
+				"create",
```
svghadi, line 44: what do you think on limiting this only to get? If I am not wrong, Redis will only read the secret and creation will happen in operator manager.

iam-veeramalla: Agreed. Made changes accordingly.

svghadi (approved):
Ran some manual test as well as ha kuttl test. LGTM. Thanks @iam-veeramalla

PR #1375, opened 2024-05-22
[code-refactoring] fix: correct appset controller status and update tests

anandrkskd:
**What type of PR is this?**
 /kind bug
 /kind failing-test

This PR fix missing application status report of argocd instance 
This PR fixes tests that due to miss-configuration

Fixes [GITOPS-4670](https://issues.redhat.com/browse/GITOPS-4670)


tests/k8s/1-001_validate_basic/01-assert.yaml
```diff
@@ -52,7 +52,7 @@ metadata:
 kind: ServiceAccount
 apiVersion: v1
 metadata:
-  name: example-argocd-argocd-redis-ha
+  name: example-argocd-argocd-redis
```
saumeya, line 55: Since the naming convention has changed in the current code, we can update all the names of service account and rolebindings in this test file to skip `argocd` duplication, 
For example - `name: example-argocd-redis`

tests/k8s/1-035_validate_applicationset_reconcile_enabled_set_true/07-install.yaml
saumeya, line 1: Why are we removing this test file?

tests/k8s/1-002_validate_cluster_config/README.md
saumeya, line 1: We should create a step to add a script patch the subscription before running this test. If needed we can can add a script to remove the env variable after the test has run.

saumeya (commented):
Thanks Anand, the changes look good. I have added a few comments, PTAL

As discussed, we have some other test failures but we can address them in subsequent PRs.

saumeya:
We also have 2 tests names 1-032, can we rename them to avoid confusion?

tests/k8s/1-001_validate_basic/01-assert.yaml
anandrkskd: updated

tests/k8s/1-035_validate_applicationset_reconcile_enabled_set_true/07-install.yaml
anandrkskd: That was unintentional, fixed it.

tests/k8s/1-002_validate_cluster_config/README.md
anandrkskd: I have discussed this with Siddhesh previously, we can move this test to one of the olm related tests that requires changes in the subscription. I would say we can let it be for now, and move once all tests are passing.
wdyt?
cc @svghadi

svghadi (approved):
Thanks @saumeya for the suggestions. I am merging this PR, we can address your suggestions regarding subscription modification in tests in a separate PR. 

LGTM. Thanks @anandrkskd

PR #1383, opened 2024-05-27
fix: Add net policy rbac and remove auth from sentinel script 

svghadi:
**What type of PR is this?**

 /kind bug

Follow up PR for #1364 to fix redis issues


svghadi:
I have verified the HA test manually. We are good to merge it @iam-veeramalla

iam-veeramalla (approved):
Thanks @svghadi

PR #1396, opened 2024-06-04
Automatically download correct operator-sdk version to bin dir, if not found

jgwest:
**What type of PR is this?**
/kind chore

- Update `Makefile` to automatically downloaded the specific version of operator-sdk into `(repo root)/bin`
- Update `codegen.yaml` to automatically use the Go version defined in `go.mod`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,11 @@ BUNDLE_DEFAULT_CHANNEL := --default-channel=$(DEFAULT_CHANNEL)
 endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
+# Set the Operator SDK version to use.
+# This is useful for CI or a project to utilize a specific version of the operator-sdk toolkit.
+OPERATOR_SDK_VERSION ?= v1.11.0
+
+
 # IMAGE_TAG_BASE defines the docker.io namespace and part of the image name for remote images.
 # This variable is used to construct full image tags for bundle and catalog images.
 #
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -104,6 +109,32 @@ docker-build: test ## Build docker image with the manager.
...
 
+##@ Build Dependencies
+
+## Location to install dependencies to
+LOCALBIN ?= $(shell pwd)/bin
+$(LOCALBIN):
+	mkdir -p $(LOCALBIN)
+
+
+.PHONY: operator-sdk
```
jgwest, line 120: This block of script code is part of the standard Makefile generated by operator-sdk

Makefile
```diff
@@ -104,6 +109,32 @@ docker-build: test ## Build docker image with the manager.
...
+ifeq (,$(shell which operator-sdk 2>/dev/null))
+	@{ \
+	set -e ;\
+	mkdir -p $(dir $(OPERATOR_SDK)) ;\
+	OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
+	curl -sSLo $(OPERATOR_SDK) https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk_$${OS}_$${ARCH} ;\
+	chmod +x $(OPERATOR_SDK) ;\
+	}
+else
+OPERATOR_SDK = $(shell which operator-sdk)
```
svghadi, line 133: What if we directly use `operator-sdk` from `$PWD/bin/` instead of looking for it in the `PATH`? Recently, there has been some confusion about which `operator-sdk` version to use. Some developers have higher versions installed in their local `PATH`, resulting in differences in code generation. If we ignore the `operator-sdk` in the `PATH` and rely on the one in `$PWD/bin`, we could avoid this confusion. What do you think?

jgwest: @svghadi Done!

svghadi (approved):
LGTM. Thanks

jgwest:
Thanks @svghadi!

PR #1401, opened 2024-06-05
fix: don't change the default policy to reencrypt if the TLS secret is already present

chetan-rns:
**What type of PR is this?**

/kind bug

https://github.com/argoproj-labs/argocd-operator/pull/1363 changed the default termination policy from passthrough to reencrypt. However, there could be some users who have configured the old passthrough Route with a custom certificate before the upgrade. We don't want to overwrite their configuration once they upgrade the operator.

This PR introduces logic to update the Route to renencrypt only if the "argocd-server-tls` secret is not present.

Fixes #?

1. Install/Run an older version of the operator that still has Passthrough as default.
2. Verify that the Route is using the Passthrough policy. Configure a custom TLS secret "argocd-server-tls" using OpenSSL.
3. Run the operator with the changes in this PR
4. The route shouldn't be updated to reenecrypt.

### DIFF
--- a/controllers/argocd/route.go
+++ b/controllers/argocd/route.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	routev1 "github.com/openshift/api/route/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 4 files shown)


chetan-rns:
Story: https://issues.redhat.com/browse/GITOPS-4778

jgwest (approved):
LGTM, thanks @chetan-rns!

PR #1407, opened 2024-06-06
0.10.1 release

reginapizza:
**What type of PR is this?**
 /kind chore

**What does this PR do / why we need it**:
Release for operator v0.10.1 which also updates argocd version to v2.10.11, which includes a fix for a critical CVE

**Have you updated the necessary documentation?**

* [ ] Documentation update is required by this PR.
* [ ] Documentation has been updated.

**Which issue(s) this PR fixes**:

Fixes #?

**How to test changes / Special notes to the reviewer**:

PR #1418, opened 2024-06-14
cherry-pick  0.10.1 release to master branch (#1407)

reginapizza:
* update argocd version to 2.10.11
* operator v0.10.1

**What type of PR is this?**
 /kind chore

pulls in the changes from the 0.10.1 release back to master. Note: this does not update any argocd versions because the 0.10.1 argocd version is less than what master is currently at. 

Fixes #?


svghadi (approved):
Looks like code-gen job is expected to fail. LGTM. Thanks

reginapizza:
@svghadi I've run make bundle and updated the PR... even though the changes I had were correct based on the what was in the release-0.10 branch, when we sync back to master I forgot that we should be running make bundle again and keeping it as is (ie; without all the changes to control planes, etc), that way it doesn't mess up the CI for all other PR's

PR #970, opened 2023-08-21
[WIP] ci: Run CI e2e tests on operator installed via OLM

svghadi:
**What type of PR is this?**

/kind ci

**What does this PR do / why we need it**:

Current CI e2e tests are running on an operator ran using `make run` command. This works great for most of the testing. However, certain features like conversion webhook (#964) or operator metrics (#928) cannot be properly tested when the operator is running via `make run`.

This PR updates CI configurations to build and install operator via OLM.

Changes:
- Update GHA configuration to setup a k3d local image registry
- Add olm-install.sh script to install operator via OLM
- Enable e2e tests from tests/olm directory

PR #1419, opened 2024-06-14
feat: Add support for Aggregated ClusterRole for cluster scoped instances

jparsai:
**What type of PR is this?**
/kind enhancement

This PR enables user to use aggregated cluster roles for application controller in a cluster scoped instances of Argo CD. 
This will create a base cluster role (argocd-argocd-application-controlle) which can inherit permissions from these two other cluster roles.
- argocd-application-controller-view = It configures View permissions.
- gitops-application-controller-admin = It is also an  aggregated cluster role to configures Admin permissions. This can inherit permissions from cluster role that is created by users and add those into base cluster role. 

Fixes #?
This is tracked in Red Hat's issue database here: https://issues.redhat.com/browse/GITOPS-2615

- Create a cluster-scoped Argo CD instance
- Create Argo CD CR in namespace and set `Spec.aggregatedClusterRolesEnabled` to true.
- Create a user-defined cluster role having these two labels  `argocd/aggregate-to-admin: 'true'` and `app.kubernetes.io/managed-by: argocd` and configure permission you want to add for application controller component.

### DIFF
--- a/controllers/argocd/role.go
+++ b/controllers/argocd/role.go
@@ -9,6 +9,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -121,7 +122,7 @@ func (r *ReconcileArgoCD) reconcileRole(name string, policyRules []v1.PolicyRule
 		}
 		// only skip creation of dex and redisHa roles for namespaces that no argocd instance is deployed in
 		if len(list.Items) < 1 {
-			// namespace doesn't contain argocd instance, so skipe all the ArgoCD internal roles
+			// namespace doesn't contain argocd instance, so skip all the ArgoCD internal roles
 			if cr.ObjectMeta.Namespace != namespace.Name && (name != common.ArgoCDApplicationControllerComponent && name != common.ArgoCDServerComponent) {
 				continue
 			}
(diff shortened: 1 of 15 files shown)


controllers/argocd/role.go
```diff
@@ -343,3 +372,141 @@ func deleteClusterRoles(c client.Client, clusterRoleList *v1.ClusterRoleList) er
...
+		if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: GenerateUniqueResourceName(componentName, cr)}, existingClusterRole); err == nil {
+
+			// default ClusterRole exists, now delete it
+			if err := r.Client.Delete(context.TODO(), existingClusterRole); err != nil {
+				return true, fmt.Errorf("failed to delete existing cluster role for the service account associated with %s : %s", componentName, err)
+			}
+		}
+
+		// don't create a default ClusterRole and return
+		return true, nil
```
jgwest, line 392: In this function we are assuming that any error is a 'resource not found' error, which is not necessarily true.

We should explicitly check that the err is a not found error using `apierrors.IsNotFound`. For any other errors, we should return them.

controllers/argocd/role.go
```diff
@@ -279,59 +279,88 @@ func (r *ReconcileArgoCD) reconcileRoleForApplicationSourceNamespaces(name strin
...
+		configureAggregatedClusterRole(cr, expectedClusterRole, componentName)
+	} else {
+		// if current mode is default mode, but last one was aggregated mode, then delete ClusterRoles for View and Admin permissions
+		if componentName == common.ArgoCDApplicationControllerComponentView || componentName == common.ArgoCDApplicationControllerComponentAdmin {
+			if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: expectedClusterRole.Name}, expectedClusterRole); err == nil {
+				if err := r.Client.Delete(context.TODO(), expectedClusterRole); err != nil {
+					return nil, fmt.Errorf("failed to delete aggregated ClusterRoles: %s", expectedClusterRole.Name)
+				}
+				return nil, nil
+			}
```
jgwest, line 316: If err is non-nil, and error is NOT a 'resource not found' error (which we can check using apierrors.IsNotFound), we should return it.

jgwest (commented):
Looks great, thanks @jparsai! A couple more requests:
- Can you add some text to the documentation that described the feature, for example, proving examples of the roles that will be created and then how a user could add to them (I presume to here would be the best place for it: https://github.com/argoproj-labs/argocd-operator/blob/master/docs/usage/custom_roles.md)
    - We will probably also have to document this exact same thing for the OpenShift docs, so we will be able to reuse any text that we write here for that
- We have extensive unit tests, which is excellent! Would you be able to add a simple E2E (e.g. kuttl) tests for this, that ensures the roles are created correctly when the feature is enabled?

controllers/argocd/role.go
```diff
@@ -279,59 +279,88 @@ func (r *ReconcileArgoCD) reconcileRoleForApplicationSourceNamespaces(name strin
...
 	}
 
-	// Check if it is cluster-scoped instance namespace and user doesn't want to use default ClusterRole
-	if allowed && cr.Spec.DefaultClusterScopedRoleDisabled {
-
-		// In case DefaultClusterScopedRoleDisabled was false earlier and default ClusterRole was created, then delete it.
-		existingClusterRole := &v1.ClusterRole{}
-		if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: GenerateUniqueResourceName(name, cr)}, existingClusterRole); err == nil {
+	if allowed && cr.Spec.DefaultClusterScopedRoleDisabled && cr.Spec.AggregatedClusterRolesEnabled {
+		return nil, fmt.Errorf("Custom Cluster Roles and Aggregated Cluster Roles can not be used together.")
```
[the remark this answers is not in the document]
jgwest: For this one, I was thinking something like this, here:

```go
	if allowed && cr.Spec.DefaultClusterScopedRoleDisabled && cr.Spec.AggregatedClusterRolesEnabled {
		log.Error(fmt.Errorf("Unable to create ClusterRole: custom Cluster Roles and Aggregated Cluster Roles can not be used together in '%s' of '%s. Update ArgoCD CR to ensure that either .spec.defaultClusterScopedRoleDisabled or .spec.aggregatedClusterRoleEnabled is true, but NOT both at the same time.'", cr.Name, cr.Namespace), "")
		return nil, nil
	}
```

Unfortunately at the moment we don't have `.status.conditions` in ArgoCD CR, but I agree they would 100% be helpful.  The best we can do in this case is just log errors to the console, and expect the user to check there :shrug:. 

Now, since we are not returning an error, what this means is that the rest of the code will continue. So a  `ClusterRoleBinding` will be created, and point to a `ClusterRole` that doesn't exist, but I think this is fine (not perfect, but fine, Kubernetes is OK when this happenes).

jgwest (changes requested):
For E2E tests, I think you mentioned that we wouldn't be able to add them without [Siddesh's Operator Image E2E Test PR](https://github.com/argoproj-labs/argocd-operator/pull/970/files).  Can you provide more info on that? For this current PR, #1419, since we are not reading from a subscription, I would think we could at least add E2E test enabling/disabling `AggregatedClusterRoles` for a single `ArgoCD` instance, and perhaps disabling of cluster roles. (You're right we wouldn't be able to test with some of the other subscription-based features)

docs/usage/aggregated_roles.md
```diff
@@ -0,0 +1,122 @@
...
+
+Example: Configure permissions using aggregated cluster role model for application controller:
+
+**1 ).** Base cluster role
+
+```yaml
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: argocd-argocd-argocd-application-controller
```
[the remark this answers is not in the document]
jparsai: This is because names of cluster roles are generated using function [GenerateUniqueResourceName](https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/role.go#L50). It has naming convention `<Instance Name>-<Namespace>-<Component Name>` where component name is taken from variables defined [here](https://github.com/argoproj-labs/argocd-operator/blob/master/common/defaults.go#L19), so name generated by function is `argocd-argocd-argocd-application-controller`,

jparsai:
> For E2E tests, I think you mentioned that we wouldn't be able to add them without [Siddesh's Operator Image E2E Test PR](https://github.com/argoproj-labs/argocd-operator/pull/970/files). Can you provide more info on that? For this current PR, #1419, since we are not reading from a subscription, I would think we could at least add E2E test enabling/disabling `AggregatedClusterRoles` for a single `ArgoCD` instance, and perhaps disabling of cluster roles. (You're right we wouldn't be able to test with some of the other subscription-based features)

@jgwest both aggregated or custom cluster roles features are applicable only for cluster-scoped instances, which is defined via `Subscription` resource env variable `ARGOCD_CLUSTER_CONFIG_NAMESPACES`, so even we set `AggregatedClusterRoles` and `DefaultClusterScopedRoleDisabled` fields in ArgoCD resource, it wont have any effect because cluster roles will not be reconciled. That means even if we write E2E tests for these they will not be able to test feature.
https://github.com/jparsai/argocd-operator/blob/aggregated-clusterrole/controllers/argocd/role.go#L310C2-L310C48
https://github.com/jparsai/argocd-operator/blob/aggregated-clusterrole/controllers/argocd/role.go#L388C5-L388C12

jgwest:
@jparsai Thanks for the detailed reply @jparsai, this makes perfect sense! I forget how much of the operator functionality is tied to those variables :sweat_smile:.

deploy/olm-catalog/argocd-operator/0.10.0/argoproj.io_argocds.yaml
```diff
@@ -49,6 +49,10 @@ spec:
           spec:
             description: ArgoCDSpec defines the desired state of ArgoCD
             properties:
+              aggregatedClusterRoles:
```
jgwest, line 52: I think we can just remove the contents of this file entirely from the PR, unless you are convinced it is beneficial. I removed, the file, ran `make bundle`, and it was not generated, which to me says it is not needed (or should not be included). WDYT?

jgwest (changes requested):
LGTM, thanks @jparsai! One final question about one of the OLM files, let me know, then I will merge.

deploy/olm-catalog/argocd-operator/0.10.0/argoproj.io_argocds.yaml
jparsai: Yes @jgwest it was generated because of old version, it have removed it. PTAL.

jgwest (approved):
LGTM, thanks again @jparsai!

PR #1463, opened 2024-07-16
fix: tls cert name for applicationSet gitlab scm

saumeya:
**What type of PR is this?**

> /kind bug
> /kind chore
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

Fixes #?


controllers/argocd/util.go
```diff
@@ -1083,9 +1103,11 @@ func (r *ReconcileArgoCD) setResourceWatches(bldr *builder.Builder, clusterResou
...
 	bldr.Watches(&v1.ClusterRole{}, clusterResourceHandler)
 
-	bldr.Watches(&corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{
-		Name: common.ArgoCDAppSetGitlabSCMTLSCertsConfigMapName,
-	}}, appSetGitlabSCMTLSConfigMapHandler)
+	if applicationSetSCMTLSConfigMapName != "" {
+		bldr.Watches(&corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{
+			Name: applicationSetSCMTLSConfigMapName,
+		}}, appSetGitlabSCMTLSConfigMapHandler)
+	}
```
svghadi, line 1110: I think the watch will be setup only for initial value of `applicationSetSCMTLSConfigMapName` when the operator manager starts. Updates to `applicationSetSCMTLSConfigMapName` won't reset the watch with new value as the watches are set only at the start of the manager. 

What is the purpose of this watch? Is it to reflect changes in configmap to the appset-controller deployment or to point the appset-controller deployment to the new cm name whenever it changes in `ArgoCD` CR? 
IMO, reflecting configmap content changes should be handled in upstream Argo CD. I think it already does for `argocd-server` deployments where changes to `argocd-cm` configmap are automatically reflected without a deployment restart. However, I don't know if it handles tls content. 
The name change in ArgoCD CR doesn't necessarily require a separate watch, existing ArgoCD CR watch should be able to handle this update in configmap name change.

saumeya: As discussed, not updating the watch logic and updated the doc to use a specific config map name so that the watch can be specified.

svghadi:
Can we test a scenario where configmap content changes are automatically reflected to appset controller deployment? I think we need to update the configmap watch code to handle the restart of appset deployment for this.

saumeya:
> Can we test a scenario where configmap content changes are automatically reflected to appset controller deployment? I think we need to update the configmap watch code to handle the restart of appset deployment for this.

I checked and verified that the configmap content is updated in the appset controller.

docs/reference/argocd.md
```diff
@@ -136,8 +136,22 @@ metadata:
     example: applicationset
 spec:
   applicationSet:
-    SCMRootCAConfigMap: example-gitlab-scm-tls-cert
+    scmRootCAConfigMap: argocd-appset-gitlab-scm-tls-certs-cm
 ```
+Note: This is a sample config-map to use to mount your certificate. The key value should be 'cert' , this is used as the filename that is mounted. Other key names will not work, and this bug will be addressed later.
```
svghadi, line 141:
```suggestion
!!! important
    Please note that the key in the ConfigMap should be named 'cert', as this is used as the filename that is mounted. Other key names will not work due to an upstream bug that will be addressed later. Additionally, ensure that the ConfigMap is named argocd-appset-gitlab-scm-tls-certs-cm. Below is a sample ConfigMap that can be used to mount your certificate.
```

svghadi (approved):
LGTM. Thanks.

PR #1469, opened 2024-07-19
fix: race condition when the TLS secret is created before setting the default policy

chetan-rns:
**What type of PR is this?**

/kind bug

Currently, there is a race condition when users upgrade from v1.12.x to v1.13. This occurs when the TLS secret gets created before updating the default policy. Because of this, the Route stays in the Passthrough policy assuming that the user created this TLS secret. In this PR, we check if the TLS secret has the OpenShift service CA annotations to verify if the secret was created by the user or by the OpenShift Service CA. 

We remove the check to find the previous Route's policy since it can change if the user manually deletes the secret. Instead, we rely on the existence of the TLS secret to determine the previous policy.

Fixes https://issues.redhat.com/browse/GITOPS-4947

### DIFF
--- a/controllers/argocd/route.go
+++ b/controllers/argocd/route.go
@@ -236,10 +236,13 @@ func (r *ReconcileArgoCD) reconcileServerRoute(cr *argoproj.ArgoCD) error {
 			TargetPort: intstr.FromString("https"),
 		}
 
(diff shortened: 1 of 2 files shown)


jgwest (approved):
LGTM, thanks @chetan-rns!

PR #1474, opened 2024-07-24
Extra command args for argocd controller

varshab1210:
**What type of PR is this?**

/kind enhancement

Extra command args functionality for application controller deployment in operator

Fixes #?


jgwest (commented):
Could use some unit tests for this functionality, if these were not already in your plan for this PR.

Also, my two cents, it seems perhaps unnecessary to add support for args when we already support env vars which do the exact same thing.  The JIRA is a perfect example: There is no need to use ` --repo-error-grace-period-seconds` because `ARGOCD_REPO_ERROR_GRACE_PERIOD_SECONDS` is the same thing.  :shrug:

svghadi:
@jgwest - Are all Argo CD CLI flags supported via env vars? Is it a standard practice in the Argo CD community to have all CLI flags configurable via env var? If so, we could also consider deprecating the already supported `extraCommandArgs` for repo-server, applicationset-controller, and argocd-server in the `ArgoCD` CR.

jgwest:
@svghadi I don't know if it's officially documented anywhere, but it definitely seems like standard practice: the vast vast majority of cobra `command.Flags().(...)` statements (used to get values from params)  in the application controller/repo server/applicationset controller commands will  first read a default value from env var using `env.ParseNumFromEnv` or `env.ParseDurationFromEnv` or `env.StringFromEnv`.

cmd/argocd-repo-server/commands/argocd_repo_server.go
cmd/argocd-applicationset-controller/commands/applicationset_controller.go
cmd/argocd-application-controller/commands/argocd_application_controller.go

(the only except appears to be metrics/probe/webhook ports?)

svghadi:
Okay. I don't have any strong opinion about this. The only added advantage I see is from a UX point of view. I glanced at the Argo CD documentation and noticed that in many places, only the CLI argument is mentioned without its corresponding environment variable, so users would have to search around to find the appropriate environment variable name to set.

@iam-veeramalla , @anandf, @gnunn1 - What are your thoughts on deprecating `extraCommandArgs` field in `ArgoCD` CR as the args can be set via `env` field of the CR?

jgwest:
> I glanced at the Argo CD documentation and noticed that in many places, only the CLI argument is mentioned without its corresponding environment variable

Hmm, this is a good point! Perhaps I am thinking about it too much like an Argo CD dev who knows where to look this up in the code, and not enough like a user just following the upstream Argo CD docs :thinking:. Making it easy for users to follow upstream docs/blogs/guides would be a benefit.

gnunn1:
Not a strong opinion either however I do wonder if in the future a case ever arises where something uses a CLI argument but isn't exposed as an ENV variable if lost flexibility. 

The argument for discoverability is good as well, there is a command reference for the upstream controllers but I don't see anything for environment variables? Googling `ARGOCD_REPO_ERROR_GRACE_PERIOD_SECONDS` to use @jgwest example doesn't return much for me?

svghadi:
Thanks @jgwest  and @gnunn1 for your inputs. Based on our discussion, I believe the `extraCommandArgs` field provides value and we should continue supporting it.

tests/k8s/1-038_validate_controller_extra_command_args/01-basic-argocd-install.yaml
```diff
@@ -0,0 +1,15 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: test-1-38-controller-extra-command
+---
```
svghadi, line 6: I think there is no need to create a namespace. Let kuttl handle this.

svghadi, line 6: I think there is no need to create a namespace. Let kuttl handle this.

tests/k8s/1-038_validate_controller_extra_command_args/01-basic-argocd-install.yaml
```diff
@@ -0,0 +1,15 @@
...
+apiVersion: argoproj.io/v1alpha1
+kind: ArgoCD
+metadata:
+  name: example-argocd
+  namespace: test-1-38-controller-extra-command
```
svghadi, line 11: If kuttl is managing the namespace, there is no need to explicitly set it.

tests/k8s/1-038_validate_controller_extra_command_args/02-add-extraCommandArgs.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: argoproj.io/v1beta1
+kind: ArgoCD
+metadata:
+  name: example-argocd
+  namespace: test-1-38-controller-extra-command
+spec:
+  controller:
+    extraCommandArgs:
+    - --app-hard-resync
```
svghadi, line 9: Can we update the test to verify addition and removal of `extraCommandArgs`? Maybe move this to `01-basic-argocd-install.yaml` i.e create ArgoCD with `extraCommandArgs` and later remove it here.

svghadi (changes requested):
Thanks @varshab1210. Overall looks good. 

Left some comments on kuttl tests. Could you also address @jgwest comment regarding unit test? Maybe add a test here: https://github.com/argoproj-labs/argocd-operator/blob/554d8aa8c2831be3367a94973be357c99e25918a/controllers/argocd/util_test.go#L357

varshab1210:
Thanks for the valuable inputs @jgwest @svghadi @gnunn1!

tests/k8s/1-038_validate_controller_extra_command_args/01-basic-argocd-install.yaml
varshab1210: Hi Siddhesh, `redis` and `repo-server` options in the the controller StatefulSet need the namespace value. We cannot replace it with `$NAMESPACE`. Not sure if there is a way to escape and consider it as an environment variable rather than a string in the test? 

```
        - --redis
        - example-argocd-redis.test-1-38-controller-extra-command.svc.cluster.local:6379
        - --repo-server
        - example-argocd-repo-server.test-1-38-controller-extra-command.svc.cluster.local:8081
```

svghadi: Oh! Okay. Overlooked that. Let's create the namespace then.

controllers/argocd/util_test.go
```diff
@@ -474,6 +480,28 @@ func TestGetArgoApplicationControllerCommand(t *testing.T) {
...
+			"--kubectl-parallelism-limit",
+			"10",
+			"--loglevel",
+			"info",
+			"--logformat",
+			"text",
+			"--app-hard-resync",
+			"--app-resync",
+		}
+	}
```
svghadi, line 503:
```suggestion
	extraCommandArgsChangedResult := func(l []string) []string {
		return append(defaultResult, l...)
	}
```

controllers/argocd/util_test.go
```diff
@@ -564,6 +592,16 @@ func TestGetArgoApplicationControllerCommand(t *testing.T) {
 			[]argoCDOpt{logLevel("error")},
 			logLevelChangedResult("error"),
 		},
+		{
+			"configured extraCommandArgs",
+			[]argoCDOpt{extraCommandArgs([]string{"--app-hard-resync", "--app-resync"})},
+			extraCommandArgsChangedResult("--app-hard-resync, --app-resync"),
```
svghadi, line 598:
```suggestion
			extraCommandArgsChangedResult([]string{"--app-hard-resync", "--app-resync"}),
```

svghadi (commented):
Just a suggestion to simplify the test and avoid duplication.

svghadi:
Also please add the new field in controller table in docs: https://github.com/argoproj-labs/argocd-operator/blob/master/docs/reference/argocd.md#controller-options

svghadi (approved):
LGTM. 
Great work 🎉  @varshab1210. Thanks

jgwest (approved):
LGTM, thanks @varshab1210!

PR #1482, opened 2024-07-26
Check API registration if API is not available

anandf:
**What type of PR is this?**

/kind bug

When there is an issue with the API server pods, or the API server pods are just coming up after a restart, the current check for API availability done with a discovery client will fail. This makes the operator code to assume that the server does not support the Route API. To fix this, whenever the API availability check fails, the operator checks if the API is registered. If registered, then the operator assumes that the Route API is supported in the cluster and would eventually become available.

NA

Fixes https://issues.redhat.com/browse/GITOPS-4358

Steps to reproduce the problem in OCP cluster.
1. Delete the API server pods and restart the `gitops-operator` pod
```
oc delete pod --all --wait=false -n openshift-apiserver --all
oc delete pods --all -n openshift-gitops-operator
```
2. Now try logging in to the ArgoCD UI, you will see an error like below
```
oidc: issuer did not match the issuer returned by provider, expected "https://openshift-gitops-server-openshift-gitops.apps.<DNS_DOMAIN_NAME>/api/dex" got "https://openshift-gitops-server/api/dex"
```


controllers/argocd/util.go
```diff
@@ -348,7 +348,7 @@ func (r *ReconcileArgoCD) getArgoServerURI(cr *argoproj.ArgoCD) string {
 	}
 
 	// Use Route host if available, override Ingress if both exist
-	if IsRouteAPIAvailable() {
+	if cr.Spec.Server.Ingress.Enabled || cr.Spec.Server.Route.Enabled {
```
svghadi, line 351: Do you think we can remove the ingress check to avoid extra route check k8s call when route is not enabled?

anandf: Ok. I will remove the extra check. If ingress is enabled in OpenShift, a route is created with a random name. That is the reason why I had it in the check. But since the random name won't match anyway, it does not make sense to have it in this check. So removing it.

svghadi (approved):
LGTM. Thanks.

svghadi (approved):
Looks good. Thanks.

PR #1489, opened 2024-08-01
Rebase master with 0.11.0  changes

reginapizza:
**What type of PR is this?**
 /kind chore

- pulls in changes from release of 0.11.0
- fixes incomplete update of argocd upstream version to 2.11.6


build/util/Dockerfile
```diff
@@ -17,12 +17,18 @@ RUN groupadd -g $ARGOCD_USER_ID argocd && \
...
-# Install the AWS CLI
-RUN pip3 install awscli
+# Create a symlink for python to point to python3
+RUN ln -s /usr/bin/python3 /usr/bin/python
+
+# Install AWS CLI v2
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
+    unzip awscliv2.zip && \
+    ./aws/install && \
+    rm -rf awscliv2.zip aws
```
svghadi, line 31: Why is awscli required?

svghadi (approved):
Looks good. Just have a question regarding aws cli.

build/util/Dockerfile
reginapizza: we've had it from a while back, but I'm not exactly sure why it's required? I had to change it here by installing it directly instead of through pip3 because of a breaking change in pip3. I'll send the link for the convo we had about it to you on slack

Issue #1492, opened 2024-08-02
[v0.11.0] argocd-operator deployment fails with PodSecurity "restricted:latest", seccompProfile missing


hamidos:
**Describe the bug**

Deploying the argocd-operator fails in a Kubernetes cluster with the PodSecurity policy "restricted:latest" enabled. 
The error message indicates a violation due to the missing securityContext.seccompProfile.type in the manager container.

```
LAST SEEN   TYPE    REASON        OBJECT                                                   MESSAGE
4m35s       Warning FailedCreate  replicaset/argocd-operator-controller-manager-6855dff84f Error creating: pods "argocd-operator-controller-manager-6855dff84f-wmqrn" is forbidden: violates PodSecurity "restricted:latest": seccompProfile (pod or container "manager" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

**To Reproduce**
1. Apply Pod Security Standards restricted at [the Cluster Level](https://kubernetes.io/docs/tutorials/security/cluster-level-pss/) or at [ the Namespace](https://kubernetes.io/docs/tutorials/security/ns-level-pss/) where the operator is going to be installed
2. Follow the installation [documentation](https://argocd-operator.readthedocs.io/en/latest/install/olm/)

**Expected behavior**
argocd-operator should deploy successfully without violating the "restricted:latest" PodSecurity policy.

svghadi:
Did you encounter similar errors for any of argocd-operator component deployments like redis, dex, appset, etc?

hamidos:
Hello @svghadi , 
After further testing, the appset and dex deployments reported similar errors.  
Adding the `seccompProfile` to those Pods addressed the issue.

PR #1493, opened 2024-08-02
fix: add missing seccompProfile to comply with restricted policy

hamidos:
**What type of PR is this?**

 /kind bug

Add missing seccompProfile to comply with restricted policy

#1492


hamidos:
Hello @svghadi , Thank you for your feedback. 
I did fix the code-gen CI job, and also added the seccompProfile to the argocd components `applicationset`, `dex`, `notifications`, `redis` and `exporter job` .

svghadi:
Awesome. I will run some local tests to see if any other deployment needs seccompProfile.

svghadi:
Ran some tests, the change looks good. We are still missing `securityContext` for keycloak deployment, can we handle that in this PR?

```bash
$ kubectl label --overwrite --dry-run=server ns test pod-security.kubernetes.io/enforce=restricted
Warning: existing pods in namespace "test" violate the new PodSecurity enforce level "restricted:latest"
Warning: keycloak-6b8bbb786d-zrssz: allowPrivilegeEscalation != false, unrestricted capabilities, runAsNonRoot != true, seccompProfile
namespace/test labeled (server dry run)
```

The changes will probably need to be made in 2 places for keycloak: 
- https://github.com/argoproj-labs/argocd-operator/blob/f8b8fab31f9df9f1a47f40e11b1b09879fe19e21/controllers/argocd/keycloak.go#L199
- https://github.com/argoproj-labs/argocd-operator/blob/f8b8fab31f9df9f1a47f40e11b1b09879fe19e21/controllers/argocd/keycloak.go#L593

hamidos:
Hello @svghadi, 
I've added the security context for keycloak deployment.
It would be a good idea to enable restricted profile in the namespace while doing the E2E KUTTL tests.

svghadi:
Yes, sounds good. However, it seems that KUTTL doesn't support adding labels to the namespaces it automatically creates during tests, which happens for most of our tests. Maybe we could create a new test that manually creates a namespace with the required pod security labels, deploys an ArgoCD CR in it, and then runs the tests.

svghadi (approved):
LGTM. Thanks @hamidos for the contribution.

PR #1510, opened 2024-08-16
Check API registration if API is not available (cherry-pick)

anandf:
**What type of PR is this?**
 /kind bug

cherry pick of PR: https://github.com/argoproj-labs/argocd-operator/pull/1482

When there is an issue with the API server pods, or the API server pods are just coming up after a restart, the current check for API availability done with a discovery client will fail. This makes the operator code to assume that the server does not support the Route API. To fix this, whenever the API availability check fails, the operator checks if the API is registered. If registered, then the operator assumes that the Route API is supported in the cluster and would eventually become available.

Fixes https://issues.redhat.com/browse/GITOPS-4358

Steps to reproduce the problem in OCP cluster.

Delete the API server pods and restart the gitops-operator pod
```
oc delete pod --all --wait=false -n openshift-apiserver --all
oc delete pods --all -n openshift-gitops-operator
```
Now try logging in to the ArgoCD UI, you will see an error like below
```
oidc: issuer did not match the issuer returned by provider, expected "https://openshift-gitops-server-openshift-gitops.apps.<DNS_DOMAIN_NAME>/api/dex" got "https://openshift-gitops-server/api/dex"
```


svghadi (approved):
LGTM. Thanks.

PR #1513, opened 2024-08-17
upgrade argo-cd version to 2.12.1

ishitasequeira:
**What type of PR is this?**
/kind chore

upgrade argo-cd version to 2.12.1


bundle/manifests/argoproj.io_argocdexports.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.14.0
+    controller-gen.kubebuilder.io/version: v0.6.1
```
anandf, line 5: Can you run `make bundle` command with the correct version of `controller-gen` - `0.14.0`

anandf (changes requested):
Update controller-gen version to 0.14.0

saumeya (approved):
LGTM

svghadi (changes requested):
Dockerfile also needs an update, rest all looks good. Thanks.
https://github.com/argoproj-labs/argocd-operator/blob/6a5b0817a758bdc531369f7eb7c20a118640762a/build/util/Dockerfile#L1-L2

anandf (commented):
Are the YAML files in https://github.com/argoproj-labs/argocd-operator/tree/master/bundle/manifests updated to take the upstream ArgoCD v2.12.1 ?

svghadi:
Oh, yes. We also need to update the Argo CD Application, Project & ApplicationSet CRD in https://github.com/argoproj-labs/argocd-operator/tree/master/config/crd/bases. Just copy these files from upstream Argo CD source tag for new version.

anandf (commented):
LGTM

anandf (approved):

svghadi:
> Because when I replace the CRDs with 2.12.1 and execute make bundle, the CRDs revert back to it's current state. 

Interesting. I see some latest changes in Application and Appset CRD in upstream https://github.com/argoproj/argo-cd/tree/v2.12.1/manifests/crds which are not present in our copy of CRDs here: https://github.com/argoproj-labs/argocd-operator/tree/master/config/crd/bases .

svghadi (approved):
LGTM.. Thanks

go.mod
```diff
@@ -1,9 +1,11 @@
 module github.com/argoproj-labs/argocd-operator
 
-go 1.21
+go 1.21.0
+
+toolchain go1.21.9
```
anandf, line 5: I think we can remove this.

svghadi: Not sure how to do it. `go mod tidy` is adding it again even if I remove it and set to just `go 1.21`.

Invincibear:
Can we get this pushed to https://operatorhub.io/operator/argocd-operator?

Invincibear:
Is this ready for a new release so we can get this pushed to https://operatorhub.io/operator/argocd-operator?

svghadi:
Hi @Invincibear, it will still take a few weeks to get a release out. In the meantime, you can use the `.spec.image` and `.spec.version` fields in the `ArgoCD` CR to specify the new image to unblock yourself. Since you mentioned that there are some bug fixes in the new images that you're interested in, you likely won't run into any issues using the new image with the old operator, as the fixes, most of the times, don't require new Argo CD CRDs.

Invincibear:
@svghadi Thanks, that worked!

Issue #1517, opened 2024-08-22
Appset controller deployment doesn't respect `.spec.image` & `.spec.version` fields in ArgoCD CR


svghadi:
**Describe the bug**
Operator supports overriding default image for Argo CD components(server, repo-server, application-controller, applicationset-controller, notifications-controller) using `.spec.image` & `.spec.version` fields in ArgoCD CR. However, the `applicationset-controller` deployment doesn't respect this configuration. The image value is not reflected in the deployment. 

**To Reproduce**
Create ArgoCD CR
```yaml
apiVersion: argoproj.io/v1beta1
kind: ArgoCD
metadata:
  name: example
spec:
  image: quay.io/argoproj/argocd
  version: v2.12.0
  applicationSet:
    enabled: true
```

**Expected behavior**
The `applicationset-controller` deployment should use image specified in `.spec.image` & `.spec.version fields` from ArgoCD CR. 

**Actual behavior**
The image in `applicationset-controller` deployment  point to default value set by the operator.

**Additional context**
ApplicationSet controller image can be configured using `.spec.applicationSet.image` & `.spec.applicationSet.version` fields of the `ArgoCD` CR. I believe, the operator should proritize the image configuration in following order
> `default value` < `.spec.image` < `.spec.applicationSet.image`

PR #1523, opened 2024-08-27
fix: allow argocd spec.image to override default image for appset controller

anandrkskd:
**What type of PR is this?**

/kind bug

Current behaviour doesnt allow image to be overriden by `argocdSpec.image` for applicationset-controller pod, with this PR update application-set  controller image can be overide with priority set `default value < .spec.image < .spec.applicationSet.image`

Fixes https://github.com/argoproj-labs/argocd-operator/issues/1517

### DIFF
--- a/controllers/argocd/applicationset.go
+++ b/controllers/argocd/applicationset.go
@@ -758,28 +758,24 @@ func (r *ReconcileArgoCD) reconcileApplicationSetRoleBinding(cr *argoproj.ArgoCD
 }
 
 func getApplicationSetContainerImage(cr *argoproj.ArgoCD) string {
+
 	defaultImg, defaultTag := false, false
-
-	img := ""
-	tag := ""
-
-	// First pull from spec, if it exists
-	if cr.Spec.ApplicationSet.Image != "" && cr.Spec.ApplicationSet.Version != "" {
-		img = cr.Spec.ApplicationSet.Image
-		tag = cr.Spec.ApplicationSet.Version
-	} else if cr.Spec.Image != "" && cr.Spec.Version != "" && cr.Spec.ApplicationSet.Image == "" && cr.Spec.ApplicationSet.Version == "" {
-		img = cr.Spec.Image
-		tag = cr.Spec.Version
-	}
-
-	// If spec is empty, use the defaults
+	img := cr.Spec.ApplicationSet.Image
 	if img == "" {
-		img = common.ArgoCDDefaultArgoImage
-		defaultImg = true
+		img = cr.Spec.Image
+		if img == "" {
+			img = common.ArgoCDDefaultArgoImage
+			defaultImg = true
+		}
 	}
+
+	tag := cr.Spec.ApplicationSet.Version
 	if tag == "" {
-		tag = common.ArgoCDDefaultArgoVersion
-		defaultTag = true
+		tag = cr.Spec.Version
+		if tag == "" {
+			tag = common.ArgoCDDefaultArgoVersion
+			defaultTag = true
+		}
 	}
 
 	// If an env var is specified then use that, but don't override the spec values (if they are present)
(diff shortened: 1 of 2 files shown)


controllers/argocd/applicationset.go
```diff
@@ -764,9 +764,12 @@ func getApplicationSetContainerImage(cr *argoproj.ArgoCD) string {
 	tag := ""
 
 	// First pull from spec, if it exists
-	if cr.Spec.ApplicationSet != nil {
+	if cr.Spec.ApplicationSet.Image != "" && cr.Spec.ApplicationSet.Version != "" {
 		img = cr.Spec.ApplicationSet.Image
 		tag = cr.Spec.ApplicationSet.Version
+	} else if cr.Spec.Image != "" && cr.Spec.Version != "" && cr.Spec.ApplicationSet.Image == "" && cr.Spec.ApplicationSet.Version == "" {
+		img = cr.Spec.Image
+		tag = cr.Spec.Version
```
svghadi, line 772: Can this be simplified? Refer https://github.com/argoproj-labs/argocd-operator/blob/f8b8fab31f9df9f1a47f40e11b1b09879fe19e21/controllers/argocd/util.go#L222-L240

anandrkskd: From the above code it seems it can take image and tag from two different places, e.g. if only `cr.Spec.Repo.Image`  field is defined, it will take the image from the field but use the __default__ version if `cr.Spec.Repo.Version`  or `cr.Spec.Version` is not defined. 

I think that should not be allowed, as it might just fail.

One thing that I think is missing in my code is logging, I think it should log the error that image is not used if the version was missing. but that might just be over engineering.

svghadi: Yeah. Let's assume the user will handle cases where it fails due to an incorrect image. Usually, the deployment will fail with an appropriate `ImagePullBackOff` error to indicate this. We should also consider cases where the user is simply interested in trying out a different version of the image, so not having a strict requirement between the image and version fields would be helpful. WDYT?

anandrkskd: Yeah, make sense...

controllers/argocd/applicationset.go
```diff
@@ -758,25 +758,24 @@ func (r *ReconcileArgoCD) reconcileApplicationSetRoleBinding(cr *argoproj.ArgoCD
...
-
-	// First pull from spec, if it exists
-	if cr.Spec.ApplicationSet != nil {
-		img = cr.Spec.ApplicationSet.Image
-		tag = cr.Spec.ApplicationSet.Version
-	}
 
-	// If spec is empty, use the defaults
+	defaultImg, defaultTag := false, false
+	img := cr.Spec.ApplicationSet.Image
```
svghadi, line 763:
```suggestion
	img := ""
	if cr.Spec.ApplicationSet != nil {
	  img = cr.Spec.ApplicationSet.Image
	}
```
Just to be on safer side let's add a nil check. In `cr.Spec.ApplicationSet.Image`,  [ApplicationSet](https://github.com/argoproj-labs/argocd-operator/blob/f6d77ef623a2cbf4d5fe481bb891f21fad10e9ee/api/v1beta1/argocd_types.go#L743) is of type pointer, so it will  result into nil pointer exception if not set and accessed. 

Same for version aswell....

anandrkskd: I think it will be handled in [here](https://github.com/argoproj-labs/argocd-operator/blob/ec69f19b9fe3047c5a8109e30988051cb5a9dcf0/controllers/argocd/applicationset.go#L170-L175)
AFAIU, If appset is not enabled or no field is there under appset, it won't be executed

svghadi (approved):
LGTM. Thanks.

PR #1533, opened 2024-09-02
fix: Add update logic to handle addition of PSS to deployments

### LINKED ISSUE #1492: [v0.11.0] argocd-operator deployment fails with PodSecurity "restricted:latest", seccompProfile missing

svghadi:
**What type of PR is this?**

> /kind bug

[#1288](https://github.com/argoproj-labs/argocd-operator/pull/1288) and [#1493](https://github.com/argoproj-labs/argocd-operator/pull/1493) introduced Pod Security Standards (PSS) to ensure all Argo CD component pods comply with the restricted security policy. However, these changes only apply to new installations.

This PR addresses the gap by adding logic to update existing installations. It ensures that the necessary `securityContext` is applied to existing pods and automatically reconciles any manual changes back to the intended configuration.

Closes #1492 

I performed an upgrade test as follows:
1. Ran `make run` with the `v0.11.0` tag.
2. Created an `ArgoCD` CR.
3. Stopped `make run` and switched to this `PR branch`.
4. Ran `make run` again.

The following command should return no warnings:
```bash
# Example of a non-compliant namespace
$ kubectl label --overwrite --dry-run=server ns test pod-security.kubernetes.io/enforce=restricted
Warning: existing pods in namespace "test" violate the new PodSecurity enforce level "restricted:latest"
namespace/test labeled (server dry run)

# Example of a compliant namespace
$ kubectl label --overwrite --dry-run=server ns test pod-security.kubernetes.io/enforce=restricted
namespace/test labeled (server dry run)
```

### DIFF
--- a/controllers/argocd/keycloak.go
+++ b/controllers/argocd/keycloak.go
@@ -21,6 +21,7 @@ import (
 	json "encoding/json"
 	"fmt"
 	"os"
+	"reflect"
 
 	argoproj "github.com/argoproj-labs/argocd-operator/api/v1beta1"
 	"github.com/argoproj-labs/argocd-operator/common"
@@ -236,18 +237,7 @@ func getKeycloakContainer(cr *argoproj.ArgoCD) corev1.Container {
 			{ContainerPort: 8443, Name: "https", Protocol: "TCP"},
 			{ContainerPort: 8888, Name: "ping", Protocol: "TCP"},
 		},
-		SecurityContext: &corev1.SecurityContext{
-			Capabilities: &corev1.Capabilities{
-				Drop: []corev1.Capability{
-					"ALL",
-				},
-			},
-			AllowPrivilegeEscalation: boolPtr(false),
-			RunAsNonRoot:             boolPtr(true),
-			SeccompProfile: &corev1.SeccompProfile{
-				Type: "RuntimeDefault",
-			},
-		},
+		SecurityContext: restrictedContainerSecurityContext(),
 		ReadinessProbe: &corev1.Probe{
 			TimeoutSeconds:      240,
 			InitialDelaySeconds: 120,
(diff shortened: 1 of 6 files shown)


saumeya:
Good solution for upgrade testing @svghadi

controllers/argocd/keycloak.go
```diff
@@ -1515,3 +1514,18 @@ func (r *ReconcileArgoCD) reconcileKeycloak(cr *argoproj.ArgoCD) error {
 
 	return nil
 }
+
+func restrictedContainerSecurityContext() *corev1.SecurityContext {
```
saumeya, line 1518: Why are we creating a function for security context only for keycloak resource?

svghadi: Actually, I wanted to create a common function for all deployments but realized that some have slightly different `securityContext` configurations. So to keep things simpler, didn't proceed with it. We can revisit this when we work on refactoring again.

saumeya (approved):
Tested with the upgrade scenario from 0.11 to this branch and works as expected.
LGTM 
Thanks @svghadi

iam-veeramalla:
LGTM @svghadi, Thanks for the PR.

PR #1538, opened 2024-09-05
chore: improve pull request template to include kind lines in comments instead of quote

chengfang:
**What type of PR is this?**
/kind chore

Current PR template uses quote (>) to list all kind values, while it probably intend to use comments so submitters can choose the right PR kind by uncommenting a certain line.

I found the current template a bit confusing. I've seen some PR descriptions still contain all kind lines, or having the selecting kind line as quote.

With the fix in this PR, submitters just need to uncomment the right kind line, leave the rest as is, and the rendered description will only display the selected kind line, while keeping the other kind lines as hidden comments. Submitters can also choose to delete the 2 instruction lines and unused kind lines.

Fixes #?


svghadi (approved):
LGTM. Thanks @chengfang.

PR #1541, opened 2024-09-09
Fix issue in extraCommandArgs field

Rizwana777:
JIRA - https://issues.redhat.com/browse/GITOPS-5200

### DIFF
--- a/controllers/argocd/deployment.go
+++ b/controllers/argocd/deployment.go
@@ -345,46 +345,50 @@ func isMergable(extraArgs []string, cmd []string) error {
 
 // appendUniqueArgs appends extraArgs to cmd while ignoring any duplicate flags.
 func appendUniqueArgs(cmd []string, extraArgs []string) []string {
-	// Parse cmd into a map to track both flags and their associated values
-	existingArgs := make(map[string]string)
+	// Parse cmd into a map to track flags and their indices
+	existingArgs := make(map[string]int) // Map to track index of each flag in cmd
 	for i := 0; i < len(cmd); i++ {
 		arg := cmd[i]
 		if strings.HasPrefix(arg, "--") {
 			// Check if the next item is a value (not another flag)
 			if i+1 < len(cmd) && !strings.HasPrefix(cmd[i+1], "--") {
-				existingArgs[arg] = cmd[i+1] // Store flag-value pair
-				i++                          // Skip the value
+				existingArgs[arg] = i
+				i++ // Skip the value
 			} else {
-				existingArgs[arg] = "" // Store flag without value
+				existingArgs[arg] = i
 			}
 		}
 	}
 
-	// Iterate over extraArgs and append only if the flag and value do not already exist in cmd
+	// Iterate over extraArgs to append or override existing flags
 	for i := 0; i < len(extraArgs); i++ {
 		arg := extraArgs[i]
 		if strings.HasPrefix(arg, "--") {
-			// If this flag already exists in cmd, skip it
-			if _, exists := existingArgs[arg]; exists {
+			if index, exists := existingArgs[arg]; exists {
+				// If the flag exists, check if we need to override its value
 				if i+1 < len(extraArgs) && !strings.HasPrefix(extraArgs[i+1], "--") {
-					i++ // Skip the associated value
+					// If the flag has a value in extraArgs,update it in cmd
+					if index+1 < len(cmd) && !strings.HasPrefix(cmd[index+1], "--") {
+						cmd[index+1] = extraArgs[i+1] // Override the existing value
+					} else {
+						// Insert the new value if it didn't previously have one
+						cmd = append(cmd[:index+1], append([]string{extraArgs[i+1]}, cmd[index+1:]...)...)
+					}
+					i++ // Skip the value in extraArgs
 				}
-				continue
-			}
-
-			// Append the flag to cmd
-			cmd = append(cmd, arg)
-
-			// Check if this flag has an associated value
-			if i+1 < len(extraArgs) && !strings.HasPrefix(extraArgs[i+1], "--") {
-				cmd = append(cmd, extraArgs[i+1])
-				existingArgs[arg] = extraArgs[i+1] // Track flag-value pair
-				i++                                // Skip the value
 			} else {
-				existingArgs[arg] = "" // Track flag without a value
+				// Append the new flag and its value if not present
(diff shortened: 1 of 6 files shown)


controllers/argocd/deployment.go
```diff
@@ -350,6 +343,55 @@ func isMergable(extraArgs []string, cmd []string) error {
...
+	}
+
+	// Iterate over extraArgs and append only if the flag and value do not already exist in cmd
+	for i := 0; i < len(extraArgs); i++ {
+		arg := extraArgs[i]
+		if strings.HasPrefix(arg, "--") {
+			// If this flag already exists in cmd, skip it
+			if _, exists := existingArgs[arg]; exists {
+				if i+1 < len(extraArgs) && !strings.HasPrefix(extraArgs[i+1], "--") {
+					i++ // Skip the associated value
```
saumeya, line 370: What if an existing flag is added with a new value, will that also be skipped?

controllers/argocd/deployment_test.go
```diff
@@ -1317,6 +1317,34 @@ func TestArgoCDServerDeploymentCommand(t *testing.T) {
...
+	assert.NoError(t, r.reconcileServerDeployment(a, false))
+	assert.NoError(t, r.Client.Get(
+		context.TODO(),
+		types.NamespacedName{
+			Name:      "argocd-server",
+			Namespace: a.Namespace,
+		},
+		deployment))
+
+	// Non-duplicate argument "--newarg" should be added, duplicate "--redis" is ignored
```
saumeya, line 1344: This is probably a typo of --redis, I don't see that in the test case.

controllers/argocd/deployment.go
Rizwana777: yes, it is skipped

saumeya: Is that the expected behavior, what if the user wants to change the current flag value?

Rizwana777: based on the existing unit test https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/deployment_test.go#L1304  which is updating existing cmd i.e --redis command, I think this is expected behaviour, I might be wrong and open for suggestions

jopit (changes requested):
I don't see any logic to address the 3rd problem reported in the Jira issue
```
3. Adding argument and value as a single element causes the pod to crash
```
Other than that, it looks good to me.

Rizwana777:
> I don't see any logic to address the 3rd problem reported in the Jira issue
> 
> ```
> 3. Adding argument and value as a single element causes the pod to crash
> ```
> 
> Other than that, it looks good to me.

Thanks John, for 3rd point -  
```
Pod crashes as mentioned in description because ,

{.spec.server.extraCommandArgs}
'
["foo bar"] , here extraCommandArgs is of type []string and we need to pass the flag value i.e, foo as --foo and bar as value like this ["--foo" "bar"] and pod crashes with error : Error: unknown flag: --foo
Usage:
argocd-server [flags]
argocd-server [command], as --foo is a flag and we can use available flags
```

Rizwana777:
`FAIL: kuttl/harness/1-036_validate_role_rolebinding_for_source_namespace ` this e2e test is failing in other PRs as well

saumeya:
Re-ran the tests

controllers/argocd/deployment.go
saumeya: Ok, maybe we can confirm with others as well. Otherwise looks good.

Rizwana777:
> `FAIL: kuttl/harness/1-036_validate_role_rolebinding_for_source_namespace ` this e2e test is failing in other PRs as well

All the tests are passing now

controllers/argocd/deployment.go
svghadi: I think users should be allowed to override default flags.

svghadi (changes requested):
@Rizwana777 - Can we allow users to override params? There may be cases where tuning of default params could be useful.

controllers/argocd/deployment.go
```diff
@@ -350,6 +343,59 @@ func isMergable(extraArgs []string, cmd []string) error {
 	return nil
 }
 
+// appendUniqueArgs appends extraArgs to cmd while ignoring any duplicate flags.
+func appendUniqueArgs(cmd []string, extraArgs []string) []string {
+	// Parse cmd into a map to track flags and their indices
+	existingArgs := make(map[string]int) // Map to track index of each flag in cmd
```
svghadi, line 349: Based on our Slack discussion, could you try using a `map[string]string` to simplify the logic?

Rizwana777: @svghadi I tried updating the logic to use map[string]string but found issues as mentioned-
- map[string]string creates a map of flags to values, meaning it's expecting each flag to be associated with a value (or an empty string if no value is provided) this approach has the risk of treating arguments like --staticassets as executable commands if there is an issue in the flag parsing. If the function expects this argument to be passed as part of a command and it's incorrectly handled (i.e., not paired with a valid value), this could lead to the error like - create failed: unable to start container process: exec: \"--staticassets\": executable file not found in $PATH"

- On the other hand, map[string]int  keeps track of the exact positions of flags in the original cmd array, so any updates or additions to cmd are carefully done by index. This prevents the accidental misinterpretation of arguments and ensures the correct structure for the final command.

I may be mistaken, but PTAL and let me know your thoughts. I'm open to any suggestions

controllers/argocd/deployment.go
```diff
@@ -350,6 +343,59 @@ func isMergable(extraArgs []string, cmd []string) error {
 	return nil
 }
 
+// appendUniqueArgs appends extraArgs to cmd while ignoring any duplicate flags.
```
jgwest, line 346: I see this PR has similar unit tests for other functions, but can we also add a unit test for this function? 

The logic in this function is complex (given the large possible variation in inputs), so I'd like to ensure we have an exhaustive list of test cases that verify the possible cases we can think of that this function might be asked to handle.

jgwest (changes requested):
In addition, it appears that `ExtraRepoCommandArgs`, which controls extra repo server args,  is missing from this PR.

jgwest:
Test failure is unrelated to PR, I've opened another PR for the fix to that.

/retest

jgwest (approved):
LGTM, thanks @Rizwana777!

PR #1545, opened 2024-09-20
fix: delete/don't create deployment resources if using .remote field

anandrkskd:
**What type of PR is this?**

/kind bug

Before this PR, operator used to create resources even if remote redis/repo server were defined to be used in argocd CR, that caused unnecessary resource consumption.
This PR resolve it by checking if remote is defined in argocd CR, if defined then it doesn't create if absent or deletes if present.

Fixes #?


controllers/argocd/deployment.go
```diff
@@ -533,6 +533,9 @@ func (r *ReconcileArgoCD) reconcileRedisDeployment(cr *argoproj.ArgoCD, useTLS b
 			// Deployment exists but component enabled flag has been set to false, delete the Deployment
 			log.Info("Redis exists but should be disabled. Deleting existing redis.")
 			return r.Client.Delete(context.TODO(), deploy)
+		} else if cr.Spec.Redis.Remote != nil && *cr.Spec.Redis.Remote != "" {
```
svghadi, line 536: Could we define a new `func IsRemote()`, similar to `IsEnabled()`, for this check? This would enhance code readability.
https://github.com/argoproj-labs/argocd-operator/blob/603addacfb9d4c00e33d13bc561374b8a041366b/api/v1beta1/argocd_types.go#L464-L466

docs/usage/enabling-disabling-argocd-core-components.md
```diff
@@ -31,7 +31,7 @@ In this example, only the controller component is disabled, while all other comp
 
 # Specifying External URLs for Redis and RepoServer Components
 When disabling core components like Redis or Repo Server, you may wish to provide an external URL for components running in external clusters. The remote URL can be set using the `spec.<component>.remote` flag (where the component can only be `redis` or `repo`).
-
+Using `spec.<component>.remote` will skip the deployment.
```
svghadi, line 34:
```suggestion
Using `spec.<component>.remote` will skip the deployment & service creation. 
```
WDYT?

svghadi (changes requested):
Thanks @anandrkskd. Left few suggestions. Overall, looks good.

controllers/argocd/service.go
```diff
@@ -361,6 +369,10 @@ func (r *ReconcileArgoCD) reconcileRepoService(cr *argoproj.ArgoCD) error {
 		if ensureAutoTLSAnnotation(r.Client, svc, common.ArgoCDRepoServerTLSSecretName, cr.Spec.Repo.WantsAutoTLS()) {
 			return r.Client.Update(context.TODO(), svc)
 		}
+		if cr.Spec.Repo.IsRemote() {
+			log.Info("skip creating repo server service, repo remote is enabled")
+			return r.Client.Update(context.TODO(), svc)
```
svghadi, line 374: I think you missed changing update to delete?

controllers/argocd/service.go
```diff
@@ -388,6 +400,11 @@ func (r *ReconcileArgoCD) reconcileRepoService(cr *argoproj.ArgoCD) error {
 		},
 	}
 
+	if cr.Spec.Repo.IsEnabled() && cr.Spec.Repo.IsRemote() {
```
svghadi, line 403: How is the behavior? Is there any dependency between `remote` and `enabled` field i.e  component should be `enabled` for `remote` to work?

controllers/argocd/service.go
```diff
@@ -298,6 +301,11 @@ func (r *ReconcileArgoCD) reconcileRedisService(cr *argoproj.ArgoCD) error {
 		},
 	}
 
+	if cr.Spec.Redis.IsEnabled() && cr.Spec.Redis.IsRemote() {
```
svghadi, line 304: Same as above. Any dependency between this fields?

controllers/argocd/deployment.go
```diff
@@ -575,7 +578,7 @@ func (r *ReconcileArgoCD) reconcileRedisDeployment(cr *argoproj.ArgoCD, useTLS b
 		return nil // Deployment found with nothing to do, move along...
 	}
 
-	if cr.Spec.Redis.IsEnabled() && cr.Spec.Redis.Remote != nil && *cr.Spec.Redis.Remote != "" {
+	if cr.Spec.Redis.IsEnabled() && cr.Spec.Redis.IsRemote() {
```
svghadi, line 581: Same as previous comment.

controllers/argocd/deployment.go
```diff
@@ -1187,6 +1193,11 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argoproj.ArgoCD, useTLSFor
 		return nil // Deployment found with nothing to do, move along...
 	}
 
+	if cr.Spec.Redis.IsEnabled() && cr.Spec.Repo.IsRemote() {
```
svghadi, line 1196: Same as previous comment.

controllers/argocd/service.go
svghadi: Conversation from slack chat:
> skips remote too, if enabled is set to false

svghadi (approved):
LGTM. Thanks.

Issue #1548, opened 2024-09-25
Invalid image digest for OLM catalog deployment (release-0.12)


bendudz:
**Describe the bug**
When installing the ArgoCD Operator via OLM, the docker image digest that is required by the OLM install plan is invalid.

**To Reproduce**
Following [this](https://argocd-operator.readthedocs.io/en/latest/install/olm/) guide, assuming you have OLM setup in Kubernetes when adding a subscription to the `release-0.12` version of argocd-operator the image digest of the operator is incorrect. 

Install ArgoCD Operator into OLM:
```shell
kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj-labs/argocd-operator/refs/heads/release-0.12/deploy/catalog_source.yaml
kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj-labs/argocd-operator/refs/heads/release-0.12/deploy/operator_group.yaml
kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj-labs/argocd-operator/refs/heads/release-0.12/deploy/subscription.yaml
```

Wait for the install plan to start `kubectl get install plans -n argocd`.

A deployment will appear in the argocd namespace which can't create the pod due to 'invalid checksum digest length'.

The image its trying to create the pod with is: `quay.io/argoprojlabs/argocd-operator@sha256:405e7bde25affb7644c576a7f2d41ba30b88e2ef4f2f11945b920a33517895cee`

However the image available in quay [here](https://quay.io/repository/argoprojlabs/argocd-operator/manifest/sha256:405e7bde25affb7644c576a7f2d41ba30b88e2ef4f2f11945b920a33517895ce) is `quay.io/argoprojlabs/argocd-operator@sha256:405e7bde25affb7644c576a7f2d41ba30b88e2ef4f2f11945b920a33517895ce` notice the last 'e' is missing.

**Expected behavior**
I'd expect the ArgoCD Operator to be able to be deployed via OLM without error.

The digest causing the issue is hardcoded here: https://github.com/argoproj-labs/argocd-operator/blob/403e0863849cfb89782fb559dae7d4ff679a4357/deploy/olm-catalog/argocd-operator/0.12.0/argocd-operator.v0.12.0.clusterserviceversion.yaml#L1904

On the surface it's a simple typo to fix but I was unsure if this is caused by your build process otherwise I'd have offered a PR to fix. If you can advise I'll create a fix for it. Thanks

[...]

PR #1550, opened 2024-09-30
fix: correct the registry image

saumeya:
**What type of PR is this?**

https://github.com/argoproj-labs/argocd-operator/issues/1548 

Fixes #?


iam-veeramalla (approved):

svghadi (approved):
Thanks @saumeya. I think this should fix the issue. Only manual installs are impacted due to this bug. I could successfully install  released 0.12 on openshift and k8s from redhat marketplace and operatorhub.

PR #1551, opened 2024-10-01
fix: don't create SA for Admin and View aggregated ClusterRoles

jparsai:
**What type of PR is this?**
 /kind bug 

This PR is to avoid creation of Service Accounts for Admin and View aggregated ClusterRoles.


jgwest (approved):
LGTM

PR #1553, opened 2024-10-03
Fix: Add Server Ingress checks to apply the changes in case it  exists

### LINKED ISSUE #558: Changing host does not update ingress

nmirasch:
**What type of PR is this?**
 /kind bug 

This PR adds logic to reconcile Server Ingress related changes in case it exists previously.

Fixes #558
Fixes: https://issues.redhat.com/browse/GITOPS-5386

I performed an upgrade test as follows:

1. Ran make run
2. Created an ArgoCD CR.
3. Check the Server Service Type is ClusterIP (by default)
4. Do a kubectl apply on something like:

> ```yaml
> apiVersion: argoproj.io/v1alpha1
> kind: ArgoCD
> metadata:
>   name: argocd
> spec:
>   server:
>     ingress:
>       enabled: true

5. Followed by a kubectl apply on something like:

> ```yaml
> apiVersion: argoproj.io/v1alpha1
> kind: ArgoCD
> metadata:
>   name: argocd
> spec:
>   server:
>     host: foo.bar
>     ingress:
>       enabled: true

6. Check the ingress have been updated with the new host value when describing the ingress

### DIFF
--- a/controllers/argocd/ingress.go
+++ b/controllers/argocd/ingress.go
@@ -17,6 +17,7 @@ package argocd
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


svghadi (approved):
LGTM. Thanks @nmirasch.

PR #1554, opened 2024-10-03
feat: add option to enable argo-rollout ui extension

saumeya:
**What type of PR is this?**

 /kind enhancement 

Fixes #?

On a cluster install argo-rollouts - 
`kubectl create ns argo-rollouts`
`kubectl apply -n argo-rollouts -f https://github.com/argoproj/argo-rollouts/releases/latest/download/install.yaml`
`kubectl apply -n default -f https://github.com/argoproj/argo-rollouts/releases/latest/download/install.yaml`

In the ArgoCD CR enable rollout extensions 
```
spec: 
  server:
     enableRolloutUI: true
```
Then in the argoCD UI, create an application 

```
project: default
source:
  repoURL: https://github.com/saumeya/argocd-example-apps.git
  path: rollout-basic
  targetRevision: HEAD
destination:
  server: https://kubernetes.default.svc
  namespace: default

```
Git link - https://github.com/saumeya/argocd-example-apps.git 
Path - rollout-basic

After the application is synced, you can click on the **Rollout resource** in the tree structure of the resources and see the Rollout Extension enabled in the UI -
 
Screenshot 2024-10-02 at 12 43 40 AM

To disable it, set the option the `enableRolloutUI: false` . This restarts the server pod (takes a few seconds), also the UI browser needs to be reloaded, the plugin will disappear. 

~~Note - Disabling the field in the CR is not removing the rollout extension - I have created an issue upstream and debugging it - https://github.com/argoproj-labs/rollout-extension/issues/85~~


svghadi:
What do you think about below CRD structure?
Instead of 
```yaml
spec: 
  enableArgoRolloutUI: true
```
Something like
```yaml
spec: 
  rollouts:
    ui: true
```
I don't have lot of understanding on how argo-rollouts works and integrates with operator but could be beneficial if there are future plans to support rollouts via operator. Worth checking with @jgwest on this.

svghadi:
Sounds good to me.

controllers/argocd/deployment.go
```diff
@@ -1537,3 +1562,66 @@ func updateNodePlacement(existing *appsv1.Deployment, deploy *appsv1.Deployment,
...
+
+func getRolloutInitContainer() []corev1.Container {
+	return []corev1.Container{
+		{
+			Name:  "rollout-extension",
+			Image: common.ArgoCDRolloutExtensionImage,
+			Env: []corev1.EnvVar{
+				{
+					Name:  "EXTENSION_URL",
+					Value: "https://github.com/argoproj-labs/rollout-extension/releases/download/v0.3.5/extension.tar",
```
svghadi, line 1574: Can we move this string to `common/defaults.go`? Probably below `ArgoCDRolloutExtensionImage` var. It would be easy to update these values in future.

controllers/argocd/deployment.go
```diff
@@ -1537,3 +1562,66 @@ func updateNodePlacement(existing *appsv1.Deployment, deploy *appsv1.Deployment,
...
+			},
+			SecurityContext: &corev1.SecurityContext{
+				AllowPrivilegeEscalation: boolPtr(false),
+				Capabilities: &corev1.Capabilities{
+					Drop: []corev1.Capability{
+						"ALL",
+					},
+				},
+				RunAsNonRoot: boolPtr(true),
+				RunAsUser:    int64Ptr(999),
```
svghadi, line 1591: Is `999` user required? Does it fail without this?

svghadi (approved):
LGTM. Thanks.

api/v1alpha1/argocd_types.go
```diff
@@ -520,6 +520,9 @@ type ArgoCDServerSpec struct {
 	// Autoscale defines the autoscale options for the Argo CD Server component.
 	Autoscale ArgoCDServerAutoscaleSpec `json:"autoscale,omitempty"`
 
+	// EnableRolloutUI will add the argo rollout UI extension in ArgoCD Dashboard.
```
iam-veeramalla, line 523: Can we change this to `ArgoRollouts` from `ArgoRollout`

common/defaults.go
```diff
@@ -231,6 +231,12 @@ const (
 	// ArgoCDDefaultResourceInclusions is the default resource inclusions.
 	ArgoCDDefaultResourceInclusions = ""
 
+	// ArgoCDRolloutExtensionImage is the default image for ArgoCD Rollout Extension Installer
```
iam-veeramalla, line 234: The name of the controller is `Argo Rollouts`. So, we should change this to `Argo Rollouts`.

PR #1591, opened 2024-11-06
chore: Upgrade argocd version to 2.12.7

ishitasequeira:
**What type of PR is this?**
/kind chore

Upgrade ArgoCD version to 2.12.7

### DIFF
--- a/bundle/manifests/argocd-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/argocd-operator.clusterserviceversion.yaml
@@ -247,7 +247,7 @@ metadata:
     capabilities: Deep Insights
     categories: Integration & Delivery
     certified: "false"
-    createdAt: "2024-09-30T10:55:59Z"
+    createdAt: "2024-11-06T02:57:10Z"
     description: Argo CD is a declarative, GitOps continuous delivery tool for Kubernetes.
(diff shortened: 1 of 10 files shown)


bundle/manifests/argocd-operator.clusterserviceversion.yaml
```diff
@@ -1901,7 +1901,7 @@ spec:
                       fieldPath: metadata.annotations['olm.targetNamespaces']
                 - name: ENABLE_CONVERSION_WEBHOOK
                   value: "true"
-                image: quay.io/argoprojlabs/argocd-operator@sha256:405e7bde25affb7644c576a7f2d41ba30b88e2ef4f2f11945b920a33517895ce
+                image: quay.io/argoprojlabs/argocd-operator:v0.12.0
```
ishitasequeira, line 1904: running `make bundle` seems to be automatically replacing the sha with tag. Do we see any issues with this update?

svghadi: I think it should be fine. The CSV is updated with sha during operator release so shouldn't be a issue.

svghadi (approved):
LGTM. Thanks

PR #1593, opened 2024-11-06
chore: create release 0.12.2

ishitasequeira:
**What type of PR is this?**
/kind chore

create release 0.12.2


bundle/manifests/argocd-operator.clusterserviceversion.yaml
```diff
@@ -1901,7 +1901,7 @@ spec:
                       fieldPath: metadata.annotations['olm.targetNamespaces']
                 - name: ENABLE_CONVERSION_WEBHOOK
                   value: "true"
-                image: quay.io/argoprojlabs/argocd-operator:v0.12.0
+                image: quay.io/argoprojlabs/argocd-operator@sha256:f15781e2314d6baedcfc3bfa12181213d8d609803f73faecb2b8feddf3da4383
```
svghadi, line 1904: Looks like image is built for arm, should be amd64
```bash
$ docker inspect quay.io/argoprojlabs/argocd-operator@sha256:f15781e2314d6baedcfc3bfa12181213d8d609803f73faecb2b8feddf3da4383 |grep Arch
          "Architecture": "arm64",
```

svghadi (commented):
Are we planning a new upstream operator release? If it is just for downstream release, I think we can ignore this PR for now. Your previous changes (#1591) should be sufficient for downstream.
We can make an exception this time i.e not release a z-stream upstream. I am working on a release automation so in future we should be able to do quick z-stream release of upstream operator.

bundle/manifests/argocd-operator.clusterserviceversion.yaml
anandf: Can we build multi-arch images ?

anandf (changes requested):
Please update the argocd-operator image either as multi-arch or amd64 supported. Other changes look good.

bundle/manifests/argocd-operator.clusterserviceversion.yaml
ishitasequeira: Do we have a make command to build a multi-arch image? I did not configure any arch-type while building the image. I think it took my default laptop arch.

ishitasequeira:
> Are we planning a new upstream operator release?

I think the fix that was merged should have a specific argocd-operator release as well for the feature to work if used only via argocd-operator. Although I am okay to skip publishing the image if needed.

bundle/manifests/argocd-operator.clusterserviceversion.yaml
svghadi: Right now, no. You need to modify image build commands and add `--platform linux/amd64` flag.
```diff
diff --git a/Makefile b/Makefile
index 45c02a62..634356bb 100644
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
        REDIS_CONFIG_PATH="build/redis" go run -ldflags=$(LD_FLAGS) ./cmd/main.go
 
 docker-build: test ## Build docker image with the manager.
-       $(CONTAINER_RUNTIME) build --build-arg LD_FLAGS=$(LD_FLAGS) -t ${IMG} .
+       $(CONTAINER_RUNTIME) build --platform linux/amd64 --build-arg LD_FLAGS=$(LD_FLAGS) -t ${IMG} .
 
 docker-push: ## Push docker image with the manager.
        $(CONTAINER_RUNTIME) push ${IMG}
```

ishitasequeira: Thanks @svghadi !! Will build the image

ishitasequeira: The image is updated on quay to be of type `amd64`.

svghadi (approved):
LGTM. Thanks

olivergondza:
Closing as outdated.

PR #1594, opened 2024-11-06
chore: Update golang to 1.22

svghadi:
**What type of PR is this?**

/kind cleanup

Argo CD 2.13 shipped with latest version of operator is built with golang 1.22


iam-veeramalla (approved):
Thanks @svghadi

Issue #923, opened 2023-06-01
Reuse argocd repo server image with config management plugins


jimmyjones2:
**Is your feature request related to a problem? Please describe.**
I'm porting a config management plugin from the [old ConfigMap](https://argocd-operator.readthedocs.io/en/latest/reference/argocd/#config-management-plugins) method. This used the existing repo server image as it had all the necessary tools - helm etc. However with the [new method](https://argocd-operator.readthedocs.io/en/latest/usage/config_management_2.0/) of creating config management plugins, I'll need to create or find an image with the necessary tools, and keep it updated.

**Describe the solution you'd like**
I'd like my plugin to be able to use the existing repo server image associated with the version of ArgoCD it's being deployed into. For example if `spec.repo.sidecarContainers[].image` is null, the repo server image is used. When the operator / ArgoCD is updated, the image will be updated.

**Describe alternatives you've considered**
* Hardcode the image to the current argocd repo server image - but won't get upgraded when ArgoCD is, and image address depends on if deployed in Openshift or in vanilla Kubernetes
* Build a new image from scratch -  but then have to take responsibility of keeping it updated

svghadi:
Implemented in #1604

PR #1604, opened 2024-11-19
feat: make specifying config-management sidecar plugin image optional

jopit:
**What type of PR is this?**

/kind enhancement

Make specifying the sidecar image for a config-management plugin optional. When the image is not specified the operator will use the product image being used for the repo server.

### DIFF
--- a/api/v1alpha1/argocd_types.go
+++ b/api/v1alpha1/argocd_types.go
@@ -468,7 +468,9 @@ type ArgoCDRepoSpec struct {
 	// InitContainers defines the list of initialization containers for the repo server deployment
 	InitContainers []corev1.Container `json:"initContainers,omitempty"`
 
-	// SidecarContainers defines the list of sidecar containers for the repo server deployment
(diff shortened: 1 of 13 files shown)


jgwest (changes requested):
Looks great, thanks @jopit!

Two quick requests: 
- Can you add a short description of the new behaviour to `docs/usage/config_management_2.0.md`? This looks like the best place to document the change in the code itself.
- Can you likewise add a short description of the behaviour as a comment to `SidecarContainers` field in https://github.com/argoproj-labs/argocd-operator/blob/9174e68766ed748642e0c3207043709a75f7a3a7/api/v1beta1/argocd_types.go#L113 (and also the v1alpha1 API).

I presume we are ALSO documenting this downstream, but it couldn't hurt to include it here too.

svghadi (approved):
LGTM.

svghadi:
Fixes #923

jgwest (approved):
LGTM too, thanks again @jopit!

PR #1605, opened 2024-11-20
Add More Banner Options to Argocd CR

neevnuv:
**What type of PR is this?**
/kind enhancement

It add the banner permanent and banner position options to the argocd cr

This PR adds an enhancement that allows easily to configure a banner's position and permanence 

### DIFF
--- a/api/v1alpha1/argocd_types.go
+++ b/api/v1alpha1/argocd_types.go
@@ -929,6 +929,10 @@ type Banner struct {
 	Content string `json:"content"`
 	// URL defines an optional URL to be used as banner message link
 	URL string `json:"url,omitempty"`
+	// Permanent defines if the banner is sticky (permanent) using a boolean
+	Permanent bool `json:"permanent,omitempty"`
+	// Position defines the banner's position - top, bottom or both
+	Position string `json:"position,omitempty"`
 }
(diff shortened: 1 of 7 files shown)


svghadi:
Hi @neevnuv , operator supports `.spec.extraConfig` field which can be used for add extra configs in `argocd-cm` configmap. Do you think that could be an option for you? 
https://argocd-operator.readthedocs.io/en/latest/reference/argocd/#extra-config 

We are trying to reduce `ArgoCD` CRD size so we might deprecate & remove some of the less used fields like `.banner` in future.  Let me know if you find `.spec.banner` field useful. You can achieve the same result by using `.spec.extraConfig`.

Example:
```yaml
kind: ArgoCD
metadata:
  name: example-argocd
spec:
  extraConfig:
    "ui.bannercontent": "value"
    "ui.bannerurl": "value"
    "ui.bannerpermanent": "value"
    "ui.bannerposition": "value"
```

neevnuv:
Hello @svghadi,
Our team uses the .spec.banner option in the Argo cr, but the extraConfig option does also work for us, so we’ll be okay either way.
You may close this issue if the banner field is planned to be deprecated in the future.

PR #1532, opened 2024-08-30
feat: Added custom label and annotations to the pods of the server, repo, application set controller, and application controller

enriqueav99:
**What type of PR is this?**

> /kind enhancement

**What does this PR do / why we need it**:

Add the feature to add custom labels and annotations  to the pods of the server, repo, application set controller, and application controller.

**Have you updated the necessary documentation?**

* [x] Documentation update is required by this PR.
* [x] Documentation has been updated.

**Which issue(s) this PR fixes**:

Fixes #1530 #637

**How to test changes / Special notes to the reviewer**:
Apply the following yaml
```
apiVersion: argoproj.io/v1beta1
kind: ArgoCD
metadata:
  name: argocd-sample
spec:
  server:
    customPodLabels:
      custom: label
      custom2: server
    customPodAnnotations:
      custom: annotation
      custom2: server
  repo:
    customPodLabels:
      custom: label
      custom2: repo
    customPodAnnotations:
      custom: annotation
      custom2: repo
  controller:
    customPodLabels:
      custom: label
      custom2: controller
    customPodAnnotations:
      custom: annotation
      custom2: controller
  applicationSet:
    customPodLabels:
      custom: label
      custom2: applicationSet
    customPodAnnotations:
      custom: annotation
      custom2: applicationSet
```

PR #1616, opened 2024-12-05
fix: update logic to preserve k8s specific labels & annotations

svghadi:
**What type of PR is this?**

/kind bug 

Operator overwrites k8s added labels or annotations. This sometimes results into unexpected behaviour. Eg. A kubectl rollout restart on deployment adds an annotation "kubectl.kubernetes.io/restartedAt" with timestamp. However operator detects this as a config drift and removes the annotations resulting into k8s terminating the rollout due to change in config. This commit fixes the issue by preserving such annotations & labels added onto live object during comparison.

**Special notes to the reviewer:**
This is a regression introduced by #1532

### DIFF
--- a/controllers/argocd/util.go
+++ b/controllers/argocd/util.go
@@ -1705,12 +1705,7 @@ func addKubernetesData(source map[string]string, live map[string]string) {
 	for key, value := range live {
 		found := glob.MatchStringInList(patterns, key, glob.GLOB)
 		if found {
-			// Don't override values already present in the source object.
-			// This ensures users have control over Kubernetes-managed data
-			// if they have intentionally set or modified these values in the source object.
-			if _, ok := source[key]; !ok {
-				source[key] = value
-			}
+			source[key] = value
 		}
 	}
 }
--- a/controllers/argocd/util.go
+++ b/controllers/argocd/util.go
@@ -1693,7 +1693,7 @@ func UseServer(name string, cr *argoproj.ArgoCD) bool {
 // in the live object and updates the source object to ensure critical metadata
 // (like scheduling, topology, or lifecycle information) is retained.
 // This helps avoid loss of important Kubernetes-managed metadata during updates.
-func addKubernetesData(source map[string]string, live map[string]string) map[string]string {
+func addKubernetesData(source map[string]string, live map[string]string) {
 
 	// List of Kubernetes-specific substrings (wildcard match)
 	patterns := []string{
@@ -1713,6 +1713,4 @@ func addKubernetesData(source map[string]string, live map[string]string) map[str
 			}
 		}
 	}
-
-	return source
 }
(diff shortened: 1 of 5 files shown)


controllers/argocd/util.go
```diff
@@ -1688,3 +1688,31 @@ func UseServer(name string, cr *argoproj.ArgoCD) bool {
...
+	for key, value := range live {
+		found := glob.MatchStringInList(patterns, key, glob.GLOB)
+		if found {
+			// Don't override values already present in the source object.
+			// This ensures users have control over Kubernetes-managed data
+			// if they have intentionally set or modified these values in the source object.
+			if _, ok := source[key]; !ok {
+				source[key] = value
+			}
+		}
```
ishitasequeira, line 1714: QQ: Will live object have the latest data for k8s annotations/labels? If yes, should we directly update the value from live object into source instead of checking it to be present?

anandf (approved):
LGTM

controllers/argocd/util.go
```diff
@@ -1688,3 +1688,31 @@ func UseServer(name string, cr *argoproj.ArgoCD) bool {
 	}
 	return true
 }
+
+// addKubernetesData checks for any Kubernetes-specific labels or annotations
+// in the live object and updates the source object to ensure critical metadata
+// (like scheduling, topology, or lifecycle information) is retained.
+// This helps avoid loss of important Kubernetes-managed metadata during updates.
+func addKubernetesData(source map[string]string, live map[string]string) map[string]string {
```
anandf, line 1696: Do we have to return the source map ? the map is already modified inplace in this function isn't it?

controllers/argocd/util.go
svghadi: Live will always have latest data for k8s annotations/labels. The check is to ensure, users can update the value from ArgoCD CR if required. Example: by setting
```yaml
kind: ArgoCD
spec:
  server:
    annotations:
      node.kubernetes.io/pod: new-value
``` 

Without this check users cannot modify kuberentes/openshift label/annotations if they wish to.

ishitasequeira: Would kubernetes not overwrite those values again given that the values are generated by kubernetes?

controllers/argocd/util.go
svghadi: Forgot golang basics, map are passed by reference😅. Thanks. Refactored the logic.

ishitasequeira (approved):

controllers/argocd/util.go
svghadi: Okay. When I think of it again, I feel the logic is required because operator also sets some kubernetes.io & openshift.io labels & annotations. If the logic is not present, any update operation on these labels/annots from operator won't reconcile. They will always take live value which will be the old one.

ishitasequeira: In that case, we can continue with the logic you had added.

ishitasequeira (approved):
LGTM!!

PR #1626, opened 2024-12-18
fix: broken check for context in notificationController

anandrkskd:
**What type of PR is this?**

/kind bug

This PR fix the broken check for context fields in notification configmap and notificationConfiguration CR.

Fixes [GITOPS-5970](https://issues.redhat.com/browse/GITOPS-5970)

- run controller on local using `make install run`
- create an argo-cd instance with notification controller enabled
- create a notificationConfiguraion with single context and check for notification-controller pod logs
- update the context field in the notificationConfiguration with more context and check for notification-controller pod logs, there should not be indefinitely repeated logs similar to `time="2024-11-21T09:40:54Z" level=info msg="invalidated cache for resource in namespace: argocd-debug with the name: argocd-notifications-cm"` .

### DIFF
--- a/controllers/notificationsconfiguration/configmap.go
+++ b/controllers/notificationsconfiguration/configmap.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


controllers/notificationsconfiguration/configmap.go
```diff
@@ -76,10 +86,34 @@ func (r *NotificationsConfigurationReconciler) reconcileNotificationsConfigmap(c
...
 	result := ""
 	for key, value := range m {
 		result += fmt.Sprintf("%s: %s\n", key, value)
 	}
 	return result
 }
+
+// checkIfContextEquals checks if context value in NotificationConfiguration and notificationConfigMap context have same value
+// return true if there is difference, and false if no changes observed
+func checkIfContextEquals(cr *v1alpha1.NotificationsConfiguration, notificationConfigMap *corev1.ConfigMap) bool {
```
svghadi, line 100: Can we add some unit test for this function?

anandrkskd: added unit test

controllers/notificationsconfiguration/configmap_test.go
```diff
@@ -226,3 +226,38 @@ func TestReconcileNotifications_DeleteConfigMap(t *testing.T) {
...
+			"key6": "value6"},
+	}
+
+	var testmap = []struct {
+		testcase string
+		cm       corev1.ConfigMap
+		result   bool
+	}{
+		{"equal context",
+			corev1.ConfigMap{Data: map[string]string{"context": "key1: value1\nkey2: value2\nkey4: value4\nkey3: value3\nkey6: value6\n"}},
```
svghadi, line 246:
```suggestion
			corev1.ConfigMap{Data: map[string]string{"context": "key4: value4\nkey2: value2\nkey1: value1\nkey3: value3\nkey3: value3\n"}},
```
Can we reorder the string or add a new test for the same to verify context is equal even if keys & values in the configmap are reordered.

svghadi (commented):
Left a small suggestion, rest all looks good. Thanks

PR #1633, opened 2024-12-30
fix: Stabilize on-deployed notification trigger

svghadi:
**What type of PR is this?**

/kind bug

Fix false positives for on-deployed notifications trigger. Refer https://github.com/argoproj/argo-cd/pull/21333 for more details. 

The fix cannot be backported as it depends on a new upstream feature.

### DIFF
--- a/controllers/argocd/notifications_util.go
+++ b/controllers/argocd/notifications_util.go
@@ -526,8 +526,8 @@ func getDefaultNotificationsTriggers() map[string]string {
   oncePer: app.status.operationState.syncResult.revision
   send:
   - app-deployed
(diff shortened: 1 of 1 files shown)


iam-veeramalla (approved):
Thanks @svghadi, Can we add a test for this ?

svghadi:
It will be very challenging to add a test. It requires extra env setup and also the bug itself is flaky so it will be tricky to write a test. I have tested it manually. The test results are attached in upstream PR.

svghadi:
As discussed on slack, merging this as it is difficult to write tests for this fix.

PR #1637, opened 2025-01-08
feat: Support multiple ArgoCD instances with ApplicationTrackingAnnotations

aali309:
**What type of PR is this?**

 add ApplicationTrackingAnnotations to prevent resource conflicts between ArgoCD instances

see [this PR on argocd repo](https://github.com/argoproj/argo-cd/pull/20222)
Fixes #?

### DIFF
--- a/controllers/argocd/configmap.go
+++ b/controllers/argocd/configmap.go
@@ -780,12 +780,3 @@ func (r *ReconcileArgoCD) reconcileGPGKeysConfigMap(cr *argoproj.ArgoCD) error {
 	argoutil.LogResourceCreation(log, cm)
 	return r.Client.Create(context.TODO(), cm)
 }
-
-func validateApplicationTrackingAnnotations(annotations map[string]string) error {
-	if id, exists := annotations["installationID"]; exists {
-		if id == "" {
-			return fmt.Errorf("installationID cannot be empty")
-		}
-	}
-	return nil
-}
--- a/controllers/argocd/configmap.go
+++ b/controllers/argocd/configmap.go
@@ -16,7 +16,6 @@ package argocd
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"reflect"
 
@@ -389,13 +388,11 @@ func (r *ReconcileArgoCD) reconcileArgoConfigMap(cr *argoproj.ArgoCD) error {
 		cm.Data["resource.tracking.method"] = cr.Spec.ResourceTrackingMethod
 	}
 
-	// Set tracking annotations and handle installationID
+	// Set tracking annotations directly in the ConfigMap
 	if cr.Spec.ApplicationTrackingAnnotations != nil {
-		trackingAnnotations, err := json.Marshal(cr.Spec.ApplicationTrackingAnnotations)
-		if err != nil {
-			return err
+		for key, value := range cr.Spec.ApplicationTrackingAnnotations {
+			cm.Data[key] = value
 		}
-		cm.Data["resource.tracking.annotations"] = string(trackingAnnotations)
 	}
 
 	cm.Data[common.ArgoCDKeyConfigManagementPlugins] = getConfigManagementPlugins(cr)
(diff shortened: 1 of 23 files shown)


ishitasequeira (commented):
Sharing the feedback here from our chat for others to be able to advise further as well,

* We do not want to change the name of the config map being created. ArgoCD looks for a specific configMap and if it does not find the configMap, it defaults everything to default values.
* There is a limitation in Operator where we can have only 1 ArgoCD instance per namespace. I can see that you are creating both the ArgoCD instances in the same namespace [here](https://github.com/argoproj-labs/argocd-operator/pull/1637/files#diff-a7c8c403951eadffe1145a0fae0c362cfb22b6a89049c4ac2cc1d15864e9159aR19-R43). Once you have them created in different namespaces, you would not need to make changes of creating a different configMap [here](https://github.com/argoproj-labs/argocd-operator/pull/1637/files#diff-e71f34a3095d930774ccee95c7e821da357ee55a51b29b7ddaed885f16344622R382-R384).

jannfis (commented):
I've got two questions/comments

controllers/argocd/configmap.go
```diff
@@ -379,10 +379,22 @@ func (r *ReconcileArgoCD) reconcileCAConfigMap(cr *argoproj.ArgoCD) error {
...
 	cm := newConfigMapWithName(common.ArgoCDConfigMapName, cr)
-
 	cm.Data = make(map[string]string)
 	cm.Data = setRespectRBAC(cr, cm.Data)
 	cm.Data[common.ArgoCDKeyApplicationInstanceLabelKey] = getApplicationInstanceLabelKey(cr)
+
+	// Set tracking method if specified
+	if cr.Spec.ResourceTrackingMethod != "" {
+		cm.Data["resource.tracking.method"] = cr.Spec.ResourceTrackingMethod
+	}
```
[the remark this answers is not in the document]
aali309: You're right on both points. I've removed this redundant code block since the resource tracking method is already being set using:
`cm.Data[common.ArgoCDKeyResourceTrackingMethod] = getResourceTrackingMethod(cr)`

jannfis (approved):
LGTM. @iam-veeramalla or @svghadi , you might want to take a look as well.

tests/k8s/1-046_validate_application_tracking/02-label-namespaces.yaml
```diff
@@ -0,0 +1,13 @@
...
+# Set managed-by labels
+- script: |
+    kubectl label ns source-ns-1 argocd.argoproj.io/managed-by=test-1-046-argocd-1
+    kubectl label ns source-ns-2 argocd.argoproj.io/managed-by=test-1-046-argocd-2
+    sleep 30
+# Set tracking-id labels
+- script: |
+    kubectl label ns source-ns-1 argocd.argoproj.io/tracking-id=instance-1
+    kubectl label ns source-ns-2 argocd.argoproj.io/tracking-id=instance-2
+    sleep 30
```
svghadi, line 13: Why do we manually label the namespaces with tracking label? I think Argo CD automatically adds this label to the resources it manages.

svghadi (commented):
I have few questions on kuttl tests. Rest all looks good. Thanks

tests/k8s/1-046_validate_application_tracking/01-install.yaml
```diff
@@ -0,0 +1,39 @@
...
+---
+# First ArgoCD instance
+apiVersion: argoproj.io/v1beta1
+kind: ArgoCD
+metadata:
+  name: argocd-1
+  namespace: test-1-046-argocd-1
+spec:
+  installationID: "instance-1"
+  resourceTrackingMethod: "annotation"
```
svghadi, line 29: Is this feature supported only for `annotation` tracking or for `label` and `annotation+label` tracking aswell?

aali309: The installationID feature works with all tracking methods. I've updated the test to use annotation+label to demonstrate this:
```
# In 01-install.yaml
spec:
  installationID: "instance-1"
  resourceTrackingMethod: "annotation+label"  # Works with "annotation", "label", and "annotation+label"
  ```

tests/k8s/1-046_validate_application_tracking/02-label-namespaces.yaml
aali309: `[02-label-namespaces.yaml]` can be removed

svghadi (approved):
LGTM. Thanks.

PR #1642, opened 2025-01-21
Add error condition when notification replica is greater than 1

Rizwana777:
JIRA - https://issues.redhat.com/browse/GITOPS-5967

### DIFF
--- a/controllers/argocd/argocd_controller.go
+++ b/controllers/argocd/argocd_controller.go
@@ -130,6 +130,13 @@ func (r *ReconcileArgoCD) internalReconcile(ctx context.Context, request ctrl.Re
 		return reconcile.Result{}, argocd, err
 	}
 
+	// If the number of notification replicas is greater than 1, return an error
+	if argocd.Spec.Notifications.Replicas != nil && *argocd.Spec.Notifications.Replicas > 1 {
+		message := "Argo CD Notification controller does not support multiple replicas. Notification replicas cannot be greater than 1."
+		reqLogger.Info(message)
+		return reconcile.Result{}, argocd, fmt.Errorf("%s", message)
+	}
+
 	// Fetch labelSelector from r.LabelSelector (command-line option)
 	labelSelector, err := labels.Parse(r.LabelSelector)
 	if err != nil {
--- a/controllers/argocd/argocd_controller.go
+++ b/controllers/argocd/argocd_controller.go
@@ -130,11 +130,9 @@ func (r *ReconcileArgoCD) internalReconcile(ctx context.Context, request ctrl.Re
 		return reconcile.Result{}, argocd, err
 	}
 
-	// If the number of notification replicas is greater than 1, return an error
+	// If the number of notification replicas is greater than 1, display a warning.
 	if argocd.Spec.Notifications.Replicas != nil && *argocd.Spec.Notifications.Replicas > 1 {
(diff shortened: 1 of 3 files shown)


controllers/argocd/argocd_controller.go
```diff
@@ -130,6 +130,13 @@ func (r *ReconcileArgoCD) internalReconcile(ctx context.Context, request ctrl.Re
 		return reconcile.Result{}, argocd, err
 	}
 
+	// If the number of notification replicas is greater than 1, return an error
+	if argocd.Spec.Notifications.Replicas != nil && *argocd.Spec.Notifications.Replicas > 1 {
+		message := "Argo CD Notification controller does not support multiple replicas. Notification replicas cannot be greater than 1."
+		reqLogger.Info(message)
+		return reconcile.Result{}, argocd, fmt.Errorf("%s", message)
```
svghadi, line 137: Will this stop the reconciliation completely if there is problem with notifications configurations?

svghadi: What do you think about ignoring the replica field value for notifications and throwing a warning message instead of stoping the entire reconciliation process due to a single unsupported configuration? Since this is not yet supported upstream, we could inform users through a warning message and maybe also update our documentation accordingly to reflect this.

Rizwana777: i agree, but in ticket they mentioned to report an error and stop the reconciliation

Rizwana777: yes, then I will just log the error and move forward

svghadi: Cool. Lets also ignore the value if set.

Rizwana777: if we are just giving the warning , then there is no need to update the condition field of ArgoCD with error type ErrorOccurred as we will not return any error here[right](https://github.com/argoproj-labs/argocd-operator/pull/1642/files#diff-2fe0a2bb5baaf6170a23a7c08adc70117adfa25b7d481923aa8626a38cefd447R137)?

svghadi:
I think we also need to ignore the value set in replica field here
https://github.com/argoproj-labs/argocd-operator/blob/5cce9cecd80c9c3d4d580deac9ac87f7451b9b9b/controllers/argocd/notifications.go#L341-L343 

Maybe, log the message in this if condition and remove the assignment statement to ignore the value. WDYT?

Rizwana777:
> I think we also need to ignore the value set in replica field here
> 
> https://github.com/argoproj-labs/argocd-operator/blob/5cce9cecd80c9c3d4d580deac9ac87f7451b9b9b/controllers/argocd/notifications.go#L341-L343
> 
> Maybe, log the message in this if condition and remove the assignment statement to ignore the value. WDYT?

I think we can remove or ignore this function itself [getArgoCDNotificationsControllerReplicas](https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/notifications_util.go#L564) if we are removing assignment statement to ignore the value, it will be of no use calling this function if we are not using the value returned by it right? and I see this function not used in any other place as well?

svghadi:
Yes. Sounds good.

jgwest (approved):
LGTM too, I'm happy with just logging the error as requested by @svghadi . Thanks @Rizwana777 and @svghadi for reviewing!

PR #1646, opened 2025-01-23
Adding the execTimeout field's type change code along with changes in the test cases too.

sarfarazgit:
**What type of PR is this?**

Fixes #?

### DIFF
--- a/controllers/argocd/deployment.go
+++ b/controllers/argocd/deployment.go
@@ -918,8 +918,8 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argoproj.ArgoCD, useTLSFor
 	})
 	// Environment specified in the CR take precedence over everything else
 	repoEnv = argoutil.EnvMerge(repoEnv, proxyEnvVars(), false)
-	if cr.Spec.Repo.ExecTimeout != nil {
-		repoEnv = argoutil.EnvMerge(repoEnv, []corev1.EnvVar{{Name: "ARGOCD_EXEC_TIMEOUT", Value: fmt.Sprintf("%ds", *cr.Spec.Repo.ExecTimeout)}}, true)
+	if cr.Spec.Repo.ExecTimeout != "" {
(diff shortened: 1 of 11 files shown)


svghadi:
There is a e2e test failure 
```bash
    logger.go:42: 14:19:11 | 1-011_validate_repo_exec_timeout/2-change-exec-timeout | starting test step 2-change-exec-timeout
    case.go:254: failed in step 2-change-exec-timeout
    case.go:256: ArgoCD.argoproj.io "argocd" is invalid: [spec.repo.execTimeout: Invalid value: "integer": spec.repo.execTimeout in body must be of type string: "integer", <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]
```

Looks like validation error. Changing field from int to string will result into a breaking change. We will have to rethink our approach.

controllers/argocd/deployment.go
```diff
@@ -918,8 +918,8 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argoproj.ArgoCD, useTLSFor
 	})
 	// Environment specified in the CR take precedence over everything else
 	repoEnv = argoutil.EnvMerge(repoEnv, proxyEnvVars(), false)
-	if cr.Spec.Repo.ExecTimeout != nil {
-		repoEnv = argoutil.EnvMerge(repoEnv, []corev1.EnvVar{{Name: "ARGOCD_EXEC_TIMEOUT", Value: fmt.Sprintf("%ds", *cr.Spec.Repo.ExecTimeout)}}, true)
+	if cr.Spec.Repo.ExecTimeout != "" {
```
anandf, line 921: Do we handle invalid strings which is not a valid duration ? something like `abc` or `xyz` must be validated and reported early.

anandf: Probably you can use this function to validate the string and then if its valid, set it in Environment. https://pkg.go.dev/maze.io/x/duration#ParseDuration

sarfarazgit: @anandf Noted.

PR #1648, opened 2025-01-27
Only update Argo CD Cluster secret if the values have changed

jgwest:
**What type of PR is this?**

/kind bug

Currently, the Argo CD operator will update the default cluster Secret after every reconcile, regardless of whether or not the `.data.namespaces` field has actually changed.

This PR updates the logic to only update `.data.namespaces` if the field has changed.

### DIFF
--- a/controllers/argocd/secret.go
+++ b/controllers/argocd/secret.go
@@ -391,23 +391,14 @@ func (r *ReconcileArgoCD) reconcileGrafanaSecret(cr *argoproj.ArgoCD) error {
 	return nil
 }
 
-// reconcileClusterPermissionsSecret ensures ArgoCD instance is namespace-scoped
-func (r *ReconcileArgoCD) reconcileClusterPermissionsSecret(cr *argoproj.ArgoCD) error {
-	var clusterConfigInstance bool
-	secret := argoutil.NewSecretWithSuffix(cr, "default-cluster-config")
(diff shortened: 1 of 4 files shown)


controllers/argocd/secret.go
```diff
@@ -419,54 +410,132 @@ func (r *ReconcileArgoCD) reconcileClusterPermissionsSecret(cr *argoproj.ArgoCD)
...
+
+// reconcileClusterPermissionsSecret ensures ArgoCD instance is namespace-scoped
+func (r *ReconcileArgoCD) reconcileClusterPermissionsSecret(cr *argoproj.ArgoCD) error {
+
+	managedNamespaceList, err := generateSortedManagedNamespaceListForArgoCDCR(cr, r.Client)
+	if err != nil {
+		return err
+	}
+
+	// isArgoCDAClusterConfigInstance is whether on not 'cr' is a cluster config instance (mentioned in ARGOCD_CLUSTER_CONFIG_NAMESPACES)
```
Rizwana777, line 450: I think it's `isArgoCDAClusterConfigInstance is whether on not 'cr' is a cluster config instance -> isArgoCDAClusterConfigInstance indicates whether 'cr' is a cluster config instance `

Rizwana777:
minor typo, LGTM

jgwest:
Thanks @Rizwana777!

PR #1650, opened 2025-01-28
Set exec timeout repo branch

sarfarazgit:
**What type of PR is this?**

Fixes #?

### DIFF
--- a/api/v1alpha1/argocd_types.go
+++ b/api/v1alpha1/argocd_types.go
@@ -454,7 +454,7 @@ type ArgoCDRepoSpec struct {
 	Version string `json:"version,omitempty"`
 
 	// ExecTimeout specifies the timeout in seconds for tool execution
-	ExecTimeout *int `json:"execTimeout,omitempty"`
+	ExecTimeout string `json:"execTimeout,omitempty"`
 
 	// Env lets you specify environment for repo server pods
 	Env []corev1.EnvVar `json:"env,omitempty"`
--- a/controllers/argocd/deployment_test.go
+++ b/controllers/argocd/deployment_test.go
@@ -295,8 +295,8 @@ func TestReconcileArgoCD_reconcile_ServerDeployment_env(t *testing.T) {
 				Value: "FOO",
 			},
 		}
-		timeout := 600
-		a.Spec.Repo.ExecTimeout = &timeout
+		timeout := "600"
+		a.Spec.Repo.ExecTimeout = timeout
 
 		resObjs := []client.Object{a}
 		subresObjs := []client.Object{a}
@@ -336,8 +336,8 @@ func TestReconcileArgoCD_reconcileRepoDeployment_env(t *testing.T) {
 				Value: "FOO",
 			},
 		}
-		timeout := 600
-		a.Spec.Repo.ExecTimeout = &timeout
+		timeout := "600m"
+		a.Spec.Repo.ExecTimeout = timeout
 
 		resObjs := []client.Object{a}
 		subresObjs := []client.Object{a}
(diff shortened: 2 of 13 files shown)


controllers/argocd/deployment_test.go
```diff
@@ -359,14 +359,14 @@ func TestReconcileArgoCD_reconcileRepoDeployment_env(t *testing.T) {
...
-		assert.Contains(t, deployment.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{Name: "ARGOCD_EXEC_TIMEOUT", Value: "600s"})
+		assert.Contains(t, deployment.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{Name: "ARGOCD_EXEC_TIMEOUT", Value: "600m"})
 	})
 
 	t.Run("ExecTimeout set", func(t *testing.T) {
 		logf.SetLogger(ZapLogger(true))
 		a := makeTestArgoCD()
-		timeout := 600
-		a.Spec.Repo.ExecTimeout = &timeout
+		timeout := "600m"
```
anandf, line 368: Should it not be `600s` ?

controllers/argocd/deployment_test.go
```diff
@@ -386,14 +386,14 @@ func TestReconcileArgoCD_reconcileRepoDeployment_env(t *testing.T) {
...
+		assert.Contains(t, deployment.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{Name: "ARGOCD_EXEC_TIMEOUT", Value: "600m"})
 	})
 
 	t.Run("ExecTimeout set with env set explicitly", func(t *testing.T) {
 		logf.SetLogger(ZapLogger(true))
 		a := makeTestArgoCD()
-		timeout := 600
-		a.Spec.Repo.ExecTimeout = &timeout
+		timeout := "600m"
+		a.Spec.Repo.ExecTimeout = timeout
```
anandf, line 396: Can do a direct assignment `a.Spec.Repo.ExecTimeout = 600s`

controllers/argocd/deployment_test.go
```diff
@@ -419,7 +419,7 @@ func TestReconcileArgoCD_reconcileRepoDeployment_env(t *testing.T) {
 
 		// Check that the env vars are set, Count is 2 because of the default REDIS_PASSWORD env var
 		assert.Len(t, deployment.Spec.Template.Spec.Containers[0].Env, 2)
-		assert.Contains(t, deployment.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{Name: "ARGOCD_EXEC_TIMEOUT", Value: "600s"})
+		assert.Contains(t, deployment.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{Name: "ARGOCD_EXEC_TIMEOUT", Value: "600m"})
```
anandf, line 422: Let's keep the same value as `600s` instead of switching to minute.

deploy/olm-catalog/argocd-operator/0.9.1/argoproj.io_argocds.yaml
```diff
@@ -1672,7 +1672,7 @@ spec:
                       type: object
                     type: array
                   execTimeout:
-                    description: ExecTimeout specifies the timeout in seconds for
+                    description: ExecTimeout specifies the timeout in minutes for
```
anandf, line 1675: It can be seconds, minutes or hours based on the suffix user provides. This description needs a change.

anandf (changes requested):
Posted few minor comments.

svghadi:
@anandf - What are you thoughts on this? To me, looks like a breaking change.

https://github.com/argoproj-labs/argocd-operator/pull/1646#issuecomment-2611481847
> There is a e2e test failure 
```bash
    logger.go:42: 14:19:11 | 1-011_validate_repo_exec_timeout/2-change-exec-timeout | starting test step 2-change-exec-timeout
    case.go:254: failed in step 2-change-exec-timeout
    case.go:256: ArgoCD.argoproj.io "argocd" is invalid: [spec.repo.execTimeout: Invalid value: "integer": spec.repo.execTimeout in body must be of type string: "integer", <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]
```

> This appears to be a validation error. Changing the field type from int to string would result in a breaking change, so we’ll need to rethink our approach. Deprecating the field might be the best path forward, especially since the value can also be set via an environment variable.

> Alternatively, we could continue supporting int and clarify through documentation that the field’s value is interpreted as seconds.

api/v1alpha1/argocd_types.go
```diff
@@ -454,7 +454,7 @@ type ArgoCDRepoSpec struct {
 	Version string `json:"version,omitempty"`
 
 	// ExecTimeout specifies the timeout in seconds for tool execution
-	ExecTimeout *int `json:"execTimeout,omitempty"`
+	ExecTimeout string `json:"execTimeout,omitempty"`
```
anandf, line 457: For including the validation, let's define an alias for `string` called `Duration`
```
// +kubebuilder:validation:Pattern:="^(0|(([0-9]+)y)?(([0-9]+)w)?(([0-9]+)d)?(([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?(([0-9]+)ms)?)$"
type Duration string
```
Detailed example can be found here. https://github.com/prometheus-operator/prometheus-operator/blob/dff3575c55ee9e6423907e4bfdcb5ac4b8fe9c89/pkg/apis/monitoring/v1/types.go#L45

anandf (approved):
LGTM

olivergondza:
@svghadi, @anandf, how do we feel about integrating this? I see LGTM and do-not-merge...

PR #1651, opened 2025-01-29
Add missing permission for impersonation

varshab1210:
**What type of PR is this?**

/kind enhancement

Because of missing permission in application controller clusterrole, impersonation feature does not work as expected. This PR aims to fix the same

Fixes #?


svghadi (commented):
Looks good. Is it possible to add a unit or e2e test?

tests/k8s/1-047_validate_impersonation_namespace_scoped_instance/01-assert.yaml
varshab1210, line 1: Hi @anandf, the current steps mentioned [here](https://github.com/argoproj/argo-cd/pull/17403) do not work for a ns scoped instance. Could you please suggest the required modifications?

Even after adding the managed-by label, I see this error:

```
message: 'ComparisonError: Failed to load live state: Namespace "guestbook" for Deployment "guestbook-ui" is not managed'
```

anandf: Instead of `guestbook` namespace, can you try using `argocd-test-impersonation` as the destination ns ? For a namespace scoped `ArgoCD` instance, you can deploy manifests only in the same ns as the RBAC permits only that by default. Otherwise, you have to create a ns `guestbook` and apply the `managed-by` label. You cannot test the `CreateNamespace=true` sync policy in that case.

tests/k8s/1-047_validate_impersonation_namespace_scoped_instance/01-install.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: test-impersonation-ns
+  labels:
+    argocd.argoproj.io/managed-by: openshift-gitops
```
Mangaal, line 6: The label `argocd.argoproj.io/managed-by` is set to `openshift-gitops`. However, since this ArgoCD instance is deployed in the `test-impersonation-ns` namespace, I suggest updating the label value to reflect that. The updated label should be `argocd.argoproj.io/managed-by: test-impersonation-ns.`

Mangaal (changes requested):

tests/k8s/1-047_validate_impersonation_namespace_scoped_instance/01-assert.yaml
```diff
@@ -0,0 +1,26 @@
...
+status:
+  phase: Available
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: argocd-cm
+  namespace: argocd-test-ns-scoped
+data:
+  application.sync.impersonation.enabled: 'true'
```
Mangaal, line 26: Once we have added application.sync.impersonation.enabled: "true" in the ArgoCD CR under extraConfig, do we need to redeclare it in argocd-cm? The operator will update the ConfigMap automatically.

Mangaal (approved):

svghadi (approved):
LGTM. Thanks.

PR #1661, opened 2025-02-11
chore: update argo-cd version

jparsai:
/kind chore

This PR is to update argocd CRDs to version to 2.13.5, since there are no changes in CRDs just updating bundle.


deploy/olm-catalog/argocd-operator/0.13.0/argocd-operator.v0.13.0.clusterserviceversion.yaml
```diff
@@ -1901,7 +1901,7 @@ spec:
                       fieldPath: metadata.annotations['olm.targetNamespaces']
                 - name: ENABLE_CONVERSION_WEBHOOK
                   value: "true"
-                image: quay.io/argoprojlabs/argocd-operator@sha256:d7f62482426bd8a1ff99f193f199b11e295a1f9093a8b65fa14ada7eec77e1a3
```
anandf, line 1904: I think we need to have the SHA based digest here.

PR #1667, opened 2025-02-18
Preserve RBAC scope conf in RBAC CM in when updating to SSO keycloak

nmirasch:
**What type of PR is this?**
 /kind bug 

This PR adds logic to preserve the RBAC scope configuration in the RBAC ConfigMap when updating to SSO Keycloak. Previously, it was set to '[groups,email]' in the RBAC ConfigMap without updating the Spec.RBAC.Scopes, which caused a conflict during the ConfigMap sync when switching to Keycloak SSO and updating the ArgoCDConfiguration.

Fixes: https://issues.redhat.com/browse/GITOPS-5977

Steps to reproduce:

1. Ran make run
2. Created an ArgoCD CR.
3. Set the RBAC scope
4. Do a kubectl apply on something like:

> ```yaml
> apiVersion: argoproj.io/v1beta1
> kind: ArgoCD
> metadata:
>   name: argocd
> spec:
>   rbac:
>    scopes: '[groups]'

5. Followed by a kubectl apply on something like:

> ```yaml
> apiVersion: argoproj.io/v1beta1
> kind: ArgoCD
> metadata:
>  name: openshift-gitops
>  namespace: openshift-gitops
> spec:
>  extraConfig:
>    oidc.tls.insecure.skip.verify: "true"
>  sso:
>    provider: keycloak
>    keycloak:
>      rootCA : ---BEGIN---END---

6. After the keycloak container is created, Previously this change, logs keep showing "RBAC ConfigMap 'argocd-rbac-cm' updated"

### DIFF
--- a/controllers/argocd/configmap.go
+++ b/controllers/argocd/configmap.go
@@ -579,12 +579,16 @@ func (r *ReconcileArgoCD) reconcileRBACConfigMap(cm *corev1.ConfigMap, cr *argop
 
 	// Scopes
 	if cr.Spec.RBAC.Scopes != nil && cm.Data[common.ArgoCDKeyRBACScopes] != *cr.Spec.RBAC.Scopes {
-		cm.Data[common.ArgoCDKeyRBACScopes] = *cr.Spec.RBAC.Scopes
-		if changed {
-			explanation += ", "
+		if cr.Spec.SSO != nil && cr.Spec.SSO.Provider.ToLower() == argoproj.SSOProviderTypeKeycloak {
+			log.Info("cr.Spec.RBAC.Scopes value could be out of sync with the RBACConfigMap, since keycloak sso is enabled and scopes are fixed to [groups,mails]")
+		} else {
+			cm.Data[common.ArgoCDKeyRBACScopes] = *cr.Spec.RBAC.Scopes
+			if changed {
+				explanation += ", "
+			}
+			explanation += "rbac scopes"
+			changed = true
 		}
-		explanation += "rbac scopes"
-		changed = true
 	}
 
 	if changed {
--- a/controllers/argocd/configmap_test.go
+++ b/controllers/argocd/configmap_test.go
@@ -1159,6 +1159,47 @@ func Test_reconcileRBAC(t *testing.T) {
 
 	assert.NoError(t, err)
 	assert.Equal(t, cm.Data["policy.matchMode"], matcherMode)
+
+	// Verify when SSO different from keycloak it sync
+	rbacScopes := "[groups,email]"
+	cmRbacScopes := ""
+	a.Spec.RBAC.Scopes = &rbacScopes
+	cm.Data["scopes"] = cmRbacScopes
+	a.Spec.SSO = &argoproj.ArgoCDSSOSpec{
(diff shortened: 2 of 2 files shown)


controllers/argocd/keycloak.go
```diff
@@ -1090,8 +1090,19 @@ func (r *ReconcileArgoCD) updateArgoCDConfiguration(cr *argoproj.ArgoCD, kRouteU
...
 	}
-
-	argoRBACCM.Data["scopes"] = "[groups,email]"
+	if cr.Spec.RBAC.Scopes != nil && *cr.Spec.RBAC.Scopes != "" {
+		argoRBACCM.Data["scopes"] = *cr.Spec.RBAC.Scopes
+	} else {
+		argoRBACCM.Data["scopes"] = "[groups,email]"
+		scopes := argoRBACCM.Data["scopes"]
+		cr.Spec.RBAC.Scopes = &scopes
+		err = r.Client.Update(context.TODO(), cr) // Update the Spec.RBAC.Scopes with the default scopes added by keycloak.
```
svghadi, line 1099: I think we should handle this in code instead of updating the rbac field in ArgoCD CR. I feel operator should respect `spec.rbac` when configuring keycloak. If the value is set, operator shouldn't update the configmap with keycloak related this. But if the `[groups,email]` is necessary for keycloak, then we have no option but to override the configmap field with keycloak related things. In this case, we should log it properly in the operator logs and handle it accordingly in `.spec.rbac` logic to avoid conflicting write.

iam-veeramalla:
LGTM, @nmirasch can you please add a log as per @svghadi suggestion above.

@svghadi yes, when working with keycloak, those rbac scopes are important. so we can log a message to the users.

iam-veeramalla (approved):

svghadi:
Thanks @iam-veeramalla for confirming. 

Since `[groups,email]` is necessary for keycloak we can keep the keycloak code as is and handle the conflicting cm write in `Spec.RBAC.Scopes` code instead of writing back to ArgoCD CR. I would suggest we ignore the update when `cr.Spec.SSO.Provider=keycloak` here and log a message so that user is aware that `cr.Spec.RBAC.Scopes` field value is not respected since keycloak sso is enabled. 

https://github.com/argoproj-labs/argocd-operator/blob/69c5a0d33bf30528c9a55e78ec7dde0d400813fd/controllers/argocd/configmap.go#L581-L588

nmirasch:
@svghadi @iam-veeramalla I moved the change to configmap.go as @svghadi suggested. As the user should be aware about this change, I just added this, I don't know is the message have to be different on if it has to be logged in another way: 
` log.Info("cr.Spec.RBAC.Scopes value could be out of sync with the RBACConfigMap, since keycloak sso is enabled and scopes are fixed to [groups,mails]")`
The spec.RBAC.scopes values will be preserved but the rbacConfigmap scopes will be updated by the keycloak.  
Could you take another look here? Thanks!!!

svghadi (approved):
LGTM. Thanks @nmirasch

iam-veeramalla:
LGTM @nmirasch, Thanks for identifying and fixing this. Also, Thanks to @svghadi for valuable suggestions.

PR #1671, opened 2025-02-23
Upgrade argocd to 2.14.13 and debug upgrade issues

anandf:
**What type of PR is this?**

Fixes #?


bundle/manifests/argoproj.io_notificationsconfigurations.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.14.0
+    controller-gen.kubebuilder.io/version: v0.17.2
```
svghadi, line 5: Is controller-gen upgrade required? If so, we will also need to update it in makefile
https://github.com/argoproj-labs/argocd-operator/blob/58532cebff27bc97f94d9671bf1cf7fbe4d4906e/Makefile#L169-L171

config/rbac/role.yaml
svghadi, line 1: Are these permission changes done by code-gen?

bundle/manifests/argoproj.io_notificationsconfigurations.yaml
anandf: Yes, IIRC, this change was required for the compilation to succeed. Let me cross check and revert if not required.

config/rbac/role.yaml
anandf: Yes, these were generated by code-gen automatically.

bundle/manifests/argoproj.io_notificationsconfigurations.yaml
anandf: Yes, this was required to make the builds succeed with kubernetes API `0.31.0`. Updated the `Makefile` too.

svghadi:
Superseded by https://github.com/argoproj-labs/argocd-operator/pull/1773

PR #1675, opened 2025-02-24
chore: Upgrade to Argo CD v2.12.10

jgwest:
**What type of PR is this?**

/kind chore

Upgrade from v2.12.3 to v2.12.10. This PR picks up security issue in v2.12.10 (which we're likely not vulnerable too, but it still shows up in dependabot scans). 

This PR should be non-disruptive. There is another PR to upgrade to v2.14, but is in draft as of this writing.


svghadi (approved):
LGTM. Thanks

jgwest:
Thanks @svghadi!

PR #1682, opened 2025-03-07
revert: "fix: Add PostStart lifecycle hook to reset Redis sentinel"

svghadi:
Reverts argoproj-labs/argocd-operator#1674

The PostStart hook results into a crashLoopBackOff for state redis when custom certificates are used with redis. Reverting the change for now to unblock the release.


anandf (approved):
LGTM

PR #1687, opened 2025-03-10
Add NamespaceManagement reconciliation

Rizwana777:
**What type of PR is this?**

Add NamespaceManagement reconciliation
Add NamespaceManagement CR
Add .Spec.NamespaceManagement field in ArgoCDSpec CR
Unit tests
e2e for glob match on .Spec.NamespaceManagement field

JIRA - https://issues.redhat.com/browse/GITOPS-6179


controllers/argocd/util.go
```diff
@@ -1879,3 +1883,194 @@ func appendUniqueArgs(cmd []string, extraArgs []string) []string {
...
+				// Delete the NamespaceManagement CR
+				if err := r.Client.Delete(ctx, &nsMgmt); err != nil && !apierrors.IsNotFound(err) {
+					log.Error(err, fmt.Sprintf("Failed to delete NamespaceManagement CR %s in namespace %s", nsMgmt.Name, nsMgmt.Namespace))
+					return err
+				}
+			}
+		}
+	}
+
+	return nil
```
jgwest, line 2075: As above, this can be removed.

controllers/argocd/custommapper.go
```diff
@@ -266,3 +266,34 @@ func (r *ReconcileArgoCD) applicationSetSCMTLSConfigMapMapper(ctx context.Contex
 
 	return result
 }
+
+// namespaceResourceMapper maps a watch event on a namespaceManagement, back to the
+// ArgoCD object that we want to reconcile.
+func (r *ReconcileArgoCD) nmMapper(ctx context.Context, o client.Object) []reconcile.Request {
```
jgwest, line 272: At first I thought this logic was correct, but I think instead this function should reconcile EVERY `ArgoCD` CR on the cluster, rather than just the one that is pointed to by `.spec.ManagedBy`.

This allows us to catch the case where `.spec.managedBy` is changed from Argo CD namespace 'A', to Argo CD namespace 'B'.

controllers/argocd/util.go
```diff
@@ -1058,7 +1058,7 @@ func (r *ReconcileArgoCD) setResourceWatches(bldr *builder.Builder, clusterResou
 	}
 
 	// Watch for changes to primary resource ArgoCD
-	bldr.For(&argoproj.ArgoCD{}, builder.WithPredicates(deleteSSOPred, deleteNotificationsPred))
+	bldr.For(&argoproj.ArgoCD{}, builder.WithPredicates(deleteSSOPred, deleteNotificationsPred, argoCDNamespaceManagementFilterPredicate()))
```
jgwest, line 1061: With the change I have proposed to `nmMapper` above, I don't think `argoCDNamespaceManagementFilterPredicate()` is required here.

controllers/argocd/util.go
```diff
@@ -1079,6 +1079,10 @@ func (r *ReconcileArgoCD) setResourceWatches(bldr *builder.Builder, clusterResou
 
 	bldr.Owns(&v1.RoleBinding{})
 
+	nmMapperResourceHandler := handler.EnqueueRequestsFromMapFunc(nmMapper)
+
+	bldr.Watches(&argoproj.NamespaceManagement{}, nmMapperResourceHandler, builder.WithPredicates(namespaceManagementFilterPredicate()))
```
jgwest, line 1084: With the change I have proposed to `nmMapper` above, I don't think `namespaceManagementFilterPredicate()` is required here.

controllers/argocd/util.go
```diff
@@ -1879,3 +1883,194 @@ func appendUniqueArgs(cmd []string, extraArgs []string) []string {
...
+				log.Error(err, fmt.Sprintf("unable to delete namespace %s from cluster secret", e.Object.GetNamespace()))
+			} else {
+				log.Info(fmt.Sprintf("Successfully deleted namespace %s from cluster secret", e.Object.GetNamespace()))
+			}
+			return false
+		},
+	}
+}
+
+func argoCDNamespaceManagementFilterPredicate() predicate.Predicate {
```
jgwest, line 1962: As above, I think this can be removed.

controllers/argocd/util.go
```diff
@@ -1879,3 +1883,194 @@ func appendUniqueArgs(cmd []string, extraArgs []string) []string {
 
 	return cmd
 }
+
+func namespaceManagementFilterPredicate() predicate.Predicate {
```
jgwest, line 1887: As above, I think this can be removed.

jgwest (changes requested):

controllers/argocd/argocd_controller.go
```diff
@@ -237,6 +243,24 @@ func (r *ReconcileArgoCD) internalReconcile(ctx context.Context, request ctrl.Re
 		return reconcile.Result{}, argocd, err
 	}
 
+	// Handle NamespaceManagement reconciliation and check if Namespace Management is enabled via the Subscription env variable.
+	if isNamespaceManagementEnabled() {
+		if err := r.reconcileNamespaceManagement(argocd); err != nil {
+			return reconcile.Result{}, argocd, err
+		}
+	} else if argocd.Spec.NamespaceManagement != nil {
+		k8sClient, err := initK8sClient()
```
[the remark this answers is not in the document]
Rizwana777: [deleteRBACsForNamespace](https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/util.go#L1292) uses kubernetes.Interface so I have called initK8sClient here

controllers/argocd/namespaceManagement.go
```diff
@@ -0,0 +1,178 @@
...
+			log.Error(err, fmt.Sprintf("failed to delete RBACs for namespace: %s", ns.Name))
+		} else {
+			log.Info(fmt.Sprintf("Successfully removed the RBACs for namespace: %s", ns.Name))
+		}
+
+		err := deleteManagedNamespaceFromClusterSecret(argocd.Namespace, ns.Name, k8sClient)
+		if err != nil {
+			log.Error(err, fmt.Sprintf("unable to delete namespace %s from cluster secret", ns.Name))
+		} else {
+			log.Info(fmt.Sprintf("Successfully deleted namespace %s from cluster secret", ns.Name))
```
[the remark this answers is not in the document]
Rizwana777: I might be mistaken, please correct me if I am wrong. I dont see any log related to success message in deleteManagedNamespaceFromClusterSecret. I only this [log](https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/util.go#L1384) which says removing/updating. Shall I keep the the existing log or remove it?

controllers/argocd/util.go
Rizwana777: In namespaceManagementFilterPredicate() I am deleting the Rbacs whenever the namespacemanagement field is updated or deleted. Whenever .spec.managedBy field is updated or removed , it will delete the respective rbacs

controllers/argocd/namespaceManagement.go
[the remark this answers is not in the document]
jgwest: I see this log statement in `deleteManagedNamespaceFromClusterSecret`: `argoutil.LogResourceUpdate(log, localClusterSecret, "removing managed namespace", sourceNS)`

controllers/argocd/namespaceManagement_test.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/argoproj-labs/argocd-operator/common"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	testclient "k8s.io/client-go/kubernetes/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
```
jgwest, line 18: Can you add additional test cases for testing `reconcileNamespaceManagement`? This test is important to maintaining security of this feature, so it would be good to ensure that all of the things this functions is doing are tested.

controllers/argocd/secret_test.go
```diff
@@ -595,6 +596,41 @@ func TestGenerateSortedManagedNamespaceListForArgoCDCR(t *testing.T) {
...
+
+	err = cl.Create(context.Background(), &nsMgmt1)
+	assert.NoError(t, err)
+	err = cl.Create(context.Background(), &nsMgmt2)
+	assert.NoError(t, err)
+
+	res, err = generateSortedManagedNamespaceListForArgoCDCR(a, r.Client)
+	assert.NoError(t, err)
+	assert.Equal(t, res, []string{managedByNamespace.Name, nsMgmt1.Namespace, a.Namespace, nsMgmt2.Namespace})
+
```
jgwest, line 633: Can you also verify that if there exists a NamespaceManagement that has `.spec.managedBy` that points to a different namespace, that the namespace of the NamespaceManagement does NOT appear in the list?

It would also be good to verify that namespace management namespaces don't appear in the list when the feature is disabled.

controllers/argocd/util_test.go
```diff
@@ -1384,3 +1384,49 @@ func TestAppendUniqueArgs(t *testing.T) {
 		})
 	}
 }
+
+func TestRemoveNamespaceManagementCRs(t *testing.T) {
```
jgwest, line 1388: What is this test testing?

controllers/argocd/util.go
```diff
@@ -1879,3 +1874,168 @@ func appendUniqueArgs(cmd []string, extraArgs []string) []string {
...
+}
+
+// updateStatusConditionOfArgoCD calls Set Condition of NamespaceManagement status
+func updateStatusConditionOfNamespaceManagement(ctx context.Context, condition metav1.Condition, cr *argoproj.NamespaceManagement, k8sClient client.Client, log logr.Logger) error {
+	changed, newConditions := insertOrUpdateConditionsInSlice(condition, cr.Status.Conditions)
+
+	if changed {
+		// get the latest version of namespacemanagement before updating
+		if err := k8sClient.Get(ctx, types.NamespacedName{Name: cr.Name, Namespace: cr.Namespace}, cr); err != nil {
+			return err
```
jgwest, line 2032: Here you can add:
```
if apierrors.IsNotfound(err) {
    return nil
}
```
If the `NamespaceManagement` no longer exists, then we can just `return nil` since there is no more work to do. (I am fixing this bug in another part of the code with a PR)

controllers/argocd/util.go
```diff
@@ -1879,3 +1874,168 @@ func appendUniqueArgs(cmd []string, extraArgs []string) []string {
...
+				log.Error(err, fmt.Sprintf("unable to delete namespace %s from cluster secret", e.Object.GetNamespace()))
+			} else {
+				log.Info(fmt.Sprintf("Successfully deleted namespace %s from cluster secret", e.Object.GetNamespace()))
+			}
+			return false
+		},
+	}
+}
+
+func argoCDNamespaceManagementFilterPredicate() predicate.Predicate {
```
jgwest, line 1953: Need unit tests for the logic defined in `argoCDNamespaceManagementFilterPredicates`. 

In order to test the functions, you may need to move the functions outside of `argoCDNamespaceManagementFilterPredicates` and into their own functions (which you can then reference from ``argoCDNamespaceManagementFilterPredicates`.`)

controllers/argocd/util.go
```diff
@@ -1879,3 +1874,168 @@ func appendUniqueArgs(cmd []string, extraArgs []string) []string {
 
 	return cmd
 }
+
+func namespaceManagementFilterPredicate() predicate.Predicate {
```
jgwest, line 1878: Need unit tests for this function, same comment as `argoCDNamespaceManagementFilterPredicate`

jgwest (changes requested):

controllers/argocd/util_test.go
Rizwana777: This needs to be removed from last review itself, which I missed, i will remove it now

deploy/olm-catalog/argocd-operator/0.14.0/argocd-operator-webhook-service_v1_service.yaml.bak
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
Rizwana777: Yes it needs to be removed as it has been created while running make bundle

controllers/argocd/namespaceManagement.go
```diff
@@ -0,0 +1,196 @@
...
+	nmList := &argoproj.NamespaceManagementList{}
+	if err := r.Client.List(ctx, nmList); err != nil {
+		return fmt.Errorf("failed to list NamespaceManagement resources: %w", err)
+	}
+
+	// Convert NamespaceManagement spec into a lookup map
+	allowedNamespaces := make(map[string]bool)
+	if argocd.Spec.NamespaceManagement != nil {
+		for _, nm := range argocd.Spec.NamespaceManagement {
+			allowedNamespaces[nm.Name] = nm.AllowManagedBy
```
jgwest: `nm.Name` can be a pattern; this logic isn't correct when `nm.Name` is a pattern instead of a namespace name.

controllers/argocd/namespaceManagement.go
```diff
@@ -0,0 +1,196 @@
...
+	for _, nsMgmt := range nsMgmtList.Items {
+		if nsMgmt.Spec.ManagedBy == argocd.Namespace {
+			managedNamespaces[nsMgmt.Namespace] = true
+		}
+	}
+
+	// Only act on namespaces in ArgoCD.spec.namespaceManagement that are truly managed by this ArgoCD
+	for _, ns := range argocd.Spec.NamespaceManagement {
+		nsName := ns.Name
+		if !managedNamespaces[nsName] {
```
jgwest: `nsName` can be a pattern, rather than just a name, so this logic likely isn't correct.

controllers/argocd/secret.go
```diff
@@ -391,24 +391,67 @@ func (r *ReconcileArgoCD) reconcileGrafanaSecret(cr *argoproj.ArgoCD) error {
...
 	for _, namespace := range namespaceList.Items {
-		namespaces = append(namespaces, namespace.Name)
+		uniqueNamespaces[namespace.Name] = struct{}{}
+	}
+
+	// Build a lookup map from ArgoCD .spec.namespaceManagement where allowManagedBy == true
+	allowedNamespaceMap := make(map[string]bool)
+	for _, entry := range cr.Spec.NamespaceManagement {
+		if entry.AllowManagedBy {
+			allowedNamespaceMap[entry.Name] = true
```
jgwest: `entry.Name` can be a pattern, not just a name. So this logic is likely not correct.

controllers/argocd/util.go
```diff
@@ -1921,3 +1887,187 @@ func appendUniqueArgs(cmd []string, extraArgs []string) []string {
...
+func getNamespacesToDelete(oldList, newList []argoproj.ManagedNamespaces) []string {
+	// Convert old and new NamespaceManagement lists into maps for easy lookup
+	oldNamespaces := make(map[string]bool)
+	newNamespaces := make(map[string]bool)
+	oldNamespaceSettings := make(map[string]bool) // Track allowManagedBy values
+	newNamespaceSettings := make(map[string]bool)
+
+	for _, ns := range oldList {
+		oldNamespaces[ns.Name] = true
+		oldNamespaceSettings[ns.Name] = ns.AllowManagedBy // Store old setting
```
jgwest: Given that `ns.name` can be a pattern, I think this logic isn't correct. (The function that calls this assumes that the []string result are actual namespace names)

controllers/argocd/namespaceManagement_test.go
```diff
@@ -0,0 +1,348 @@
...
+			reconciledCondition = &cond
+			break
+		}
+	}
+	assert.NotNil(t, reconciledCondition)
+	assert.Equal(t, metav1.ConditionFalse, reconciledCondition.Status)
+	assert.Equal(t, "ErrorOccurred", reconciledCondition.Reason)
+}
+
+func TestHandleFeatureDisable_NoNamespaceManagement(t *testing.T) {
```
jgwest, line 112: Could use a few more tests that verify the logic in `disableNamespaceManagement`

controllers/argocd/util.go
```diff
@@ -1921,3 +1887,187 @@ func appendUniqueArgs(cmd []string, extraArgs []string) []string {
...
+		if err := k8sClient.Status().Update(ctx, cr); err != nil {
+			log.Error(err, "unable to update NamespaceManagement status condition")
+			return err
+		}
+	}
+	return nil
+}
+
+// getNamespacesToDelete determines which namespaces were removed or had allowManagedBy changed.
+func getNamespacesToDelete(oldList, newList []argoproj.ManagedNamespaces) []string {
```
jgwest: Could use some unit tests for `getNamespacesToDelete`

jgwest (approved):
LGTM, thanks @Rizwana777!

PR #1692, opened 2025-03-17
Copying argocd-cmp-server binary instead of argocd which can be dynamically linked for FIPS

anandf:
**What type of PR is this?**
/kind bug
Fix for issue [GITOPS-6495](https://issues.redhat.com/browse/GITOPS-6495)

Error message when starting cmp server with busybox image and a FIPS compliant Argo CD binary.
```
/var/run/argocd/argocd-cmp-server: error while loading shared libraries: libdl.so.2: cannot open shared object file: No such file or directory
```
####Root Cause:
For FIPS compliance, `argocd` binary must be compiled with `CGO_ENABLED=1`, and this mandates it to run with the same OS version in which it was complied so that the libraries linked during compilation are available at runtime as well. So if the binary was compiled in RHEL 8, it may not run on a RHEL 9/10. Since the `argocd-cmp-server` is designed to be run on different images, we need to have a separate binary for `argocd-cmp-server` which is compiled with `CGO_ENABLED=0` so that the binary is statically compiled and can work across different flavours of Linux images.

#### Note:
If the same FIPS compliant image is used, then the `argocd-cmp-server` would work without any issues. Only if you copy the binary image to a volume and run it from a different container image, this issue would occur.

Fixes [GITOPS-6495](https://issues.redhat.com/browse/GITOPS-6495)

Run cmp server using a FIPS compliant Argo CD base image and binary.

### DIFF
--- a/controllers/argocd/deployment.go
+++ b/controllers/argocd/deployment.go
@@ -276,7 +276,7 @@ func getArgoCmpServerInitCommand() []string {
 	cmd := make([]string, 0)
 	cmd = append(cmd, "cp")
 	cmd = append(cmd, "-n")
-	cmd = append(cmd, "/usr/local/bin/argocd")
+	cmd = append(cmd, "/usr/local/bin/argocd-cmp-server")
 	cmd = append(cmd, "/var/run/argocd/argocd-cmp-server")
 	return cmd
 }


svghadi (approved):
The change looks good to me. I am just trying to understand why upstream copies `argocd` binary as argocd-cmp-server when there is a separate `argocd-cmp-server` binary present [here](https://github.com/argoproj/argo-cd/tree/master/cmd/argocd-cmp-server/commands).

https://github.com/argoproj/argo-cd/blob/9c2ffb74c91d722e2d474f28b76a60293ab92a4f/manifests/ha/install.yaml#L26659-L26664
```yaml
      initContainers:
      - command:
        - /bin/cp
        - -n
        - /usr/local/bin/argocd
        - /var/run/argocd/argocd-cmp-server
        image: quay.io/argoproj/argocd:latest
        name: copyutil
```

anandf:
> I am just trying to understand why upstream copies argocd binary as argocd-cmp-server when there is a separate argocd-cmp-server binary present [here](https://github.com/argoproj/argo-cd/tree/master/cmd/argocd-cmp-server/commands).
@svghadi If argocd container image is used for the sidecar container, then the `argocd-cmp-binary` present in the image can be directly used. But if customer chooses to use a different image for reasons like a to have a more recent version of helm or kustomize, then that image may not contain this binary. The test case that is failing in the JIRA ticket uses `busybox` as the sidecar container image. The binary is mounted on a shared volume and run from there.

svghadi:
Okay. Makes sense.

svghadi (approved):
LGTM

PR #1693, opened 2025-03-17
feat: add extra volume and volumeMounts spec to Dex

### LINKED ISSUE #1597: Allow setting volume mounts for dex container
For now it's not possible to mount custom root-ca certificates to the dex container for allowing OIDC connections to servers secured by a private CA.

Add new parameters in _ArgoCD_ crd for modifying the dex deployment:
- _.spec.sso.dex.volumes_
- _.spec.sso.dex.volumeMounts_

Similar settings are already existing for other argocd deployments e.g. 
- _.spec.controller.volumes_ and 
- _.spec.controller.volumeMounts_

As a workaround it is possible:
-  to disable verifying certificates in the dex config. But this isn't secure.
- Volume mounting is also possible with a manual customization of the deployment manifest, after it was created by the operator. But this could lead to a loss of the mount after changing dex config in the _ArgoCD_ crd

miinsun:
**What type of PR is this?**
/kind enhancement

This PR includes the following fields in the DexSpec type:

- .spec.sso.dex.volumes
- .spec.sso.dex.volumeMounts

This ensures that Dex deployment can properly handle volumes and volume mounts.

Fixes #1597 


svghadi:
Hi @miinsun , thanks for the PR. Can you run `make generate manifests bundle` to fix the codegen CI check?

miinsun:
Hi @svghadi , sure! I ran make generate manifests bundle and pushed the updated files.

Just a quick note — on macOS, the sed -i command in the Makefile was causing an error due to the sed syntax. I was able to fix it by changing:

```
❌ sed -i 's/control-plane: argocd-operator/control-plane: ...   

✅ sed -i '' 's/control-plane: argocd-operator/control-plane: ... 
```

miinsun:
Hi @svghadi,

I’m encountering some SSO configuration errors in my E2E runs:

```
2025-05-10T04:27:38Z ERROR controller_argocd  
  Failed to update existing SSO Configuration for ArgoCD “argocd”  
  in namespace kuttl-test-guiding-lacewing  
  {"error":"illegal SSO configuration: cannot supply dex configuration when requested SSO provider is keycloak"}

2025-05-10T04:32:49Z ERROR controller_argocd  
  Failed to update existing SSO Configuration for ArgoCD “argocd”  
  in namespace test-1-042-restricted-pss-compliant  
  {"error":"illegal SSO configuration: must supply valid dex configuration when requested SSO provider is dex"}
```

I originally assumed that simply adding the following fields would be sufficient:
- `spec.sso.dex.volumes`
- `spec.sso.dex.volumeMounts`

However, the E2E failures suggest there’s more to it.

**Do we need to update the ArgoCD SSOConfig CRD (or controller logic) to properly handle these new Dex volumes and volumeMounts fields?** Any guidance or a reference fixture/example would be greatly appreciated!

Thanks 😊

tests/k8s/1-007_validate_volume_mounts/03-install.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: argoproj.io/v1beta1
+kind: ArgoCD
+metadata:
+  name: argocd
+spec:
+  sso:
+    provider: dex
+    dex:
+      volumes:
```
svghadi:
```suggestion
    dex:
      config: test-config
      volumes:
```
This should fix your test failure. The controller is not able to find any dex config hence it fails.

svghadi:
> Just a quick note — on macOS, the sed -i command in the Makefile was causing an error due to the sed syntax. I was able to fix it by changing:

Ah, yes. It's a known issue. We're planning to remove the sed command soon.

tests/k8s/1-007_validate_volume_mounts/03-install.yaml
miinsun: Ah, I see the config setting for the Dex container was missing—thanks for pointing that out!

miinsun:
Hello @svghadi , I’ve updated the test code configuration and run make build.
When you have a moment, could you please take a look?

svghadi (approved):
LGTM!
Just pushed a few small commits to rebase and add some docs.

svghadi:
Thanks a lot for your contribution, @miinsun!

PR #1694, opened 2025-03-18
Copying argocd-cmp-server binary instead of argocd which can be dynamically linked for FIPS (#1692)

anandf:
**What type of PR is this?**
/kind bug

Copy the `argocd-cmp-server` instead of the `argocd` binary which can be compiled dynamically for FIPS compliance and may not run with other container images.

Fixes #?


svghadi (approved):
LGTM

PR #1703, opened 2025-04-02
 Redis HA Server StatefulSet SecurityContext Not Updated During Upgrade

Mangaal:
Description:
After upgrading from GitOps 1.15 to 1.16, the redis service account in GitOps 1.16 is assigned lower SecurityContextConstraints (SCC), and the operator fails to update the securityContext of the redis-ha-server StatefulSet. As a result, the container's user is hardcoded instead of being randomly assigned as required by the restricted-v2 SCC. This prevents the new configuration of the StatefulSet from being applied correctly.

Impact:
The redis-ha-server StatefulSet pods will not be updated with the new settings, causing them to retain old configurations.


nmirasch (approved):
lgtm!

svghadi (approved):
LGTM. Thanks.

PR #1707, opened 2025-04-11
Upgrade to latest prometheus-operator

jgwest:
**What type of PR is this?**
/kind enhancement

We're using an extremely old version of prometheus-operator, which does not contain some of the more recent changes in the various Prometheus operator CRDs.

This PR upgrades to a newer prometheus-operator operator version which I picked so that it should be as compatible as possible with our other dependencies (same go version, same controller-runtime version, compatible kubernetes version).


chetan-rns (approved):
LGTM!

jgwest:
Thanks @chetan-rns!

PR #1712, opened 2025-04-15
fix: ArgoCD SSO status is not updated

jparsai:
**What type of PR is this?**
/kind bug

This PR is to fix a bug where, ArgoCD `Status.Phase` is not updated properly when `SSO` fails.

Fixes https://issues.redhat.com/browse/GITOPS-6667


jgwest (approved):
LGTM, thanks @jparsai!

PR #1715, opened 2025-04-16
Enable JSON logging option for all GitOps components

Rizwana777:
JIRA - https://issues.redhat.com/browse/GITOPS-6669

logFormat is not supported for redis and dex

### DIFF
--- a/api/v1beta1/argocd_types.go
+++ b/api/v1beta1/argocd_types.go
@@ -218,6 +218,7 @@ type ArgoCDApplicationSet struct {
 	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
 
 	// LogFormat refers to the log format used by the ApplicationSet component. Defaults to ArgoCDDefaultLogFormat if not configured. Valid options are text or json.
+	// +kubebuilder:validation:Enum=text;json
 	LogFormat string `json:"logformat,omitempty"`
 }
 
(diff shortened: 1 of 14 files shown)


api/v1beta1/argocd_types.go
```diff
@@ -216,6 +216,9 @@ type ArgoCDApplicationSet struct {
 
 	// VolumeMounts adds volumeMounts to the Argo CD ApplicationSet Controller container.
 	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
+
+	// LogFormat refers to the log format used by the ApplicationSet component. Defaults to ArgoCDDefaultLogFormat if not configured. Valid options are text or json.
+	LogFormat string `json:"logformat,omitempty"`
```
svghadi, line 221: Can we try using kubebuilder marker to validate the input for `text` & `json` values?
- https://book.kubebuilder.io/reference/markers/crd-validation 
- https://github.com/kubernetes-sigs/kubebuilder/issues/2812#issuecomment-1179936665

```suggestion
	// LogFormat refers to the log format used by the ApplicationSet component. Defaults to ArgoCDDefaultLogFormat if not configured. Valid options are text or json.
	// +kubebuilder:validation:Enum=text;json
	LogFormat string `json:"logformat,omitempty"`
```

jgwest (approved):
LGTM too, thanks @Rizwana777!

PR #1722, opened 2025-04-24
Add support in appendUniqueArgs to allow flag with multiple values

Rizwana777:
**What type of PR is this?**
`appendUniqueArgs` func update
- This function merges two slices of CLI arguments (cmd and extraArgs) while avoiding duplication and ensuring correct override behavior.
It handles the following scenarios:
- Preserves argument order, starting with all args from cmd.
- Flags in cmd are treated as non-repeatable — if the same flag appears in extraArgs, it replaces the one from cmd. (To update the existing flags I have added this logic orelse the multiple values for single flag logic will cause bug here)
- Skips duplicate flag+value pairs if already present.
- Allows repeatable flags with different values, only if the flag is not already in cmd. (flags in cmd are preexisting and is non-repeatable)
- Correctly pairs flags with their values (e.g., --flag value).
- Safely handles flags with no values (e.g., --flag alone). eg - --hydrator-enabled

Fixed and updated unit tests and e2e test to work accordingly

JIRA - https://issues.redhat.com/browse/GITOPS-6703

### DIFF
--- a/controllers/argocd/applicationset_test.go
+++ b/controllers/argocd/applicationset_test.go
@@ -1035,12 +1035,12 @@ func TestArgoCDApplicationSetCommand(t *testing.T) {
 	wantCmd := []string{
 		"entrypoint.sh",
 		"argocd-applicationset-controller",
-		"--argocd-repo-server",
-		"foo.scv.cluster.local:6379",
 		"--loglevel",
 		"info",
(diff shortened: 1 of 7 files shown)


jgwest (approved):
LGTM, thanks @Rizwana777!

PR #1726, opened 2025-04-30
Fix: Process regular expressions at getSourceNamespaces

nmirasch:
**What type of PR is this?**

/kind bug 

Updates the  operator code to use new regex option using glob.REGEXP instead of glob.GLOB when processing getSourceNamespaces

Fixes https://issues.redhat.com/browse/GITOPS-6675


svghadi (approved):
LGTM. Thanks

PR #1727, opened 2025-05-01
Fix linter warnings including deprecation warning SA1019

jgwest:
**What type of PR is this?**
/kind chore

- This PR fixes a number of warnings within the argocd-operator codebase.
- The main one being 'SA1019', which is reported against our own deprecated APIs.
    - We were previously using `//nolint:staticcheck` for this, but `//lint:ignore SA1019 known to be deprecated` appears to be more effective
    - Let me know if this differs for you in your IDE of choice (I tested from VS Code)
- Also fixed ginkgo import warning, sprintf warning, and more.


jgwest:
Thanks @jparsai!

PR #1728, opened 2025-05-03
IsObjectFound function should not bury unexpected errors

jgwest:
**What type of PR is this?**
/kind bug 

Currently, the `IsObjectFound` utility function is used to check whether a particular object exists on the cluster. It is used across the codebase, and with many different types of objects.

Unfortunately, `IsObjectFound` is making an incorrect assumption: this function assumes that ANY error that occurs (besides is not found) should be treated the same as the object not existing.

For example:

| Cause      | Return value      |
| ------------- | ------------- |
|A) No error on Get object call|`IsObjectFound` returns true|
|B) Not found error on Get object call|`IsObjectFound` returns false|
|C) Any other error occurs on Get object call (network error, permission error, TLS error, etc)|`IsObjectFound` returns false|

Notice that B and C, despite very different causes, return the same false of `false`.

This PR updates the function signature of `IsObjectFound` to return (rather than burying) the `error`, and updates all locations in the code where it is called.


svghadi (approved):
LGTM.

Thanks @jgwest.

jgwest:
Thanks @svghadi!

PR #1736, opened 2025-05-15
[release-0.13] Avoid spec.RBAC.scope sync with rbac configmap when sso keycloak (#1667)

nmirasch:
cherry-pick https://github.com/argoproj-labs/argocd-operator/pull/1667

(cherry picked from commit b1ec29f77a3741f07355d554c7f8ea4e7d3a4528) 

This PR adds logic to preserve the RBAC scope configuration in the RBAC ConfigMap when updating to SSO Keycloak. Previously, it was set to '[groups,email]' in the RBAC ConfigMap without updating the Spec.RBAC.Scopes, which caused a conflict during the ConfigMap sync when switching to Keycloak SSO and updating the ArgoCDConfiguration.


svghadi (approved):
LGTM

PR #1737, opened 2025-05-15
GITOPS-6285: reconcile operator owned labels for configMap and ignore non operator labels

alkakumari016:
This PR adds reconcile logic for operator owned labels of ConfigMaps. The labels are reconciled to it's desired values whenever their is any change and ignores the labels that are not owned by the operator

https://issues.redhat.com/browse/GITOPS-6285

Fixes #?


controllers/argocd/util.go
```diff
@@ -1752,6 +1752,20 @@ func addKubernetesData(source map[string]string, live map[string]string) {
 	}
 }
 
+func AddExistingLabels(existing *map[string]string, expected map[string]string) bool {
```
svghadi: Can we make the function name generic so that it can be used with annotations aswell? Maybe

```suggestion
func UpdateMapValues(existing *map[string]string, expected map[string]string) bool {
```

svghadi: Also can you add a unit test for this function?

svghadi (changes requested):
Thanks @alkakumari016. Left some suggestions. 

Let's remove the `addKubernetesData` function and replace its references with your new function. I'm also thinking of limiting the changes to only those places where we're already updating labels and annotations, to keep the testing scope small.
We can introduce label/annotation checks on resources that don’t currently have them in a separate PR.

controllers/argocd/util.go
alkakumari016: Changed the func name and added test case

controllers/argocd/util.go
```diff
@@ -1727,29 +1727,21 @@ func UseServer(name string, cr *argoproj.ArgoCD) bool {
...
-		found := glob.MatchStringInList(patterns, key, glob.GLOB)
-		if found {
-			// Don't override values already present in the source object.
-			// This allows the operator to update Kubernetes specific data when needed.
-			if _, ok := source[key]; !ok {
-				source[key] = value
-			}
+// UpdateMapValues checks for any Kubernetes-specific labels or annotations
+// in the existing and source object and updates the existing object.
+// This helps avoid loss of important Kubernetes-managed or non-Kubernetes-managed labels/annotations during updates.
```
svghadi: The comment doesn't describe the function properly. It checks 2 maps and update values in existing map with value from source if they differ. Can you update the comment to reflect this?

controllers/argocd/statefulset.go
```diff
@@ -972,8 +972,8 @@ func (r *ReconcileArgoCD) reconcileApplicationControllerStatefulSet(cr *argoproj
 		}
 
 		// Add Kubernetes-specific labels/annotations from the live object in the source to preserve metadata.
```
svghadi: We can drop this comment or else just mention about adding expected labels from source object onto live. We are no longer handing k8s specific data explicitly.

controllers/argocd/deployment.go
```diff
@@ -1345,9 +1345,9 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argoproj.ArgoCD, useTLSFor
 			changed = true
 		}
 
-		// Add Kubernetes-specific labels/annotations from the live object in the source to preserve metadata.
-		addKubernetesData(deploy.Spec.Template.Labels, existing.Spec.Template.Labels)
-		addKubernetesData(deploy.Spec.Template.Annotations, existing.Spec.Template.Annotations)
+		//  Add Kubernetes-specific labels/annotations from the source object in the live and preserve non-operator specific labels/anotations.
```
svghadi: Same as previous comment.

controllers/argocd/configmap.go
```diff
@@ -332,6 +332,22 @@ func (r *ReconcileArgoCD) reconcileConfigMaps(cr *argoproj.ArgoCD, useTLSForRedi
 // This ConfigMap holds the CA Certificate data for client use.
 func (r *ReconcileArgoCD) reconcileCAConfigMap(cr *argoproj.ArgoCD) error {
 	cm := newConfigMapWithName(getCAConfigMapName(cr), cr)
+	existingCM := &corev1.ConfigMap{}
+	if argoutil.IsObjectFound(r.Client, cr.Namespace, cm.Name, existingCM) {
```
svghadi: You will need to update this call according to new func signature. See #1728 .

controllers/argocd/configmap.go
```diff
@@ -332,6 +332,22 @@ func (r *ReconcileArgoCD) reconcileConfigMaps(cr *argoproj.ArgoCD, useTLSForRedi
...
+		changed := false
+		//Check if labels have changed
```
svghadi, line 344: This is a nice comment. We can use this at places mentioned in previous comments.

controllers/argocd/applicationset.go
```diff
@@ -269,18 +269,17 @@ func (r *ReconcileArgoCD) reconcileApplicationSetDeployment(cr *argoproj.ArgoCD,
 	AddSeccompProfileForOpenShift(r.Client, podSpec)
 
 	if exists {
-		// Add Kubernetes-specific labels/annotations from the live object in the source to preserve metadata.
-		addKubernetesData(deploy.Spec.Template.Labels, existing.Spec.Template.Labels)
-		addKubernetesData(deploy.Spec.Template.Annotations, existing.Spec.Template.Annotations)
+		// Add Kubernetes-specific annotations from the source object in the live object and preserve non-operator labels.
```
svghadi: Update or drop this comment.

controllers/argocd/deployment.go
```diff
@@ -1730,9 +1729,9 @@ func (r *ReconcileArgoCD) reconcileServerDeployment(cr *argoproj.ArgoCD, useTLSF
 			}
 		}
 
-		// Add Kubernetes-specific labels/annotations from the live object in the source to preserve metadata.
-		addKubernetesData(deploy.Spec.Template.Labels, existing.Spec.Template.Labels)
-		addKubernetesData(deploy.Spec.Template.Annotations, existing.Spec.Template.Annotations)
+		// Add Kubernetes-specific labels/annotations from the source object in the live and preserve non-operator specific labels/anotations.
```
svghadi: Same as previous comment

PR #1738, opened 2025-05-15
Add ARGOCD_API_SERVER_REPLICAS env when .spec.server.replicas is conf…

Rizwana777:
JIRA - https://issues.redhat.com/browse/GITOPS-6674

### DIFF
--- a/controllers/argocd/deployment.go
+++ b/controllers/argocd/deployment.go
@@ -1596,6 +1596,15 @@ func (r *ReconcileArgoCD) reconcileServerDeployment(cr *argoproj.ArgoCD, useTLSF
 
 	if replicas := getArgoCDServerReplicas(cr); replicas != nil {
 		deploy.Spec.Replicas = replicas
+
+		// Add ARGOCD_API_SERVER_REPLICAS env var to the argocd-server container
+		deploy.Spec.Template.Spec.Containers[0].Env = append(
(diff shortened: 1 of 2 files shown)


controllers/argocd/deployment.go
```diff
@@ -1596,6 +1596,15 @@ func (r *ReconcileArgoCD) reconcileServerDeployment(cr *argoproj.ArgoCD, useTLSF
 
 	if replicas := getArgoCDServerReplicas(cr); replicas != nil {
 		deploy.Spec.Replicas = replicas
+
+		// Add ARGOCD_API_SERVER_REPLICAS env var to the argocd-server container
+		deploy.Spec.Template.Spec.Containers[0].Env = append(
+			deploy.Spec.Template.Spec.Containers[0].Env,
+			corev1.EnvVar{
+				Name:  "ARGOCD_API_SERVER_REPLICAS",
+				Value: fmt.Sprintf("%d", *replicas),
```
svghadi, line 1605: Append will take care of edge case where `ARGOCD_API_SERVER_REPLICAS` is set by both user and operator. The last value take effect in case of conflict in pods. Here, operator adds the value after user supplied env (`.spec.server.env`) are added  to deployment.

svghadi (approved):
LGTM. Thanks.

jgwest:
Thanks @Rizwana777 and @svghadi

PR #1742, opened 2025-05-28
Update redis-ha-haproxy replicas similar to redis

Rizwana777:
JIRA - https://issues.redhat.com/browse/GITOPS-6806

### DIFF
--- a/controllers/argocd/deployment.go
+++ b/controllers/argocd/deployment.go
@@ -650,6 +650,8 @@ func (r *ReconcileArgoCD) reconcileRedisHAProxyDeployment(cr *argoproj.ArgoCD) e
 		},
 	})
 
+	deploy.Spec.Replicas = getRedisHAReplicas()
+
 	deploy.Spec.Template.Spec.Affinity = &corev1.Affinity{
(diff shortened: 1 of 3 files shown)


svghadi (commented):
Have some questions, other than that looks good. Thanks.
- Do we need any updates in https://github.com/argoproj-labs/argocd-operator/blob/master/build/redis/haproxy.cfg.tpl  to support multiple replicas? 
- Can you run some tests to verify that traffic is routed correctly when one HAProxy replica goes down?

Rizwana777:
> Have some questions, other than that looks good. Thanks.
> 
> * Do we need any updates in https://github.com/argoproj-labs/argocd-operator/blob/master/build/redis/haproxy.cfg.tpl  to support multiple replicas?
> * Can you run some tests to verify that traffic is routed correctly when one HAProxy replica goes down?

tested

svghadi (approved):
LGTM. Thanks

jgwest:
Thanks @Rizwana777 and thanks @svghadi and @jparsai for reviewing!

PR #1749, opened 2025-06-05
Enable ApplicationFineGrainedRBACInheritance by default

nmirasch:
In Argo CD 3.0, fine grained RBAC inheritance of Applications have been disable.  This PR updates the operator codebase to set the new flag `server.rbac.disableApplicationFineGrainedRBACInheritance=false` in `argocd-cm` configmap to preserve the old RBAC behaviour. Users can override this behaviour using .spec.extraConfig in ArgoCD CR if they want to use the new fine grained rbac functionality.

https://issues.redhat.com/browse/GITOPS-6889

### DIFF
--- a/common/keys.go
+++ b/common/keys.go
@@ -193,6 +193,10 @@ const (
 	// to used for the Redis container.
 	ArgoCDRedisImageEnvName = "ARGOCD_REDIS_IMAGE"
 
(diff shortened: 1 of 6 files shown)


svghadi (approved):
LGTM. Thanks

PR #1751, opened 2025-06-10
Add default resource.exclusions in ConfigMap

Rizwana777:
JIRA - https://issues.redhat.com/browse/GITOPS-6890

As mentioned [here](https://github.com/argoproj-labs/argocd-operator/blob/master/api/v1alpha1/argocd_types.go#L786) customizations field is deprecated in v1beta1  and in the ticket I see they have added customizations - https://github.com/argoproj/argo-cd/pull/21760/files#diff-d4871b2556f24756487ca6be13230ab1bcba17792d8fca91ca138d9b9727f614R13, does this mean this field is not deprecated? If not i will add the logic to add defaults for customizations as well

### DIFF
--- a/controllers/argocd/configmap.go
+++ b/controllers/argocd/configmap.go
@@ -216,13 +216,25 @@ func getResourceActions(cr *argoproj.ArgoCD) map[string]string {
 	return action
 }
 
-// getResourceExclusions will return the resource exclusions for the given ArgoCD.
+// getResourceExclusions returns resource exclusions from the CR or defaults if not set.
 func getResourceExclusions(cr *argoproj.ArgoCD) string {
-	re := common.ArgoCDDefaultResourceExclusions
+	var resources []filteredResource
+
+	// Use CR value if provided
 	if cr.Spec.ResourceExclusions != "" {
-		re = cr.Spec.ResourceExclusions
+		return cr.Spec.ResourceExclusions
+	} else {
+		// Use defaults
+		resources = getDefaultResourceExclusions()
 	}
-	return re
+
+	finalYaml, err := marshalWithYAML(resources)
+	if err != nil {
+		log.Error(err, "Failed to marshal resource exclusions")
+		return ""
+	}
+
+	return finalYaml
 }
 
 // getResourceInclusions will return the resource inclusions for the given ArgoCD.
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -30,6 +30,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.17.2
 )
 
+require sigs.k8s.io/yaml v1.4.0
```
anandf: Can this be moved to the already existing `require` block

Rizwana777: Done

controllers/argocd/configmap.go
```diff
@@ -29,6 +31,8 @@ import (
 	argoproj "github.com/argoproj-labs/argocd-operator/api/v1beta1"
 	"github.com/argoproj-labs/argocd-operator/common"
 	"github.com/argoproj-labs/argocd-operator/controllers/argoutil"
+
+	k8syaml "sigs.k8s.io/yaml"
```
anandf: can we not use `gopkg.in/yaml.v2` which is already imported ?

controllers/argocd/configmap.go
```diff
@@ -216,13 +220,32 @@ func getResourceActions(cr *argoproj.ArgoCD) map[string]string {
...
 	if cr.Spec.ResourceExclusions != "" {
-		re = cr.Spec.ResourceExclusions
+		if err := k8syaml.Unmarshal([]byte(cr.Spec.ResourceExclusions), &userExclusions); err != nil {
+			// Log error but fallback to defaults
+			log.Info("Failed to parse ResourceExclusions from CR; using defaults", "error", err)
+		}
 	}
-	return re
+
+	merged := mergeResourceExclusions(defaults, userExclusions)
```
anandf: We can skip this merge resource exclusions.

controllers/argocd/configmap.go
```diff
@@ -945,3 +968,58 @@ func (r *ReconcileArgoCD) reconcileArgoCmdParamsConfigMap(cr *argoproj.ArgoCD) e
...
+
+		if _, exists := seen[key]; !exists {
+			seen[key] = struct{}{}
+			merged = append(merged, r)
+		}
+	}
+	return merged
+}
+
+func marshalWithJSONTags(resources []filteredResource) (string, error) {
```
anandf: We can remove this I think.

controllers/argocd/configmap.go
```diff
@@ -945,3 +968,58 @@ func (r *ReconcileArgoCD) reconcileArgoCmdParamsConfigMap(cr *argoproj.ArgoCD) e
...
+		{APIGroups: []string{"cilium.io"}, Kinds: []string{"CiliumIdentity", "CiliumEndpoint", "CiliumEndpointSlice"}},
+		{APIGroups: []string{"kyverno.io", "reports.kyverno.io", "wgpolicyk8s.io"},
+			Kinds: []string{
+				"PolicyReport", "ClusterPolicyReport", "EphemeralReport", "ClusterEphemeralReport",
+				"AdmissionReport", "ClusterAdmissionReport", "BackgroundScanReport",
+				"ClusterBackgroundScanReport", "UpdateRequest"}},
+	}
+}
+
+func mergeResourceExclusions(defaults, user []filteredResource) []filteredResource {
```
anandf: This can be removed I think, now that we will override whatever user provides in the `ArgoCD` CR.

anandf: If user provides a resource exclusion value in the CR, that will be used as it is with no merge with default values. If customers want the merge behaviour, they can still update the merged value.

controllers/argocd/configmap.go
Rizwana777: I have updated this func to `marshalWithYAML` as we need this because getResourceExclusions returns string and cr.spec.resourceExclusions is string as well, we need to convert the yaml to string to get the expected output from defaults

jgwest (approved):
LGTM, thanks @Rizwana777!

PR #1756, opened 2025-06-16
Create argocd-cmd-params-cm configMap

Rizwana777:
JIRA - https://issues.redhat.com/browse/GITOPS-6326


controllers/argocd/configmap.go
```diff
@@ -789,3 +793,41 @@ func (r *ReconcileArgoCD) reconcileGPGKeysConfigMap(cr *argoproj.ArgoCD) error {
...
+	if err := controllerutil.SetControllerReference(cr, cm, r.Scheme); err != nil {
+		return err
+	}
+	existingCM := &corev1.ConfigMap{}
+	if argoutil.IsObjectFound(r.Client, cr.Namespace, cm.Name, existingCM) {
+		changed := false
+		if !reflect.DeepEqual(cm.Data, existingCM.Data) {
+			existingCM.Data = cm.Data
+			changed = true
+		}
```
svghadi: We’ll either need to allow the data change directly in the ConfigMap or provide a new field in the ArgoCD CR to update the data in ConfigMap.

Rizwana777: I will add a new key to the ArgoCD CR to manage the data in the ConfigMap. I believe allowing direct/manual edits to the ConfigMap is not ideal? WDYT @svghadi 
New field in CR -
```
spec:
  cmdParams:
    application.namespaces: "default"
    redis.compression: gzip
```

svghadi: Yes. Makes sense to disallow direct edits. 

What do you think about below structure? It would allow us to move the extraConfig field (which currently corresponds to `argocd-cm`) here in a future release. Right now, we have a number of fields scattered across the ArgoCD CRD, and I am thinking of reorganizing them into a more grouped and structured format.
```yaml
spec:
  configMaps:
    argocd-cmd-params-cm:
      application.namespaces: "default"
      controller.log.level: "debug"
    argocd-cm:
      timeout.reconciliation: "180s"
      kustomize.buildOptions: "--load-restrictor LoadRestrictionsNone"
```

@anandf @jgwest - Thoughts? I was thinking of to keeping the structure similar to upstream to avoid confusion for users, since we refer them to the upstream docs for configuration.

anandf: Rather than exposing the underlying CM name, I prefer something abstract like below:
```
spec:
  config: (instead of extraConfig)
    application.namespaces: "default"
    controller.log.level: "debug"
  cmd-params:
    timeout.reconciliation: "180s"
    kustomize.buildOptions: "--load-restrictor LoadRestrictionsNone"
```
This can help avoid changes to ArgoCD CR spec if the upstream decides to change the ConfigMap names although its very unlikely.

svghadi: Let's go with `spec.cmdParams` for now to unblock things. I will put together a proposal for reorganizing the fields, and we can revisit the discussion then.

Rizwana777: Added the new field and I will add logic to avoid duplicates and overriding default values when I work on this - https://issues.redhat.com/browse/GITOPS-7009 , here I will add the defaults and then I will update the logic and add unit test as well for the same

PR #1761, opened 2025-06-19
fix(bug): Truncate SourceNamespaces with long project name causing reconciler error

aali309:
**What type of PR is this?**

/kind bug

This PR fixes the issue where ArgoCD Operator fails to create RoleBindings when using long namespace names in sourceNamespaces. The issue occurs because Kubernetes labels have a 63-character limit, but the operator was trying to use the full namespace name as a label value, causing validation errors.

The fix implements a truncateWithHash function that:

- Truncates long names to fit within the 63-character limit

- Appends a deterministic hash suffix to maintain uniqueness

- Ensures RoleBinding names remain unique even when truncated

Fixes #?

Create an ArgoCD CR with a long namespace name (>63 chars) in sourceNamespaces and observe the RoleBinding creation and logs
 verify that:

- RoleBinding is created successfully with a truncated name

- The name is exactly 63 characters

- No validation errors occur


controllers/argocd/rolebinding.go
```diff
@@ -19,6 +20,32 @@ import (
...
 )
 
+// truncates a string to maxLabelLength characters and adds a hash suffix to ensure uniqueness
+func truncateWithHash(input string) string {
+	if len(input) <= maxLabelLength {
+		return input
+	}
+
+	// Calculate hash of the original string
+	hash := sha256.Sum256([]byte(input))
```
svghadi: Any specific reason to use `SHA-256` here? Since we only need uniqueness (not security), I think `SHA-1` should be sufficient, it's faster and has a smaller output. Also, would it make sense to append a 7-character hash instead of just 4 to reduce the chance of collisions?

tests/k8s/1-051_validate_long_namespace_rolebinding/02-verify-rolebindings.yaml
```diff
@@ -0,0 +1,83 @@
...
+    if [ "$phase" == "Available" ]; then
+      echo "✅ ArgoCD is Available."
+      break
+    fi
+    if [ $i -eq 30 ]; then
+      echo "❌ Timed out waiting for ArgoCD to become Available"
+      exit 1
+    fi
+    sleep 10
+  done
```
svghadi:
```suggestion
```
You can remove this check by setting `phase: Available` in `01-assert.yaml`

tests/k8s/1-051_validate_long_namespace_rolebinding/01-assert.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ArgoCD
+metadata:
+  name: argocd-long-namespace
+  namespace: argocd-long-namespace-test
+status:
+  phase: Pending
```
svghadi:
```suggestion
  phase: Available
```

tests/k8s/1-051_validate_long_namespace_rolebinding/02-verify-rolebindings.yaml
svghadi, line 1: I believe the unit tests already cover most of the logic. To simplify the e2e tests, we could update the existing `tests/k8s/1-024_validate_apps_in_any_namespace` test to include this case. Since SHA hashes are deterministic, we can compute them locally and assert against the shortened RoleBinding name. This would validate both the deterministic behavior and the name shortening logic.

svghadi (commented):
Thanks @aali309 . Left some comments.

controllers/argocd/rolebinding.go
aali309: yes make sense!

tests/k8s/1-024_validate_apps_in_any_namespace/02-verify-long-namespace-rolebindings.yaml
svghadi, line 1: Let's simplify this check. You can delete this file and instead add a new entry to tests/k8s/1-024_validate_apps_in_any_namespace/01-assert.yaml to validate the RoleBinding. The following should be sufficient to verify the naming:

```yaml
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: example-argocd_longns-abcdefghijklmnopqrstuvwxyz-123456789012-9a19a95
  namespace: longns-abcdefghijklmnopqrstuvwxyz-123456789012345
```

aali309: I tried this, but it fails. The problem is that static assertions in KUTTL expect resources to exist immediately, but the operator needs time to reconcile and create the RoleBinding. That's why I chose to got with the dynamic bash script approach.

svghadi: Oh, got it.  Learnt something new! Looks good then. Thanks!

svghadi (approved):
LGTM

controllers/argocd/rolebinding.go
```diff
@@ -19,6 +20,32 @@ import (
 	"github.com/argoproj-labs/argocd-operator/controllers/argoutil"
 )
 
+// truncates a string to maxLabelLength characters and adds a hash suffix to ensure uniqueness
```
anandf: nit pick
// truncateWithHash truncates a string to a maximum of 63 characters and adds a hash suffix to ensure uniqueness

controllers/argocd/rolebinding.go
```diff
@@ -19,6 +20,32 @@ import (
...
+func truncateWithHash(input string) string {
+	if len(input) <= maxLabelLength {
```
anandf, line 29: Can `maxLabelLength` be kept a constant of value `63` ? and hash label length as `7`. We can avoid checks like 
``` if maxBaseLength <= 0```

aali309: `maxLabelLength` had been defined as a constant in `route.go`

anandf (approved):

svghadi (approved):

PR #1763, opened 2025-06-23
Add default health value in CM to get Health status in the Application CR

Rizwana777:
JIRA - https://issues.redhat.com/browse/GITOPS-7009

Steps to test -
Update [this](https://github.com/argoproj-labs/argocd-operator/blob/master/common/defaults.go#L73) value to ArgoCD version 3.0.0  
then apply :
kind: ArgoCD
metadata:
  name: example-argocd
spec:
  cmdParams:
    controller.resource.health.persist: "true"

Create an Application CR to see the CR status of application is working as expected

### DIFF
--- a/controllers/argocd/configmap.go
+++ b/controllers/argocd/configmap.go
@@ -800,18 +800,15 @@ func (r *ReconcileArgoCD) reconcileArgoCmdParamsConfigMap(cr *argoproj.ArgoCD) e
 	cm := newConfigMapWithName(common.ArgoCDCmdParamsConfigMapName, cr)
 	cm.Data = make(map[string]string)
 
+	// Set default for controller.resource.health.persist to "true"
+	const healthPersistKey = "controller.resource.health.persist"
+	cm.Data[healthPersistKey] = "true"
+
 	// Copy user-specified command parameters if any
 	if len(cr.Spec.CmdParams) > 0 {
 		for k, v := range cr.Spec.CmdParams {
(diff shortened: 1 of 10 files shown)


controllers/argocd/configmap.go
```diff
@@ -800,16 +800,24 @@ func (r *ReconcileArgoCD) reconcileArgoCmdParamsConfigMap(cr *argoproj.ArgoCD) e
...
 		}
 	}
 
+	// Set default for controller.resource.health.persist to "true"
+	// only if user hasn't provided it explicitly
+	const healthPersistKey = "controller.resource.health.persist"
+	if _, exists := cm.Data[healthPersistKey]; !exists {
+		cm.Data[healthPersistKey] = "true"
+	}
+
```
svghadi:
```suggestion
         // Set default for controller.resource.health.persist to "true"
         const healthPersistKey = "controller.resource.health.persist"
         cm.Data[healthPersistKey] = "true"
         
	// Copy user-specified command parameters if any
	if len(cr.Spec.CmdParams) > 0 {
		for k, v := range cr.Spec.CmdParams {
			cm.Data[k] = v
		}
	}
```

Just a nit to remove the conditional logic, not a blocker for merging. Everything else looks good!

Rizwana777: Done

jgwest (approved):
LGTM too, thanks @Rizwana777 and thanks @svghadi for reviewing!

PR #1765, opened 2025-06-23
feat: Support Argo-agent Principal Component in operator

jparsai:
**What type of PR is this?**
/kind enhancement

This PR is to add functionality to enable Argo CD Agent's Principal component from Argo CD resource.

**Jira Ticket link**:
https://issues.redhat.com/browse/GITOPS-7010


controllers/argocdagent/configmap.go
```diff
@@ -0,0 +1,145 @@
...
+func buildData() map[string]string {
+	expectedData := make(map[string]string)
+
+	// Basic configuration
+	expectedData[PrincipalAllowedNamespaces] = getPrincipalAllowedNamespaces()
+	expectedData[PrincipalAuth] = getPrincipalAuth()
+	expectedData[PrincipalJwtAllowGenerate] = getPrincipalJwtAllowGenerate()
+	expectedData[PrincipalLogLevel] = getPrincipalLogLevel()
+	expectedData[PrincipalMetricsPort] = getPrincipalMetricsPort()
+	expectedData[PrincipalNamespace] = getPrincipalNamespace()
```
jgwest: If I'm not mistaken, principal namespace should always be the namespace that the `ArgoCD` CR is defined in, by default

controllers/argocdagent/utils.go
```diff
@@ -0,0 +1,279 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	logr "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var log = logr.Log.WithName("controller_agent")
+
+const (
```
jgwest: IMHO you can move all these values/functions into `configmap.go` since they are only relevant for the ConfigMap logic (correct me if wrong)

controllers/argocdagent/README.md
```diff
@@ -0,0 +1,33 @@
...
+kind: ArgoCD
+metadata:
+  name: argocd
+spec: 
+  argoCDAgent:
+    principal:
+      enabled: true
+```
+
+5. Copy Principal LoadBalancer host and use it in Agent's ConfigMap `agent.server.address` to allow it to communicate with principal.
```
jgwest: In the agent config map, the loadbalancer for principal argo cd repo server, and principal argo cd redis, should also be updated:

For example, [from dev-env](https://github.com/argoproj-labs/argocd-agent/blob/3346a14d9cf0ec2856952c08a5b545b8aa9170b1/hack/dev-env/agent-managed/argocd-cmd-params-cm.yaml#L6):
```
  repo.server: (load balancer address of prinicpal's argo cd's repo-server-address):8081
  redis.server: (load balancer address of principal's argo cd redis):6379
```

controllers/argocdagent/create-agent-config.sh
```diff
@@ -0,0 +1,66 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eo pipefail
+
+RECREATE="$1"
+
+export ARGOCD_AGENT_PRINCIPAL_NAMESPACE=argocd
```
jgwest: Doesn't this require the principal to be installed in `argocd` namespace?

controllers/argocdagent/create-agent-config.sh
```diff
@@ -0,0 +1,66 @@
...
+if ! argocd-agentctl pki inspect >/dev/null 2>&1; then
+	argocd-agentctl pki init
+	echo "  -> PKI initialized."
+else
+	echo "  -> Reusing existing agent PKI."
+fi
+
+echo "[*] Creating principal TLS configuration"
+argocd-agentctl pki issue principal --upsert \
+	--principal-namespace argocd \
```
jgwest: Assumes principal namespace is `argocd`

controllers/argocdagent/create-agent-config.sh
```diff
@@ -0,0 +1,66 @@
...
+	--ip "127.0.0.1,${IPADDR}"
+echo "  -> Principal TLS config created."
+
+echo "[*] Creating resouce proxy TLS configuration"
+argocd-agentctl pki issue resource-proxy --upsert \
+	--principal-namespace argocd \
```
jgwest: Assumes principal namespace is `argocd`

controllers/argocdagent/create-agent-config.sh
```diff
@@ -0,0 +1,66 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eo pipefail
+
+RECREATE="$1"
+
+export ARGOCD_AGENT_PRINCIPAL_NAMESPACE=argocd
+
+IPADDR=$(ip r show default | sed -e 's,.*\ src\ ,,' | sed -e 's,\ metric.*$,,')
```
jgwest: If I'm not mistaken, when principal is running on the cluster, the `IPADDR` value should be the name of the principal Service e.g. `( crName )-agent-principal`.

The current line assumes that the principal is running on the local machine.

api/v1alpha1/argocd_types.go
```diff
@@ -990,6 +993,22 @@ type WebhookServerSpec struct {
...
 }
 
+type ArgoCDAgentSpec struct {
+
+	// Principal defines configurations for the Principal component of ArgoCD Agent.
+	Principal *PrincipalSpec `json:"principal,omitempty"`
+}
+
+type PrincipalSpec struct {
+
```
jgwest, line 1004: We should add some configuration paramters to `.spec.argoCDAgent.principal`:
e.g.
```yaml
kind: ArgoCD
spec:
  argoCDAgent:
    principal:
      allowedNamespaces: "*"
```

At a minimum, we should add configuration parameters for the following parameters to `.spec.argoCDAgent.principal`
- ARGOCD_PRINCIPAL_ALLOWED_NAMESPACES / --allowed-namespaces / principal.allowed-namespaces
- ARGOCD_PRINCIPAL_JWT_ALLOW_GENERATE / --insecure-jwt-generate / principal.jwt.allow-generate
- ARGOCD_PRINCIPAL_LOG_LEVEL / --log-level / principal.log.level
- ARGOCD_PRINCIPAL_AUTH / --auth / principal.auth

There may be others but these are the ones we are using in `start-principal.sh` in dev-env.


(Unless you were thinking the user would use the ConfigMap for this? In which case, the logic for ConfigMap reconcile needs to be updated)

controllers/argocdagent/deployment.go
```diff
@@ -0,0 +1,476 @@
...
+
+func buildPrincipalImage() string {
+	image := os.Getenv("ARGOCD_AGENT_PRINCIPAL_IMAGE")
+	if image == "" {
+		image = "quay.io/openshift-gitops/argocd-agent-principal:v1"
+	}
+	return image
+}
+
+func isDeploymentChanged(compName, saName string, cr *argoproj.ArgoCD, deployment *appsv1.Deployment) (*appsv1.Deployment, bool) {
```
jgwest: Can you update this function name to something like `updateDeploymentIfChanged` to indicate that it will update Deployment, not just detect if it has changed?

controllers/argocdagent/role.go
```diff
@@ -0,0 +1,164 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func ReconcilePrincipalRole(client client.Client, compName string, cr *argoproj.ArgoCD, scheme *runtime.Scheme) (*v1.Role, error) {
+	role := buildRole(compName, cr)
+	expectedPolicyRule := buildPolicyRuleForPrincipal()
```
jgwest: It seems like we are creating both a Role and a ClusterRole that have the exact same content? 

Should they instead contain the content from https://github.com/argoproj-labs/argocd-agent/blob/main/install/kubernetes/principal/principal-clusterrole.yaml and https://github.com/argoproj-labs/argocd-agent/blob/main/install/kubernetes/principal/principal-role.yaml?

controllers/argocdagent/role.go
```diff
@@ -0,0 +1,164 @@
...
+func buildClusterRole(compName string, cr *argoproj.ArgoCD) *v1.ClusterRole {
+	return &v1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   generateAgentResourceName(cr.Name+"-"+cr.Namespace, compName),
+			Labels: buildLabelsForAgentPrincipal(cr.Name),
+		},
+	}
+}
+
+func buildPolicyRuleForPrincipal() []v1.PolicyRule {
```
jgwest: As above, this seems different from what is define in `install/kubernetes/principal` for roles

controllers/argocdagent/create-agent-config.sh
```diff
@@ -0,0 +1,66 @@
...
+
+if ! test -x argocd-agentctl; then
+	echo "Please build argocd-agentctl first by running 'make cli'" >&2
```
jgwest: This message should probably be something more like 'Please ensure argocd-agentctl is on your path', since this repo (argocd-operator) won't be able to build `argocd-agentctl` via `make cli`

controllers/argocdagent/create-agent-config.sh
```diff
@@ -0,0 +1,66 @@
...
+	if test "$RECREATE" = "--recreate"; then
+		echo "  -> Deleting existing cluster secret, if it exists"
+		kubectl -n ${ARGOCD_AGENT_PRINCIPAL_NAMESPACE} delete --ignore-not-found secret cluster-${agent}
+	fi
+	if ! argocd-agentctl agent inspect ${agent} >/dev/null 2>&1; then
+		echo "  -> Creating cluster secret for agent configuration"
+		argocd-agentctl agent create ${agent} \
+			--resource-proxy-username ${agent} \
+			--resource-proxy-password ${agent} \
+			--resource-proxy-server ${IPADDR}:9090
```
jgwest: If I'm not mistaken, when principal is running on the cluster, the `IPADDR` value should be the name of the principal Service e.g. `( crName )-agent-principal`.

controllers/argocdagent/create-agent-config.sh
```diff
@@ -0,0 +1,66 @@
...
+echo "[*] Creating principal TLS configuration"
+argocd-agentctl pki issue principal --upsert \
+	--principal-namespace argocd \
+	--ip "127.0.0.1,${IPADDR}"
+echo "  -> Principal TLS config created."
+
+echo "[*] Creating resouce proxy TLS configuration"
+argocd-agentctl pki issue resource-proxy --upsert \
+	--principal-namespace argocd \
+	--ip "127.0.0.1,${IPADDR}"
```
jgwest: I _think_ this IPADDR should be the loadbalancer hostname for principal

controllers/argocdagent/create-agent-config.sh
```diff
@@ -0,0 +1,66 @@
...
+	argocd-agentctl pki init
+	echo "  -> PKI initialized."
+else
+	echo "  -> Reusing existing agent PKI."
+fi
+
+echo "[*] Creating principal TLS configuration"
+argocd-agentctl pki issue principal --upsert \
+	--principal-namespace argocd \
+	--ip "127.0.0.1,${IPADDR}"
```
jgwest: I _think_ this IPADDR should be the loadbalancer hostname for principal

controllers/argocdagent/deployment.go
```diff
@@ -0,0 +1,476 @@
...
+func buildArgs() []string {
+	args := make([]string, 0)
+	args = append(args, "/usr/local/bin/argocd-agent-principal")
+	return args
+}
+
+func buildPrincipalImage() string {
+	image := os.Getenv("ARGOCD_AGENT_PRINCIPAL_IMAGE")
+	if image == "" {
+		image = "quay.io/openshift-gitops/argocd-agent-principal:v1"
```
jgwest: `openshift-gitops` -> `argoproj-labs`, since this is an `argoproj-labs`-based repo.

controllers/argocdagent/deployment.go
```diff
@@ -0,0 +1,476 @@
...
+	if !reflect.DeepEqual(deployment.Spec.Template.Spec.ServiceAccountName, saName) {
+		log.Info("deployment service account name is being updated")
+		changed = true
+		deployment.Spec.Template.Spec.ServiceAccountName = saName
+	}
+
+	return deployment, changed
+}
+
+func boolPtr(val bool) *bool {
```
jgwest: Can delete this function and use `ptr.To` instead

jgwest (changes requested):

controllers/argocdagent/README.md
jparsai: Yes that is also required, I am yet to work on it to, so that cluster's connection status could be updated in Cluster secret. Because of connection drop issue I didnt configure it yet.

https://github.com/argoproj-labs/argocd-operator/blob/1a60aade6e39eb5e16324ec43a8a8a4ecd648010/controllers/argocdagent/deployment.go#L300

controllers/argocdagent/create-agent-config.sh
jparsai: Yes, this will be ArgoCD namespace, I have to change it, when I was running the script, it was throwing the error that argocd namespace not found, for now I hard coded to have a working environment for development, I have to check agent cli or scripts to find out why it is not taking any other namespace.

controllers/argocdagent/create-agent-config.sh
jparsai: Yes, yet to look into this as well. The connection drop issue was the highest priority, so didnt allow me to look into other problems.
But please confirm one thing, I thought IPADDR has to be resource proxy host not the Principal host, am I wrong here?

api/v1alpha1/argocd_types.go
jparsai: Yes, my understanding was that all configurations will be done via configmap. GitOps operator would set values for mandatory env variables (which you mentioned above), which will be used to create principal configmap.
Should I need use ArgoCD CR for configurations of these fields?

controllers/argocdagent/role.go
jparsai: Sorry, again same reason, I am yet to fix it, for initial development I have given all permissions, I will use same permissions in cluster role, that are given to a Cluster-scoped Argocd instance in GitOps operator and similarly for role.

jparsai:
Thanks @jgwest for you review and suggestions. I have few doubts PTAL at those. 
Yes, there are few changes required which I am yet to do. Because I was caught up with connection drop issue, which was blocking all this functionalities. So I focused on fixing that first. There are few corrections/improvements needed (some of them you already mentioned above) which I wanted to do after connection issue is fixed. As if it done now, I will start looking into these and change code accordingly.

api/v1alpha1/argocd_types.go
jgwest: It's fine for now if you want to use the `ConfigMap`. Jann et al might have a preference here, but for TP, ConfigMap is fine with me.

The reason I didn't think you were using the `ConfigMap` is because, if I'm not mistaken, any changes the user makes to the `ConfigMap` will be reverted?

jparsai: Yes, configmap will be reverted if it is changed directly, if value is updated in env variable then it is fine.
On second thought I think it is better to add in ArgoCD CR, as it will be easy to update will be less confusing. I will add it in ArgoCD CR.

controllers/argocdagent/create-agent-config.sh
jgwest: > argocd-agentctl pki issue principal --upsert \
> 	--principal-namespace argocd \
> 	--ip "127.0.0.1,${IPADDR}"

Ah, OK, looking at the code for this command, the '--ip' field requires an actual IP (which I didn't realize).

When agent/principal are being served via load balancer, we wouldn't be connecting via IP, but via host name (the host name of the loadbalancer).

Thus, perhaps for this case, we don't want to use '--ip' at all, but rather use '--dns (load balancer hostname)'?

(Including the '--ip' parameter _probably_ can't hurt. So if you want to leave it in, that's fine. But perhaps once everything else is working as expected, we can remove it, and see if anything breaks. Jann might likewise have a better idea of whether this is useful for anything.)

controllers/argocdagent/create-agent-config.sh
jgwest: I've reviewed the code, and I believe this comment is still correct: we shouldn't use IPADDR here, but rather resource proxy Service name. 

This allows the principal's Argo CD to connect to the principal's resource proxy port

controllers/argocdagent/create-agent-config.sh
jgwest: I re-reviewed the code, and [see my comment above](https://github.com/argoproj-labs/argocd-operator/pull/1765#discussion_r2166240704). Seems like we should be using --dns, rather than --ip, but it's probably fine to use both for now (that is, '--ip' probably isn't hurting anything).

controllers/argocdagent/create-agent-config.sh
jgwest: Same as above.

jgwest:
Yup, thanks @jparsai! It is draft PR so there are always things that are  TODO: As I went through the PR I added comments for things I noticed, so I/we don't forget to check them once the final PR is up (and maybe in the mean time at least some of them are useful/helpful!)

I've responded to your comments re: `create-agent-config.sh`:
- My guess is that Jann was using IP addressses ('--ip`') because he was using microk8s with IP-based Services. 
- In our case, since we are using loadbalancers, we can likely use the hostname in `--dns`.

controllers/argocdagent/configmap.go
```diff
@@ -0,0 +1,452 @@
...
+func getPrincipalMetricsPort() string {
+	if value := os.Getenv(EnvArgoCDPrincipalMetricsPort); value != "" {
+		return value
+	}
+	return "8000"
+}
+
+func getPrincipalAllowedNamespaces(cr *argoproj.ArgoCD) string {
+	if cr.Spec.ArgoCDAgent != nil && cr.Spec.ArgoCDAgent.Principal != nil {
+		return cr.Spec.ArgoCDAgent.Principal.AllowedNamespaces
```
jparsai: Should I use values of `cr.Spec.SourceNamespaces`?

jparsai:
1) Do I need to create reconciler for AppProject? In existing implementation operator does not reconcile it and it is created in ArgoCD upstream, that is why I didn't add reconciler for AppProject and assuming users would update `sourceNamespaces` filed in AppProject spec.
2) Steps to configure agent are not added here, as it will be done via helm chart. Once it is delivered link to doc can be added here. 
3) This is the Agent version used for testing changes of PR  https://github.com/argoproj-labs/argocd-agent/commit/16badb5ccb8a96095f5afba825f710c11ffc4cdf

jparsai:
@jgwest thanks for your initial reviews on draft PR. This PR in open now. PTAL

controllers/argocdagent/configmap.go
jgwest: To answer your question above, IMHO my preference is what you've already done here: to use `.spec.argocdAgent.prinicpal.allowedNamespaces` for this, rather than automatically using all sourceNamespaces.

api/v1beta1/argocd_types.go
```diff
@@ -1098,6 +1101,37 @@ type WebhookServerSpec struct {
 	Route ArgoCDRouteSpec `json:"route,omitempty"`
 }
 
+type ArgoCDAgentSpec struct {
+
+	// Principal defines configurations for the Principal component of Argo CD Agent.
+	Principal *PrincipalSpec `json:"principal,omitempty"`
```
jgwest, line 1107: Can you create a simple E2E test for this feature? (I recall I said they aren't necessarily required, but this should be quick to add :crossed_fingers: )

Basically a kuttl test that does this:
1) Create `ArgoCD` CR, something like this:
```
kind: ArgoCD
metadata:
  name: argocd
spec:
  controller:
    enabled: false
  argoCDAgent:
    principal:
      enabled: true
      allowed-namespaces: "*"
      jwt-allow-generate: true
      auth: "mtls:CN=([^,]+)"
      logLevel: "trace"
      image: "quay.io/user/argocd-agent:v1"
  sourceNamespaces:
    - "agent-managed"
    - "agent-autonomous"
```
(This example stolen from `controllers/argocdagent/example/argocd.yaml`)

2) Verify ConfigMap, Deployment, Role, RoleBinding, etc, that are created for principal are correct. 
- For now, don't worry about examining any of the Argo CD resources as part of this test. (for example, don't need to check that argocd has expected service types)
- For now, don't need to worry about handling delete/update case (for example, don't need to test whether `.spec.argocdAgent.principal.enabled: false` will delete old resources)
- Likewise, don't worry about checking whether Deployment/Pod are running/ready. You can just verify that they exist and have the expected fields.

WDYT?

api/v1beta1/argocd_types.go
```diff
@@ -1098,6 +1101,37 @@ type WebhookServerSpec struct {
...
+}
+
+type PrincipalSpec struct {
+
+	// Enabled is the flag to enable the Principal component during Argo CD installation. (optional, default `false`)
+	Enabled *bool `json:"enabled,omitempty"`
+
+	// AllowedNamespaces is the list of namespaces that the Principal component is allowed to access.
+	AllowedNamespaces string `json:"allowed-namespaces,omitempty"`
```
jgwest: Would it make more sense to have this as a `[]string`, similar to source namespaces?

Also `allowed-namespaces` -> `allowedNamespaces` to be consistent with other fields in the CR

api/v1beta1/argocd_types.go
```diff
@@ -1098,6 +1101,37 @@ type WebhookServerSpec struct {
...
+
+	// JWTAllowGenerate is the flag to enable the JWT generation during Argo CD installation.
+	JWTAllowGenerate bool `json:"jwt-allow-generate,omitempty"`
```
jgwest: `jwt-allow-generate` -> `jwtAllowGenerate` to be consistent with other fields

deploy/olm-catalog/argocd-operator/0.14.0/argoproj.io_argocds.yaml
```diff
@@ -446,6 +446,40 @@ spec:
                         type: object
                     type: object
                 type: object
+              argoCDAgent:
+                description: ArgoCDAgent defines configurations for the ArgoCD Agent
+                  component.
+                properties:
+                  principal:
+                    description: Principal defines configurations for the Principal
```
jgwest: One more thing, I think possibly these `v0.14.0` version of the files shouldn't be modified? You can try reverting the changes and running `make bundle` etc, again, to see if the files are regenerated.

jgwest:
> 1. Do I need to create reconciler for AppProject? In existing implementation operator does not reconcile it and it is created in ArgoCD upstream, that is why I didn't add reconciler for AppProject and assuming users would update `sourceNamespaces` filed in AppProject spec.
> 
> 2. Steps to configure agent are not added here, as it will be done via helm chart. Once it is delivered link to doc can be added here.
> 
> 3. This is the Agent version used for testing changes of PR  [argoproj-labs/argocd-agent@16badb5](https://github.com/argoproj-labs/argocd-agent/commit/16badb5ccb8a96095f5afba825f710c11ffc4cdf)

1) I don't think we need to create a reconciler for AppProject for now (or create the AppProject for the user). Users can create/update`AppProject` themselves (until we determine, through experience, the best behaviour here)
2) :+1:
3) :+1:

controllers/argocd/service.go
```diff
@@ -35,6 +35,12 @@ func getArgoServerServiceType(cr *argoproj.ArgoCD) corev1.ServiceType {
 	if len(cr.Spec.Server.Service.Type) > 0 {
 		return cr.Spec.Server.Service.Type
 	}
+
+	// If Principal is enabled, use LoadBalancer service type
```
svghadi, line 39: This assumes the cluster has LoadBalancer configured. Is that documented somewhere?

controllers/argocdagent/docs/quickstart.md
```diff
@@ -0,0 +1,132 @@
+# Argo CD Agent Setup Guide
+
+## Overview
+
+This guide provides step-by-step instructions for setting up and configuring the Argo CD Agent in a dev environment.
+
+## Prerequisites
+
+- The `argocd-agentctl` CLI tool installed
```
svghadi: It would be helpful to include the download location of this tool.

api/v1beta1/argocd_types.go
jparsai: I added a very basic e2e test. I doesnt check every details.

controllers/argocd/service.go
jparsai: Here I am actually creating service of LoadBalancer type because we need to expose it for agent to allow it to communicate with principal.

controllers/argocdagent/docs/quickstart.md
jparsai: Yes, that is correct but as of now we didn't make any release of agent CLI, once we have that, then I will add link to it. For development we are manually building agent CLI.

jparsai:
> Looks great, some minor updates requested then good to go!

Thanks @jgwest, I addressed you review comments, PTAL

controllers/argocd/service.go
jgwest: Hey @svghadi, at the moment we're only supporting `Service` with 'type: loadbalancer', but I recognize this is a limitation since there may be other mechanisms that users want to use to expose their services.

My other concern with enabling load balancer is potential security implications, e.g. exposing that endpoint to (potententially) external traffic.

I have some ideas for how we can improve things (on separate PRs), and will followup on these.

At the moment, with this PR, we are following the current 'golden path' defined in the argo agent repo for installing argo cd agent.

jgwest (commented):
LGTM, thanks @jparsai!

PR #1768, opened 2025-06-30
GITOPS-6284: refactor schema registration code in main.go

alkakumari016:
Assisted-by: < Cursor/Gemini etc>

Refactor schema registration code in main.go

Fixes #?
GITOPS-6284

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ COPY api/ api/
 COPY common/ common/
 COPY controllers/ controllers/
 COPY version/ version/
+COPY pkg/ pkg/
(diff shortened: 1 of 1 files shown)


controllers/schema/scheme.go
svghadi, line 1: Can we move this to the `pkg` directory at the root, so it can be imported directly in the `gitops-operator` instead of duplicating the logic there?
Let's discuss further on Slack.

alkakumari016: @svghadi I have incorporated the required changes.

PR #1778, opened 2025-07-08
feat: Add position and permanent properties to Banner

alkakumari016:
**What type of PR is this?**

/kind enhancement

GITOPS-7207: Add position and permanent properties to Banner

Fixes #? GITOPS-7207

### DIFF
--- a/api/v1alpha1/argocd_types.go
+++ b/api/v1alpha1/argocd_types.go
@@ -959,6 +959,10 @@ type Banner struct {
 	Content string `json:"content"`
 	// URL defines an optional URL to be used as banner message link
 	URL string `json:"url,omitempty"`
+	// Permanent defines if the banner should be displayed permanently or only for a certain period of time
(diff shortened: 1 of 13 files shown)


svghadi (commented):
Looks good. Just need to perform a rebase, revert changes from `deploy/olm-catalog/argocd-operator/0.15.0/` directory and run `make generate manifests bundle` again so that the crds are updated for correct version i.e 0.16.0.

svghadi:
Can you manually test a corner case? What happens if the same fields are defined in both `.spec.extraConfig` and `.spec.banner`?

alkakumari016:
> Can you manually test a corner case? What happens if the same fields are defined in both `.spec.extraConfig` and `.spec.banner`?

In that case  `.spec.extraConfig` takes precedence and `.spec.banner` is ignored

svghadi:
Okay. Behaviour is as expected. Thanks.

PR #1780, opened 2025-07-08
Implement local user management via API tokens

jopit:
**What type of PR is this?**
/kind enhancement

This PR adds the ability to configure local users via the argocd CR.

**Jira Ticket:**
https://issues.redhat.com/browse/GITOPS-1973


controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
+	return token.SignedString(serverSignature)
+}
+
+// localUsersInExtraConfig returns the names of all users explicitly defined via
+// extraConfig
+func localUsersInExtraConfig(cr *argoproj.ArgoCD) map[string]bool {
+	localUsers := make(map[string]bool)
+	for k := range cr.Spec.ExtraConfig {
+		if strings.HasPrefix(k, "accounts.") && !strings.HasSuffix(k, ".enabled") {
```
jgwest, line 519: Maybe this is overly cautious :grin: , but can we add a sanity test logic here that if a key begins with `accounts.`, AND contains a `.` after `accounts`., but does NOT have `.enabled` suffix, that we return an error?

Basically I want to catch the case where Argo CD adds a new suffix besides `.enabled`, like:
`accounts.alice.secure`, or something. In this case, the current logic would assume the user is named `alice.secure`

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+	code()
+}
+
+type tokenRenewalTimer struct {
+	timer *time.Timer
+	stop  bool
+}
+
+var (
+	tokenRenewalTimers = map[string]*tokenRenewalTimer{}
```
jgwest: ### A) 
Can we move these variables into `ReconcileArgoCD` (either directly moving these vars into `ReconcileArgoCD`  as fields, or instead via a child struct e.g. `struct localUsersInfo{}` and `localUsers *localUsersInfo` in `ArgoCD` CR)?

Why? IMHO we should always seek to avoid mutable package-scope  variables. Mutable package-scope variables make it challenging to reason about function behaviour.

### B) 
Can you add a brief comment describing what the `tokenRenewalTimers` variable is , and what the key format is (AFAICT key format is `(ArgoCD CR namespace)+(user name)`)?

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+	} else if err != nil {
+		return err
+	}
+
+	// If the ApiKey setting has been changed to false, delete the secret and
+	// the token and the timer.
+	if secretExists && user.ApiKey != nil && !*user.ApiKey {
+		var err error
+		userTokensLock.protect(func() {
+			key := cr.Namespace + "/" + user.Name
```
jgwest: I noticed there are a numberof places  where we have `key := cr.Namespace + "/" + user.Name`. It seems like this would be a good candidate for centralizing the logic  to a `generateKey(cr, user)` func.

api/v1beta1/argocd_types.go
```diff
@@ -774,6 +774,32 @@ type KustomizeVersionSpec struct {
...
+	// Login defines whether or not the user is configured to be able to login. Default is false
+	Login bool `json:"login,omitempty"`
+
+	// TokenLifetime defines the how long the token issued to this user is valid
+	// for. An empty string or the value 0 indicates an infinite lifetime.
+	// Examples: "30m", "8760h"
+	TokenLifetime string `json:"tokenLifetime,omitempty"`
+
+	// AutoRenewToken specifies if a new token is to be issued once the existing
+	// one has expired
```
jgwest: Similar to comments about, we can add a brief comment re: whether the default is true or false (default appears to be true AFAICT)

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+
+	var expAt int64
+	if secretExists {
+		expAt, err = strconv.ParseInt(string(userSecret.Data[localUserExpiresAt]), 10, 64)
+		if err != nil {
+			return fmt.Errorf("failed to convert \"%s\" value to int64 from user secret %s: %w", localUserExpiresAt, userSecret.Name, err)
+		}
+	}
+
+	// If the use secret already exists and neither the token lifetime nor the
```
jgwest: `use secret` -> `user Secret`, I suspect

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+	"github.com/argoproj-labs/argocd-operator/common"
+	"github.com/argoproj-labs/argocd-operator/controllers/argoutil"
+
+	"github.com/argoproj/argo-cd/v3/util/settings"
+
+	"github.com/golang-jwt/jwt/v5"
+	"github.com/google/uuid"
+)
+
+const (
```
jgwest, line 42: It would be helpful to add short comments describing what each of these constants are used for. (AFAICT they are used in the non-Argo CD Secret that we create to create track of the Argo CD Secret lifetime, renewal, etc)

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+	if secretExists && !tokenLifetimeChanged && !autoRenewChanged {
+		if autoRenew == "true" && tokenDuration > 0 {
+			// If the secret exists and autoRenew is true, but there is no timer for the
+			// user, we need to add a timer for the remaining duration. This would
+			// happen when the operator is restared.
+			userTokensLock.protect(func() {
+				key := cr.Namespace + "/" + user.Name
+				if _, ok := tokenRenewalTimers[key]; !ok {
+					renewalTimer := &tokenRenewalTimer{}
+					when := time.Until(time.Unix(expAt, 0))
```
jgwest: Would it be beneficial to have `if expAt == 0 { return fmt.errof("...")}`  here to sanity check that expAt is defined?

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+	localUserID            = "ID"
+	localUserTokenLifetime = "tokenLifetime"
+	localUserUser          = "user"
+)
+
+type lock struct {
+	lock sync.Mutex
+}
+
+func (l *lock) protect(code func()) {
```
jgwest: An optional suggestion:

Could add an `error` return to `protect` function signature. This would be an alternative to the following the pattern I see in the PR:
```go
		var err error

		userTokensLock.protect(func() {
			// (...)
			err = r.Update(context.TODO(), &userSecret)
		})

		return err
```

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+
+	// If auto-renew has been changed to true but the token lifetime has not
+	// changed, add a timer for the remaining duration. There is no need to
+	// re-issue the token, so just update the secret.
+	if autoRenewChanged && autoRenew == "true" && !tokenLifetimeChanged && tokenDuration > 0 {
+		var err error
+
+		userTokensLock.protect(func() {
+			renewalTimer := &tokenRenewalTimer{}
+			when := time.Until(time.Unix(expAt, 0))
```
jgwest: Similar comment re: checking expAt is defined, here

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+
+func (l *lock) protect(code func()) {
+	l.lock.Lock()
+	defer l.lock.Unlock()
+	code()
+}
+
+type tokenRenewalTimer struct {
+	timer *time.Timer
+	stop  bool
```
jgwest: Nit: could do `stop` -> `stopped`, which might be a better name (as it better fits the description of its role?)

controllers/argocd/configmap.go
```diff
@@ -441,6 +441,30 @@ func (r *ReconcileArgoCD) reconcileArgoConfigMap(cr *argoproj.ArgoCD) error {
...
+	// Find all users explicitly defined via extraConfig
+	legacyUsers := localUsersInExtraConfig(cr)
+
+	// Create local users
+	for _, user := range cr.Spec.LocalUsers {
+		// Ignore any user defined via extraConfig
+		if legacyUsers[user.Name] {
+			continue
+		}
+		key := "accounts." + user.Name
```
jgwest, line 475: It may be worth sanity test here that user.Name does NOT contain any spaces (and/or leading/trailing whitespace)? I presume spaces in user names are not supported?

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+			explanation += ", "
+		}
+		explanation += fmt.Sprintf("auto-renew changed from %s to %s for user %s", string(userSecret.Data[localUserAutoRenew]), autoRenew, cr.Namespace+"/"+user.Name)
+	}
+
+	err = nil
+	userTokensLock.protect(func() {
+		err = r.issueToken(*cr, user, userSecret, secretExists, tokenLifetime, tokenDuration, explanation, signingKey)
+	})
+	return nil
```
jgwest: Looks like the `err` should be returned here. Otherwise it's not used.

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+
+	return nil
+}
+
+func (r *ReconcileArgoCD) cleanupLocalUsers(cr *argoproj.ArgoCD) error {
+	// Get a list of the local user secrets
+	secrets := corev1.SecretList{}
+	options := client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{
+			common.ArgoCDKeyComponent: "local-users",
```
jgwest: `local-users` string would be a good candidate to convert to a string constant, since it's important that it's consistent throughout the code

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+	secretExists := true
+
+	// Get the user secret if it exists, else create a new one
+	userSecret := corev1.Secret{}
+	secretName := user.Name + "-local-user"
+	err = r.Get(context.TODO(), types.NamespacedName{Name: secretName, Namespace: cr.Namespace}, &userSecret)
+	if apierrors.IsNotFound(err) {
+		secretExists = false
+		userSecret = *argoutil.NewSecretWithName(cr, secretName)
+		userSecret.Labels[common.ArgoCDKeyComponent] = "local-users"
```
jgwest: WDYT about adding an ownerRef on the user secret, which points back to the parent `ArgoCD` CR?

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+
+	err = nil
+	userTokensLock.protect(func() {
+		err = r.issueToken(*cr, user, userSecret, secretExists, tokenLifetime, tokenDuration, explanation, signingKey)
+	})
+	return nil
+}
+
+// *** THIS METHOD NEEDS TO BE CALLED UNDER THE PROTECTION OF userTokensLock ***
+func (r *ReconcileArgoCD) issueToken(cr argoproj.ArgoCD, user argoproj.LocalUserSpec, userSecret corev1.Secret, secretExists bool, tokenLifetime string, tokenDuration time.Duration, explanation string, signingKey []byte) error {
```
jgwest: ### A) We likely want to remove `secretExists` as a parameter, and instead (re-)acquire the Secret in this function. 

As an example of why, I see this code block elsewhere in the PR:
```go
			renewalTimer := &tokenRenewalTimer{}
			when := time.Until(time.Unix(expAt, 0))
			timer := time.AfterFunc(when, func() {
				userTokensLock.protect(func() {
					if renewalTimer.stop {
						return
					}
					err := r.issueToken(*cr, user, userSecret, secretExists, tokenLifetime, tokenDuration, "token automatically re-issued after expiration", signingKey)
					if err != nil {
						log.Error(err, "when trying to re-issue token for user", cr.Namespace+"/"+user.Name)
					}
				})
			})
```
Notice here that `issueToken` may be called MONTHS after this reconcile function is called. And yet, we are passing `secretExists`, which is a stale state that is potentially months out of date by the time the timer triggers.


### B) Likewise with userSecret; it seems highly likely be stale K8s resource when we go to use it in this function.

tests/k8s/1-052_validate_local_user_management/01-install.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: argoproj.io/v1beta1
+kind: ArgoCD
+metadata:
+  name: example-argocd
+spec:
+  localUsers:
```
jgwest, line 6: IMHO we could use more E2E tests to verify functionality, since this is a security critical feature. BUT, since kuttl is not the best medium for these tests, I am happy to wait for ginkgo port to merge (and contribute new E2E tests on a separate PR)

api/v1beta1/argocd_types.go
```diff
@@ -774,6 +774,32 @@ type KustomizeVersionSpec struct {
 	Path string `json:"path,omitempty"`
 }
 
+// LocalUserSpec is used to specify information about an ArgoCD local user to be created by the operator.
```
jgwest, line 784: If someone deletes a user secret or Argo CD secret, will it be re-reconciled by the operator? This would be a good E2E test candidate.

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+
+	var autoRenew string
+	if user.AutoRenewToken == nil || *user.AutoRenewToken {
+		autoRenew = "true"
+	} else {
+		autoRenew = "false"
+	}
+	userSecret.Data = map[string][]byte{
+		localUserUser:          []byte(user.Name),
+		localUserID:            []byte(uniqueId),
```
jgwest: It appears localUserID field is written but never read (correct me if wrong!)

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+		localUserExpiresAt:     []byte(strconv.FormatInt(expiresAt, 10)),
+		localUserTokenLifetime: []byte(tokenLifetime),
+		localUserAutoRenew:     []byte(autoRenew),
+		localUserApiToken:      []byte(jwtToken),
```
jgwest, line 347: It appears `localUserApiToken` is written and never read? (correct me if wrong!)

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+
+package argocd
+
+import (
+	"context"
+	json "encoding/json"
+	"fmt"
+	"strconv"
+	"strings"
+	"sync"
```
jgwest: Since this is a security sensitive feature, I looked at the code coverage of the unit tests for this file. It seems some blocks are not covered by the unit tests. 

Looking at what those blocks are testing, it _shoud_ be straightforward to add. WDYT?

controllers/argocd/localusers.go
```diff
@@ -0,0 +1,517 @@
...
+	if tokenDuration > 0 && autoRenew == "true" {
+		key := cr.Namespace + "/" + user.Name
+		existingTimer, ok := tokenRenewalTimers[key]
+		if ok {
+			existingTimer.stop = true
+			existingTimer.timer.Stop()
+			delete(tokenRenewalTimers, key)
+		}
+		renewalTimer := &tokenRenewalTimer{}
+		timer := time.AfterFunc(tokenDuration, func() {
```
jgwest, line 401: At every point in the code where we schedule a renewal in the future, it would be advantageous to log it via log statement. Something like `Scheduled token renewal for user '(...)' to (time)'.

controllers/argocd/localusers.go
jopit: Changed this to log a warning and ignore the entry.  Not sure it warrants aborting the reconcile.

controllers/argocd/localusers.go
jopit: Moved the variables into `ReconcileArgoCD`

controllers/argocd/localusers.go
jopit: Done

api/v1beta1/argocd_types.go
jopit: Done

controllers/argocd/localusers.go
jopit: Done

controllers/argocd/localusers.go
jopit: Done

controllers/argocd/localusers.go
jopit: Done

controllers/argocd/localusers.go
jopit: I'd rather leave this as-is. I think the change makes the code harder to read.

controllers/argocd/localusers.go
jopit: Done

controllers/argocd/localusers.go
jopit: Done

controllers/argocd/configmap.go
jopit: ArgoCD doesn't check the name for spaces or anything else, so I left this alone.

controllers/argocd/localusers.go
jopit: Fixed

controllers/argocd/localusers.go
jopit: Done

controllers/argocd/localusers.go
jopit: The ownerRef was being added at the time the secret was created on the cluster.  I've moved it to here.

controllers/argocd/localusers.go
jopit: Done

tests/k8s/1-052_validate_local_user_management/01-install.yaml
jopit: I'd prefer to wait for the ginkgo port.

api/v1beta1/argocd_types.go
jopit: Yes the secret will be recreated with a new token.  I've updated the E2E tests to add a test for this.

controllers/argocd/localusers.go
jopit: I've removed the field.

controllers/argocd/localusers.go
jopit: This is the JWT token that the user will retrieve from the secret, so it needs to remain.

controllers/argocd/localusers.go
jopit: Added some unit tests to increase the coverage

controllers/argocd/localusers.go
jopit: Done

jopit: Log message look like this: `Scheduled token renewal for user 'argocd/alice' to Mon, 18 Aug 2025 14:27:57 EDT`

jgwest (approved):
LGTM, thanks @jopit!

PR #1794, opened 2025-07-23
chore: Use SkipNameValidation to skip controller name validation

svghadi:
**What type of PR is this?**
 /kind chore

With controller-runtime v0.19.0, unique controller name validation is enforced. The operator may fail to start due to this as we don't have unique names. Use SkipNameValidation to ignore the uniqueness check and prevent 

Fixes #?


anandf (approved):
LGTM

PR #1795, opened 2025-07-24
Optimize caching for ConfigMaps and Secrets using custom ClientWrapper

Mangaal:
**What type of PR is this?**
/kind enhancement 

Assisted by: Cursor

This PR implements cache optimization for ConfigMaps and Secrets to significantly reduce memory usage in large Kubernetes clusters. It does so by introducing selective caching and a custom client wrapper.

**Problem**
By default, controller-runtime caches all objects of a given type when a watch is added for that resource.
For example, even if we only want to watch ConfigMaps owned by our operator with:
`bldr.Owns(&appsv1.ConfigMap{})
`
the controller-runtime will still cache every ConfigMap in the cluster.
This can result in high memory usage, especially in large environments.

**Solution**
This PR updates the default controller-runtime cache to store only metadata for Secrets and ConfigMaps, and adds an additional label-filtered cache to store full objects for those with the tracking label.

It also introduces a HybridClient that:

- `Get`: Checks the metadata cache first, then the filtered cache, and finally falls back to a live API call. On live fetch, it applies the tracking label so future requests can be served from the filtered cache.

- `List`: Returns metadata-only objects for Secrets/ConfigMaps. To retrieve the full object, callers should issue a Get()—which is inexpensive after the first call, as the object will then be present in the filtered cache.

This approach reduces memory usage while ensuring full objects are available on-demand with self-healing cache population.

#### User Impact
From an end-user perspective, functionality remains unchanged. The only visible difference is that the operator will now automatically label certain externally managed Secrets and ConfigMaps—for example:
- Argo CD cluster secrets.
- User-provided secrets for route TLS configuration.
- ConfigMaps used for webhook configurations.

These will be patched with: `operator.argoproj.io/tracked-by: argocd`
This ensures they are stored in the filtered cache for faster subsequent access. In Kubernetes, such labeling is generally considered acceptable, but if this is a concern, auto-labeling can be made configurable via a flag.

#### Developer Notes
- Watches on Secrets/ConfigMaps should always use metadata-only mode (e.g., `builder.OnlyMetadata`) to avoid loading large .data fields into memory unnecessarily.
- `List()` calls for Secrets/ConfigMaps will return metadata-only. If full object data is needed, follow the list with a `Get()`—this will be inexpensive after the first call because the object will then be present in the filtered cache.

## Improvements

These metrics were collected from a test cluster containing 100 ConfigMaps and 100 Secrets, each approximately 1 MB in size. The cluster was running four Argo CD instances and no other workload operators. We observed a drop in startup memory usage from ~750 MB to ~60 MB, which remained consistent afterward. In comparison, the unoptimized operator consumed ~300 MB.

##### UnOptimized Operator:
UnOptimized

##### Optimized Operator:
Optimized

### DIFF
--- a/pkg/hybridcache/cache.go
+++ b/pkg/hybridcache/cache.go
@@ -17,9 +17,7 @@ import (
 )
 
 // LabelFilteredCache provides a controller-runtime cache and client that only
-// cache/read corev1 Secrets and ConfigMaps bearing a specific label selector.
-// Reads go through the filtered cache; writes and status updates go to the API
-// server directly via the underlying client.
+// cache/read Secrets and ConfigMaps bearing a specific label selector.
 type LabelFilteredCache struct {
 	Cache  crcache.Cache
 	Client crclient.Client
--- a/pkg/hybridcache/client.go
+++ b/pkg/hybridcache/client.go
@@ -98,9 +98,12 @@ func (hc *HybridClient) List(ctx context.Context, list crclient.ObjectList, opts
 		}
 		out.Items = make([]corev1.Secret, len(metaList.Items))
 		for i := range metaList.Items {
+			// NOTE: Metadata-only list; data/spec omitted.
+			// TODO: If full objects are needed, call Get() for each before returning.
+			// For now, callers should Get() each item as needed.
 			out.Items[i] = corev1.Secret{ObjectMeta: metaList.Items[i].ObjectMeta}
 		}
-		// NOTE: metadata-only list by design
+		
 		return nil
 
 	case *corev1.ConfigMapList:
@@ -111,6 +114,9 @@ func (hc *HybridClient) List(ctx context.Context, list crclient.ObjectList, opts
 		}
 		out.Items = make([]corev1.ConfigMap, len(metaList.Items))
 		for i := range metaList.Items {
+			// NOTE: Metadata-only list; data/spec omitted.
+			// TODO: If full objects are needed, call Get() for each before returning.
+			// For now, callers should Get() each item as needed.
 			out.Items[i] = corev1.ConfigMap{ObjectMeta: metaList.Items[i].ObjectMeta}
 		}
 		return nil


controllers/argoutil/client.go
svghadi, line 1: Can we move this to a new package(maybe `clientwrapper`?) and in `pkg` directory at the root, so it can be imported directly in the gitops-operator?

controllers/argocd/argocd_controller.go
```diff
@@ -86,6 +86,7 @@ var ActiveInstanceMap = make(map[string]string)
 //+kubebuilder:rbac:groups="oauth.openshift.io",resources=oauthclients,verbs=get;list;watch;create;delete;patch;update
 //+kubebuilder:rbac:groups=argoproj.io,resources=notificationsconfigurations;notificationsconfigurations/finalizers,verbs=*
 //+kubebuilder:rbac:groups="apiregistration.k8s.io",resources="apiservices",verbs=get;list
+//+kubebuilder:rbac:groups=argoproj.io,resources=namespacemanagements;namespacemanagements/finalizers;namespacemanagements/status,verbs=*
```
anandf, line 89: is this a new argoproj resource ?

svghadi: It's actually unrelated to this PR. We introduced a new resource in #1687 for namespace management. When running with OLM, it fails with an permission error. This change is to fix that. Should probably go as a separate PR. 
cc: @Rizwana777

Rizwana777: Yes its related to NamespaceMangement, I will raise a seperate PR for this

pkg/hybridcache/cache.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, fmt.Errorf("failed to create client for filtered cache: %w", err)
+	}
+	fc.Client = cl
+
+	return fc, nil
+}
+
+// AttachToManager starts the cache with the manager and blocks manager start
+// until caches have synced. Any start error from the cache is returned to the mgr.
+func (fc *LabelFilteredCache) AttachToManager(mgr manager.Manager) error {
```
anandf, line 69: Should the `mgr` instance be passed as reference `mgr *manager.Manager` ?

svghadi: `mgr` is an interface. We can pass interfaces without a reference.

svghadi:
Superseded by #1866

PR #1800, opened 2025-07-29
feat: Make agent configs first class citizens of ArgoCD CR

jparsai:
/kind enhancement

- Currently only few agent configurations are part of ArgoCD CR and rest are configured via env variables. This is bit complex as all users may not have access to Subscription resource in GitOps operator. This is changed and all fields are now part of ArgoCD CR which makes it very easy to configure the principal via operator.
- Also removed configmap reconciler from operator as it is just used to initialize env variables in principal container, which could be done directly.

Fixes: 
https://issues.redhat.com/browse/GITOPS-7676
https://issues.redhat.com/browse/GITOPS-7678

### DIFF
--- a/tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
+++ b/tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package parallel
+package sequential
 
 import (
 	"context"
@@ -48,12 +48,12 @@ import (
 	fixtureUtils "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/utils"
 )
 
-const (
-	argoCDName               = "argocd"
-	argoCDAgentPrincipalName = "argocd-agent-principal"
-)
+var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 
-var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
+	const (
+		argoCDName               = "argocd"
+		argoCDAgentPrincipalName = "argocd-agent-principal"
+	)
 
 	Context("1-051_validate_argocd_agent_principal", func() {
 
@@ -76,7 +76,7 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
 		)
 
 		BeforeEach(func() {
-			fixture.EnsureParallelCleanSlate()
+			fixture.EnsureSequentialCleanSlate()
 			k8sClient, _ = fixtureUtils.GetE2ETestKubeClient()
 			ctx = context.Background()
 			ns, cleanupFunc = fixture.CreateNamespaceWithCleanupFunc("argocd-agent-principal-1-051")
@@ -192,6 +192,13 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
 				argocdagent.EnvArgoCDPrincipalResourceProxySecretName:   "argocd-agent-resource-proxy-tls",
 				argocdagent.EnvArgoCDPrincipalResourceProxyCaSecretName: "argocd-agent-ca",
 				argocdagent.EnvArgoCDPrincipalJwtSecretName:             "argocd-agent-jwt",
+			}
+		})
+
+		AfterEach(func() {
+			By("Cleanup namespace")
+			if cleanupFunc != nil {
+				cleanupFunc()
 			}
 		})
 
@@ -636,11 +643,5 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
 			}
 		})
 
-		AfterEach(func() {
-			By("Cleanup namespace")
-			if cleanupFunc != nil {
-				cleanupFunc()
-			}
-		})
 	})
 })


controllers/argocdagent/scripts/create-agent-config.sh
```diff
@@ -19,12 +19,15 @@ set -eo pipefail
 RECREATE="$1"
 
 export ARGOCD_AGENT_PRINCIPAL_NAMESPACE=argocd # Should be same as ArgoCD instance namespace
-KUBECTL=$(which kubectl)
+OC=$(which oc)
```
jgwest: Should be `kubectl`, not `oc`, for projects on argoproj-labs org

controllers/argocdagent/deployment.go
```diff
@@ -267,6 +272,9 @@ func updateDeploymentIfChanged(compName, saName string, cr *argoproj.ArgoCD, dep
 	if !reflect.DeepEqual(deployment.Spec.Template.Spec.Containers[0].Ports, buildPorts(compName)) {
 		log.Info("deployment container ports is being updated")
 		changed = true
+		fmt.Println("------------------------------------------------------------------------------")
+		fmt.Println("deployment.Spec.Template.Spec.Containers[0].Ports", deployment.Spec.Template.Spec.Containers[0].Ports)
+		fmt.Println("buildPorts(compName)", buildPorts(compName))
```
jgwest: I'm guessing these lines were some development  logs that were unintentionally left in.

controllers/argocdagent/deployment.go
```diff
@@ -279,258 +287,509 @@ func updateDeploymentIfChanged(compName, saName string, cr *argoproj.ArgoCD, dep
...
+	if cr.Spec.ArgoCDAgent != nil &&
+		cr.Spec.ArgoCDAgent.Principal != nil &&
+		cr.Spec.ArgoCDAgent.Principal.JWT != nil &&
+		cr.Spec.ArgoCDAgent.Principal.JWT.KeyPath != "" {
+		return cr.Spec.ArgoCDAgent.Principal.JWT.KeyPath
+	}
+	return ""
+}
+
+func getPrincipalJwtAllowGenerate(cr *argoproj.ArgoCD) string {
```
jgwest: `Jwt` -> `JWT`

controllers/argocdagent/deployment.go
```diff
@@ -279,258 +287,509 @@ func updateDeploymentIfChanged(compName, saName string, cr *argoproj.ArgoCD, dep
...
+func getPrincipalPprofPort(cr *argoproj.ArgoCD) string {
+	if cr.Spec.ArgoCDAgent != nil &&
+		cr.Spec.ArgoCDAgent.Principal != nil &&
+		cr.Spec.ArgoCDAgent.Principal.PprofPort != 0 {
+		return strconv.Itoa(cr.Spec.ArgoCDAgent.Principal.PprofPort)
+	}
+	return "0"
+}
+
+func getPrincipalJwtSecretName(cr *argoproj.ArgoCD) string {
```
jgwest: `Jwt`->`JWT`

controllers/argocdagent/deployment.go
```diff
@@ -279,258 +287,509 @@ func updateDeploymentIfChanged(compName, saName string, cr *argoproj.ArgoCD, dep
...
+	EnvArgoCDPrincipalJWTKeyPath                = "ARGOCD_PRINCIPAL_JWT_KEY_PATH"
+	EnvArgoCDPrincipalJWTAllowGenerate          = "ARGOCD_PRINCIPAL_JWT_ALLOW_GENERATE"
+	EnvArgoCDPrincipalAuth                      = "ARGOCD_PRINCIPAL_AUTH"
+	EnvArgoCDPrincipalEnableWebSocket           = "ARGOCD_PRINCIPAL_ENABLE_WEBSOCKET"
+	EnvArgoCDPrincipalEnableResourceProxy       = "ARGOCD_PRINCIPAL_ENABLE_RESOURCE_PROXY"
+	EnvArgoCDPrincipalKeepAliveMinInterval      = "ARGOCD_PRINCIPAL_KEEP_ALIVE_MIN_INTERVAL"
+	EnvArgoCDPrincipalRedisServerAddress        = "ARGOCD_PRINCIPAL_REDIS_SERVER_ADDRESS"
+	EnvArgoCDPrincipalRedisCompressionType      = "ARGOCD_PRINCIPAL_REDIS_COMPRESSION_TYPE"
+	EnvArgoCDPrincipalPprofPort                 = "ARGOCD_PRINCIPAL_PPROF_PORT"
+	EnvArgoCDPrincipalTlsSecretName             = "ARGOCD_PRINCIPAL_TLS_SECRET_NAME"
```
jgwest: `Tls` -> `TLS` to be consistent with other uses of TLS, AFAICT

jparsai:
Please hold this PR

api/v1beta1/argocd_types.go
```diff
@@ -1164,20 +1164,158 @@ type PrincipalSpec struct {
...
+	NamespaceCreatePattern string `json:"namespaceCreatePattern,omitempty"`
+
+	// NamespaceCreateLabels is the set of labels to apply to namespaces created for agents. Ex: "foo=bar,bar=baz"
+	NamespaceCreateLabels []string `json:"namespaceCreateLabels,omitempty"`
+}
+
+type PrincipalResourceProxySpec struct {
+
+	// Enable is the flag to enable the resource proxy.
+	Enable *bool `json:"enable,omitempty"`
```
jgwest: My expectation is that principal resource proxy is an not optional component, that it is an essential component of argocd-agent. 

If this sounds right, we can likely remove `Enable`.

If we do want to keep this field, it should likely be `Enabled` to be consistent with other fields of the same function.

api/v1beta1/argocd_types.go
```diff
@@ -1164,20 +1164,158 @@ type PrincipalSpec struct {
...
+	KeyPath string `json:"keyPath,omitempty"`
+}
+
+type PrincipalTLSClientSpec struct {
+
+	// RootCASecretName is the name of the secret containing the root CA TLS certificate
+	RootCASecretName string `json:"rootCASecretName,omitempty"`
+
+	// RootCAPath is the path to a TLS root certificate authority to be used to validate agent's client certificates against.
+	RootCAPath string `json:"rootCAPath,omitempty"`
```
jgwest: Same question as for other paths re: whether it should be included

controllers/argocdagent/deployment.go
```diff
@@ -293,322 +293,473 @@ func updateDeploymentIfChanged(compName, saName string, cr *argoproj.ArgoCD, dep
...
+	}
+
+	// Add custom environment variables if specified in the CR
+	if hasServer(cr) && cr.Spec.ArgoCDAgent.Principal.Server.Env != nil {
+		env = append(env, cr.Spec.ArgoCDAgent.Principal.Server.Env...)
+	}
 
 	return env
 }
+
```
jgwest, line 378: Can add a comment at the top here, something like 'These constants are environment variables that correspond to the environment variables used to configure Argo CD agent, and should match the names exactly from the agent."

controllers/argocdagent/scripts/create-agent-config.sh
```diff
@@ -22,6 +22,9 @@ export ARGOCD_AGENT_PRINCIPAL_NAMESPACE=argocd # Should be same as ArgoCD instan
 KUBECTL=$(which kubectl)
 OPENSSL=$(which openssl)
 
+# Create a secret for the redis password
+${KUBECTL} create secret generic argocd-redis -n argocd --from-literal=auth="$(${KUBECTL} get secret argocd-redis-initial-password -n argocd -o jsonpath='{.data.admin\.password}' | base64 -d)"
```
jgwest, line 26: Does the operator already generate a secret for redis (which is what this command is reading from)? Could it use that?

deploy/olm-catalog/argocd-operator/0.16.0/argoproj.io_argocds.yaml
```diff
@@ -454,32 +454,202 @@ spec:
...
-                        description: AllowedNamespaces is the list of namespaces that
-                          the Principal component is allowed to access.
-                        items:
-                          type: string
-                        type: array
                       auth:
                         description: Auth is the authentication method for the Principal
                           component.
                         type: string
+                      enableWebSocket:
```
jgwest: If I'm not mistaken, the contents of the file can be reverted, since the `main` branch is mainly focused on `0.17.0`

api/v1beta1/argocd_types.go
```diff
@@ -1164,20 +1164,158 @@ type PrincipalSpec struct {
...
+
+type PrincipalServerSpec struct {
 
 	// Auth is the authentication method for the Principal component.
 	Auth string `json:"auth,omitempty"`
 
-	// LogLevel refers to the log level used by the Principal component. Defaults to info if not configured. Valid options are debug, info, trace, error, and warn.
+	// MetricsPort is the port number for the Principal component metrics.
+	MetricsPort int `json:"metricsPort,omitempty"`
+
```
jgwest: I think more thought may be required into which values we should allow to be customized. 

For example, setting this field to any other value will break metrics, because we've hardcoded the actual value we use in `buildPrincipalMetricsServiceSpec` to 8080.

We should either:
- A) remove this field and not allow it to be customized in the CR (for now)
- B) keep this field, but ensure that the value is also used in the `Service` we generate that points to the `Deployment`.

My vote is A), because AFAICT there's no purpose to changing the port.

api/v1beta1/argocd_types.go
```diff
@@ -1164,20 +1164,158 @@ type PrincipalSpec struct {
...
+	PprofPort int `json:"pprofPort,omitempty"`
+
+	// HealthzPort is the port number for the Principal component healthz.
+	HealthzPort int `json:"healthzPort,omitempty"`
+
+	// ListenHost is the interface address to listen on.
+	ListenHost string `json:"listenHost,omitempty"`
+
+	// ListenPort is the port gRPC server should listen on.
+	ListenPort int `json:"listenPort,omitempty"`
```
jgwest: Not sure these fields are worth customizing at this time. If we do, we need to update Service.

api/v1beta1/argocd_types.go
```diff
@@ -1164,20 +1164,158 @@ type PrincipalSpec struct {
...
+	SecretName string `json:"secretName,omitempty"`
+
+	// TLSCertPath is the path to the TLS certificate to be used by the resource proxy.
+	TLSCertPath string `json:"tlsCertPath,omitempty"`
+
+	// TLSKeyPath is the path to the TLS private key to be used by the resource proxy.
+	TLSKeyPath string `json:"tlsKeyPath,omitempty"`
+
+	// TLSCAPath is the path to the CA certificate to be used by the resource proxy.
+	TLSCAPath string `json:"tlsCAPath,omitempty"`
```
jgwest: What is the goal of allowing these paths to be customized via the CR?

For example, these are paths in the agent container, but how would a use actually get the files they want at these paths? We don't currently allow custom volumes/volume mounts via the CR.

api/v1beta1/argocd_types.go
```diff
@@ -1164,20 +1164,158 @@ type PrincipalSpec struct {
...
+	SecretName string `json:"secretName,omitempty"`
+
+	// AllowGenerate is the flag to allow the principal to generate its own set of TLS cert and key on startup when none are configured
+	AllowGenerate *bool `json:"allowGenerate,omitempty"`
+
+	// CertPath is the path to the TLS certificate to be used by the gRPC server.
+	CertPath string `json:"certPath,omitempty"`
+
+	// KeyPath is the path to the TLS private key to be used by the gRPC server
+	KeyPath string `json:"keyPath,omitempty"`
```
jgwest: Same comment above re: what the goal is of allowing these fields to be customized.

api/v1beta1/argocd_types.go
```diff
@@ -1164,20 +1164,158 @@ type PrincipalSpec struct {
...
+	RootCASecretName string `json:"rootCASecretName,omitempty"`
+
+	// RootCAPath is the path to a TLS root certificate authority to be used to validate agent's client certificates against.
+	RootCAPath string `json:"rootCAPath,omitempty"`
+
+	// RequireClientCert is the flag to indicate whether to require client certs from agents upon connection.
+	RequireClientCert *bool `json:"requireClientCert,omitempty"`
+
+	// ClientMatchSubject is the flag to mention whether to match the subject field in a client certificate presented by an agent to the agent's name.
+	ClientMatchSubject *bool `json:"clientMatchSubject,omitempty"`
```
jgwest: We should ask Jann whether he thinks it's worth exposing these fields at this time. My concern is they may unintentionally make the argo cd agent configuration insecure.

jgwest (changes requested):

controllers/argocdagent/scripts/create-agent-config.sh
jparsai: As discussed over Slack, I will change argocd agent logic to use  `argocd-redis-initial-password`. I will raise another operator PR for this after agent logic is changed.

api/v1beta1/argocd_types.go
jparsai: @jannfis should I drop these fields as well?

jparsai:
Thanks @jgwest  addressed your review comments PTAL.

api/v1beta1/argocd_types.go
jgwest: Thanks @jparsai, I pinged him on Slack. My guess is he gets a million GitHub notifications a day :sweat_smile:

controllers/argocdagent/scripts/create-agent-config.sh
jgwest: Yup, sounds good to me :+1:

tests/ginkgo/parallel/1-051_validate_argocd_agent_principal_test.go
```diff
@@ -208,87 +304,125 @@ data:
...
-						"agent-autonomous",
-					},
-				},
-			}
+			By("Verify environment variable is set correctly")
+			Expect(container.Env).To(ContainElement(corev1.EnvVar{Name: "ARGOCD_PRINCIPAL_LOG_LEVEL", Value: "trace"}))
+		})
+
+		It("should create argocd agent principal resources, and pod should start successfully with default image", func() {
+			defer cleanupFunc()
```
jgwest: Rather than calling `defer cleanupFunc()` at the beginning of each test block, even easier is  to just add a call to `AfterEach`, e.g.
```
		AfterEach(func() {
	
			if nsCleanup != nil {
				nsCleanup()
			}

		})
```

Just search for `AfterEach` for existing examples in other tests.

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
```diff
@@ -43,16 +51,164 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
 	Context("1-051_validate_argocd_agent_principal", func() {
```
jgwest, line 59: Thanks for handling the port from kuttl to ginkgo, your tests look great!

Two more simple tests I think would be useful (correct me if they already exist, I didn't do a thorough examination!):
- Verify the delete case: test that setting `.spec.argoCDAgent.principal` to `nil` causes the various argo cd agent workloads to be cleaned up and any changes principal-specific to argo cd instance to be reverted.
- Verify the field update case: test that changing the various `.spec.argocdAgent.*` values in an existing ArgoCD CR  causes the corresponding value to be set in Deployment (or wherever it needs to be set)
    - Basically just making sure our field update logic is correct. We already have logic that verifies the fields are set to a default value, this makes sure they can be changed away from the default.
    - If it's too much work to test them all, feel free to test only the important fields (by your judgement)

jgwest (changes requested):

api/v1beta1/argocd_types.go
jparsai: dropped these fields as well.

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
jparsai: Added check to verify principal resources are deleted when `.spec.argoCDAgent.principal.enabled` is set to `false` and `nil` as well as check for ArgoCD CR updates. I didnt add these new tests in kuttl as we are using ginkgo now.
PTAL

jgwest:
Pushed a commit switching from `k8sClient.Update` -> `argocdFixture.Update` to avoid race condition and intermittent test failure

PR #1801, opened 2025-07-29
feat: Remove configmap dependency for principal deployment

jparsai:
/kind enhancement

This PR is to remove dependency of `argocd-agent-params` configmap for creating principal deployment. This will also fix the bug where cm is deleted by perator on agent cluster because of the same name.

Fixes https://issues.redhat.com/browse/GITOPS-7361


jgwest (changes requested):
Same comments as provided on https://github.com/argoproj-labs/argocd-operator/pull/1800

PR #1802, opened 2025-07-30
fix: argocd-agent-params cm gets deleted in workload cluster

jparsai:
**What type of PR is this?**

Fixes #?


controllers/argocdagent/configmap.go
```diff
@@ -110,6 +110,8 @@ const (
 	EncArgoCDPrincipalJwtSecretName             = "ARGOCD_PRINCIPAL_JWT_SECRET_NAME"
 )
 
+var cmSuffix = "-agent-principal-params"
```
jgwest: This can likely be a const, rather than a var

jgwest (changes requested):
One tiny change then LGTM

jgwest:
After minor change is made, feel free to merge if I am not around (it is late here :smile:). The change is minor and not necessarily a merge blocker :grin:

anandf (approved):
LGTM

PR #1803, opened 2025-07-30
fix: add missing env variables in deployment

jparsai:
/kind bug

This PR is to add few missing env variables from deployment manifest, that were added in configmap later but missed from deployment.


controllers/argocdagent/configmap.go
```diff
@@ -71,6 +71,7 @@ const (
 	PrincipalKeepAliveMinInterval      = "principal.keep-alive-min-interval"
 	PrincipalRedisCompressionType      = "principal.redis-compression-type"
 	PrincipalPprofPort                 = "principal.pprof-port"
+	PrincipalRedisPassword             = "auth"
```
jgwest, line 74: Is this line correct? Variable name is `PrincipalRedisPassword` but value is `auth`?

jparsai: yes it is for this env variable reading from secret.
https://github.com/argoproj-labs/argocd-agent/blob/main/install/kubernetes/principal/principal-deployment.yaml#L186

jgwest (approved):
LGTM, thanks @jparsai!

PR #1659, opened 2025-02-08
Added ReadOnlyRootFileSystem to the Argo CD components

anandf:
**What type of PR is this?**

[//]: # (Uncomment only one  line, and delete the rest.)
[//]: # (For example,  would simply become: /kind bug  )

**What does this PR do / why we need it**:
Upstream Argo CD has all components running with `readOnlyRootFileSystem` set to `true`. For security reasons this needs to be enabled even for the components that are created by the operator.

Application Controller: 
https://github.com/argoproj/argo-cd/blob/d183d9c614d05979f1327a554942a9e656f1c2ec/manifests/base/application-controller-deployment/argocd-application-controller-deployment.yaml#L263
**Note:** Stateful set based application controller does not have this flag set. So need to validate if enabling it will cause any breaking change
https://github.com/argoproj/argo-cd/blob/master/manifests/base/application-controller-deployment/argocd-application-controller-statefulset.yaml
AppSet Controller: https://github.com/argoproj/argo-cd/blob/d183d9c614d05979f1327a554942a9e656f1c2ec/manifests/base/applicationset-controller/argocd-applicationset-controller-deployment.yaml#L202
Notification Controller : 
https://github.com/argoproj/argo-cd/blob/d183d9c614d05979f1327a554942a9e656f1c2ec/manifests/base/notification/argocd-notifications-controller-deployment.yaml#L92
Dex: 
https://github.com/argoproj/argo-cd/blob/d183d9c614d05979f1327a554942a9e656f1c2ec/manifests/base/dex/argocd-dex-server-deployment.yaml#L34
Redis: 
https://github.com/argoproj/argo-cd/blob/d183d9c614d05979f1327a554942a9e656f1c2ec/manifests/base/redis/argocd-redis-deployment.yaml#L31
https://github.com/argoproj/argo-cd/blob/d183d9c614d05979f1327a554942a9e656f1c2ec/manifests/base/redis/argocd-redis-deployment.yaml#L60
RepoServer:
https://github.com/argoproj/argo-cd/blob/d183d9c614d05979f1327a554942a9e656f1c2ec/manifests/base/repo-server/argocd-repo-server-deployment.yaml#L261
https://github.com/argoproj/argo-cd/blob/d183d9c614d05979f1327a554942a9e656f1c2ec/manifests/base/repo-server/argocd-repo-server-deployment.yaml#L295
Server:
[...]

PR #1812, opened 2025-08-05
fix: Argo Rollouts UI extension installation error in initContainer due to read-only access

anandf:
**What type of PR is this?**
/kind bug

**Description of the problem**
When Argo Rollouts UI extension was enabled, the init container failed with the following error message
```
++ basename -- file:///home/ext-installer/rollout-extension.tar
+ ext_filename=rollout-extension.tar
++ mktemp -d -t extension-XXXXXX
mktemp: failed to create directory via template '/tmp/extension-XXXXXX': Read-only file system
```
Error message was coming from this line.
https://github.com/rh-gitops-midstream/release/blob/eddad5e34e05bcd548144f65f35152377ec5c21f/containers/argocd-extensions/install.sh#L110
This breaking change was introduced by https://github.com/argoproj-labs/argocd-operator/pull/1659 when the container root file system was made `read-only` using the property setting`readOnlyRootFilesystem: true`.

Since the temp directory `/tmp` was not mounted as an empty dir, it was using the root file system and the installation step failed.

It mounts an additional volume `tmp` at the mount path `/tmp` this will allow the tar file extraction to directory under `/tmp` and the `mktemp` command to succeed. There were some gaps in the existing unit tests, which was also addressed.

Fixes #?

### DIFF
--- a/controllers/argocd/deployment.go
+++ b/controllers/argocd/deployment.go
@@ -1574,10 +1574,8 @@ func (r *ReconcileArgoCD) reconcileServerDeployment(cr *argoproj.ArgoCD, useTLSF
 
 	deploy.Spec.Template.Spec.Volumes = serverVolumes
 
-	rolloutsVolumeName := "rollout-extensions"
-
+	const rolloutsVolumeName = "rollout-extensions"
 	if cr.Spec.Server.EnableRolloutsUI {
-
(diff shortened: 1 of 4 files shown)


svghadi (approved):
LGTM

PR #1813, opened 2025-08-06
fix: Truncate Kubernetes resource names and labels to comply with 63-…

aali309:
…character limit

**What type of PR is this?**
These changes fixes length validation errors that occur when creating ArgoCD
instances with long names by implementing comprehensive truncation with hash
suffixes to ensure uniqueness while staying within Kubernetes limits.

 /kind bug 
 
 

Fixes #?

### DIFF
--- a/controllers/argocd/applicationset.go
+++ b/controllers/argocd/applicationset.go
@@ -943,7 +943,7 @@ func (r *ReconcileArgoCD) reconcileApplicationSetService(cr *argoproj.ArgoCD) er
 	}
 
 	svc.Spec.Selector = map[string]string{
-		common.ArgoCDKeyName: nameWithSuffix(common.ApplicationSetServiceNameSuffix, cr),
+		common.ArgoCDKeyName: argoutil.TruncateWithHash(nameWithSuffix(common.ApplicationSetServiceNameSuffix, cr)),
 	}
 
 	if err := controllerutil.SetControllerReference(cr, svc, r.Scheme); err != nil {
--- a/controllers/argocd/rolebinding.go
+++ b/controllers/argocd/rolebinding.go
@@ -96,7 +96,10 @@ func getRoleBindingNameForSourceNamespaces(argocdName, targetNamespace string) s
 // newRoleBindingWithname creates a new RoleBinding with the given name for the given ArgCD.
 func newRoleBindingWithname(name string, cr *argoproj.ArgoCD) *v1.RoleBinding {
 	roleBinding := newRoleBinding(cr)
-	roleBinding.Name = fmt.Sprintf("%s-%s", cr.Name, name)
+
+	// Truncate the name to stay within 63 character limit
+	fullName := fmt.Sprintf("%s-%s", cr.Name, name)
+	roleBinding.Name = truncateWithHash(fullName)
 
 	labels := roleBinding.Labels
 	labels[common.ArgoCDKeyName] = name
--- a/controllers/argoutil/resource.go
+++ b/controllers/argoutil/resource.go
@@ -16,6 +16,7 @@ package argoutil
 
 import (
 	"context"
+	"crypto/sha1"
 	"fmt"
 	"reflect"
 	"strings"
@@ -31,6 +32,11 @@ import (
 	argoprojv1alpha1 "github.com/argoproj-labs/argocd-operator/api/v1alpha1"
 	argoproj "github.com/argoproj-labs/argocd-operator/api/v1beta1"
 	"github.com/argoproj-labs/argocd-operator/common"
+)
+
+const (
+	hashLabelLength = 7
+	maxLabelLength  = 63
 )
 
 // AppendStringMap will append the map `add` to the given map `src` and return the result.
@@ -204,4 +210,21 @@ func AddTrackedByOperatorLabel(meta *metav1.ObjectMeta) {
 func IsTrackedByOperator(labels map[string]string) bool {
 	value, exists := labels[common.ArgoCDTrackedByOperatorLabel]
 	return exists && value == common.ArgoCDAppName
+}
+
+// TruncateWithHash truncates a string to a maximum of 63 characters and adds a hash suffix to ensure uniqueness
+func TruncateWithHash(input string) string {
+	if len(input) <= maxLabelLength {
+		return input
+	}
+
+	// Calculate hash of the original string
+	hash := sha1.Sum([]byte(input)) // #nosec G401
+	hashSuffix := fmt.Sprintf("-%x", hash[:hashLabelLength])
+
+	// Calculate how much we can truncate
+	maxBaseLength := maxLabelLength - len(hashSuffix)
+
+	// Truncate and add hash
+	return input[:maxBaseLength] + hashSuffix
 }
--- a/controllers/argocd/deployment.go
+++ b/controllers/argocd/deployment.go
@@ -390,23 +390,26 @@ func newDeployment(cr *argoproj.ArgoCD) *appsv1.Deployment {
 // newDeploymentWithName returns a new Deployment instance for the given ArgoCD using the given name.
 func newDeploymentWithName(name string, component string, cr *argoproj.ArgoCD) *appsv1.Deployment {
 	deploy := newDeployment(cr)
-	deploy.Name = name
+
+	// Truncate the name for both deployment name and labels to stay within 63 character limit
+	truncatedName := argoutil.TruncateWithHash(name)
+	deploy.Name = truncatedName
 
 	lbls := deploy.Labels
-	lbls[common.ArgoCDKeyName] = name
+	lbls[common.ArgoCDKeyName] = truncatedName
 	lbls[common.ArgoCDKeyComponent] = component
 	deploy.Labels = lbls
 
 	deploy.Spec = appsv1.DeploymentSpec{
 		Selector: &metav1.LabelSelector{
 			MatchLabels: map[string]string{
-				common.ArgoCDKeyName: name,
+				common.ArgoCDKeyName: truncatedName,
 			},
 		},
 		Template: corev1.PodTemplateSpec{
 			ObjectMeta: metav1.ObjectMeta{
 				Labels: map[string]string{
-					common.ArgoCDKeyName: name,
+					common.ArgoCDKeyName: truncatedName,
 				},
 				Annotations: make(map[string]string),
 			},
--- a/controllers/argocd/dex.go
+++ b/controllers/argocd/dex.go
@@ -513,7 +513,7 @@ func (r *ReconcileArgoCD) reconcileDexService(cr *argoproj.ArgoCD) error {
 	}
 
 	svc.Spec.Selector = map[string]string{
-		common.ArgoCDKeyName: nameWithSuffix("dex-server", cr),
+		common.ArgoCDKeyName: argoutil.TruncateWithHash(nameWithSuffix("dex-server", cr)),
 	}
 
 	svc.Spec.Ports = []corev1.ServicePort{
--- a/controllers/argocd/networkpolicies.go
+++ b/controllers/argocd/networkpolicies.go
@@ -55,7 +55,7 @@ func (r *ReconcileArgoCD) ReconcileRedisNetworkPolicy(cr *argoproj.ArgoCD) error
 		Spec: networkingv1.NetworkPolicySpec{
 			PodSelector: metav1.LabelSelector{
 				MatchLabels: map[string]string{
-					"app.kubernetes.io/name": fmt.Sprintf("%s-%s", cr.Name, "redis"),
+					"app.kubernetes.io/name": argoutil.TruncateWithHash(fmt.Sprintf("%s-%s", cr.Name, "redis")),
 				},
 			},
 			PolicyTypes: []networkingv1.PolicyType{
@@ -67,21 +67,21 @@ func (r *ReconcileArgoCD) ReconcileRedisNetworkPolicy(cr *argoproj.ArgoCD) error
 						{
 							PodSelector: &metav1.LabelSelector{
 								MatchLabels: map[string]string{
-									"app.kubernetes.io/name": fmt.Sprintf("%s-%s", cr.Name, "application-controller"),
+									"app.kubernetes.io/name": argoutil.TruncateWithHash(fmt.Sprintf("%s-%s", cr.Name, "application-controller")),
 								},
 							},
 						},
 						{
 							PodSelector: &metav1.LabelSelector{
 								MatchLabels: map[string]string{
-									"app.kubernetes.io/name": fmt.Sprintf("%s-%s", cr.Name, "repo-server"),
+									"app.kubernetes.io/name": argoutil.TruncateWithHash(fmt.Sprintf("%s-%s", cr.Name, "repo-server")),
 								},
 							},
 						},
 						{
 							PodSelector: &metav1.LabelSelector{
 								MatchLabels: map[string]string{
-									"app.kubernetes.io/name": fmt.Sprintf("%s-%s", cr.Name, "server"),
+									"app.kubernetes.io/name": argoutil.TruncateWithHash(fmt.Sprintf("%s-%s", cr.Name, "server")),
 								},
 							},
 						},
@@ -101,7 +101,7 @@ func (r *ReconcileArgoCD) ReconcileRedisNetworkPolicy(cr *argoproj.ArgoCD) error
 		networkPolicy.Spec.Ingress[0].From = append(networkPolicy.Spec.Ingress[0].From, networkingv1.NetworkPolicyPeer{
 			PodSelector: &metav1.LabelSelector{
 				MatchLabels: map[string]string{
-					"app.kubernetes.io/name": fmt.Sprintf("%s-agent-%s", cr.Name, "principal"),
+					"app.kubernetes.io/name": argoutil.TruncateWithHash(fmt.Sprintf("%s-agent-%s", cr.Name, "principal")),
 				},
 			},
 		})
@@ -186,7 +186,7 @@ func (r *ReconcileArgoCD) ReconcileRedisHANetworkPolicy(cr *argoproj.ArgoCD) err
 		Spec: networkingv1.NetworkPolicySpec{
 			PodSelector: metav1.LabelSelector{
 				MatchLabels: map[string]string{
-					"app.kubernetes.io/name": fmt.Sprintf("%s-%s", cr.Name, "redis-ha-haproxy"),
+					"app.kubernetes.io/name": argoutil.TruncateWithHash(fmt.Sprintf("%s-%s", cr.Name, "redis-ha-haproxy")),
 				},
 			},
 			PolicyTypes: []networkingv1.PolicyType{
@@ -198,21 +198,21 @@ func (r *ReconcileArgoCD) ReconcileRedisHANetworkPolicy(cr *argoproj.ArgoCD) err
 						{
 							PodSelector: &metav1.LabelSelector{
 								MatchLabels: map[string]string{
-									"app.kubernetes.io/name": fmt.Sprintf("%s-%s", cr.Name, "application-controller"),
+									"app.kubernetes.io/name": argoutil.TruncateWithHash(fmt.Sprintf("%s-%s", cr.Name, "application-controller")),
 								},
 							},
 						},
 						{
 							PodSelector: &metav1.LabelSelector{
 								MatchLabels: map[string]string{
-									"app.kubernetes.io/name": fmt.Sprintf("%s-%s", cr.Name, "repo-server"),
+									"app.kubernetes.io/name": argoutil.TruncateWithHash(fmt.Sprintf("%s-%s", cr.Name, "repo-server")),
 								},
 							},
 						},
 						{
 							PodSelector: &metav1.LabelSelector{
 								MatchLabels: map[string]string{
-									"app.kubernetes.io/name": fmt.Sprintf("%s-%s", cr.Name, "server"),
+									"app.kubernetes.io/name": argoutil.TruncateWithHash(fmt.Sprintf("%s-%s", cr.Name, "server")),
 								},
 							},
 						},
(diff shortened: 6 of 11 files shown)


tests/ocp/1-007_verify_hostname_with_route/04-delete.yaml
ranakan19, line 1: Can you explain what is the intention of this file here? 
I was seeing a warning `Warning: unknown field "$delete"`

tests/ocp/1-007_verify_hostname_with_route/01-assert.yaml
```diff
@@ -1,8 +1,11 @@
...
-spec: {}
+  name: "argocd-long-name-for-route-testiiiiiiiiiiiiiiiiiiiiiiiing-appse"
+spec:
+  to:
+    kind: Service
+    name: argocd-long-name-for-route-testiiiiiiiiiiiiiiiiiiiiiiiing-applicationset-controller
 status:
   ingress:
-  - host: argocd-long-name-for-route-testiiiiiiiiiiiiiiiiiiiiiiiing-appset.apps.testing.lab.upshift.rdu2.redhat.com
+  - host: argocd-long-name-for-route-testiiiiiiiiiiiiiiiiiiiiiiiing-appse.apps.psi-418-4.ocp-gitops-qe.com
```
ranakan19: this url will change depending on the cluster being run on

ranakan19 (changes requested):
I think we should preserve the suffix of generated resources. In you current approach, a suffix is added and then the name is truncated, the suffix itself will cut off when the argocd cr name is already above 63 characters. This could make it difficult for users to search for resources using those suffixes.

A better approach would be to:

1. Truncate the custom resource (CR) name first. We can store this truncated name as an annotation on the ArgoCD CR.
2. Append the full suffix to the already-truncated name.

for example on applicationset:
```
truncatedCrName := cr.Annotations[truncatedNameKey]
...
svc.Spec.Selector = map[string]string{
		common.ArgoCDKeyName: nameWithSuffix(common.ApplicationSetServiceNameSuffix, truncatedCrName),
	}

```

So for the examples mentioned:
argocd-cr name: argocd-long-for-route-testing-long-very-long-very-very-long-name
truncated cr-name: argocd-long-for-route-testing-long-very(39 chars after removing the trailing -)
secrets would be:  argocd-long-for-route-testing-long-very-cluster (47 characters)
and argocd-long-for-route-testing-long-very-redis-initial-password (62 characters)


To determine the correct truncation length, we need to find the maximum possible suffix length and subtract that from the Kubernetes limit of 63 characters. A common suffix, such as -redis-initial-password, is 23 characters long. Therefore, the truncated CR name should be limited to 40 characters (63 - 23 = 40). You can cross-check if there is indeed the longest suffix.

Feel free to use something other than annotation as well, whatever is easier.

tests/ocp/1-007_verify_hostname_with_route/04-delete.yaml
aali309: Intention is to clean up the ArgoCD Custom Resource  for proper test isolation but the file used invalid $delete: true syntax instead of proper KUTTL TestStep structure. I have updated to use proper KUTTL TestStep with delete field.

controllers/argocd/applicationset.go
```diff
@@ -943,7 +943,7 @@ func (r *ReconcileArgoCD) reconcileApplicationSetService(cr *argoproj.ArgoCD) er
 	}
 
 	svc.Spec.Selector = map[string]string{
-		common.ArgoCDKeyName: nameWithSuffix(common.ApplicationSetServiceNameSuffix, cr),
+		common.ArgoCDKeyName: argoutil.TruncateWithHash(nameWithSuffix(common.ApplicationSetServiceNameSuffix, cr)),
```
ranakan19: this will still truncate the suffix

ranakan19: actually no, looking further you have the truncated CR logic in nameWithSuffix. 
So is `TruncateWithHash` needed here if `nameWithSuffix` already takes care of having the name under 63 chars? 
It is a little confusing with the names, maybe you could consider renaming them so its clearer what the function is doing?

controllers/argocd/rolebinding.go
```diff
@@ -96,7 +96,10 @@ func getRoleBindingNameForSourceNamespaces(argocdName, targetNamespace string) s
 // newRoleBindingWithname creates a new RoleBinding with the given name for the given ArgCD.
 func newRoleBindingWithname(name string, cr *argoproj.ArgoCD) *v1.RoleBinding {
 	roleBinding := newRoleBinding(cr)
-	roleBinding.Name = fmt.Sprintf("%s-%s", cr.Name, name)
+
+	// Truncate the name to stay within 63 character limit
+	fullName := fmt.Sprintf("%s-%s", cr.Name, name)
+	roleBinding.Name = truncateWithHash(fullName)
```
ranakan19: this is using https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/rolebinding.go#L28, which looks redundant, Could this be replaced with argoutil.TruncateWithHash() instead? And thus removing the function from rolebinding.go, would need to replace its occurance in tests as well.

controllers/argocd/applicationset.go
ranakan19: There is also another function with the same name - https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argoutil/resource.go#L118 - so should definitely rename `nameWithSuffix` you are adding.

ranakan19 (changes requested):
Thank you for making these changes! I had a few follow-ups we can consider:

1. We could consolidate and remove the redundant truncateWithHash() function.
2. The function nameWithSuffix might be clearer if renamed. (and there is an existing function with the same name)
3. Some tests seem to be missing — I see coverage for routes and secrets, but since the changes affect multiple k8s resources, it may help to add tests for a long-name Argo CD instance in the relevant resource files. If not kuttl tests, unit tests would also work (e.g., networkpolicies_test.go).
4. Not blocking, but it could be nice to add tests for the new functions TruncateCRName(), GetTruncatedCRName(), and nameWithSuffix().
5. I noticed that references to argoutil.TruncateWithHash() in networkpolicies.go may truncate the suffix when the Argo CD instance name is long. To keep things consistent, might consider using the nameWithSuffix approach, since it preserves the suffix.

aali309:
> * The function nameWithSuffix might be clearer if renamed. (and there is an existing function with the same name)

On your 2nd point, the function already existed in the codebase, and renaming it is a separate concern that should be addressed in another PR.  I will create another follow up PR for this once this is merged.

controllers/argocd/rolebinding_test.go
```diff
@@ -325,7 +326,7 @@ func TestReconcileArgoCD_reconcileRoleBinding_forSourceNamespaces(t *testing.T)
 	expectedName := getRoleBindingNameForSourceNamespaces(a.Name, sourceNamespace)
 
 	// Verify the name is truncated to 63 characters
-	assert.LessOrEqual(t, len(expectedName), maxLabelLength, "RoleBinding name should not exceed maxLabelLength")
+	assert.LessOrEqual(t, len(expectedName), 63, "RoleBinding name should not exceed maxLabelLength")
```
ranakan19: any reason to remove the maxLabelLength usage here? Its better to not hardcode it

aali309: was doing some testing, forgot to restore it

controllers/argocd/rolebinding_test.go
```diff
@@ -369,25 +370,25 @@ func TestTruncateWithHash(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			result := truncateWithHash(tt.input)
+			result := argoutil.TruncateWithHash(tt.input)
 
 			// Check length constraint
-			assert.LessOrEqual(t, len(result), maxLabelLength, "Result should not exceed maxLabelLength")
+			assert.LessOrEqual(t, len(result), 63, "Result should not exceed maxLabelLength")
```
ranakan19: leftover hardcoding

controllers/argocd/rolebinding_test.go
```diff
@@ -369,25 +370,25 @@ func TestTruncateWithHash(t *testing.T) {
...
 
 			// Check that result is deterministic
-			result2 := truncateWithHash(tt.input)
+			result2 := argoutil.TruncateWithHash(tt.input)
 			assert.Equal(t, result, result2, "Function should be deterministic")
 
 			// For short strings, should be unchanged
-			if len(tt.input) <= maxLabelLength {
+			if len(tt.input) <= 63 {
```
ranakan19: leftover hardcoding

controllers/argocd/rolebinding_test.go
```diff
@@ -369,25 +370,25 @@ func TestTruncateWithHash(t *testing.T) {
...
 				assert.Equal(t, tt.input, result, "Short strings should not be modified")
 			} else {
 				// For long strings, should be different and shorter
 				assert.NotEqual(t, tt.input, result, "Long strings should be modified")
-				assert.LessOrEqual(t, len(result), maxLabelLength, "Result should be within length limit")
+				assert.LessOrEqual(t, len(result), 63, "Result should be within length limit")
```
ranakan19: leftover hardcoding

controllers/argocd/rolebinding_test.go
```diff
@@ -369,25 +370,25 @@ func TestTruncateWithHash(t *testing.T) {
...
 
 				// Should contain hash suffix if truncated
-				if len(tt.input) > maxLabelLength {
+				if len(tt.input) > 63 {
```
ranakan19: leftover hardcoding

controllers/argocd/rolebinding_test.go
```diff
@@ -434,19 +435,19 @@ func TestGetRoleBindingNameForSourceNamespaces(t *testing.T) {
 			result := getRoleBindingNameForSourceNamespaces(tt.argocdName, tt.targetNamespace)
 
 			// Check length constraint
-			assert.LessOrEqual(t, len(result), maxLabelLength, "RoleBinding name should not exceed maxLabelLength")
+			assert.LessOrEqual(t, len(result), 63, "RoleBinding name should not exceed maxLabelLength")
```
ranakan19: leftover hardcoding

controllers/argocd/rolebinding_test.go
```diff
@@ -434,19 +435,19 @@ func TestGetRoleBindingNameForSourceNamespaces(t *testing.T) {
...
 
 			// Check that result is deterministic
 			result2 := getRoleBindingNameForSourceNamespaces(tt.argocdName, tt.targetNamespace)
 			assert.Equal(t, result, result2, "Function should be deterministic")
 
 			// For short namespaces, should contain original namespace name
-			if len(tt.argocdName)+len(tt.targetNamespace)+1 <= maxLabelLength {
+			if len(tt.argocdName)+len(tt.targetNamespace)+1 <= 63 {
```
ranakan19: leftover hardcoding

controllers/argocd/rolebinding_test.go
```diff
@@ -434,19 +435,19 @@ func TestGetRoleBindingNameForSourceNamespaces(t *testing.T) {
...
 				expected := fmt.Sprintf("%s_%s", tt.argocdName, tt.targetNamespace)
 				assert.Equal(t, expected, result, "Short namespaces should not be truncated")
 			} else {
 				// For long namespaces, should be truncated and contain hash
-				assert.LessOrEqual(t, len(result), maxLabelLength, "Long namespaces should be truncated")
+				assert.LessOrEqual(t, len(result), 63, "Long namespaces should be truncated")
```
ranakan19: leftover hardcoding

controllers/argocd/rolebinding_test.go
```diff
@@ -468,11 +469,11 @@ func TestTruncateWithHashUniqueness(t *testing.T) {
...
 
 	for _, input := range inputs {
-		result := truncateWithHash(input)
+		result := argoutil.TruncateWithHash(input)
 		assert.False(t, results[result], "Hash should be unique for different inputs: %s", input)
 		results[result] = true
 
 		// Verify length constraint
-		assert.LessOrEqual(t, len(result), maxLabelLength, "Result should not exceed maxLabelLength")
+		assert.LessOrEqual(t, len(result), 63, "Result should not exceed maxLabelLength")
```
ranakan19: leftover hardcoding

controllers/argocd/applicationset.go
```diff
@@ -943,7 +943,7 @@ func (r *ReconcileArgoCD) reconcileApplicationSetService(cr *argoproj.ArgoCD) er
 	}
 
 	svc.Spec.Selector = map[string]string{
-		common.ArgoCDKeyName: nameWithSuffix(common.ApplicationSetServiceNameSuffix, cr),
+		common.ArgoCDKeyName: argoutil.TruncateWithHash(fmt.Sprintf("%s-%s", cr.Name, common.ApplicationSetServiceNameSuffix)),
```
ranakan19:
```suggestion
		common.ArgoCDKeyName: nameWithSuffix(common.ApplicationSetServiceNameSuffix, cr),
```

This should be keeping the suffix as well.
For the long argocd name - `argocd-long-name-for-route-testiiiiiiiiiiiiiiiiiiiiiiiing` the selector with your current changes would have the value `argocd-long-name-for-route-testiiiiiiiiiiiiiiiii-2a814fa2e3b158` instead of the expected `argocd-long-name-for-r-2a814fa2e3b158-applicationset-controller` 
Correct me if wrong.

aali309: No, I actually missed this!

ranakan19 (approved):

controllers/argoutil/resource.go
```diff
@@ -33,6 +34,19 @@ import (
 	"github.com/argoproj-labs/argocd-operator/common"
 )
 
+const (
+	hashLabelLength = 7
+	// MaxLabelLength is the maximum length for Kubernetes labels and names
+	MaxLabelLength = 63
```
anandf: Why `MaxLabelLength` has to have public access ? If all the truncation logic is within this utility file, we can make it private isn't it ? Can it be changed to `maxLabelLength` ? If this constant is required to be used else where then its better to make it a global constant in file `constants.go`.

controllers/argoutil/resource.go
```diff
@@ -33,6 +34,19 @@ import (
...
+	// Maximum suffix length is "applicationset-controller" = 25 characters
+	// So CR name should be limited to 63 - 25 - 1 (hyphen) = 37 characters
+	maxSuffixLength = 25
+	// MaxCRNameLength is the maximum length for ArgoCD CR names to accommodate longest suffix
+	MaxCRNameLength = MaxLabelLength - maxSuffixLength - 1 // -1 for hyphen separator
```
anandf: Same comment. Can it be made private access?

controllers/argoutil/resource.go
```diff
@@ -33,6 +34,19 @@ import (
...
+	// TruncatedNameAnnotation is the annotation key to store truncated CR name
+	TruncatedNameAnnotation = "argoproj.io/truncated-name"
```
anandf: Can it be made private access ?

controllers/argocd/util.go
```diff
@@ -678,9 +679,41 @@ func (r *ReconcileArgoCD) redisShouldUseTLS(cr *argoproj.ArgoCD) bool {
 	return false
 }
 
+// reconcileTruncatedNameAnnotation ensures that the ArgoCD CR has the truncated name annotation
+// if its name exceeds the maximum length for proper suffix handling
+func (r *ReconcileArgoCD) reconcileTruncatedNameAnnotation(cr *argoproj.ArgoCD) error {
```
anandf: if the name is already truncated ? but if the user removes the annotation, then reconcile would not add it right ? or this scenario is already handled in the code ?

anandf (changes requested):
Posted few minor comments.

controllers/argoutil/resource.go
aali309: Thanks @anandf ,  made the constant private as suggested.

controllers/argoutil/resource.go
```diff
@@ -205,3 +219,70 @@ func IsTrackedByOperator(labels map[string]string) bool {
...
+	// First check if we have a stored truncated name in annotations
+	if cr.Annotations != nil {
+		if truncatedName, exists := cr.Annotations[truncatedNameAnnotation]; exists {
+			return truncatedName
+		}
+	}
+
+	// Otherwise, truncate the current name
+	return TruncateCRName(cr.Name)
+}
```
svghadi, line 262: I feel we should rely on the internal truncate function to generate names rather than using the annotation. The annotation can still be added for end-user visibility and debugging, but it shouldn’t be treated as the source of truth. Since Kubernetes object names are immutable, we can safely rely on the truncate function to generate deterministic names every time it runs.

controllers/argoutil/resource.go
```diff
@@ -205,3 +219,70 @@ func IsTrackedByOperator(labels map[string]string) bool {
...
+	// Calculate hash of the original CR name for uniqueness
+	hash := sha1.Sum([]byte(crName)) // #nosec G401 - SHA1 used for non-cryptographic name hashing only
+	hashSuffix := fmt.Sprintf("-%x", hash[:hashLabelLength])
+
+	// Calculate how much we can truncate (accounting for hash suffix)
+	maxBaseLength := maxCRNameLength - len(hashSuffix)
+
+	// Truncate and add hash
+	return crName[:maxBaseLength] + hashSuffix
+}
```
svghadi: Any specific reason we need two separate truncate functions? From what I understand, the goal is to produce names in the format `<truncated-argocd-cr-name>-<hash(7)>-<suffix>` whenever the `ArgoCD CR name` length exceeds `38 (maxCRNameLength)`.

tests/ha/1-020_validate_redis_ha_nonha/02-assert.yaml
svghadi, line 1: Can we migrate all these kuttl tests to ginko?

controllers/argoutil/resource.go
```diff
@@ -205,3 +219,70 @@ func IsTrackedByOperator(labels map[string]string) bool {
...
+	// Calculate hash of the original string
+	hash := sha1.Sum([]byte(input)) // #nosec G401 - SHA1 used for non-cryptographic name hashing only
+	hashSuffix := fmt.Sprintf("-%x", hash[:hashLabelLength])
+
+	// Calculate how much we can truncate
+	maxBaseLength := maxLabelLength - len(hashSuffix)
+
+	// Truncate and add hash
+	return input[:maxBaseLength] + hashSuffix
+}
```
svghadi: Correct me if I’m wrong, but I don’t see any tests for `TruncateWithHash`. Could we add unit tests to verify that the generated names follow the expected format? Since the SHA is deterministic, we can pre-compute the expected names and compare them directly as strings.

tests/ocp/1-007_verify_hostname_with_route/02-validate-secrets.yaml
```diff
@@ -0,0 +1,25 @@
+# Test that validates the secrets are created with better truncated names (preserving suffixes)
+# This addresses GITOPS-7315: Secret names with preserved suffixes for better user experience
+apiVersion: v1
+kind: Secret
+metadata:
+  # The cluster secret name: CR name truncated, suffix preserved
+  name: argocd-long-name-for-r-9800df6ee28bb4-cluster
```
svghadi: The computed hash is supposed to be 7 characters long, right?

controllers/argocd/util.go
```diff
@@ -678,9 +679,46 @@ func (r *ReconcileArgoCD) redisShouldUseTLS(cr *argoproj.ArgoCD) bool {
 	return false
 }
 
+// reconcileTruncatedNameAnnotation ensures that the ArgoCD CR has the truncated name annotation
+// if its name exceeds the maximum length for proper suffix handling
```
svghadi: Adding an annotation on the ArgoCD CR could potentially impact some users. For example, users who store the ArgoCD CR in Git and sync it back with Argo CD may see drift if ignore differences aren’t configured properly. AFAIK, we don’t currently add any annotations or labels on the ArgoCD CRs, so this would be the first case. Not all users will have long names, but just wanted to highlight this.

What do you think about adding something on the created resources instead of the ArgoCD CR itself? Just a thought.

tests/ha/1-020_validate_redis_ha_nonha/02-assert.yaml
aali309: Yes, I think the migration is already complete?

controllers/argoutil/resource.go
aali309: updated

controllers/argoutil/resource.go
aali309: I've consolidated the two functions into one parameterized TruncateWithHash(input, maxLength) function.

controllers/argocd/util.go
aali309: Great catch! I've removed the `reconcileTruncatedNameAnnotation` function entirely. The operator now uses the deterministic `TruncateCRName` function as the source of truth without modifying the ArgoCD CR at all.

controllers/argocd/util.go
```diff
@@ -681,6 +682,11 @@ func (r *ReconcileArgoCD) redisShouldUseTLS(cr *argoproj.ArgoCD) bool {
 // reconcileResources will reconcile common ArgoCD resources.
 func (r *ReconcileArgoCD) reconcileResources(cr *argoproj.ArgoCD, argocdStatus *argoproj.ArgoCDStatus) error {
 
+	log.Info("reconciling status")
+	if err := r.reconcileStatus(cr, argocdStatus); err != nil {
+		log.Info(err.Error())
+	}
+
```
svghadi: Why is this change required?

aali309: I think I did this for debugging and appears to have been accidentally added.

svghadi:
@aali309 @ranakan19 - any objection to merging this one?

PR #1822, opened 2025-08-11
fix for setting GODEBUG environment variable to fips140=on in a FIPS cluster

anandf:
**What type of PR is this?**
/king bug

Fixes https://issues.redhat.com/browse/GITOPS-7424

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -24,7 +24,6 @@ import (
 	goruntime "runtime"
 	"strings"
 
-	"github.com/argoproj-labs/argocd-operator/controllers/argoutil"
 	"github.com/argoproj/argo-cd/v3/util/env"
 	appsv1 "github.com/openshift/api/apps/v1"
 	configv1 "github.com/openshift/api/config/v1"
@@ -39,6 +38,7 @@ import (
 	"github.com/argoproj-labs/argocd-operator/common"
 	"github.com/argoproj-labs/argocd-operator/controllers/argocd"
 	"github.com/argoproj-labs/argocd-operator/controllers/argocdexport"
+	"github.com/argoproj-labs/argocd-operator/controllers/argoutil"
 
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 


controllers/argocd/deployment.go
```diff
@@ -965,6 +965,21 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argoproj.ArgoCD, useTLSFor
 		repoEnv = argoutil.EnvMerge(repoEnv, []corev1.EnvVar{{Name: "ARGOCD_EXEC_TIMEOUT", Value: fmt.Sprintf("%ds", *cr.Spec.Repo.ExecTimeout)}}, true)
 	}
 
+	// if running in a FIPS enabled host, set the GODEBUG env wit appropriate value
+	fipsConfigChecker := r.FipsConfigChecker
+	if fipsConfigChecker != nil {
+		fipsEnabled, err := fipsConfigChecker.IsFipsEnabled()
+		if err != nil {
+			return err
+		}
```
svghadi, line 974: Do we want to fail on error or just log and continue? On non-FIPS clusters this might be disruptive if restricted policies prevent reading `/proc/sys/crypto/fips_enabled`. It’s a rare case but just highlighting.

svghadi (commented):
Left a comment. Rest all looks good.

controllers/argocd/deployment.go
anandf: if file is not existent, we don't return an error. We check for non existent file path and ignore the error.  In the `fileExists` function, I have put this logic.
```
if err != nil {
    if os.IsNotExist(err) {
	    return false, nil
    }
    return false, err
}
```

svghadi: Okay. And are permission issues handled in similar way?

anandf: permission errors are thrown as errors. If the file does not exist, it is not treated as an error, but if the file exists but not readable, it is treated as an error. I have added unit tests for both these scenarios.

PR #1823, opened 2025-08-11
fix: added logic to remove custom labels from live objects if removed from argocd spec

alkakumari016:
Assisted-by: < Cursor/Gemini etc>

**What type of PR is this?**

/kind bug

When we remove custom labels from the ArgoCD spec (set them to {}), the new code iterates over an empty map, so it doesn't add any new keys, but it also doesn't remove the existing custom keys that should be cleaned up. This PR fixes this bug.

Fixes #?
GITOPS-7405


svghadi:
@coderabbitai full review

tests/ginkgo/parallel/1-043_validate_custom_labels_annotations_test.go
```diff
@@ -16,198 +16,199 @@ limitations under the License.
...
-
-			}
-
-			expectLabelsAndAnnotationsRemovedFromDepl(appsetDepl)
-			expectLabelsAndAnnotationsRemovedFromDepl(serverDepl)
-			expectLabelsAndAnnotationsRemovedFromDepl(repoDepl)
-
-			// Evaluate the controller statefulset on its own, since it's a StatefulSet not a Deployment
-			Eventually(controllerSS).Should(k8sFixture.ExistByName())
+//TODO
```
olivergondza, line 19: Pointing out the unresolved TODO. Github hides the hunk for its length...

PR #1827, opened 2025-08-12
feat: Update apiVersion argoproj.io/v1alpha1 to v1beta1 for ArgoCD in docs and tests

olivergondza:
**What type of PR is this?**

/kind chore

The examples in the docs are outdated

None

### DIFF
--- a/docs/developer-guide/development.md
+++ b/docs/developer-guide/development.md
@@ -129,7 +129,7 @@ For more infomation see [build operator images to test on a cluster.](https://ar
 
 Once the operator is installed, you would need to configure an ArgoCD instance that the operator would manage. The sample instance configuration is below:
 ``` yaml
-apiVersion: argoproj.io/v1alpha1
+apiVersion: argoproj.io/v1beta1
 kind: ArgoCD
 metadata:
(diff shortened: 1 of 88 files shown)


jgwest (changes requested):
For the kuttl tests, IMHO we can leave them as is and remove those changes from this PR. Why?
- The kuttl tests will likely be entirely removed shortly, anyways.
- The kuttl tests are no longer run as part of PRs, so this PR may unintentionally break them and we would not know. (Unless, for example, you have personally run those kuttl tests yourself from your own machine)

olivergondza:
> For the kuttl tests, IMHO we can leave them as is and remove those changes from this PR. Why?
> 
> * The kuttl tests will likely be entirely removed shortly, anyways.
> * The kuttl tests are no longer run as part of PRs, so this PR may unintentionally break them and we would not know. (Unless, for example, you have personally run those kuttl tests yourself from your own machine)

Yeah, makes sense since we want to throw them away. For my complete kuttl ignorance, it will that be all the references in `tests/` of just the ones with kuttl CRs? Thanks!

PR #1836, opened 2025-08-22
fix: argocd-agent: remove LoadBalancer from repo-server

jgwest:
**What type of PR is this?**
/kind bug

At present, when argocd-agent is enabled, the repo server `Service` will be converted to a LoadBalancer. As discussed on internal Slack, repo server no longer need to be exposed via a LoadBalancer.


jparsai:
LGTM

jgwest:
Thanks @jparsai!

PR #1841, opened 2025-08-26
bug: fix redis fails after upgrade

### LINKED ISSUE #1804: ArgoCD Redis container fails with "runAsNonRoot and image will run as root" after upgrade to argocd-operator 0.14.1
**Description:**

After upgrading argocd-operator from version 0.13.1 to 0.14.1, the `argocd-redis` pod fails to start with the following error:

```
state:
  waiting:
    message: 'container has runAsNonRoot and image will run as root (pod: "argocd-redis-64656fd77c-5b58n_argocd(4433a4b7-dc74-4ab6-b270-5f47ce031643)", container: redis)'
    reason: CreateContainerConfigError
```

**Environment:**
- ArgoCD Operator version: 0.14.1 (previously 0.13.1)
- Kubernetes version: v1.32.4-gke.1698000
- Redis image: `redis@sha256:8061ca607db2a0c80010aeb5fc9bed0253448bc68711eaa14253a392f6c48280`

**Root Cause:**
The security context configuration changed between versions. The working configuration (v0.13.1) included:
```yaml
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop:
    - ALL
  runAsNonRoot: true
  runAsUser: 999  # This line is missing in v0.14.1
  seccompProfile:
    type: RuntimeDefault
```

The new configuration (v0.14.1) is missing the `runAsUser: 999` field:
```yaml
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop:
    - ALL
  runAsNonRoot: true  # Without runAsUser, defaults to root (UID 0)
  seccompProfile:
    type: RuntimeDefault
```

**Expected Behavior:**
The Redis container should start successfully with the same security context as the previous version.

**Actual Behavior:**
The container fails to start because `runAsNonRoot: true` is specified but no `runAsUser` is set, causing the container to attempt running as root (UID 0), which violates the `runAsNonRoot` constraint.

anandrkskd:
**What type of PR is this?**

/kind bug 

Fixes #1804

### DIFF
--- a/controllers/argocd/deployment.go
+++ b/controllers/argocd/deployment.go
@@ -616,6 +616,15 @@ func (r *ReconcileArgoCD) reconcileRedisDeployment(cr *argoproj.ArgoCD, useTLS b
 			changed = true
 		}
 
+		if !reflect.DeepEqual(deploy.Spec.Template.Spec.SecurityContext, existing.Spec.Template.Spec.SecurityContext) {
+			existing.Spec.Template.Spec.SecurityContext = deploy.Spec.Template.Spec.SecurityContext
+			if changed {
+				explanation += ", "
+			}
+			explanation += "pod security context"
+			changed = true
+		}
+
 		if !reflect.DeepEqual(deploy.Spec.Template.Spec.ServiceAccountName, existing.Spec.Template.Spec.ServiceAccountName) {
 			existing.Spec.Template.Spec.ServiceAccountName = deploy.Spec.Template.Spec.ServiceAccountName
 			if changed {


svghadi:
@anandrkskd - is this ready for review?

controllers/argocd/deployment.go
```diff
@@ -531,6 +525,10 @@ func (r *ReconcileArgoCD) reconcileRedisDeployment(cr *argoproj.ArgoCD, useTLS b
 		},
 	}}
 
+	if !IsOpenShiftCluster() {
+		deploy.Spec.Template.Spec.Containers[0].SecurityContext.RunAsUser = int64Ptr(999)
```
anandf: Is the change required because the init containers must not run as `999` user ?

svghadi:
I was discussing this issue with @Mangaal. He suggested to add the missing update logic for `deploy.Spec.Template.Spec.SecurityContext` instead of dropping `1000` userid.  

Currently, the update logic only checks for  `SecurityContext` at container level, it should also check at template level in deployment
https://github.com/argoproj-labs/argocd-operator/blob/0882af31483c3779ea6dafa394db109eaede6ffd/controllers/argocd/deployment.go#L610-L617

controllers/argocd/deployment.go
anandrkskd: Removed this part, pod Security context will now compare and update.

PR #1842, opened 2025-08-26
fix: ArgoCD agent principal image configurations

jparsai:
/kind bug

This PR is to add logic for principal where image name is first taken from ArgoCD resource, if not available then it is taken from an env variable. If that is also not available then default is used.

Fixes: https://issues.redhat.com/browse/GITOPS-7605


jgwest (approved):
LGTM, thanks @jparsai!

PR #1847, opened 2025-08-28
fix: ArgoCD agent principal configurations update

jparsai:
**What type of PR is this?**

Fixes #?


controllers/argocdagent/configmap.go
```diff
@@ -71,6 +71,7 @@ const (
 	PrincipalKeepAliveMinInterval      = "principal.keep-alive-min-interval"
 	PrincipalRedisCompressionType      = "principal.redis-compression-type"
 	PrincipalPprofPort                 = "principal.pprof-port"
+	PrincipalRedisPassword             = "auth"
```
jannfis, line 74: This constant name may be a little misleading, in that people can think `auth` is the actual password used (I hope it isn't 😅 )

controllers/argocdagent/configmap.go
```diff
@@ -167,7 +172,7 @@ func ReconcilePrincipalConfigMap(client client.Client, compName string, cr *argo
 func buildConfigMap(cr *argoproj.ArgoCD, compName string) *corev1.ConfigMap {
 	return &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      cr.Name + "-agent-params",
+			Name:      cr.Name + cmSuffix,
```
jannfis, line 175: To be honest, I'm not sure if this will work. The agent uses a hard-coded ConfigMap name `argocd-agent-params` right now. If you name that ConfigMap after the CR's name, it wouldn't use that ConfigMap.

controllers/argocdagent/configmap.go
jparsai: I think it has to be same as ArgoCD Redis secret https://github.com/argoproj-labs/argocd-agent/blob/main/hack/dev-env/common/redis-secret.yaml#L3

controllers/argocdagent/configmap.go
jparsai: Yes @jann, but for principal deployed through operator are having manifests according to this CM [here](https://github.com/argoproj-labs/argocd-operator/blob/release-0.15/controllers/argocdagent/deployment.go#L295), this only for now, once my PR with updated ArggoCD CR is merged, ConfigMap will not be used any more and all the code related to CM will be removed as you suggested that it is just to initialize env variables in deployment. I wanted to merge it in 1.18 but I got busy with Cluster Cache feature as that was higher priority for 1.18 and this got delayed.

jannfis: Ah, OK I see. I forgot that this ConfigMap is mounted to the pod by name we chose. All good, I guess.

jannfis (approved):
LGTM

controllers/argocdagent/configmap.go
jannfis: It really should be `PrincipalRedisPasswordKey` to remove confusion. But it's only cosmetical.

jgwest:
Thanks @jparsai and @jannfis for reviewing!

controllers/argocdagent/configmap.go
jparsai: I will make sure to change it in other PR which is for ArgoCD CR update

PR #1849, opened 2025-08-28
fix: ArgoCD agent CM TLS allow-generate False by default

jparsai:
/kind bug

This PR is to change default value of `principal.tls.server.allow-generate` field in `argocd-agent-params` configmap.


jparsai:
This is blocked by https://github.com/argoproj-labs/argocd-operator/pull/1847 because it is adding a new check in e2e (https://github.com/argoproj-labs/argocd-operator/pull/1847/files#diff-69c765a7f507dc1f36d0ba214a162853a99fd50340d11cea8c7da6228877b9b0R122) where principal.tls.server.allow-generate is set to true. I will update that check in this PR once that is merged.

PR #1851, opened 2025-08-29
Added related images to CSV

### LINKED ISSUE #1643: relatedimages issue in olm
I'm using ArgoCD from OperatorHub in a private/air-gapped environment and need to know which images should be synced to my private registry. This information should be defined in the `relatedImages` section of the CSV (ClusterServiceVersion), but the ArgoCD operator currently lacks this definition.

This makes it challenging to:
1. Identify all required images for air-gapped installations
2. Ensure all necessary images are mirrored to private registries
3. Deploy ArgoCD operator reliably in disconnected environments

Could we add the missing `relatedImages` section to properly document all required images for the ArgoCD operator?

This would help users in private/air-gapped environments to properly prepare their environments for ArgoCD deployment.

enriqueav99:
**What type of PR is this?**

/kind enhancement 

Add to the csv the related images

Fixes #1643 

The images with tags and not with digests are because they were specified that way in the file commons/defaults.go


enriqueav99:
Hello! That’s strange. When generating the bundle, the operator-sdk seems to ignore the relatedImages defined in the bases.

Looking at this issue [https://github.com/operator-framework/operator-sdk/issues/5763](https://github.com/operator-framework/operator-sdk/issues/5763), It should be able to add them.

jgwest:
Rather than updating the `relatedImages` once (in this PR), it would perhaps make more sense to have a script that instead detects and injects these fields for every release

PR #1860, opened 2025-09-09
bug: fix redis ha proxy pod fails to restart after upgrade

alkakumari016:
**What type of PR is this?**

This PR fixes the bug where redis ha proxy pod fails to restart with error related to security context after upgrade

Fixes #?
[GITOPS-7180](https://issues.redhat.com/browse/GITOPS-7180)


Mangaal:
@alkakumari016 , Could you please check if other workloads are also missing this security context check?

nmirasch (approved):
LGTM

alkakumari016:
> @alkakumari016 , Could you please check if other workloads are also missing this security context check?

@Mangaal as Discussed we will add the fix to other components on another PR.

PR #1866, opened 2025-09-11
feat: Cache transform for Secrets and ConfigMaps to reduce memory

svghadi:
**What type of PR is this?**
/kind enhancement 

This PR reduces memory usage in the Argo CD Operator by introducing a cache transform and client wrapper for Secrets and ConfigMaps.
- Cache transform: Strips large data fields from non-operator Secrets/ConfigMaps before storing them in the controller-runtime cache, while retaining full content for operator-tracked objects.
- Client wrapper: Falls back to the live client when a stripped object is accessed and applies a tracking label so future updates are fully cached.
- Integration: Wired into main.go via cache.Options.ByObject and applied across reconcilers.

See the attached proposal in this PR for design details, trade-offs, PoC results, and future scope.
- [proposal](https://github.com/argoproj-labs/argocd-operator/blob/013277d4e3dfbb0f425a5346a2b69a4e7816a0c7/docs/proposals/002-controller-runtime-cache-transforms-for-secrets-and-configmaps.md)

**Results**
These metrics were collected from a test cluster containing **100 ConfigMaps** and **100 Secrets**, each approximately **1 MB** in size. The cluster was running four Argo CD instances and no other workload operators.

With the optimization enabled, operator memory usage dropped from **~350 MB** to **~100 MB**.

UnOptimized Operator Manager Memory:

unoptimized-manager-memory

Optimized Operator Manager Memory:

optimized-manager-memory

However, we could not reduce the startup memory consumption, which remained at **~750 MB** in both cases.

We previously attempted another approach in [#1795](https://github.com/argoproj-labs/argocd-operator/pull/1795), but it introduced significant complexity and restricted how watches could be set up. Compared to that, this solution provides a better balance between complexity, maintenance overhead, and outcome.

### DIFF
--- a/pkg/clientwrapper/client.go
+++ b/pkg/clientwrapper/client.go
@@ -41,7 +41,7 @@ func (cw *ClientWrapper) Get(ctx context.Context, key types.NamespacedName, obj 
 
 	switch o := obj.(type) {
 	case *corev1.Secret:
-		if secretNeedsLiveRefresh(o) {
+		if !cacheutils.IsTrackedByOperator(o) && secretNeedsLiveRefresh(o) {
 			if err := cw.liveClient.Get(ctx, key, obj, opts...); err != nil {
 				return err
 			}
@@ -53,7 +53,7 @@ func (cw *ClientWrapper) Get(ctx context.Context, key types.NamespacedName, obj 
 			}
 		}
 	case *corev1.ConfigMap:
-		if configmapNeedsLiveRefresh(o) {
+		if !cacheutils.IsTrackedByOperator(o) && configmapNeedsLiveRefresh(o) {
 			if err := cw.liveClient.Get(ctx, key, obj, opts...); err != nil {
 				return err
 			}
@@ -72,10 +72,6 @@ func (cw *ClientWrapper) Get(ctx context.Context, key types.NamespacedName, obj 
 
 // secretNeedsLiveRefresh returns true if the cached secret looks stripped or untracked.
 func secretNeedsLiveRefresh(s *corev1.Secret) bool {
-	if !cacheutils.IsTrackedByOperator(s) {
-		return true
-	}
-
 	// Heuristic: a "stripped" Secret from our transform has nil Data/StringData.
 	// A truly empty Secret may also match, but that only triggers an extra live GET,
 	// which is rare and acceptable.
@@ -87,10 +83,6 @@ func secretNeedsLiveRefresh(s *corev1.Secret) bool {
 
 // configmapNeedsLiveRefresh returns true if the cached cm looks stripped or untracked.
 func configmapNeedsLiveRefresh(cm *corev1.ConfigMap) bool {
-	if !cacheutils.IsTrackedByOperator(cm) {
-		return true
-	}
-
 	// Heuristic: a "stripped" ConfigMap from our transform has nil Data/BinaryData.
 	// A truly empty ConfigMap may also match, but that only triggers an extra live GET,
 	// which is rare and acceptable.


pkg/cacheutils/utlis.go
```diff
@@ -0,0 +1,69 @@
...
+// when caching Secrets that are not managed by the operator.
+func StripSecretDataTransform() clientgotools.TransformFunc {
+	return func(in interface{}) (interface{}, error) {
+		if s, ok := in.(*v1.Secret); ok {
+			if IsTrackedByOperator(s.Labels) {
+				// Keep full secret
+				return in, nil
+			}
+			// Strip data for non-operator secrets
+			return &v1.Secret{
```
anandf: Rather than creating a new Object, can we not set the `Data` field to `nil` on the same object ? When we create new Secret object. The old object with the data field also exists in memory and waits for Garbage collection to clean it up, isn't it ?

pkg/cacheutils/utlis.go
```diff
@@ -0,0 +1,69 @@
...
+			}, nil
+		}
+		return in, nil
+	}
+}
+
+// StripConfigMapDataTransform returns a TransformFunc that strips the data from ConfigMaps
+// that are not tracked by the operator. This is useful for reducing memory usage
+// when caching ConfigMaps that are not managed by the operator.
+func StripConfigMapDataTransform() clientgotools.TransformFunc {
```
anandf: Can we have a single generic function that can handle both `Secret` and `ConfigMap` ? Anyways the function takes in an generic `interface{}` struct.

pkg/cacheutils/utlis.go
```diff
@@ -0,0 +1,69 @@
...
+				Data:       nil,
+				BinaryData: nil,
+			}, nil
+		}
+		return in, nil
+	}
+}
+
+// IsTrackedByOperator checks if the given labels indicate that the resource is tracked by the operator.
+func IsTrackedByOperator(labels map[string]string) bool {
```
anandf: Can it take in a `runtime.Object` as input instead of labels ? That would be more meaningful as the method checks if a given resource is tracked by operator or not.

pkg/cacheutils/utlis.go
anandrkskd: Hey @anandf I have updated it according to how you have mentioned in comment

pkg/cacheutils/utlis.go
anandrkskd: Unified both function

pkg/clientwrapper/client.go
svghadi, line 1: Is it possible to add some tests for the wrapper?

docs/usage/environment_variables.md
```diff
@@ -9,6 +9,8 @@ The following environment variables are available in `argocd-operator`:
 | `REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION` | false | When an Argo CD instance is deleted, namespaces managed by that instance (via the `argocd.argoproj.io/managed-by` label ) will retain the label by default. Users can change this behavior by setting the environment variable `REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION` to `true` in the Subscription. |
 | `ARGOCD_LABEL_SELECTOR` | none | The label selector can be set on argocd-opertor by exporting `ARGOCD_LABEL_SELECTOR` (eg: `export ARGOCD_LABEL_SELECTOR=foo=bar`). The labels can be added to the argocd instances using the command `kubectl label argocd test1 foo=bar -n test-argocd`. This will enable the operator instance to be tailored to oversee only the corresponding ArgoCD instances having the matching label selector. |
 | `LOG_LEVEL` | info | This sets the logging level of the manager (operator) pod. Valid values are "debug", "info", "warn", "error", "panic" and "fatal". |
+| `MEMORY_OPTIMIZATION_ENABLED` | true | Enables memory optimization by stripping data from Secrets and ConfigMaps that are not tracked by the operator, reducing memory usage. Set to `false` to disable this optimization. |
```
svghadi:
```suggestion
| `DISABLE_MEMORY_OPTIMIZATION` | false | When set to `true`, disables the memory optimization that strips data from Secrets and ConfigMaps that are not tracked by the operator.  This optimization helps reduce memory usage. |
```

cmd/main.go
```diff
@@ -182,18 +187,40 @@ func main() {
 		},
 	}
 
-	if watchedNsCache := getDefaultWatchedNamespacesCacheOptions(); watchedNsCache != nil {
+	// Use transformers to strip data from Secrets and ConfigMaps
+	// that are not tracked by the operator to reduce memory usage.
+	if strings.ToLower(os.Getenv("MEMORY_OPTIMIZATION_ENABLED")) != "false" {
```
svghadi, line 192: Lets use `DISABLE_MEMORY_OPTIMIZATION` env var which users should explicitly set to `true` to disable optimization

cmd/main.go
```diff
@@ -182,18 +187,40 @@ func main() {
...
+		setupLog.Error(err, "unable to create live client")
+		os.Exit(1)
+	}
+
+	// Wraps the controller runtime's default client to provide:
+	//   1. Fallback to the live client when a Secret/ConfigMap is stripped in the cache.
+	//   2. Automatic labeling of fetched objects, so they are retained in full form
+	//      in subsequent cache updates and avoid repeated live lookups.
+	client := cw.NewClientWrapper(mgr.GetClient(), liveClient)
+
```
svghadi: This logic can also be placed behind the `DISABLE_MEMORY_OPTIMIZATION` check. Use the default `client := mgr.GetClient()` when memory optimization is disabled.

pkg/cacheutils/utlis.go
svghadi: @anandrkskd - can we incorporate this suggestion aswell?

anandrkskd: updated

cmd/main.go
anandrkskd: updated

cmd/main.go
anandrkskd: updated

docs/usage/environment_variables.md
anandrkskd: updated

pkg/cacheutils/utlis.go
anandrkskd: updated

cmd/main.go
```diff
@@ -194,6 +209,22 @@ func main() {
 		os.Exit(1)
 	}
 
+	var client crclient.Client
+	if strings.ToLower(os.Getenv("DISABLE_MEMORY_OPTIMIZATION")) == "true" {
```
svghadi:
```suggestion
	if strings.ToLower(os.Getenv("DISABLE_MEMORY_OPTIMIZATION")) != "true" {
```

svghadi: Updated

pkg/clientwrapper/client.go
svghadi: unit tests added

pkg/clientwrapper/client.go
```diff
@@ -0,0 +1,122 @@
...
+func (cw *ClientWrapper) Get(ctx context.Context, key types.NamespacedName, obj ctrlclient.Object, opts ...ctrlclient.GetOption) error {
+
+	if err := cw.Client.Get(ctx, key, obj, opts...); err != nil {
+		return err
+	}
+
+	switch o := obj.(type) {
+	case *corev1.Secret:
+		if secretNeedsLiveRefresh(o) {
+
```
anandf: nitpick: Additional empty lines can be removed.

docs/usage/environment_variables.md
```diff
@@ -9,6 +9,8 @@ The following environment variables are available in `argocd-operator`:
 | `REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION` | false | When an Argo CD instance is deleted, namespaces managed by that instance (via the `argocd.argoproj.io/managed-by` label ) will retain the label by default. Users can change this behavior by setting the environment variable `REMOVE_MANAGED_BY_LABEL_ON_ARGOCD_DELETION` to `true` in the Subscription. |
 | `ARGOCD_LABEL_SELECTOR` | none | The label selector can be set on argocd-opertor by exporting `ARGOCD_LABEL_SELECTOR` (eg: `export ARGOCD_LABEL_SELECTOR=foo=bar`). The labels can be added to the argocd instances using the command `kubectl label argocd test1 foo=bar -n test-argocd`. This will enable the operator instance to be tailored to oversee only the corresponding ArgoCD instances having the matching label selector. |
 | `LOG_LEVEL` | info | This sets the logging level of the manager (operator) pod. Valid values are "debug", "info", "warn", "error", "panic" and "fatal". |
+| `DISABLE_MEMORY_OPTIMIZATION` | false | When set to `true`, disables the memory optimization that strips data from Secrets and ConfigMaps that are not tracked by the operator.  This optimization helps reduce memory usage. |
```
anandf: Instead of having a negative notion with `DISABLE_` prefix having a `false` value as default, it is better to have a positive notion something like - `ENABLE_MEMORY_OPTIMIZATION` or `MEMORY_OPTIMIZATION_ENABLED`.

pkg/cacheutils/utils.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/argoproj-labs/argocd-operator/common"
+)
+
+// StripDataFromSecretOrConfigMap returns a TransformFunc that strips data from both
+// Secrets and ConfigMaps that are not tracked by the operator. This unified function
+// handles both resource types in a single transform. This is useful for reducing memory usage
+// when caching Secrets and ConfigMaps that are not managed/used by the operator.
+func StripDataFromSecretOrConfigMapTransform() clientgotools.TransformFunc {
+	return func(in interface{}) (interface{}, error) {
+
```
anandf: nitpick: remove unwanted empty lines.

docs/usage/environment_variables.md
svghadi: Okay. Should the optimization be disabled by default? and let users opt in if needed

pkg/clientwrapper/client.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+
+	case *corev1.ConfigMap:
+		if configmapNeedsLiveRefresh(o) {
+
+			if err := cw.liveClient.Get(ctx, key, obj, opts...); err != nil {
+				return err
+			}
+
+			cw.ensureTrackedLabel(ctx, obj)
```
anandf: Is this code adding labels to all objects ? Ideally it should be doing it only for objects that is managed by the operator isn't it ? Also do we have `patch` verb added  to the operator service account for configmap/secrets across all namespaces ? this call might fail and there is no error handling done in this code.

docs/usage/environment_variables.md
anandrkskd: It will be enabled by default, We will only disable it when users set it to false explicitly, and we can mention the same in the documentation.

pkg/clientwrapper/client.go
```diff
@@ -0,0 +1,122 @@
...
+
+	}
+
+	return nil
+}
+
+// secretNeedsLiveRefresh returns true if the cached secret looks stripped or untracked.
+func secretNeedsLiveRefresh(s *corev1.Secret) bool {
+
+	if !cacheutils.IsTrackedByOperator(s) {
```
anandf: We are checking two things here. whether its an operator tracked resource and if so whether it needs a live refresh. Its better to track only if it needs live refresh in this method and do the additional check in the caller.

pkg/clientwrapper/client.go
```diff
@@ -0,0 +1,122 @@
...
+		labels = make(map[string]string, 1)
+	}
+	labels[common.ArgoCDTrackedByOperatorLabel] = common.ArgoCDAppName
+	obj.SetLabels(labels)
+
+	// Best-effort patch to add the operator tracking label.
+	// Non-fatal: a later reconcile will reattempt if this fails.
+	err := cw.liveClient.Patch(ctx, obj, ctrlclient.MergeFrom(orig))
+	if err != nil {
+		log.Error(err, "failed to add operator tracking label to object", "name", obj.GetName(), "namespace", obj.GetNamespace())
```
anandf: better to return error and handle it in the calling method.

pkg/clientwrapper/client_test.go
```diff
@@ -0,0 +1,332 @@
...
+		t.Fatalf("expected cached data to be returned, got %v", got.Data)
+	}
+
+	// Label should remain
+	if got.Labels[common.ArgoCDTrackedByOperatorLabel] != common.ArgoCDAppName {
+		t.Fatalf("expected tracking label to be preserved")
+	}
+}
+
+func TestClientWrapper_Get_ConfigMap_Stripped_ShouldLiveRefreshAndLabel(t *testing.T) {
```
anandf, line 176: Can we add a test case where the patch fails ?

pkg/clientwrapper/client.go
svghadi: > Is this code adding labels to all objects ? Ideally it should be doing it only for objects that is managed by the operator isn't it ? 

No, it only patches the Secrets and ConfigMaps for which the operator makes a GET call. The idea behind this is that the operator fetches certain resources only when required. So, if there’s a GET call, it indicates that the operator needs that resource. In such cases, if the tracking labels are missing, we add them to ensure the full object is stored in the cache for subsequent calls.

> Also do we have patch verb added to the operator service account for configmap/secrets across all namespaces ? 

Yes. The operator has `*` for `configmaps` and `secrets`

pkg/clientwrapper/client.go
```diff
@@ -0,0 +1,122 @@
...
+		liveClient: live,
+	}
+}
+
+// Get first tries to get from the cached client, and only falls back to live GET
+// if the cached object appears stripped or is missing required labels.
+// After a live GET, it also ensures the object has the tracking label (best-effort).
+func (cw *ClientWrapper) Get(ctx context.Context, key types.NamespacedName, obj ctrlclient.Object, opts ...ctrlclient.GetOption) error {
+
+	if err := cw.Client.Get(ctx, key, obj, opts...); err != nil {
```
anandf, line 38: `obj` is getting overridden here. Is that expected ?

pkg/clientwrapper/client.go
anandrkskd: IIUC, this function is right as live check will be required for those secret/configmaps that are tracked and have empty data fields and resources that are not labelled.  Both the checks are required to determine if it needs live refresh.

pkg/clientwrapper/client.go
svghadi: Yes. This follows the client-go implementation, where the same ob is passed around across calls.

pkg/clientwrapper/client.go
svghadi: Removed

docs/usage/environment_variables.md
svghadi: Changed it to `MEMORY_OPTIMIZATION_ENABLED`. It will be enabled by default.

pkg/clientwrapper/client.go
svghadi: Updated

pkg/clientwrapper/client_test.go
svghadi: Added

pkg/cacheutils/utils.go
svghadi: Removed

pkg/clientwrapper/client.go
anandf: Ok. The check is correct. I thought, In the calling function, we can do something like this 
```
if cacheutils.IsTrackedByOperator(s) && secretNeedsLiveRefresh(s) {..}
```

svghadi: Okay. Makes sense.

svghadi: Updated

PR #1871, opened 2025-09-15
fix: ArgoCD CR .status  is intermittently out of sync with actual Argo CD workloads

jgwest:
**What type of PR is this?**
/kind bug

- Primary purpose is to fix this status race condition, which has been causing periodic test failures: 
    - https://github.com/redhat-developer/gitops-operator/issues/909
- The fix modifies how status updating working in the operator: 
    - Functions which wish to modify the ArgoCD CR .status field, are instead passed a pointer to a `ArgoCDStatus` object. Functions should update this status object, not the `ArgoCD` CR status directly. This status object is set on the `ArgoCD` CR on completion of reconciliation.
- This PR:
    - Updates to .status field of `ArgoCD` CR only happen once, from one function (`updateStatusConditionOfArgoCD`), rather than many different functions.
    - Remove some old dex deletion logic which is no longer needed, and added a unit/ test to verify this is true.
        - We previously removed keycloak, which allows us to make this logic much simpler.
    - Some dex errors which were previously ignored, are no longer ignored (and IMHO never should have been).
    - Remove global variable from SSO reconciliation logic
    - Update other misc code, but no changes to the code logic, just adding the above pointers when needed.
    - Added a new E2E test to verify nothing has broken

Fixes https://github.com/redhat-developer/gitops-operator/issues/909

### DIFF
--- a/controllers/argocd/argocd_controller.go
+++ b/controllers/argocd/argocd_controller.go
@@ -124,22 +124,32 @@ var ActiveInstanceMap = make(map[string]string)
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.2/pkg/reconcile
 func (r *ReconcileArgoCD) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
 
-	result, argocd, err := r.internalReconcile(ctx, request)
+	result, argocd, argocdStatus, err := r.internalReconcile(ctx, request)
 
 	message := ""
 	if err != nil {
 		message = err.Error()
+		argocdStatus.Phase = "Failed" // Any error should reset phase back to Failed
 	}
 
-	if error := updateStatusConditionOfArgoCD(ctx, createCondition(message), argocd, r.Client, log); error != nil {
-		log.Error(error, "unable to update status of ArgoCD")
-		return reconcile.Result{}, error
+	log.Info("reconciling status")
+	if reconcileStatusErr := r.reconcileStatus(argocd, argocdStatus); reconcileStatusErr != nil {
+		log.Error(reconcileStatusErr, "Unable to reconcile status")
+		argocdStatus.Phase = "Failed"
+		message = "unable to reconcile ArgoCD CR .status field"
+	}
+
+	if updateStatusErr := updateStatusConditionOfArgoCD(ctx, createCondition(message), argocd, argocdStatus, r.Client, log); updateStatusErr != nil {
+		log.Error(updateStatusErr, "unable to update status of ArgoCD")
+		return reconcile.Result{}, updateStatusErr
 	}
 
 	return result, err
 }
 
-func (r *ReconcileArgoCD) internalReconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, *argoproj.ArgoCD, error) {
+func (r *ReconcileArgoCD) internalReconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, *argoproj.ArgoCD, *argoproj.ArgoCDStatus, error) {
+
+	argoCDStatus := &argoproj.ArgoCDStatus{} // Start with a blank canvas
 
 	reconcileStartTS := time.Now()
 	defer func() {
@@ -156,10 +166,10 @@ func (r *ReconcileArgoCD) internalReconcile(ctx context.Context, request ctrl.Re
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
-			return reconcile.Result{}, argocd, nil
+			return reconcile.Result{}, argocd, argoCDStatus, nil
 		}
 		// Error reading the object - requeue the request.
-		return reconcile.Result{}, argocd, err
+		return reconcile.Result{}, argocd, argoCDStatus, err
 	}
 
 	// If the number of notification replicas is greater than 1, display a warning.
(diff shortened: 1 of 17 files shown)


controllers/argocd/argocd_controller.go
```diff
@@ -124,22 +124,32 @@ var ActiveInstanceMap = make(map[string]string)
...
 	message := ""
 	if err != nil {
 		message = err.Error()
+		argocdStatus.Phase = "Failed" // Any error should reset phase back to Failed
 	}
 
-	if error := updateStatusConditionOfArgoCD(ctx, createCondition(message), argocd, r.Client, log); error != nil {
-		log.Error(error, "unable to update status of ArgoCD")
-		return reconcile.Result{}, error
+	log.Info("reconciling status")
```
jgwest, line 135: New logic is:  we call `reconcileStatus` to calculate what the `.status` field contents should be, and then set it once.
- If `.status` was already modified by `internalReconcile` call above, then we don't override that value

controllers/argocd/dex.go
```diff
@@ -556,78 +555,24 @@ func (r *ReconcileArgoCD) reconcileDexResources(cr *argoproj.ArgoCD) error {
...
-
-	return nil
-}
-
-// The code to create/delete notifications resources is written within the reconciliation logic itself. However, these functions must be called
-// in the right order depending on whether resources are getting created or deleted. During creation we must create the role and sa first.
-// RoleBinding and deployment are dependent on these resouces. During deletion the order is reversed.
-// Deployment and RoleBinding must be deleted before the role and sa. deleteDexResources will only be called during
-// delete events, so we don't need to worry about duplicate, recurring reconciliation calls
-func (r *ReconcileArgoCD) deleteDexResources(cr *argoproj.ArgoCD) error {
```
jgwest, line 586: `deleteDexResources` is no longer needed: it is handled by existing logic, elsewhere (and I added unit test to verify this)

controllers/argocd/repo_server.go
```diff
@@ -692,7 +693,7 @@ func (r *ReconcileArgoCD) reconcileStatusRepo(cr *v1beta1.ArgoCD) error {
 				status = "Running"
 			} else if deploy.Status.Conditions != nil {
 				for _, condition := range deploy.Status.Conditions {
-					if condition.Type == v2.DeploymentReplicaFailure && condition.Status == v1.ConditionTrue {
+					if condition.Type == appsv1.DeploymentReplicaFailure && condition.Status == corev1.ConditionTrue {
```
jgwest, line 696: `v2` is a dupe of `appsv1`, I fixed it here

controllers/argocd/sso.go
```diff
@@ -28,125 +28,94 @@ const (
 	illegalSSOConfiguration string = "illegal SSO configuration: "
 )
 
-var (
```
jgwest, line 31: No longer required; this is handled by `argocdStatus` object which is passed into `reconcileSSO`

controllers/argocd/sso.go
```diff
@@ -28,125 +28,94 @@ const (
...
+func (r *ReconcileArgoCD) reconcileSSO(cr *argoproj.ArgoCD, argocdStatus *argoproj.ArgoCDStatus) error {
 
 	// case 1
 	if cr.Spec.SSO == nil {
 		// no SSO configured, nothing to do here
 		return nil
 	}
 
-	if cr.Spec.SSO != nil {
+	// After case 1, cr.Spec.SSO is necessarily non-nil
```
jgwest, line 44: The logic for this function in this PR is unchanged: 
- I removed an unnecessary `if cr.Spec.SSO != nil { }` if statement, which has caused GitHub to no longer diff the two blocks.
- Thus it looks different, but it's the same.

controllers/argocd/sso.go
```diff
@@ -28,125 +28,94 @@ const (
...
 		// Keycloak functionality has been removed, skipping reconciliation
 	} else if UseDex(cr) {
 		// dex
 		if err := r.reconcileDexResources(cr); err != nil {
+			argocdStatus.SSO = "Failed"
 			return err
 		}
 	}
 
-	_ = r.reconcileStatusSSO(cr)
```
jgwest, line 130: Intentionally removed: this is handled by a final call to reconcileStatusSSO at the end of the process.

controllers/argocd/sso.go
```diff
@@ -28,125 +28,94 @@ const (
...
-
-	return nil
-}
-
-func (r *ReconcileArgoCD) deleteSSOConfiguration(newCr *argoproj.ArgoCD, oldCr *argoproj.ArgoCD) error {
```
jgwest, line 135: No longer required as keycloak has been removed, and deleting dex resources is handled elsewhere. I added a new unit test to verify this.

controllers/argocd/status.go
```diff
@@ -124,101 +152,98 @@ func (r *ReconcileArgoCD) reconcileStatusDex(cr *argoproj.ArgoCD) error {
...
-func (r *ReconcileArgoCD) reconcileStatusPhase(cr *argoproj.ArgoCD) error {
+func (r *ReconcileArgoCD) reconcileStatusPhase(cr *argoproj.ArgoCD, argocdStatus *argoproj.ArgoCDStatus) error {
 	var phase string
 
-	if ((!cr.Spec.Controller.IsEnabled() && cr.Status.ApplicationController == "Unknown") || cr.Status.ApplicationController == "Running") &&
-		((!cr.Spec.Redis.IsEnabled() && cr.Status.Redis == "Unknown") || cr.Status.Redis == "Running" || (cr.Spec.Redis.IsEnabled() && cr.Spec.Redis.Remote != nil && *cr.Spec.Redis.Remote != "")) &&
-		((!cr.Spec.Repo.IsEnabled() && cr.Status.Repo == "Unknown") || cr.Status.Repo == "Running") &&
-		((!cr.Spec.Server.IsEnabled() && cr.Status.Server == "Unknown") || cr.Status.Server == "Running") &&
-		((!cr.Spec.SSO.IsEnabled() && cr.Status.SSO == "Unknown") || cr.Status.SSO == "Running") {
+	appControllerAvailable := (!cr.Spec.Controller.IsEnabled() && argocdStatus.ApplicationController == "Unknown") || argocdStatus.ApplicationController == "Running"
```
jgwest, line 213: The logic in this function is the same: I just moved the logic blocks out of the uber `if` statement and into separate variables, for readability.

controllers/argocd/util.go
```diff
@@ -926,38 +921,6 @@ func removeString(slice []string, s string) []string {
 // setResourceWatches will register Watches for each of the supported Resources.
 func (r *ReconcileArgoCD) setResourceWatches(bldr *builder.Builder, clusterResourceMapper, tlsSecretMapper, namespaceResourceMapper, clusterSecretResourceMapper, applicationSetGitlabSCMTLSConfigMapMapper, nmMapper handler.MapFunc) *builder.Builder {
 
-	deleteSSOPred := predicate.Funcs{
-		UpdateFunc: func(e event.UpdateEvent) bool {
```
jgwest, line 930: No longer required, this is handled by existing logic in `Reconcile` function. I likewise added a unit test to verify the existing logic works as expected.

jgwest:
Thanks for reviewing @chetan-rns!

Issue #1830, opened 2025-08-14
Populate repo-server container with ClusterTrustBundle CA certificates


olivergondza:
Permit Argo CD administrators to declare additional CA trust anchors for repo-server.

Repository server and its sidecars issues a TLS connection to external hosts where Argo's certificate pinning does not work very well.

**Is your feature request related to a problem? Please describe.**

Use Cases:

1. As an Argo CD administrator, I would like to have a possibility to accept Trust Anchors from the cluster in the same way as for other workloads (instead of re-declaring them in Argo's `argocd-tls-certs-cm`).
2. As an Argo CD administrator, I would like not to worry about explicitly managing trust to well known hosts (raw.githubusercontent.com, etc.) while adding custom CAs.
3. As an Argo CD administrator, I would like my repo-server (and its plugin sidecars) trust the hosts on OS level. This is important when my executed commands reach out to further TLS hosts. Such as kustomize, fetching files over TLS, etc.

**Describe the solution you'd like**

1. Introduce a new element in the `ArgoCD` configuration:

```yaml
apiVersion: argoproj.io/v1beta1
kind: ArgoCD
spec:
  repo:
    trustAnchors:
      clusterTrustBundles:
        - "name-of-the-ClusterTrustBundles"
```

2. Have the operator decorate the repo-server pod with needed `initContainer`s and `volumeMount`s (see *Additional context*) to update the system CA trust.
3. Have the operator decorate the repo-server pod sidecar container with the needed `volumeMount`s.

**Describe alternatives you've considered**

- Certificate pinning through `argocd-tls-certs-cm`
  - Does not satisfy UC1 and UC3)
- Mounting the pre-build CA bundle, with well known CA + the custom ones.
  - This delegates a fair deal of the complexity to the user side
  - Implementation is not significantly simpler on the Argo CD side, albeit it will not require new config option
- Not using TLS
  - :-/

**Additional context**

The described can currently be achieved through:

(Though it is long, ugly, and Ubuntu specific. It will also not receive any fixes if users start copy-pasting this)

```yaml
apiVersion: argoproj.io/v1beta1
kind: ArgoCD
metadata:
  name: example-argocd
  labels:
    example: basic
spec:
  repo:
    volumeMounts:
      - name: ca-trust-target
        mountPath: "/etc/ssl/certs/"
      - name: ca-trust-source
        mountPath: "/usr/local/share/ca-certificates/"
    volumes:
      - name: ca-trust-source
        projected:
[...]

PR #1876, opened 2025-09-18
feat(repo-server): Declare custom trust certs for repo-server and plugins

olivergondza:
**What type of PR is this?**

/kind enhancement

Permit users to trust CAs on a repo-server system level

Screenshot From 2025-09-23 13-17-05

Fixes https://github.com/argoproj-labs/argocd-operator/issues/1830

Can be tested against:
```
k3d cluster create localtest \
              --k3s-arg "--kube-apiserver-arg=feature-gates=ClusterTrustBundle=true,ClusterTrustBundleProjection=true@server:*" \
              --k3s-arg "--kube-apiserver-arg=runtime-config=certificates.k8s.io/v1beta1/clustertrustbundles=true@server:*" \
              --k3s-arg "--kubelet-arg=feature-gates=ClusterTrustBundle=true,ClusterTrustBundleProjection=true@agent:*" \
              --image rancher/k3s:v1.33.0-k3s1
```

## Summary by CodeRabbit

* **New Features**
  * System CA trust injection for the repo server and plugins via ClusterTrustBundles, Secrets, and ConfigMaps; supports mounting, optional dropping of image certificates, init-container CA installation, and automatic rollouts when trust changes.

* **Documentation**
  * New reference pages and YAML examples for SystemCATrust configuration and merging behavior.

* **Tests**
  * Large E2E suite and new test helpers validating trust projection, updates, and reconciliations.

* **Chores**
  * CI matrix expanded for additional k3s versions; operator RBAC and permissions updated to support ClusterTrustBundle access.

### DIFF
--- a/controllers/argocd/repo_server.go
+++ b/controllers/argocd/repo_server.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
+	"hash"
 	"reflect"
 	"strings"
 	"time"
@@ -28,6 +29,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/utils/ptr"
@@ -594,7 +597,16 @@ func (r *ReconcileArgoCD) injectCATrustToContainers(cr *argocdoperatorv1beta1.Ar
 		return
 	}
 
-	sources, sourceNames := r.caTrustVolumes(cr)
+	sources, sourceNames, sourcesChecksum := r.caTrustVolumes(cr)
+
+	// Update pod checksum so eventual update to sources causes deployment rollout
+	if deploy.Spec.Template.Annotations == nil {
+		deploy.Spec.Template.Annotations = make(map[string]string)
+	}
+	if old, ok := deploy.Spec.Template.Annotations[common.ArgoCDCATrustChecksum]; ok && old != sourcesChecksum {
+		log.Info("repo-server " + common.ArgoCDCATrustChecksum + " changed")
+	}
+	deploy.Spec.Template.Annotations[common.ArgoCDCATrustChecksum] = sourcesChecksum
 
 	volumeSource := "argocd-ca-trust-source"
 	volumeTarget := "argocd-ca-trust-target"
@@ -625,10 +637,10 @@ func (r *ReconcileArgoCD) injectCATrustToContainers(cr *argocdoperatorv1beta1.Ar
 		}
 	}
 
-	argoImage := getArgoContainerImage(cr)
+	repoServerImage := getRepoServerContainerImage(cr)
 	deploy.Spec.Template.Spec.InitContainers = append(
 		deploy.Spec.Template.Spec.InitContainers,
-		caTrustInitContainer(cr, argoImage, volumeSource, volumeTarget, volumeTmp),
+		caTrustInitContainer(cr, repoServerImage, volumeSource, volumeTarget, volumeTmp),
 	)
 
 	prodVolumeMounts := func() []corev1.VolumeMount {
@@ -642,7 +654,7 @@ func (r *ReconcileArgoCD) injectCATrustToContainers(cr *argocdoperatorv1beta1.Ar
 	var containerNames []string
 	for i, container := range deploy.Spec.Template.Spec.Containers {
 		// This can only work with ubuntu or compatible, so do not inject to potentially incompatible containers
-		if container.Image == argoImage {
+		if container.Image == repoServerImage {
 			// Accessing by index because the container is a copy of the original struct
 			deploy.Spec.Template.Spec.Containers[i].VolumeMounts = append(deploy.Spec.Template.Spec.Containers[i].VolumeMounts, prodVolumeMounts()...)
 			containerNames = append(containerNames, container.Name)
@@ -658,7 +670,7 @@ func (r *ReconcileArgoCD) injectCATrustToContainers(cr *argocdoperatorv1beta1.Ar
 	deploy.Spec.Template.Spec.Volumes = append(deploy.Spec.Template.Spec.Volumes, repoServerVolumes...)
 }
 
-func (r *ReconcileArgoCD) caTrustVolumes(cr *argocdoperatorv1beta1.ArgoCD) (sources []corev1.VolumeProjection, sourceNames []string) {
+func (r *ReconcileArgoCD) caTrustVolumes(cr *argocdoperatorv1beta1.ArgoCD) (sources []corev1.VolumeProjection, sourceNames []string, checksum string) {
 	// The projected file needs to have the `.crt` suffix for the update-ca-certificates to work correctly. Add it if not present.
 	ensureValidPath := func(path string) string {
 		if strings.HasSuffix(path, ".crt") {
@@ -675,33 +687,68 @@ func (r *ReconcileArgoCD) caTrustVolumes(cr *argocdoperatorv1beta1.ArgoCD) (sour
 		sourceNames = append(sourceNames, path)
 	}
 
-	for _, bundle := range cr.Spec.Repo.SystemCATrust.ClusterTrustBundles {
-		bundle = *bundle.DeepCopy()
+	// Create checksum from the source projection spec and the source contents.
+	// This is to detect that either of them has changed and rollout is needed.
+	objectChecksum := newObjectChecksum()
+
+	// Relevant bundles cannot be queried independently and need to be looked up through `isRelevantCtb`, so pulled as a list ahead of time
+	bundles := certificatesv1beta1.ClusterTrustBundleList{}
+	if len(cr.Spec.Repo.SystemCATrust.ClusterTrustBundles) > 0 {
+		err := r.List(context.TODO(), &bundles)
+		if err != nil {
+			log.Error(err, "failed querying ClusterTrustBundles for System CA Trust checksum")
+		}
+	}
+
+	for i, bundleProjection := range cr.Spec.Repo.SystemCATrust.ClusterTrustBundles {
+		bundleProjection = *bundleProjection.DeepCopy()
 		// Using .Path, because .Name might not be specified
-		trackSource("ClusterTrustBundle", bundle.Path, bundle.Optional)
+		trackSource("ClusterTrustBundle", bundleProjection.Path, bundleProjection.Optional)
 
-		bundle.Path = ensureValidPath(bundle.Path)
-		sources = append(sources, corev1.VolumeProjection{ClusterTrustBundle: &bundle})
-	}
-	for _, secret := range cr.Spec.Repo.SystemCATrust.Secrets {
-		secret = *secret.DeepCopy()
-		trackSource("Secret", secret.Name, secret.Optional)
+		bundleProjection.Path = ensureValidPath(bundleProjection.Path)
+		sources = append(sources, corev1.VolumeProjection{ClusterTrustBundle: &bundleProjection})
 
-		for i, item := range secret.Items {
-			secret.Items[i].Path = ensureValidPath(item.Path)
+		objectChecksum.sprintf("ctb:%d:%s", i, bundleProjection.Path)
+		for _, bundle := range findRelevantBundles(bundles, bundleProjection) {
+			objectChecksum.writeObject(&bundle)
 		}
-		sources = append(sources, corev1.VolumeProjection{Secret: &secret})
 	}
-	for _, cm := range cr.Spec.Repo.SystemCATrust.ConfigMaps {
-		cm = *cm.DeepCopy()
-		trackSource("ConfigMap", cm.Name, cm.Optional)
+	for i, secretProjection := range cr.Spec.Repo.SystemCATrust.Secrets {
+		secretProjection = *secretProjection.DeepCopy()
+		trackSource("Secret", secretProjection.Name, secretProjection.Optional)
 
-		for i, cmi := range cm.Items {
-			cm.Items[i].Path = ensureValidPath(cmi.Path)
+		for i, item := range secretProjection.Items {
+			secretProjection.Items[i].Path = ensureValidPath(item.Path)
 		}
-		sources = append(sources, corev1.VolumeProjection{ConfigMap: &cm})
+		sources = append(sources, corev1.VolumeProjection{Secret: &secretProjection})
+
+		objectChecksum.sprintf("secret:%d:%s", i, secretProjection.Name)
+		var secret corev1.Secret
+		err := r.Get(context.TODO(), types.NamespacedName{Namespace: cr.Namespace, Name: secretProjection.Name}, &secret)
+		if err != nil && !errors.IsNotFound(err) {
+			log.Error(err, "failed querying Secret for System CA Trust checksum", "name", secretProjection.Name, "namespace", cr.Namespace)
+		}
+		objectChecksum.writeObject(&secret)
 	}
-	return sources, sourceNames
+	for i, cmProjection := range cr.Spec.Repo.SystemCATrust.ConfigMaps {
+		cmProjection = *cmProjection.DeepCopy()
+		trackSource("ConfigMap", cmProjection.Name, cmProjection.Optional)
+
+		for i, cmi := range cmProjection.Items {
+			cmProjection.Items[i].Path = ensureValidPath(cmi.Path)
+		}
+		sources = append(sources, corev1.VolumeProjection{ConfigMap: &cmProjection})
+
+		objectChecksum.sprintf("cm:%d:%s", i, cmProjection.Name)
+		var cm corev1.ConfigMap
+		err := r.Get(context.TODO(), types.NamespacedName{Namespace: cr.Namespace, Name: cmProjection.Name}, &cm)
+		if err != nil && !errors.IsNotFound(err) {
+			log.Error(err, "failed querying ConfigMap for System CA Trust checksum", "name", cmProjection.Name, "namespace", cr.Namespace)
+		}
+		objectChecksum.writeObject(&cm)
+	}
+
+	return sources, sourceNames, objectChecksum.hexSum()
 }
 
 func caTrustInitContainer(cr *argocdoperatorv1beta1.ArgoCD, argoImage string, volumeSource string, volumeTarget string, tmpVolume string) corev1.Container {
@@ -733,6 +780,7 @@ func caTrustInitContainer(cr *argocdoperatorv1beta1.ArgoCD, argoImage string, vo
                 update-ca-certificates --verbose --certsdir "$IMAGE_CERT_PATH"
                 echo "Resulting /etc/ssl/certs/"
                 ls -l /etc/ssl/certs/
+
                 echo "Done!"
         `},
 		VolumeMounts: []corev1.VolumeMount{
@@ -956,16 +1004,13 @@ func (r *ReconcileArgoCD) reconcileRepoServerTLSSecret(cr *argocdoperatorv1beta1
 }
 
 // systemCATrustMapper triggers reconciliation of repo-server Deployment if some of the tracked Secrets, ConfigMaps or ClusterTrustBundles have changed
-func (r *ReconcileArgoCD) systemCATrustMapper(ctx context.Context, o client.Object) []reconcile.Request {
-	// Track Argo CDs whose repo-servers need a rollout, and id of the resource that changed
-	rolloutBecause := make(map[*argocdoperatorv1beta1.ArgoCD]string)
-
+func (r *ReconcileArgoCD) systemCATrustMapper(ctx context.Context, o client.Object) (toReconcile []reconcile.Request) {
 	// For cluster-wide resources, it is needed to consult all argos. For cluster-scoped ones, only the argos in the same NS.
 	argoNamespace := client.InNamespace(o.GetNamespace())
 	var argoCDs argocdoperatorv1beta1.ArgoCDList
 	if err := r.List(ctx, &argoCDs, argoNamespace); err != nil {
 		log.Error(err, "unable to list ArgoCD instances")
-		return []reconcile.Request{}
+		return
 	}
 
 	for _, argocd := range argoCDs.Items {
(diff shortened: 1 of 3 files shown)


olivergondza:
Compared to the proposal in https://github.com/argoproj-labs/argocd-operator/pull/1876, it turned out 1 init container is enough. Also, this implements `DropImageAnchors` to suppress whatever CAs was in the image originally.

olivergondza:
The "[Code scans / Run golangci-lint and gosec (pull_request)](https://github.com/argoproj-labs/argocd-operator/actions/runs/17917840135/job/50944661667?pr=1876)" failure to be adressed by https://github.com/argoproj-labs/argocd-operator/pull/1880

olivergondza:
The test failures are related to the fact the code depends on a tech-preview features. Any advise on how to handle such functionality?

https://github.com/argoproj-labs/argocd-operator/actions/runs/17917840047/job/50944661583?pr=1876#step:11:45

api/v1beta1/argocd_types.go
```diff
@@ -586,6 +589,18 @@ func (a *ArgoCDRepoSpec) IsRemote() bool {
...
 }
 
+// ArgoCDSystemCATrustSpec defines custom certificates to inject into the repo server container and its plugins to trust source hosting sites
+type ArgoCDSystemCATrustSpec struct {
+	// DropImageCertificates will remove all certs that are present in the image, leaving only those explicitly configured here.
+	DropImageCertificates bool `json:"dropImageCertificates,omitempty"`
+	// ClusterTrustBundles is a list of projected ClusterTrustBundle volume definitions from where to take the trust certs.
+	ClusterTrustBundles []corev1.ClusterTrustBundleProjection `json:"clusterTrustBundles,omitempty"`
+	// Secrets is a list of projected Secret volume definitions from where to take the trust certs.
+	Secrets []corev1.SecretProjection `json:"secrets,omitempty"`
```
anandf, line 614: Can it be `SecretRefs` instead of `SecretProjection` ? What is the advantage we get by using `SecretProjection` instead of referring a `Secret` directly which can be read directly by the operator code.

olivergondza: Good question, projections are the only way (I know off), to create a single volume from multiple sources. So this can combine all three kinds of sources in any quantity, and the entries will be merged in a single directory making the init container completely agnostic of files' origin.

If customers are to create the Secret or CM manually, using only one resource might not be that much to ask. But ability to merge them permits seemless integration with signed `ClusterTrustBundles` that are likely to have more than one resource.

.github/workflows/ci-build.yaml
```diff
@@ -48,7 +48,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        k3s-version: [ v1.27.1 ]
+        k3s-version: [ v1.33.0-k3s1 ]
```
olivergondza: Version bump is needed to enable `ClusterTrustBundle` (maybe even slightly older would work as well, <1.30 would not). I am not sure what is the support requirement here.

If desired, we can keep 1.27 and start it without the feature gate args below.

svghadi: From a downstream perspective, we need to continue supporting Kubernetes 1.27.x (OCP 4.14). I think we can set up a separate CI job with the required Kubernetes version.

olivergondza: Ah, ok. Let me bring it back through the matrix...

olivergondza: Done, they are selectively skipped.

olivergondza:
@jannfis, after your review, this is the added feature of change detection: https://github.com/argoproj-labs/argocd-operator/pull/1876/commits/9c3639cbd5eb2e103da5cbc9ebaf3f610843b846

controllers/argocd/repo_server.go
```diff
@@ -559,6 +573,193 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argocdoperatorv1beta1.Argo
...
+	// This is where the image keeps its vendored CAs, look elsewhere if DropImageCertificates
+	imageCertPath := "/usr/share/ca-certificates"
+	if cr.Spec.Repo.SystemCATrust.DropImageCertificates {
+		imageCertPath = "/systemCATrust.dropImageCertificates"
+	}
+
+	return corev1.Container{
+		Name:            "update-ca-certificates",
+		Image:           argoImage,
+		ImagePullPolicy: corev1.PullAlways,
```
anandf: Instead of hardcoding , Can be set to `argoutil.GetImagePullPolicy(cr.Spec.ImagePullPolicy)`

controllers/argocd/repo_server.go
```diff
@@ -559,6 +573,193 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argocdoperatorv1beta1.Argo
...
+		Resources: getArgoRepoResources(cr),
+		SecurityContext: &corev1.SecurityContext{
+			AllowPrivilegeEscalation: boolPtr(false),
+			Capabilities: &corev1.Capabilities{
+				Drop: []corev1.Capability{
+					"ALL",
+				},
+			},
+			// Needed by update-ca-certificates for /tmp/
+			ReadOnlyRootFilesystem: boolPtr(false),
```
anandf: Any specific reason why `ReadOnlyRootFilesystem` is set to `false`. This has security implications and must be avoided. Since we are mounting the required volumes, root filesystem can be readonly IMO. Let me know if you think otherwise.

controllers/argocd/repo_server.go
```diff
@@ -559,6 +573,193 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argocdoperatorv1beta1.Argo
...
+				Name: volumeSource,
+				// Source path for user additional certificates - empty in the image, so not shadowing anything.
+				MountPath: "/usr/local/share/ca-certificates/",
+				ReadOnly:  true,
+			}, {
+				Name:      volumeTarget,
+				MountPath: "/etc/ssl/certs/",
+			},
+		},
+		Resources: getArgoRepoResources(cr),
```
anandf: Repo server resources are generally quite high. For init container, which is just updating the certs, this can be reduced IMO.

controllers/argocd/repo_server.go
anandf: IIRC, We had issues in the past using `Always` pull policy in air gapped environments.

controllers/argocd/repo_server.go
```diff
@@ -559,6 +573,193 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argocdoperatorv1beta1.Argo
...
+		Env: []corev1.EnvVar{
+			{
+				Name:  "IMAGE_CERT_PATH",
+				Value: imageCertPath,
+			},
+		},
+		Command: []string{"/bin/bash", "-c"},
+		Args: []string{`
+                #!/usr/bin/env bash
+                # https://github.com/olivergondza/bash-strict-mode
```
anandf: Is this comment required ?

controllers/argocd/repo_server.go
olivergondza: Definitely. Though I see `copyutil` init container is using the same.

I am thinking if this is not actually desirable to occupy the resources early in the container start because a) the prod container is going to occupy the resources anyway, and b) init container are executed serially.

controllers/argocd/repo_server.go
olivergondza: Good call.

controllers/argocd/repo_server.go
olivergondza: Technically, no. It points back to the source to make sure different scripts do not use different strict modes. :shrug:

controllers/argocd/repo_server.go
olivergondza: From the comment, it is because the `update-ca-certificates` writes to `/tmp`. I need to retest if it is still the case after all the changes, but we can mount an empty one over `/tmp` if we believe this is an issue for init container.

olivergondza:
@anandf, thanks for the review. I have addressed all expect for the resources, as the discussion there is ongoing...

svghadi:
@coderabbitai full review

tests/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
```diff
@@ -0,0 +1,852 @@
...
+		case strings.Contains(line, "BEGIN CERTIFICATE"):
+			currentBlock.Reset()
+		case strings.Contains(line, "END CERTIFICATE"):
+			seen[currentBlock.String()] = true
+		default:
+			currentBlock.WriteString(line)
+		}
+	}
+	return len(seen)
+}
```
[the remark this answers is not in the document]
olivergondza: Ehm https://cs.opensource.google/go/go/+/refs/tags/go1.24.6:src/strings/iter.go;l=18

controllers/argocd/repo_server.go
```diff
@@ -759,3 +946,103 @@ func (r *ReconcileArgoCD) reconcileRepoServerTLSSecret(cr *argocdoperatorv1beta1
...
+					break
+				}
+			}
+		case *v1beta1.ClusterTrustBundle:
+			for _, trustSource := range argocd.Spec.Repo.SystemCATrust.ClusterTrustBundles {
+				if isRelevantCtb(trustSource, obj) {
+					rolloutBecause[&argocd] = fmt.Sprintf("ClusterTrustBundle %s", obj.Name)
+					break
+				}
+			}
```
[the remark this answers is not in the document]
olivergondza: Interesting observation. It would be a bug with 1.21 or earlier. No point in complicating the code unless we want to get back to the old go versions.

controllers/argocd/repo_server.go
```diff
@@ -559,6 +573,179 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argocdoperatorv1beta1.Argo
...
+				MountPath: "/usr/local/share/ca-certificates/",
+				ReadOnly:  true,
+			}, {
+				Name:      volumeTarget,
+				MountPath: "/etc/ssl/certs/",
+			}, {
+				Name:      "tmp",
+				MountPath: "/tmp",
+			},
+		},
```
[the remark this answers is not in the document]
olivergondza: As per `reconcileRepoDeployment`, the `/tmp` is always there. Either user-provided, or the operator default one.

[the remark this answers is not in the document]
olivergondza: @anandf, Rabbit here is correct ☝️. I did depend on existence of volume named `tmp` - that is not guaranteed. We can either:

- Get back to writable `/`
- Reuse pod temp volume, no matter what its name is.
- Use dedicated empty volume as `/tmp` for this init container

Any preference?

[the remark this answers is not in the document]
anandf: Getting back to writable root `/`must be avoided if alternatives exists as it is not security best practice. Can we check if the pod already has a tmp volume and add it only if it is not there ? Anyways all the containers in that pod shares the same volume provided by the pod, no need to add a seperate volume just for the cert handling.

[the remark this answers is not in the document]
svghadi: Some documentation around /tmp volume mount in repo server: https://argocd-operator.readthedocs.io/en/latest/usage/ha/repo-server/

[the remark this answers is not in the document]
olivergondza: Yeah, the volume for `/tmp` always exists, it is just it might have a name chosen by the user. If it is ok to share the volume across container, I will take a look at option 2.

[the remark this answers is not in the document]
olivergondza: Implemented

olivergondza:
@anandf, what would you say are the next steps here?

controllers/argocd/util.go
```diff
@@ -688,6 +704,9 @@ func InspectCluster() error {
 	if err := verifyVersionAPI(); err != nil {
 		return err
 	}
+	if err := verifyClusterTrustBundleAPI(); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
olivergondza: Ignore. The `verifyClusterTrustBundleAPI` not when the api is not present, but when we fail checking.

controllers/argocd/repo_server.go
```diff
@@ -759,3 +954,103 @@ func (r *ReconcileArgoCD) reconcileRepoServerTLSSecret(cr *argocdoperatorv1beta1
...
+		case *v1beta1.ClusterTrustBundle:
+			for _, trustSource := range argocd.Spec.Repo.SystemCATrust.ClusterTrustBundles {
+				if isRelevantCtb(trustSource, obj) {
+					rolloutBecause[&argocd] = fmt.Sprintf("ClusterTrustBundle %s", obj.Name)
+					break
+				}
+			}
+		default:
+			panic(fmt.Errorf("systemCATrustMapper called for unknown type %t", o))
+		}
```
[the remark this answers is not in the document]
olivergondza: No, this code should be unreachable unless there is a bug in the code. Log line can be overlooked, panic can not.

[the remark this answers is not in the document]
olivergondza: Spot on with the `T`. Fixed.

controllers/argocd/repo_server.go
olivergondza: This is now using the pod's temp partition.

olivergondza:
@anandf, @svghadi, can you please take a look at reviewing/integrating this? Thanks

controllers/argocd/repo_server.go
```diff
@@ -559,6 +573,187 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argocdoperatorv1beta1.Argo
...
+                echo "User defined CA files:"
+                ls -l /usr/local/share/ca-certificates/
+
+                # Make sure the file exist even when the update-ca-certificates produces no pem blocks
+                echo "" > /etc/ssl/certs/ca-certificates.crt
+
+                update-ca-certificates --verbose --certsdir "$IMAGE_CERT_PATH"
+                echo "Resulting /etc/ssl/certs/"
+                ls -l /etc/ssl/certs/
+                echo "Done!"
```
svghadi, line 784: Is this Debian-specific, or will it work on RHEL-based systems as well?

olivergondza: Correct, it is going to work only in debian-style distros, such as Ubuntu here. There are 2 precautions applied:

- It is only applied to repo-server sidecars, IFF they use the same image (https://github.com/argoproj-labs/argocd-operator/pull/1876/changes#diff-f3e1a767efe24a5843567747fe06dc944b52299accf98b4210cb86aa2402ee51R645). This is to prevent breakage if incompatible image is used for user provided plugin.
- in gitops-operator, this is replaced with RHEL/Fedora/UBI specific logic.

tests/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
```diff
@@ -0,0 +1,860 @@
...
+			Expect(k8sClient.Create(ctx, argoCD)).To(Succeed())
+			Eventually(argoCD, "5m", "5s").Should(argocdFixture.BeAvailable())
+
+			Expect(repoServerSystemCaTrust(ns)).Should(trustCerts(Equal(0), Not(BeEmpty())))
+
+			trustedHelmApp := createHelmApp(ns, trustedHelmAppSource)
+			Expect(k8sClient.Create(ctx, trustedHelmApp)).To(Succeed())
+
+			// Sleep to make sure the apps sync took place - otherwise there might be no conditions _yet_
+			time.Sleep(20 * time.Second)
```
jgwest: Avoid using `time.Sleep()` in E2E tests when an `Eventually` can be used instead (which is true in nearly all cases)

tests/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
```diff
@@ -0,0 +1,860 @@
...
+	}, logMatcher)
+
+	return And(matchCount, matchLog)
+}
+
+func getTrustedCertCount(rsPod *corev1.Pod) int {
+	var out string
+	var err error
+	// retry a few times, because pod can be restarting during trust source update, get Terminating between check and use.
+	for i := 0; i < 3; i++ {
```
jgwest: Can use Eventually here, rather than time.Sleep in a for loop

tests/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
```diff
@@ -0,0 +1,860 @@
...
+			if pod.Status.Phase == "Running" && nameRegexp.MatchString(pod.Name) {
+				pods = append(pods, &pod)
+			}
+		}
+		if len(pods) == 1 {
+			return pods[0]
+		}
+	}
+
+	panic(fmt.Sprintf("Failed to find Running repo-server pod. have %+v", pods))
```
jgwest: Can use Eventually here, instead of sleep in a for loop. `panic` equivalent for Ginkgo is Fail()

tests/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
```diff
@@ -0,0 +1,860 @@
...
+	Expect(k8sClient.Create(ctx, untrustedPluginApp)).To(Succeed())
+
+	trustedHelmApp := createHelmApp(ns, trustedHelmAppSource)
+	Expect(k8sClient.Create(ctx, trustedHelmApp)).To(Succeed())
+
+	trustedPluginApp := createPluginApp(ns, "https://github.com/argoproj-labs/argocd-operator.git")
+	Expect(k8sClient.Create(ctx, trustedPluginApp)).To(Succeed())
+
+	// Sleep to make sure the apps sync took place - otherwise there might be no conditions _yet_
+	time.Sleep(20 * time.Second)
```
jgwest: Same commment as above re: use of time.Sleep in E2E tests

olivergondza:
@jgwest, it took a bit, but the tests are free of sleeps now.

tests/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
```diff
@@ -0,0 +1,899 @@
...
+	"github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture"
+	argocdFixture "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/argocd"
+	fixtureUtils "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/utils"
+)
+
+var (
+	// The differences between the upstream image using Ubuntu, and the downstream one using rhel.
+	image        = "" // argocd-operator default
+	version      = "" // argocd-operator default
+	caBundlePath = "/etc/ssl/certs/ca-certificates.crt"
```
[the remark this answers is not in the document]
olivergondza: Disregard.

> in gitops-operator, this is replaced with RHEL/Fedora/UBI specific logic.

controllers/argocd/util.go
```diff
@@ -36,6 +36,7 @@ import (
 	"github.com/argoproj/argo-cd/v3/util/glob"
 	"github.com/distribution/reference"
 	"github.com/go-logr/logr"
+	certificates "k8s.io/api/certificates/v1beta1"
```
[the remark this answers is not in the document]
olivergondza: Not true, they are in beta as well.

olivergondza:
@jgwest, do you see anything else needed to proceed with this?

controllers/argocd/repo_server.go
```diff
@@ -559,6 +573,187 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argocdoperatorv1beta1.Argo
...
+					DefaultMode: ptr.To(int32(0o444)),
+				},
+			},
+		}, {
+			Name: volumeTarget,
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
olivergondza: This is verified in scenario "ensures that empty trust keeps image certs in place". It does work as expected

tests/ginkgo/fixture/pod/fixture.go
```diff
@@ -13,6 +14,22 @@ import (
...
+	Expect(err).ToNot(HaveOccurred())
+	for _, pod := range list.Items {
+		if nameRegexp.MatchString(pod.Name) {
+			pods = append(pods, pod)
+		}
+	}
+
+	Expect(pods).Should(HaveLen(1), "expected a single pod matching "+nameRegexp.String())
+
+	return &pods[0]
```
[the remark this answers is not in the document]
olivergondza: Ignore

AdamSaleh:
As far as tests go, it looks throroug! LGTM

controllers/argocd/repo_server.go
```diff
@@ -759,3 +954,103 @@ func (r *ReconcileArgoCD) reconcileRepoServerTLSSecret(cr *argocdoperatorv1beta1
...
+			}
+		case *certificatesv1beta1.ClusterTrustBundle:
+			for _, trustSource := range argocd.Spec.Repo.SystemCATrust.ClusterTrustBundles {
+				if isRelevantCtb(trustSource, obj) {
+					rolloutBecause[&argocd] = fmt.Sprintf("ClusterTrustBundle %s", obj.Name)
+					break
+				}
+			}
+		default:
+			panic(fmt.Errorf("systemCATrustMapper called for unknown type %T", o))
```
svghadi, line 1059: I don’t think we should panic here. Panicking may abruptly stop the manager controller and trigger a restart. I believe we can safely ignore this request, since it won’t be for an object we care about.

controllers/argocd/repo_server.go
```diff
@@ -759,3 +954,103 @@ func (r *ReconcileArgoCD) reconcileRepoServerTLSSecret(cr *argocdoperatorv1beta1
...
+				"restarting repo-server pod after SystemCATrust change in "+cause,
+				"pod", pod.Name, "ns", pod.Namespace, "phase", pod.Status.Phase,
+			)
+			if err := r.Delete(context.TODO(), &pod); err != nil {
+				log.Error(err, "unable to delete repo-server pod 1", "pod", pod.Name, "ns", pod.Namespace, "phase", pod.Status.Phase)
+			}
+		}
+	}
+	// No need to reconcile. The pods have been restarted
+	return []reconcile.Request{}
```
svghadi: Instead of deleting the pod to trigger reconciliation and re-computation of the certificates, what do you think about directly requeuing the reconciliation request and letting the reconciliation loop handle the re-computation and restart?

One approach could be to compute a checksum from the certificate contents and add it as an annotation on the repo-server Deployment via `.spec.template.metadata.annotations`. This way, whenever the certificates change, the checksum changes as well, updating the deployment spec and automatically triggering a pod restart. This would keep the mapper side-effect free and avoids doing heavy work in the handler, which is generally recommended.

controllers/argocd/repo_server.go
olivergondza: I choose panic here to make sure this `switch` statement is aligned with the types watched: https://github.com/argoproj-labs/argocd-operator/pull/1876/changes#diff-30c42e628eebe205919571b78686864b3d5aaab1d82bb09bdb56a1c53bc3d81fR1112. Reaching this can only IMHO happen due to a programmer error.

I am concerned that ignoring this condition can make such mistake harder to detect.

controllers/argocd/repo_server.go
olivergondza: You are correct that letting the deployment restart the pods, is likely cleaner than deleting them here. Thinking of rolling strategy, etc. Although, I have some questions:

- Do we need to include the `Secret`/`ConfigMap` metadata into the checksum? Not sure their change can be significant.
- Where would the checksum be computed? If in the mapper, then I am concerned it might not be less work - it is `pod.List+N*pod.Delete` vs. `X*secret.Get + Y*configmap.Get + Z*ctb.Get` + data crunching.

CC @jgwest, as per our old slack discussion

svghadi: > Do we need to include the Secret/ConfigMap metadata into the checksum? Not sure their change can be significant.

If changes to referenced secrets or configMaps are expected to trigger a re-computation, then we should include them in the checksum.

> Where would the checksum be computed? 

This should live in the main repo-server deployment reconciliation logic, possibly in your existing `injectCATrustToContainers` function.
The key is to ensure we have update logic that detects changes to the deployment 's `.spec.template.metadata.annotations/labels`  (which I believe already exists). The idea is to compute the checksum on every reconciliation of the repo-server deployment and add it to the deployment object. If the checksum changes, the existing annotation update logic should pick it up and patch the live deployment. That patch will then trigger a rollout restart of the pods.

olivergondza: That is an interesting way to go. Do we do something alike in other places of the operators?

svghadi: We don’t have something exactly the same, but almost every configuration change in the operator follows the same principle, we patch the live object, which is then handled by kubernetes to perform the required restarts.

olivergondza: Cool, let me see what would it take to implement this!

olivergondza: @svghadi, Please take a look. I have implemented the checksum based approach.

controllers/argocd/repo_server.go
```diff
@@ -559,6 +576,232 @@ func (r *ReconcileArgoCD) reconcileRepoDeployment(cr *argocdoperatorv1beta1.Argo
...
+	objectChecksum := newObjectChecksum()
+
+	// Relevant bundles cannot be queried independently and need to be looked up through `isRelevantCtb`, so pulled as a list ahead of time
+	bundles := certificatesv1beta1.ClusterTrustBundleList{}
+	if len(cr.Spec.Repo.SystemCATrust.ClusterTrustBundles) > 0 {
+		err := r.List(context.TODO(), &bundles)
+		if err != nil {
+			panic(err)
+		}
+	}
```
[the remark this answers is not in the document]
olivergondza: Replaced with logging

PR #1877, opened 2025-09-19
chore: add script to update argocd dependencies and images

jgwest:
**What type of PR is this?**
/kind chore

- This script is an experimental utility that will automatically upgrade the various downstream dependencies that are referenced in the repository, including go.mod and various container images.
- I've added it as a PR check, but if we find it's breaking other PRs then we can disable it.

## Summary by CodeRabbit

* **New Features**
  * Added a make target to run automated dependency updates and a CI job to verify the resulting artifacts.

* **Chores**
  * Introduced ARGO_CD_TARGET_VERSION and made the update target public.
  * Added a dependency-update utility and internal helper tools.
  * Renamed workflow to "Validation, verifications, and lints".
  * Runtime now warns (does not fail) when no container runtime is found.
  * Bumped Argo CD target/version and updated related image and dependency references (CVE mitigations).

* **Documentation**
  * Added README describing the dependency update workflow.

### DIFF
--- a/hack/update-dependencies-script/main.go
+++ b/hack/update-dependencies-script/main.go
@@ -1,0 +1,556 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+func main() {
+
+	wd, err := os.Getwd()
+	if err != nil {
+		exitWithError(fmt.Errorf("unable to get working dir: %v", err))
+		return
+	}
+
+	argocdOperatorRoot, err := filepath.Abs(wd + "/../..")
+	if err != nil {
+		exitWithError(fmt.Errorf("unable to get absolute dir: %v", err))
+		return
+	}
+
+	// Version variable is in the form of 'v3.1.1', note the 'v' prefix
+	targetArgoCDVersion := readTargetVersionFromMakefile(argocdOperatorRoot)
+
+	fmt.Println()
+	fmt.Println("Target Argo CD version from Makefile is", targetArgoCDVersion)
+	fmt.Println()
+
+	// Clone Argo CD into a temporary directory
+	argoCDRepoRoot, err := cloneArgoCDRepoIntoTempDir(targetArgoCDVersion)
+	if err != nil {
+		exitWithError(fmt.Errorf("unable to checkout Argo CD: %v", err))
+		return
+	}
+
+	fmt.Println("Using Argo CD temporary directory:", argoCDRepoRoot)
+
+	// Sanity test that that we have the correct root path for argocd-operator repo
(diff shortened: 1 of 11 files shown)


hack/update-dependencies-script/main.go
```diff
@@ -0,0 +1,549 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+func main() {
+
+	wd, err := os.Getwd()
```
olivergondza, line 13: I see this is correct when called as `make`>`run.sh`>`main.go`, though I would like to propose passing in the repo root as a parameter. That way, we can run `go run` straight from the Makefile and drop the shell script middleman. I find it more robust because then `main.go` would not rely on being invoked in a specific way - you test the root is the real root few lines after.

tests/ginkgo/fixture/fixture.go
```diff
@@ -42,8 +43,51 @@ const (
...
+		// Split output by whitespace to parse df fields
+		fields := strings.Fields(rootEntry)
+		Expect(len(fields)).To(BeNumerically(">=", 4), "df output should have at least 4 fields")
+
+		// Parse fields[3] which is the available space in bytes
+		availableBytes, err := strconv.ParseInt(fields[3], 10, 64)
+		Expect(err).ToNot(HaveOccurred(), "failed to parse available bytes from df output")
+
+		// Convert bytes to gigabytes (1 GB = 1024^3 bytes)
+		availableGB := availableBytes / (1024 * 1024 * 1024)
```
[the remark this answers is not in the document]
olivergondza: It may not be always 1K:

> **man 1 diff**
       Display values are in units of the first available  SIZE  from
       --block-size,  and the DF_BLOCK_SIZE, BLOCK_SIZE and BLOCKSIZE
       environment variables.  Otherwise, units default to 1024 bytes
       (or 512 if POSIXLY_CORRECT is set).

Better run `df -m` to always get megabytes.

hack/update-dependencies-script/main.go
```diff
@@ -0,0 +1,556 @@
...
+// )
+//
+// This function expects only one replace block to exist, and will fail if 0 or >1 are found.
+func readReplaceBlockFromGoMod(pathToGoMod string) ([]string, int, int) {
+	lines := getFileContentsAsLines(pathToGoMod)
+
+	replaceBlockStart := -1
+	replaceBlockEnd := -1
+	for x := 0; x < len(lines); x++ {
+
```
anandf, line 343: nitpick: Wondering why there is more empty lines ? Any issue with the formatting ?

jgwest:
Thanks @anandf!

PR #1879, opened 2025-09-20
fix: Add missing services required by principal

jparsai:
This PR is to create missing resource proxy service in operator when Principal is enabled.

/kind bug

Fixes https://issues.redhat.com/browse/GITOPS-7730


jparsai:
It would be good to merge this PR after https://github.com/argoproj-labs/argocd-operator/pull/1800 to avoid conflicts in #1800

jgwest (approved):
LGTM, thanks @jparsai!

PR #1883, opened 2025-09-25
fix: Fix intermittent test fail in argocd status test (1-120)

jgwest:
**What type of PR is this?**
/kind failing-test 

On another PR, I noticed that test 1-120 was intermittently failing. It's likely because we're not waiting long enough for the status to update. 

I have updated status update logic, and added logic to output namespace state on fail.


jparsai:
LGTM

svghadi (approved):

PR #1885, opened 2025-09-29
fix: Version override in ArgoCD CR causes operator to use upstream images

akhilnittala:
**What type of PR is this?**

/kind bug

overrides the env variable if present with spec values provided by user

Documentation Needs to be updated accordingly

https://issues.redhat.com/browse/GITOPS-7789

Fixes #?

configuring Argo CD operator image via environment variable and validate the following scenarios:

No spec.image and no spec.version in the ArgoCD CR → Verify that the ArgoCD resource uses the image specified in the environment variable.

spec.version provided, but no spec.image in the ArgoCD CR → Verify that the ArgoCD resource uses the environment variable image with the user-provided tag.

Both spec.image and spec.version provided in the ArgoCD CR → Verify that the ArgoCD resource uses the user-provided image and version.

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -32,7 +32,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v6
         with:
-          go-version: '1.24'
+          go-version: '1.25'
 
       - name: Restore go build cache
         uses: actions/cache@v4
@@ -77,7 +77,7 @@ jobs:
       - name: Setup Golang
         uses: actions/setup-go@v6
         with:
-          go-version: '1.24'
+          go-version: '1.25'
       - name: GH actions workaround - Kill XSP4 process
         run: |
           sudo pkill mono || true
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v6
       with:
-        go-version: '1.24'
+        go-version: '1.25'
 
     - name: Build
       run: make build
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -22,7 +22,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v6
         with:
-          go-version: '1.24'
+          go-version: '1.25'
 
       - name: Restore go build cache
         uses: actions/cache@v4


controllers/argocd/applicationset.go
```diff
@@ -862,10 +861,18 @@ func getApplicationSetContainerImage(cr *argoproj.ArgoCD) string {
...
-	if e := os.Getenv(common.ArgoCDImageEnvName); e != "" && (defaultTag && defaultImg) {
-		return e
+	// if env variable is not set, then use the default tag common.ArgoCDDefaultArgoVersion and default img common.ArgoCDDefaultArgoImage
+	if e := os.Getenv(common.ArgoCDImageEnvName); e != "" {
+		if defaultImg && !defaultTag {
+			image, _, found := strings.Cut(e, "@")
+			if found {
+				return argoutil.CombineImageTag(image, tag)
+			}
+		} else if !defaultImg && !defaultTag {
```
anandf: This can be removed as its handled in line `877`. and instead have ```else if defaultImg && defaultTag { return e}```,  else block can be removed then.

controllers/argocd/applicationset.go
```diff
@@ -862,10 +861,18 @@ func getApplicationSetContainerImage(cr *argoproj.ArgoCD) string {
...
 	}
-
-	// If an env var is specified then use that, but don't override the spec values (if they are present)
-	if e := os.Getenv(common.ArgoCDImageEnvName); e != "" && (defaultTag && defaultImg) {
-		return e
+	// if env variable is not set, then use the default tag common.ArgoCDDefaultArgoVersion and default img common.ArgoCDDefaultArgoImage
+	if e := os.Getenv(common.ArgoCDImageEnvName); e != "" {
+		if defaultImg && !defaultTag {
+			image, _, found := strings.Cut(e, "@")
+			if found {
```
anandf: How to handle if `@` is not found ? do we assume that the env variable contains the image part alone with no tag or digest ? or should we assume if the env variable is with tag ?

controllers/argocd/applicationset.go
akhilnittala: done

controllers/argocd/applicationset.go
akhilnittala: added case for : as well

controllers/argocd/repo_server.go
```diff
@@ -586,8 +587,19 @@ func getRepoServerContainerImage(cr *argocdoperatorv1beta1.ArgoCD) string {
...
+			if found {
+				return argoutil.CombineImageTag(image, tag)
+			}
+			image, _, found = strings.Cut(e, ":")
+			if found {
+				return argoutil.CombineImageTag(image, tag)
+			}
+		} else if defaultImg && defaultTag {
+			return e
+		}
```
svghadi: Is it possible to move this to a common generic function?

I was thinking we could simplify the functions like `getRepoServerContainerImage`, `getApplicationSetContainerImage` and others. Right now they use booleans to decide the `image` and `tag`. We could simplify the logic by setting the image and tag to default values first and then override them if environment variables or values in the CR are set. Basically following:
```
image, tag = default < env < CR
```

akhilnittala: modified, done. thanks sid.

controllers/argocd/util.go
```diff
@@ -190,24 +191,86 @@ func getArgoApplicationControllerCommand(cr *argoproj.ArgoCD, useTLSForRedis boo
...
+func GetImageAndTag(envVar, containerSpecImage, containerSpecVersion, commonSpecImage, commonSpecVersion string) (string, string) {
+	// Start with defaults
+	image := common.ArgoCDDefaultArgoImage
+	tag := common.ArgoCDDefaultArgoVersion
+
+	// Check if environment variable is set
+	envVal := os.Getenv(envVar)
+
+	// If no spec values are provided and env var is set, use env var as-is
+	if envVal != "" && containerSpecImage == "" && commonSpecImage == "" &&
```
anandf, line 209: need to check `containerSpecImage`, `containerSpecVersion` as well isn't it ?

anandf:
Lgtm

PR #1888, opened 2025-09-30
fix: Redis server address is hard coded for principal deployment

jparsai:
/kind bug

This PR is to fix principal deployment configs where default redis server address is hard coded to "argocd-redis", but when ArgoCD is deployed in different namespace then redis service name would be changed and this will cause issue in principal as it will be looking for a wrong address.

Fixes https://issues.redhat.com/browse/GITOPS-7790

### DIFF
--- a/tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
+++ b/tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
@@ -23,6 +23,7 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
+	"fmt"
 	"math/big"
 	"net"
 	"strings"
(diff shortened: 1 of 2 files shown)


tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
```diff
@@ -51,8 +53,8 @@ import (
 var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 
 	const (
-		argoCDName               = "argocd"
-		argoCDAgentPrincipalName = "argocd-agent-principal"
+		argoCDName               = "openshift-gitops"
```
jgwest: The `openshift-gitops` `ArgoCD` CR name is an invention of the downstream `gitops-operator` repository. Notice that there are no other references to `openshift-gitops` within this codebase. 

We should thus not use it here, as the argocd-operator repo is upstream of `gitops-operator` (and argocd-operator is vendor neutral, whereas openshift-gitops operator is not).

jgwest (changes requested):

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
jparsai: My bad :) I used a different name now, PTAL

jgwest (approved):
LGTM, thanks @jparsai!

PR #1894, opened 2025-10-01
fix: preserve checksums in ArgoCD CR .status field

jgwest:
**What type of PR is this?**
/kind bug

Due to missing behaviour from https://github.com/argoproj-labs/argocd-operator/pull/1871, several of the downstream operator tests were failing:

```
Summarizing 4 Failures:
  [FAIL] GitOps Operator Parallel E2E Tests 1-066_validate_redis_secure_comm_no_autotls_no_ha [It] validates that Argo CD components correctly inherit 'argocd-operator-redis-tls' Secret once it is created
  /go/src/github.com/redhat-developer/gitops-operator/test/openshift/e2e/ginkgo/parallel/1-066_validate_redis_secure_comm_no_autotls_no_ha_test.go:75
  [FAIL] GitOps Operator Parallel E2E Tests 1-023_validate_repo_server_tls [It] verifying ArgoCD .spec.repo AutoTLS and verifyTLS work as expected
  /go/src/github.com/redhat-developer/gitops-operator/test/openshift/e2e/ginkgo/parallel/1-023_validate_repo_server_tls_test.go:79
  [FAIL] GitOps Operator Parallel E2E Tests 1-069_validate_redis_secure_comm_autotls_ha [It] verifying when HA is enabled that Argo CD starts successfully in HA mode, and that AutoTLS can be enabled
  /go/src/github.com/redhat-developer/gitops-operator/test/openshift/e2e/ginkgo/parallel/1-069_validate_redis_secure_comm_autotls_ha_test.go:84
  [FAIL] GitOps Operator Parallel E2E Tests 1-067_validate_redis_secure_comm_no_autotls_ha [It] ensures that redis HA can be enabled with tls with generated certificate
  /go/src/github.com/redhat-developer/gitops-operator/test/openshift/e2e/ginkgo/parallel/1-067_validate_redis_secure_comm_no_autotls_ha_test.go:87
```

This failure is due to #1871 not preserving the `.status.redisTLSChecksum` and `.status.repoTLSChecksum` between reconciles. The value of these two fields should NOT be regenerated after each reconcile, but instead should be preserved until replaced by another value (this is the existing operator behaviour before #1871).

The consequence of this is that redis/repo server logic that relies on this checksum will enter a continous reconciliation loop, which causes the above tests to fail.

This test also adds downstream tests to upstream, allow us to catch this issue in the future.


controllers/argocd/repo_server.go
```diff
@@ -762,6 +754,11 @@ func (r *ReconcileArgoCD) reconcileRepoServerTLSSecret(cr *argocdoperatorv1beta1
 		if err != nil {
 			return err
 		}
+		// We store the value early to prevent a possible restart loop, for the
```
jopit: Since we are no longer actually updating the status on the cluster right here, I don't think this comment is accurate any longer.  But perhaps I am missing something?

controllers/argocd/secret.go
```diff
@@ -713,6 +704,12 @@ func (r *ReconcileArgoCD) reconcileRedisTLSSecret(cr *argoproj.ArgoCD, useTLSFor
 		if err != nil {
 			return err
 		}
+
+		// We store the value early to prevent a possible restart loop, for the
```
jopit: Since we are no longer actually updating the status on the cluster right here, I don't think this comment is accurate any longer. But perhaps I am missing something?

tests/ginkgo/parallel/1-066_validate_redis_secure_comm_no_autotls_no_ha_test.go
```diff
@@ -0,0 +1,182 @@
...
+	statefulsetFixture "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/statefulset"
+	fixtureUtils "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/utils"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
```
jopit, line 43: Since this is the ArgCD Operator, maybe we should change the description?

tests/ginkgo/parallel/1-067_validate_redis_secure_comm_no_autotls_ha_test.go
```diff
@@ -0,0 +1,219 @@
...
+	statefulsetFixture "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/statefulset"
+	fixtureUtils "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/utils"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
```
jopit, line 42: Change `GitOps Operator` to `ArgoCD Operator`?

controllers/argocd/secret_test.go
```diff
@@ -457,13 +470,18 @@ func Test_ReconcileArgoCD_ReconcileRedisTLSSecret(t *testing.T) {
...
+
 		// Second run - no change
-		err = r.reconcileRedisTLSSecret(argocd, true)
+		err = r.reconcileRedisTLSSecret(argocd, true, argocdStatus)
 		if err != nil {
 			t.Errorf("Error should be nil, but is %v", err)
 		}
-		if shasum != argocd.Status.RedisTLSChecksum {
-			t.Errorf("Error in SHA256 sum of secret, want=%s got=%s", shasum, argocd.Status.RepoTLSChecksum)
+		if argocdStatus.RedisTLSChecksum != "" { // The checksum should be "" because it has not changed.
```
jopit, line 483: A really minor nit -- I think it would be clearer (and a better test) if the `argocdStatus .RedisTLSChecksum` field was set to something other than the empty string before being passed into `reconcileRedisTLSSecret()` and then checked to see it hadn't changed.  Up to you though.

tests/ginkgo/parallel/1-066_validate_redis_secure_comm_no_autotls_no_ha_test.go
jgwest: These test are shared between argocd-operator and gitops-operator codebase, so I needed a name that would work (kind of) for both. I went with GitOps Operator, but I agree there may be a better name. 

(I will add this to my TODO list for future ginkgo work)

jgwest:
Agree on all points and have updated PR, thanks @jopit! For the test suite rename I will defer that to another PR (there is some other non-functional cleanup work I can likely do at the same time).

PR #1897, opened 2025-10-04
feat(ImageUpdater): Add ImageUpdater controller

dkarpele:
**What type of PR is this?**

/kind enhancement

Image Updater: Provide option in Argo CD CR to enable/disable image updater

Fixes #?


controllers/argocd/image_updater.go
svghadi, line 1: Can we follow the `snake_case` naming convention for the file to stay consistent with the codebase?

controllers/argocd/image_updater.go
```diff
@@ -0,0 +1,730 @@
...
+		Spec: imageupdaterv1alpha1.ImageUpdaterSpec{
+			Namespace: cr.Namespace,
+			ApplicationRefs: []imageupdaterv1alpha1.ApplicationRef{
+				{
+					NamePattern: "app-name-pattern",
+					Images: []imageupdaterv1alpha1.ImageConfig{
+						{
+							Alias:     "image-alias",
+							ImageName: "image-name",
+						},
```
svghadi: Are these placeholder values or actual ones? I haven’t looked into the image-updater project yet, so just confirming.

controllers/argocd/image_updater.go
```diff
@@ -0,0 +1,730 @@
...
+			Namespace: sa.Namespace,
+		},
+	}
+
+	return r.reconcileRoleBindingHelper(cr, desiredClusterRoleBinding)
+}
+
+func (r *ReconcileArgoCD) reconcileImageUpdaterConfigurationCR(cr *argoproj.ArgoCD) error {
+
+	defaultImageUpdaterConfigurationCR := &imageupdaterv1alpha1.ImageUpdater{
```
svghadi: Just trying to understand the use case for having a default ImageUpdater CR?
Is it meant to serve as a catch-all configuration for all image update operations, or is the expectation that users should modify this default ImageUpdater CR to suit their needs, rather than creating new ones?

controllers/argocd/imageUpdater.go
```diff
@@ -0,0 +1,730 @@
...
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
+	}
+
+	podSpec.Containers = []corev1.Container{{
+		Command:         []string{"/manager"},
+		Args:            []string{"run"},
+		Image:           argoutil.CombineImageTag(DefaultImageUpdaterImage, DefaultImageUpdaterTag),
```
svghadi: We should make this configurable via env var so that downstream images can be set through the CSV in gitops-operator. 

For ref: https://github.com/argoproj-labs/argocd-operator/blob/ef7197954c1d2ba5b3f733406d294956a78d423a/common/keys.go#L176

controllers/argocd/image_updater.go
```diff
@@ -0,0 +1,730 @@
...
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
+	}
+
+	podSpec.Containers = []corev1.Container{{
+		Command:         []string{"/manager"},
+		Args:            []string{"run"},
+		Image:           argoutil.CombineImageTag(DefaultImageUpdaterImage, DefaultImageUpdaterTag),
+		ImagePullPolicy: corev1.PullAlways,
```
svghadi, line 409: There are ongoing efforts to make the pull policy configurable, just wanted to highlight that.

svghadi (changes requested):
Thanks @dkarpele for working on this. Looks good overall. I have left some comments. 

- We should also bundle the ImageUpdater CRD into operator, similar to how we handle others [here](https://github.com/argoproj-labs/argocd-operator/tree/master/config/crd/bases).

- In the PR description, it mentions pulling from the crd branch in go.mod:
   ```
   go get github.com/argoproj-labs/argocd-image-updater@crd
   ```
   Is there a specific reason for using the crd branch instead of main or a release branch/tag?

controllers/argocd/image_updater.go
dkarpele: Users can modify this CR or create a new one.
My initial idea was to create a basic ImageUpdater CR that can help user to start working with image updater. @chengfang do you think we should have default ImageUpdater CR or not?

controllers/argocd/image_updater.go
dkarpele: Yes, these are placeholders. `NamePattern`, `Alias`, `ImageName` are all mandatory, so we should give them some names.

dkarpele:
> * In the PR description, it mentions pulling from the crd branch in go.mod:
>   ```
>   go get github.com/argoproj-labs/argocd-image-updater@crd
>   ```
  
>   Is there a specific reason for using the crd branch instead of main or a release branch/tag?

The reason is that at the moment crd development of image updater goes in crd branch (not master). When we make the crd version "main" we will update the dependencies here to use
```
go get github.com/argoproj-labs/argocd-image-updater@main
```

dkarpele:
Thanks for the review @svghadi 
I addressed your requests and also added Image Updater's documentation.

controllers/argocd/image_updater.go
chengfang: It should be straightforward for users to create their ImageUpdater CR, and likely in their application git repo. I imagine most users will probably not actually use or modify this default CR for their purpose. I feel we can remove it to avoid the runtime overhead.

Makefile
```diff
@@ -101,6 +101,10 @@ fmt: ## Run go fmt against code.
 vet: ## Run go vet against code.
 	go vet ./...
 
+get-image-updater-crd: ## Download Image Updater CRD.
+	@echo "downloading image updater crd"
+	@curl -sSLo config/crd/bases/argocd-image-updater.argoproj.io_imageupdaters.yaml https://raw.githubusercontent.com/argoproj-labs/argocd-image-updater/crd/config/crd/bases/argocd-image-updater.argoproj.io_imageupdaters.yaml
+
```
chengfang, line 107: do we still need this target now that the ImageUpdater crd is bundled in?

dkarpele: I was thinking it could be useful in case someone wants to update `argocd-image-updater.argoproj.io_imageupdaters.yaml`. The original of this file is located in Image Updater repo and to reflect the future changes in this crd the target `get-image-updater-crd:` can be used.

dkarpele:
@svghadi 
- I removed ImageUpdater CR as discussed.
- As a result we don't need to update go.mod with ImageUpdater dependency.
- I added unit tests to `controllers/argocd/image_updater_test.go`.
- e2e tests will be created later (end of October), after ImageUpdater CRD release 1.0.0. This is because we can't create Deployment properly (image link leads to an old version of ImageUpdater), see GITOPS-8016.

I don't plan any more changes in this PR.

controllers/argocd/util.go
```diff
@@ -944,8 +978,31 @@ func (r *ReconcileArgoCD) setResourceWatches(bldr *builder.Builder, clusterResou
 		},
 	}
 
+	// Add new predicate to delete ImageUpdater Resources. The predicate watches the Argo CD CR for changes to the `.spec.ImageUpdater.Enabled`
+	// field. When a change is detected that results in image updater being disabled, we trigger deletion of image updater resources
```
svghadi: I was thinking we could let the main reconcile handle the deletion when the component is disabled.
Predicates are generally meant to be lightweight filters that decide whether an event should be queued or skipped, rather than performing side-effect operations like updates or deletions. We already have a few places in the operator doing this, but it might be something we can gradually refactor later for consistency. Thoughts?

dkarpele: Yes, that makes sense. Initially I created the same predicate as `deleteNotificationsPred`.
The main problem with deletion is that the order of resource deletion (deployment first) is different from resource creation (deployment last).  To handle this I 
- removed predicate for ImageUpdater
- created 2 functions: `reconcileImageUpdaterControllerEnabled` and `reconcileImageUpdaterControllerDisabled`.

Each reconcile function (reconcileImageUpdaterDeployment, reconcileImageUpdaterClusterRoleBinding etc.) handles deletion itself.

PR #1900, opened 2025-10-08
fix: Remove dependency on argocd-redis secret

jparsai:
/kind bug 

This PR is to use `argocd-redis-initial-password` secret instead of `argocd-redis` to get redis password  

Fixes https://issues.redhat.com/browse/GITOPS-7837

### DIFF
--- a/tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
+++ b/tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
@@ -29,7 +29,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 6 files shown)


tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
```diff
@@ -547,7 +548,7 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 
 			container := deploymentFixture.GetTemplateSpecContainerByName(argoCDAgentPrincipalName, *principalDeployment)
 			Expect(container).ToNot(BeNil())
-			Expect(container.Image).To(Equal("quay.io/argoprojlabs/argocd-agent:v0.3.2"))
+			Expect(container.Image).To(Equal("quay.io/jparsai/argocd-agent:test"))
```
jparsai, line 542: I have to use my image, since latest agent image is not having this fix yet. Once we have new image I will change it.

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
```diff
@@ -313,11 +313,11 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 			// Create argocd-redis secret
 			redisSecret := &corev1.Secret{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      "argocd-redis",
+					Name:      argocdagent.PrincipalRedisSecretname,
```
jgwest: Would it make sense to remove this code where we create the Redis `Secret`?  I presume the Secret will be eventually be created by the `ArgoCD` operator code, and we just need to wait for it. WDYT?

PR #1906, opened 2025-10-14
feat: Add Route for Principal

jparsai:
/kind enhancement

This PR is to add Route for Principal component.

Fixes: https://issues.redhat.com/browse/GITOPS-7729

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -216,7 +216,7 @@ func main() {
 	}
 
 	// Setup Scheme for OpenShift Routes if available.
-	if argocd.IsRouteAPIAvailable() {
+	if argoutil.IsRouteAPIAvailable() {
 		if err := routev1.Install(mgr.GetScheme()); err != nil {
 			setupLog.Error(err, "")
 			os.Exit(1)
--- a/controllers/argocd/argocd_controller_test.go
+++ b/controllers/argocd/argocd_controller_test.go
@@ -25,6 +25,8 @@ import (
 	testclient "k8s.io/client-go/kubernetes/fake"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	configv1 "github.com/openshift/api/config/v1"
+	routev1 "github.com/openshift/api/route/v1"
 	"github.com/stretchr/testify/assert"
 	appsv1 "k8s.io/api/apps/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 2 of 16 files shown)


mikeshng:
@jparsai will the operator keep enforcing the `Type: corev1.ServiceTypeClusterIP`? If I want to use `Type: corev1.ServiceTypeLoadBalancer` (ie. my cluster doesn't have openshift route api), is it possible to change that (do we need to introduce a new spec for this)?

jparsai:
Yes @mikeshng , according to current implementation it will enforce service type to `ClusterIP`. I can change the logic so that default is `ClusterIP` and if user intentionally changes it to `LoadBalancer`, operator wont revert it until Service is deleted by user.
@jgwest WDYT?

jgwest:
Hey @jparsai @mikeshng, for the Argo CD part of argocd-operator, we have the ability to specify either Service or Route [via the corresponding field of the component.](https://github.com/argoproj-labs/argocd-operator/blob/f7c3f5b0cc957cf0ebed7fd3ceaf9997223d4f3f/api/v1beta1/argocd_types.go#L686)

My vote is that it makes sense to do the same thing for principal of Argo CD Agent: add a `route` field, and add a `service` field, and then when one is defined we handle that K8s resource.
- Note that Route needs to point to a Service, so if Service is not defined we will still create it, but it will default to being ClusterIP

Because we are short on time for the next release, we can add new Spec structs, but they can be very simple, for example:
```go
type ArgoCDAgentServiceSpec struct {
	// Type is the ServiceType to use for the Service resource.
	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Service Type'",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:fieldGroup:Server","urn:alm:descriptor:com.tectonic.ui:text"}
}


// ArgoCDRouteSpec defines the desired state for an OpenShift Route.
type ArgoCDAgentRouteSpec struct {
	// Enabled will toggle the creation of the OpenShift Route.
	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Route Enabled'",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:fieldGroup:Grafana","urn:alm:descriptor:com.tectonic.ui:fieldGroup:Prometheus","urn:alm:descriptor:com.tectonic.ui:fieldGroup:Server","urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
	Enabled bool `json:"enabled"`

}
```
That is, for now we will NOT allow customization of the Route (but can add this in the future when users request it).

@jparsai WDYT?

jparsai:
Yes @jgwest this sounds good, I will make this change.

jparsai:
Raised PR for this
https://github.com/argoproj-labs/argocd-operator/pull/1925

PR #1209, opened 2024-01-31
feat: Support appset in any namespace ArgoCD feature

svghadi:
**What type of PR is this?**
> /kind enhancement

**What does this PR do / why we need it**:
[ApplicationSet in any namespace](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Appset-Any-Namespace/) feature allows users to mange `ApplicationSet` resources in namespaces other than the ArgoCD/control plane's namespace This PR adds support to configure this Argo CD feature via operator. 

Two new fields are introduced in ArgoCD CR to support ApplicationSet in any namespace. 
- `.spec.applicationSet.sourceNamespaces` - list of namespaces to allow users to create appset resources
- `.spec.applicationSet.scmProviders` - optional list of scm providers

Changes:
- Add new applicationset-controller clusterrole & clusterrolebinding for cluster-scoped ArgoCD instance
- Add new role & rolebinding in target namespace for applicationset-controller
- Update role for argocd-server in target namespace to grant access to appsets
- Add new fields under `.spec.applicationSet` in ArgoCD CR
- Modify applicationset-controlller deployment command to include new params to support appset in any ns
- Add unit & kuttl test

**Important Note:**
- This feature can only be enabled and used when Argo CD is installed as a cluster-wide instance
- ApplicationSet in any namespace requires App in any namespace configured to function properly. For instance, the applications generated by applicationset in namespace `foo` won't be created if App in any namespace for `foo` is not configured. How it would look in ArgoCD CR is
   ```yaml
   spec:
     applicationSet:
       sourceNamespaces:
         - foo
     # appset in any ns requires apps in any ns to function properly
     sourceNamespaces:
       - foo
   ```
- ApplicationSet in any namespace doesn't support default SCM providers list. SCM providers are disabled by default when this feature is enabled via operator, admins need to explicitly defined the allowed SCM providers if required. https://github.com/argoproj/argo-cd/issues/15246 

**Have you updated the necessary documentation?**

* [x] Documentation update is required by this PR.
* [ ] Documentation has been updated.

**How to test changes / Special notes to the reviewer**:
*Automated kuttl test:*
```
kubectl kuttl test tests/k8s/ --config tests/kuttl-tests.yaml  --test 1-002_validate_cluster_config
kubectl kuttl test tests/k8s/ --config tests/kuttl-tests.yaml --test 1-037_validate_applicationset_in_any_namespace
```
*Manual test:*
1. Create namespaces for testing
   ```bash
   kubectl create ns test
[...]

PR #1912, opened 2025-10-22
feat: Support notifications in any namespace 

nmirasch:
**What type of PR is this?**

/kind enhancement 

Adding support for the Operator to be able to use ArgoCD in Any Namespaces with The Notification Controller. Support upstream Argo CD in any namespace feature in operator https://argo-cd.readthedocs.io/en/stable/operator-manual/notifications/#namespace-based-configuration 

Fixes https://issues.redhat.com/browse/GITOPS-8026 & https://issues.redhat.com/browse/GITOPS-8025

### DIFF
--- a/controllers/argocd/notifications.go
+++ b/controllers/argocd/notifications.go
@@ -692,11 +692,6 @@ func (r *ReconcileArgoCD) reconcileSourceNamespaceNotificationsConfigurationCR(c
 				Name:      DefaultNotificationsConfigurationInstanceName,
 				Namespace: sourceNamespace,
 			},
-			Spec: v1alpha1.NotificationsConfigurationSpec{
-				Context:   getDefaultNotificationsContext(),
-				Triggers:  getDefaultNotificationsTriggers(),
-				Templates: getDefaultNotificationsTemplates(),
-			},
 		}
 		argoutil.LogResourceCreation(log, sourceNotifCfg, "creating NotificationsConfiguration for namespace")
 		return r.Create(context.TODO(), newCfg)
--- a/controllers/argocd/notifications_test.go
+++ b/controllers/argocd/notifications_test.go
@@ -847,57 +847,6 @@ func TestNotifications_removeUnmanagedNotificationsSourceNamespaceResources(t *t
 	assert.Equal(t, a.Namespace, val)
 }
 
-func TestReconcileNotifications_CreateNotificationsConfigurationInSourceNamespace_WithDefaults(t *testing.T) {
-	logf.SetLogger(ZapLogger(true))
-	sourceNamespace := "ns1"
-	a := makeTestArgoCD(func(a *argoproj.ArgoCD) {
-		a.Spec.Notifications.Enabled = true
-		a.Spec.Notifications.SourceNamespaces = []string{sourceNamespace}
-		a.Spec.SourceNamespaces = []string{sourceNamespace}
-	})
-
-	resObjs := []client.Object{a}
-	subresObjs := []client.Object{a}
-	runtimeObjs := []runtime.Object{}
-	sch := makeTestReconcilerScheme(argoproj.AddToScheme)
-	err := v1alpha1.AddToScheme(sch)
-	assert.NoError(t, err)
-	cl := makeTestReconcilerClient(sch, resObjs, subresObjs, runtimeObjs)
-	r := makeTestReconciler(cl, sch, testclient.NewSimpleClientset())
-
-	// Create the source namespace
-	err = createNamespace(r, sourceNamespace, "")
-	assert.NoError(t, err)
-
-	// Reconcile should create NotificationsConfiguration CR in source namespace with defaults
-	// (since no NotificationsConfiguration exists in instance namespace)
-	err = r.reconcileSourceNamespaceNotificationsConfigurationCR(a, sourceNamespace)
-	assert.NoError(t, err)
-
-	// Verify NotificationsConfiguration CR exists in source namespace
-	notifCfg := &v1alpha1.NotificationsConfiguration{}
-	err = r.Get(context.TODO(), types.NamespacedName{
-		Name:      DefaultNotificationsConfigurationInstanceName,
-		Namespace: sourceNamespace,
-	}, notifCfg)
-	assert.NoError(t, err)
-
-	// Verify it has the expected defaults
-	expectedContext := getDefaultNotificationsContext()
-	expectedTriggers := getDefaultNotificationsTriggers()
-	expectedTemplates := getDefaultNotificationsTemplates()
-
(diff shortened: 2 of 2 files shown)


svghadi:
Can have the cleanup logic similar to #1209 for consistency?

controllers/argocd/notifications.go
```diff
@@ -65,6 +70,30 @@ func (r *ReconcileArgoCD) reconcileNotificationsController(cr *argoproj.ArgoCD)
...
+	log.Info("reconciling notifications clusterroles")
+	clusterrole, err := r.reconcileNotificationsClusterRole(cr)
+	if err != nil {
+		return err
+	}
+
+	log.Info("reconciling notifications clusterrolebindings")
+	if err := r.reconcileNotificationsClusterRoleBinding(cr, clusterrole, sa); err != nil {
+		return err
+	}
```
svghadi: Can you confirm if we actually need the cluster roles? In the upstream cluster-scoped [manifests](https://github.com/argoproj/argo-cd/blob/master/manifests/install.yaml), I don’t see any cluster role for the `notifications-controller`, so we probably don’t need one.

controllers/argocd/notifications.go
```diff
@@ -537,6 +821,23 @@ func getNotificationsCommand(cr *argoproj.ArgoCD) []string {
...
+				notificationsSourceNamespaces = append(notificationsSourceNamespaces, ns)
+			} else {
+				log.V(1).Info(fmt.Sprintf("Apps in target sourceNamespace %s is not enabled, thus skipping the namespace in deployment command.", ns))
+			}
+		}
+	}
+
+	if len(notificationsSourceNamespaces) > 0 {
+		cmd = append(cmd, "--application-namespaces", fmt.Sprint(strings.Join(notificationsSourceNamespaces, ",")))
+	}
```
svghadi, line 736: Seems like you missed adding the `--self-service-notification-enabled` flag.

controllers/argocd/notifications.go
nmirasch: @svghadi That's correct, removed!

nmirasch:
@svghadi code updated!

svghadi:
I was wondering if we should consider creating a `NotificationsConfiguration` CR in the source namespaces as well, similar to what we do in the Argo CD instance namespace. This would provide a consistent UX where users manage the `argocd-notifications-cm` ConfigMap through the `NotificationsConfiguration` CR instead of editing the ConfigMap directly.

controllers/argocd/notifications.go
```diff
@@ -520,7 +549,215 @@ func (r *ReconcileArgoCD) reconcileNotificationsSecret(cr *argoproj.ArgoCD) erro
...
+	for _, sourceNamespace := range cr.Spec.Notifications.SourceNamespaces {
+
+		// source ns should be part of app-in-any-ns
+		appsNamespaces, err := r.getSourceNamespaces(cr)
+		if err != nil {
+			reconciliationErrors = append(reconciliationErrors, err)
+			continue
+		}
+		if !contains(appsNamespaces, sourceNamespace) {
+			log.Error(fmt.Errorf("skipping reconciliation of Notification resources for sourceNamespace %s as Apps in target sourceNamespace is not enabled", sourceNamespace), "Warning")
```
svghadi: Let's log this at info level

controllers/argocd/notifications.go
```diff
@@ -520,7 +549,215 @@ func (r *ReconcileArgoCD) reconcileNotificationsSecret(cr *argoproj.ArgoCD) erro
...
+			}
+			r.ManagedNotificationsSourceNamespaces[sourceNamespace] = ""
+		}
+	}
+
+	return amerr.NewAggregate(reconciliationErrors)
+}
+
+// reconcileSourceNamespaceNotificationsConfigurationCR ensures a NotificationsConfiguration CR exists in the given source namespace.
+// It propagates the NotificationsConfiguration from the Argo CD instance namespace if it exists, otherwise uses defaults.
```
svghadi: Lets create an empty `NotificationsConfiguration` CR in the source namespaces instead of propagating the configs from Argo CD instance namespace or the default values. Notifications in any namespace has a fallback to Argo CD instance's notification configs if not found in source namespaces.

controllers/argocd/argocd_controller.go
```diff
@@ -78,6 +78,10 @@ type ReconcileArgoCD struct {
 	ManagedSourceNamespaces map[string]string
 	// Stores a list of ApplicationSetSourceNamespaces as keys
 	ManagedApplicationSetSourceNamespaces map[string]string
+
+	// Stores a list of ApplicationSetSourceNamespaces as keys
```
svghadi:
```suggestion
	// Stores a list of NotificationsSourceNamespaces as keys
```

controllers/argocd/notifications.go
```diff
@@ -551,3 +806,140 @@ func getNotificationsResources(cr *argoproj.ArgoCD) corev1.ResourceRequirements
...
+			return fmt.Errorf("failed to get the NotificationsConfiguration in namespace %s : %s", namespace.Name, err)
+		}
+	} else {
+		argoutil.LogResourceDeletion(log, notifCfg, "cleaning up unmanaged notifications resources")
+		if err := r.Delete(context.TODO(), notifCfg); err != nil {
+			return fmt.Errorf("failed to delete the NotificationsConfiguration in namespace %s : %s", namespace.Name, err)
+		}
+	}
+
+	// app-in-any-ns code will handle removal of notifications permissions for argocd-server in target namespace
```
svghadi: Do we add any permissions to argocd-server resources for notifications in any namespace?

svghadi (commented):
Overall looks good. Left few comments.

controllers/argocd/notifications.go
```diff
@@ -520,7 +549,163 @@ func (r *ReconcileArgoCD) reconcileNotificationsSecret(cr *argoproj.ArgoCD) erro
...
+		// Not found in source namespace, create it with propagated/default spec
+		newCfg := &v1alpha1.NotificationsConfiguration{
+			ObjectMeta: v1.ObjectMeta{
+				Name:      DefaultNotificationsConfigurationInstanceName,
+				Namespace: sourceNamespace,
+			},
+			Spec: v1alpha1.NotificationsConfigurationSpec{
+				Context:   getDefaultNotificationsContext(),
+				Triggers:  getDefaultNotificationsTriggers(),
+				Templates: getDefaultNotificationsTemplates(),
```
svghadi: This will still populate the default notifications configs. Can you try assigning a empty map?
```suggestion
				Context:   make(map[string]string),
				Triggers:  make(map[string]string),
				Templates: make(map[string]string),
```

svghadi:
Merging this one. Let's add e2e and docs in separate PR.

PR #1925, opened 2025-10-30
feat: Add switch for Route and Service for Principal

jparsai:
/kind enhancement

This PR adds a switch to enable/disable creation of principal route and option to provide expected type (ClusterIP/LoadBalancer) of principal service.

Fixes https://issues.redhat.com/browse/GITOPS-8072


api/v1alpha1/argocd_types.go
```diff
@@ -1123,6 +1129,18 @@ type PrincipalTLSSpec struct {
 	InsecureGenerate *bool `json:"insecureGenerate,omitempty"`
 }
 
+// ArgoCDAgentPrincipalServiceSpec defines the options for the Service backing the ArgoCD Agent Principal component.
+type ArgoCDAgentPrincipalServiceSpec struct {
+	// Type is the ServiceType to use for the Service resource.
```
jgwest, line 1136: Can add a comment indicating what the default is for this field. Something like 'if type is not specified, the default of 'ClusterIP' will be used'

api/v1alpha1/argocd_types.go
```diff
@@ -1123,6 +1129,18 @@ type PrincipalTLSSpec struct {
...
+	Type corev1.ServiceType `json:"type"`
+}
+
+// ArgoCDAgentPrincipalRouteSpec defines the options for the Route backing the ArgoCD Agent Principal component.
+type ArgoCDAgentPrincipalRouteSpec struct {
+	// Enabled will toggle the creation of the OpenShift Route.
```
jgwest: Can add a comment like, 'if enabled is not defined, the default is ...'

api/v1alpha1/argocd_types.go
```diff
@@ -1067,6 +1067,12 @@ type PrincipalServerSpec struct {
 
 	// Env lets you specify environment for principal pods
 	Env []corev1.EnvVar `json:"env,omitempty"`
+
+	// Service defines the options for the Service backing the ArgoCD Agent component.
```
jgwest, line 1071: Can add a comment like: 'if Service is not defined, a default of ClusterIP will be used'

api/v1alpha1/argocd_types.go
```diff
@@ -1067,6 +1067,12 @@ type PrincipalServerSpec struct {
...
+	Service ArgoCDAgentPrincipalServiceSpec `json:"service,omitempty"`
+
+	// Route defines the options for the Route backing the ArgoCD Agent component.
```
jgwest, line 1075: Can add a comment like 'If Route is not specified, a default of ... will be used'

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
```diff
@@ -351,28 +356,31 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
...
+
+				// skip principal service
+				if serviceName != argoCDAgentPrincipalName {
+					Expect(string(service.Spec.Type)).To(Equal("ClusterIP"), "Service '%s' should have ClusterIP type, got '%s'", serviceName, service.Spec.Type)
+				}
 			}
 
-			By("verifying Route for principal exists on OpenShift")
+			if shouldExpectRoute {
+				By("verifying Route for principal exists on OpenShift")
```
jgwest: Nit: IMHO can move this `By` inside the `if fixture.RunningOnOpenShift {}`,so as not to confuse the case where it is NOT running on openshift

jparsai:
> LGTM, some minor tweaks then good to go!
> 
> I'm assuming you already have, but if you haven't already, it would be good to run the updated E2E test on an OpenShift cluster to make sure it works as expected (since the E2E tests in argocd-operator repo use k3d which does not have Route installed).

Thanks Jonathan, yes I tested all principal e2e tests in openshift. Addressed your review comments, PTAL.

PR #1945, opened 2025-11-24
fix: namespace validation

jparsai:
This PR is to validate namespace.

### DIFF
--- a/controllers/argoutil/namespace.go
+++ b/controllers/argoutil/namespace.go
@@ -1,0 +1,48 @@
+// Copyright 2025 ArgoCD Operator Developers
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


controllers/argoutil/namespace.go
```diff
@@ -0,0 +1,43 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package argoutil
+
+import (
+	"strings"
+)
+
```
jgwest, line 21: AFAICT, for all the cases in the code where we are calling `AllowedNamespace`, the second parameter is always `os.Getenv("ARGOCD_CLUSTER_CONFIG_NAMESPACES")`

Should we just have a new function like:
```go
func IsNamespaceClusterConfigNamespace(current string) bool {
    return allowedNamespace(current, os.Getenv("ARGOCD_CLUSTER_CONFIG_NAMESPACES")(
}
```

And then replace all the existing callers that are using `AllowedNamespaces` to just use the above func?

This way, we don't have to repeat the same logic over and over in a bunch of different places in the code. 
- (Yes, this is affecting code outside Argo CD agent, but the change is trivial)


And `AllowedNamespace` -> `allowedNamespace`

jparsai: It is updated, PTAL

jparsai:
It seems there is some flaky test, retriggered test

PR #1956, opened 2025-12-02
fix: namespace validation for sourcenamespace

jparsai:
Namespace validation for source namespaces.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of cluster-config namespaces so source-namespace RBAC creation/removal and cleanup behave correctly.
  * Controller now skips source-namespace reconciliation when not permitted, avoiding unintended resource changes.
  * Application controller/server command no longer includes application-namespaces when namespaces aren’t allowed.

* **Security**
  * Removed batch job create/update permissions from server policy rules.

* **Tests**
  * Added and updated unit and end-to-end tests and test helpers to validate cluster-scoped vs namespace-scoped behaviors, RBAC cleanup, and command flags.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/tests/ginkgo/sequential/1-037_validate_applicationset_in_any_namespace_test.go
+++ b/tests/ginkgo/sequential/1-037_validate_applicationset_in_any_namespace_test.go
@@ -237,15 +237,6 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 					},
 				},
 				{
-					APIGroups: []string{"batch"},
-					Resources: []string{
-						"jobs",
-						"cronjobs",
-						"cronjobs/finalizers",
-					},
-					Verbs: []string{"create", "update"},
-				},
-				{
 					APIGroups: []string{"argoproj.io"},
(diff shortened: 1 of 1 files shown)


controllers/argocd/util.go
```diff
@@ -1477,6 +1489,43 @@ func (r *ReconcileArgoCD) cleanupUnmanagedSourceNamespaceResources(cr *argoproj.
 	return nil
 }
 
+func (r *ReconcileArgoCD) cleanupAllSourceNamespaces(cr *argoproj.ArgoCD) {
```
jgwest: Do we need `cleanupAllSourceNamespaces`? It seems like it's doing the exact same thing as `removeUnmanagedSourceNamespaceResources`?

jgwest:
Pushed some minor tweaks and rebased to latest

nmirasch (approved):
LGTM!

PR #1978, opened 2025-12-11
fix: verify argocd instance matches namespace label on notifications …

nmirasch:
…cleanup

**What type of PR is this?**

/kind bug 

- This improves the logic of notifications in any namespace source cleanup, such that we verify that theArgoCD CR matches the managed namespace.
- Before, we were not checking this, and that meant some resources were leftover.
- Improves notification testing, checking "--application-namespaces" cmd arg presence in the notification deployment container args

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved notification controller cleanup to verify namespace ownership before removing resources, preventing unintended cleanup of namespaces managed by other Argo CD instances.

* **Tests**
  * Enhanced notification controller validation with active assertions for deployment configurations.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/controllers/argocd/notifications.go
+++ b/controllers/argocd/notifications.go
@@ -914,11 +914,48 @@ func (r *ReconcileArgoCD) setManagedNotificationsSourceNamespaces(cr *argoproj.A
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


svghadi (approved):
LGTM

PR #1979, opened 2025-12-12
fix: Restore tracking labels on namespaces

svghadi:
**What type of PR is this?**

Fixes #?

### DIFF
--- a/controllers/argocd/util.go
+++ b/controllers/argocd/util.go
@@ -57,6 +57,7 @@ import (
 	v1 "k8s.io/api/rbac/v1"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	amerr "k8s.io/apimachinery/pkg/util/errors"
 
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


controllers/argocd/util.go
```diff
@@ -2059,3 +2060,75 @@ func (r *ReconcileArgoCD) reconcileArgoCDAgent(cr *argoproj.ArgoCD) error {
...
+		if role.Namespace == cr.Namespace {
+			continue
+		}
+
+		missingTrackingLabels := make(map[string]string)
+
+		// Check for namespaces with ApplicationSet source namespace resources
+		appsetRoleName := getResourceNameForApplicationSetSourceNamespaces(cr)
+		if role.Name == appsetRoleName {
+			missingTrackingLabels[common.ArgoCDApplicationSetManagedByClusterArgoCDLabel] = cr.Namespace
```
anandf, line 2091: rather than naming `missingTrackingLabels` which may be confusing for future versions where these labels could already be present, it is better to name it generic, say `requiredTrackingLabels` or just `trackingLabels`. Later part of the code checks if the label is missing or not and then add it if required. WDYT ?

svghadi:
Superseded by #2017

PR #2030, opened 2026-01-20
feat: Validation webhook to reject invalid SSO configs in ArgoCD CR

akhilnittala:
**What type of PR is this?**

/kind enhancement

When both sso and oidc config is provided in argocd, we are not rejecting the request. so with this pr changes, we have implemented validation webhook which rejects the request with argocd CR sso and oidc configuration.

https://issues.redhat.com/browse/GITOPS-8560
Fixes #?
https://issues.redhat.com/browse/GITOPS-8560

- install argocd operator with changes
- create argo cd CR with both sso and oidc config
- request should be rejected.

### DIFF
--- a/bundle/manifests/argocd-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/argocd-operator.clusterserviceversion.yaml
@@ -257,7 +257,7 @@ metadata:
     capabilities: Deep Insights
     categories: Integration & Delivery
     certified: "false"
-    createdAt: "2025-12-17T12:19:06Z"
+    createdAt: "2026-01-20T06:46:19Z"
     description: Argo CD is a declarative, GitOps continuous delivery tool for Kubernetes.
     operators.operatorframework.io/builder: operator-sdk-v1.35.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v4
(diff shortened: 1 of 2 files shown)


api/v1beta1/argocd_webhook.go
```diff
@@ -17,11 +17,54 @@ limitations under the License.
...
 )
 
-func (r *ArgoCD) SetupWebhookWithManager(mgr ctrl.Manager) error {
+type ArgoCDValidator struct {
+	// TODO(user): Add more fields as needed for validation
+}
+
+var _ admission.CustomValidator = &ArgoCDValidator{}
+
+func SetupWebhookWithManager(mgr ctrl.Manager) error {
```
anandf: Why changing from a receiver function ?

svghadi:
Hi @akhilnittala , thanks for the PR. While reviewing it, I was reading about validation webhooks and came across Kubebuilder validation markers (XValidation).
It seems these validations run at the API server level and can handle many basic validation use cases without requiring us to write and maintain custom webhooks.

Have we explored using this approach here?

Ref:
- https://book.kubebuilder.io/reference/markers/crd-validation
- https://blog.rewanthtammana.com/kubernetes-crd-validation-with-cel-and-kubebuilder-marker-comments

akhilnittala:
> Hi @akhilnittala , thanks for the PR. While reviewing it, I was reading about validation webhooks and came across Kubebuilder validation markers (XValidation). It seems these validations run at the API server level and can handle many basic validation use cases without requiring us to write and maintain custom webhooks.
> 
> Have we explored using this approach here?
> 
> Ref:
> 
> * https://book.kubebuilder.io/reference/markers/crd-validation
> * https://blog.rewanthtammana.com/kubernetes-crd-validation-with-cel-and-kubebuilder-marker-comments

Thanks sid, Looks like this is the new feature from kubernetes, will try and make the changes accordingly. Thanks for sharing the information.

akhilnittala:
closing this PR, will create a new one with suggested changes

PR #2031, opened 2026-01-21
feat: refactor request-scoped reconciler variables to new struct

jgwest:
**What type of PR is this?**
/kind enhancement

TBD: refactor request-scoped reconciler variables to new struct

### DIFF
--- a/controllers/argocd/argocd_controller.go
+++ b/controllers/argocd/argocd_controller.go
@@ -57,7 +57,7 @@ type TokenRenewalTimer struct {
 }
 
 type LocalUsersInfo struct {
-	// Stores the the timers that will auto-renew the API tokens for local users
+	// Stores the timers that will auto-renew the API tokens for local users
 	// after they expire. The key format is "argocd-namespace/user-name"
 	TokenRenewalTimers map[string]*TokenRenewalTimer
 	// Protects access to the token renewal timers and the K8S resources that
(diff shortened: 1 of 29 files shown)


controllers/argocd/argocd_controller.go
```diff
@@ -160,6 +176,8 @@ func (r *ReconcileArgoCD) Reconcile(ctx context.Context, request ctrl.Request) (
 
 func (r *ReconcileArgoCD) internalReconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, *argoproj.ArgoCD, *argoproj.ArgoCDStatus, error) {
 
+	reqState := &RequestState{}
```
akhilnittala, line 179: we can make this as part of ReconcileArgoCD struct and we can use r.reqState in struct methods? it will reduce the lines of code by passing to below functions.

olivergondza: I am not sure what is the exact intent here, but one thing Jonathan's change archives, is that the datastructure is not sensitive to precisely updating per-request state when `ReconcileArgoCD.Reconcile` is called, eventually even permitting an overlapping reconciliations. This was a challenge, because figuring out which fields are reconciler specific and which are request specific required detailed understanding.

It points out the 1:N relationship between reconciler and reconciliation request. Referencing the new struct from the old one, would still enforce 1:1 relationship, albeit it would be easier to update (one field, over multiple).

I would suggest doing it the other way around: Let `RequestState` point to its `ReconcileArgoCD` (1:N relationship honored), and even make it the receiver of the follow-up calls. It would not require passing in one more parameter, because `ReconcileArgoCD` will be accessible like `req.r.`.

PR #2059, opened 2026-02-09
feat(redis): Inject creds via volume

olivergondza:
**What type of PR is this?**

/kind enhancement

Improve security by injection redis credentials to components by avoiding envvars, and cmdline arguments.

Agent/principal will be updated in follow-up PRs as it does not support `REDIS_CREDS_DIR_PATH` for now.

Fixes #?

## Summary by CodeRabbit

* **Refactor**
  * Redis authentication now uses a mounted credentials file (secret) instead of in-container environment variables; credentials are propagated to Server, Repo Server, Application Controller, Agent and HA components.
  * Redis and HAProxy init/readiness/liveness scripts and config templates updated to use the new credential placeholders and env-driven auth handling.

* **Tests**
  * New/updated end-to-end and unit tests verify secure distribution and use of Redis credentials, including HA scenarios.

### DIFF
--- a/controllers/argocd/deployment.go
+++ b/controllers/argocd/deployment.go
@@ -525,20 +525,12 @@ func (r *ReconcileArgoCD) reconcileRedisDeployment(cr *argoproj.ArgoCD, useTLS b
 
 		if !reflect.DeepEqual(deploy.Spec.Template.Spec.Containers[0].VolumeMounts, existing.Spec.Template.Spec.Containers[0].VolumeMounts) {
 			existing.Spec.Template.Spec.Containers[0].VolumeMounts = deploy.Spec.Template.Spec.Containers[0].VolumeMounts
-			if changed {
-				explanation += ", "
-			}
-			explanation += "container volume mounts"
-			changed = true
+			changes = append(changes, "container volume mounts")
 		}
 
 		if !reflect.DeepEqual(deploy.Spec.Template.Spec.Volumes, existing.Spec.Template.Spec.Volumes) {
 			existing.Spec.Template.Spec.Volumes = deploy.Spec.Template.Spec.Volumes
-			if changed {
-				explanation += ", "
-			}
-			explanation += "volumes"
-			changed = true
+			changes = append(changes, "volumes")
 		}
 
 		if !reflect.DeepEqual(existing.Spec.Template.Spec.Containers[0].Env,
--- a/tests/ginkgo/parallel/1-019_validate_volume_mounts_test.go
+++ b/tests/ginkgo/parallel/1-019_validate_volume_mounts_test.go
@@ -225,6 +225,10 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
 						Secret: &corev1.SecretVolumeSource{
 							SecretName:  "argocd-redis-initial-password",
 							DefaultMode: ptr.To(int32(420)),
+							Items: []corev1.KeyToPath{
+								{Key: "auth", Path: "auth"},
+								{Key: "auth_username", Path: "auth_username"},
+							},
 						},
 					},
 				},
@@ -296,6 +300,10 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
 						Secret: &corev1.SecretVolumeSource{
 							SecretName:  "argocd-redis-initial-password",
 							DefaultMode: ptr.To(int32(420)),
+							Items: []corev1.KeyToPath{
+								{Key: "auth", Path: "auth"},
+								{Key: "auth_username", Path: "auth_username"},
+							},
 						},
 					},
 				},


tests/ginkgo/sequential/1-067_validate_redis_secure_comm_no_autotls_ha_test.go
```diff
@@ -213,8 +214,60 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
...
+			for _, component := range expectedComponents {
+				logOutput, err := osFixture.ExecCommandWithOutputParam(false, true,
+					"kubectl", "logs", component, "-n", ns.Name,
+				)
+				Expect(err).ToNot(HaveOccurred(), "Output: "+logOutput)
+				Expect(logOutput).To(ContainSubstring(expectedMsg))
+				// Some logs how redis disconnect manifests
+				Expect(logOutput).ToNot(ContainSubstring("manifest cache error"))
+				Expect(logOutput).ToNot(ContainSubstring("WRONGPASS"))
+			}
```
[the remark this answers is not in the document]
olivergondza: Disregard. If run too early in the startup, the `expectedMsg` would not be present and the test would fail cleanly stating so...

controllers/argocd/deployment.go
```diff
@@ -429,20 +428,12 @@ func (r *ReconcileArgoCD) reconcileGrafanaDeployment(cr *argoproj.ArgoCD) error
...
-				},
-				Key: "admin.password",
-			},
-		},
-	})
+	env := proxyEnvVars()
 
 	AddSeccompProfileForOpenShift(r.Client, &deploy.Spec.Template.Spec)
 
+	redisVolume, redisMount := argoutil.MountRedisAuthToArgo(cr)
```
Mangaal: We are mounting the Redis auth secret, but how is the password configured for this standalone Redis instance?
In the HA setup, the init.sh script (from the ConfigMap) reads the password from the secret file path and generates the redis.conf along with other configurations in an init container.

olivergondza:
Thanks, @Mangaal. I am not well versed with operator's Redis use, so I will certainly appreciate it if you take a look.

build/redis/haproxy_init.sh.tpl
```diff
@@ -33,6 +33,12 @@ if [ -z "$ANNOUNCE_IP2" ]; then
...
-auth=$(cat /redis-initial-pass/admin.password)
-sed -i "s/replace-with-redis-auth/$auth/" "$HAPROXY_CONF"
+AUTH="$(cat /app/config/redis-auth/auth)"
+if [ -z "${AUTH}" ]; then
+    echo "Error: Redis password not mounted correctly"
+    exit 1
+fi
+echo "Setting redis auth values.."
+ESCAPED_AUTH=$(echo "${AUTH}" | sed -e 's/[\/&]/\\&/g');
+sed -i "s/__REPLACE_DEFAULT_AUTH__/${ESCAPED_AUTH}/" "$HAPROXY_CONF"
```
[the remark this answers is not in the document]
olivergondza: Careful, not true. In `'s/[\/&]/\\&/g'`, it replaces all three `/`, `&` and `\`. The backslash is meant literally in character group. Tested.

build/redis/init.sh.tpl
```diff
@@ -252,16 +253,19 @@ else
...
-    ESCAPED_AUTH=$(echo "${AUTH}" | sed -e 's/[\/&]/\\&/g');
-    sed -i "s/replace-default-auth/${ESCAPED_AUTH}/" "${REDIS_CONF}" "${SENTINEL_CONF}"
+AUTH="$(cat /app/config/redis-auth/auth)"
+if [ -z "${AUTH}" ]; then
+    echo "Error: Redis password not mounted correctly"
+    exit 1
 fi
+echo "Setting redis auth values.."
+ESCAPED_AUTH=$(echo "${AUTH}" | sed -e 's/[\/&]/\\&/g');
+sed -i "s/__REPLACE_DEFAULT_AUTH__/${ESCAPED_AUTH}/" "${REDIS_CONF}" "${SENTINEL_CONF}"
```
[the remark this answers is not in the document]
olivergondza: Careful, not true. In 's/[\/&]/\\&/g', it replaces all three /, & and \. The backslash is meant literally in character group. Tested.

controllers/argoutil/redis.go
```diff
@@ -0,0 +1,320 @@
...
+	if err != nil {
+		log.Error(err, "unable to execute template")
+		return "", fmt.Errorf("unable to execute template. error: %w", err)
+	}
+	return buf.String(), nil
+}
+
+func GenerateAgentPrincipalRedisProxyServiceName(crName string) string {
+	return fmt.Sprintf("%s-agent-%s", crName, "principal-redisproxy")
+}
```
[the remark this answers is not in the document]
olivergondza: Unrelated, that is what the function did even before. Tracking as https://github.com/argoproj-labs/argocd-operator/issues/2071.

controllers/argocd/deployment.go
Mangaal: I was testing out locally. I think the standalone Redis is not password protected.
```
kubectl exec -n test-argocd  example-argocd-redis-64ff4f79cf-f649r -c redis  -- redis-cli  ping 
PONG
```

Similar to Redis HA, we need to have an init container that generates the redis.conf file with the password.

I think we can add a dedicated init script for standalone in build/redis, the current init.sh looks like it can be used only for  HA. We can reuse the existing redis.conf.tpl for standalone or add a similar base config. I did something similar in the Argo CD repo.
https://github.com/argoproj/argo-cd/pull/23487/changes#diff-f0d945f2ef412b62f1a6c2fe58c9d332526fa0c685c31a28c7b85845dd4522da

build/redis/init.sh.tpl
```diff
@@ -170,7 +171,7 @@ find_master() {
         if [ "$SENTINEL_PORT" -eq 0 ]; then
             echo "  on sentinel (${SERVICE}:${SENTINEL_TLS_PORT}), sentinel grp (${MASTER_GROUP})"
             echo "  $(date).."
-            if redis-cli -h "${SERVICE}" -p "${SENTINEL_TLS_PORT}"   --tls --cacert /app/config/redis/tls/tls.crt sentinel failover "${MASTER_GROUP}" | grep -q 'NOGOODSLAVE' ; then
+            if redis-cli -h "${SERVICE}" -p "${SENTINEL_TLS_PORT}" --tls --cacert /app/config/redis/tls/tls.crt sentinel failover "${MASTER_GROUP}" | grep -q 'NOGOODSLAVE' ; then
```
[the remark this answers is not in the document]
olivergondza: Yeah, this has changed the placeholder name, but it was unreferenced even before. Original code that have introduced it is not using it either: https://github.com/argoproj-labs/argocd-operator/commit/7d0356b79d4bd4cb099605455f2b1e41582944b4.

Based on my google search we never advertise `SENTINELAUTH` support. Nor is the `replace-default-sentinel-auth` placeholder mentioned anywhere else around the internet.

Please file an issue to remove this dead code later.

controllers/argocd/deployment.go
olivergondza: Thanks! Let me take a look into this.

Good catch that missing password can manifest in fully functional redis grid. Will add tests that verify redis does not work without password.

controllers/argocd/deployment_test.go
```diff
@@ -2406,6 +2367,20 @@ func repoServerDefaultVolumes() []corev1.Volume {
...
+					SecretName: "argocd-redis-initial-password",
+					Items: []corev1.KeyToPath{
+						{
+							Key:  "admin.password",
+							Path: "auth",
+						},
+					},
+				},
+			},
+		},
```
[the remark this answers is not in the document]
olivergondza: Ignore for tests

controllers/argocd/deployment.go
olivergondza: Great catch, captured in f1b6b38.

controllers/argocd/util.go
```diff
@@ -400,276 +397,11 @@ func getArgoControllerParellismLimit(cr *argoproj.ArgoCD) int32 {
 	return pl
 }
 
-// getRedisConfigPath will return the path for the Redis configuration templates.
```
olivergondza, line 419: There redis specific methods are moved to its own file in `argoutil`.

olivergondza:
@Mangaal, please take another look. CI tests may not be passing cleanly yet...

controllers/argocd/secret.go
```diff
@@ -746,28 +746,52 @@ func (r *ReconcileArgoCD) getClusterSecrets(cr *argoproj.ArgoCD) (*corev1.Secret
...
 	}
 
+	pw := strings.TrimRight(string(redisInitialPassword), "\n")
+	usersACL := fmt.Sprintf("user default on >%s allchannels allkeys allcommands\n", pw)
+
 	secret.Data = map[string][]byte{
-		"immutable":                   []byte("true"),
+		"immutable": []byte("true"),
+		// Mapping the legacy key-name, the operator customers can depend on.
 		common.ArgoCDKeyAdminPassword: redisInitialPassword,
```
olivergondza: I would love getting rid of this now legacy and duplicated key - same content is now in `auth` (key name expected by argo-cd*). Are there some external readers of this (`admin.password` in `Secret/*-redis-initial-password`)?

@Mangaal, @svghadi? Thanks!

*) https://github.com/argoproj/argo-cd/pull/24597/changes

Mangaal: I see that only agent deployments are using the key `admin.password`.
We can standardize on the `auth` key.

olivergondza: When it comes to this repo, yes, it is possible. Although, what I am asking is if there are reasons - outside this repository - why the `admin.password` key needs to be preserved. If there are such reasons, and key old name will be kept, it is not so important to thoroughly migrate even in this repository.

Mangaal: We can add a breaking change comment for the external component to uses the standardized key. I am not sure whether any external component actually uses Redis. I don’t think any external component is currently using Redis.
++ @svghadi

Mangaal:
I hope all tests pass successfully this time. :smile:

controllers/argocd/secret.go
olivergondza: Found searching `argoproj-labs`, `argoproj` and gitops operator docs:

- https://github.com/argoproj-labs/argocd-agent/blob/33793feb39b60e9dc813cf756b15163fbfb0498d/docs/getting-started/openshift/index.md?plain=1#L90
- https://docs.redhat.com/en/documentation/red_hat_openshift_gitops/1.19/html/argo_cd_agent_installation/argocd-agent-installation#gitops-installing-argocd-agent-in-managed-mode-by-using-helm-chart_argocd-agent-installation
- https://docs.redhat.com/en/documentation/red_hat_openshift_gitops/1.19/html/argo_cd_agent_installation/argocd-agent-installation#gitops-installing-argocd-agent-in-autonomous-mode-by-using-helm-chart_argocd-agent-installation

olivergondza: @coderabbitai, please file an issue in this repo to remove this field once the external references mentioned in this thread are migrated to the new name: `auth`. Thanks!

olivergondza: @coderabbitai, please do file that issue. You have not...

controllers/argoutil/resource.go
```diff
@@ -132,6 +132,12 @@ func NameWithSuffix(meta metav1.ObjectMeta, suffix string) string {
 	return fmt.Sprintf("%s-%s", meta.Name, suffix)
 }
 
+// FqdnServiceRef will return the FQDN referencing a specific service name, as set up by the operator, with the
+// given port.
+func FqdnServiceRef(service string, port int, cr *argoproj.ArgoCD) string {
+	return fmt.Sprintf("%s.%s.svc.cluster.local:%d", NameWithSuffix(cr.ObjectMeta, service), cr.Namespace, port)
+}
```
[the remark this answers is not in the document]
olivergondza: `NameWithSuffix` is now using `TruncateCRName` to achieve that.

olivergondza:
@svghadi, @jgwest, can you please take a look? Thanks!

akhilnittala:
just a doubt, is upgrading to the version which has this PR will be a breaking change ?

akhilnittala:
small doubt, is the path for file containing creds can be in any directory ? or only should be avaialble in hardcoded location ?

akhilnittala:
All Doubts I have:
- How are existing installations upgraded? Do we need backward compatibility or migration handling?
- Should we add a checksum annotation to trigger pod restart on secret change?
- Can we add explicit file existence validation in init scripts?
- Do we need to handle fsGroup for OpenShift restricted SCC?

olivergondza:
@akhilnittala, thanks for your feedback.

> All Doubts I have:
> 
> * How are existing installations upgraded? Do we need backward compatibility or migration handling?

I believe this does not require any migration, because everything this is changing is happening inside pods and is fully deleted/recreated with their every restart (pod volume, container volume mount).

So what I expect to happen is that as soon as new version of operator kicks in, the reconciliation starts producing updated version of manifests (different envvar, + pod volume, + container volume mount) and rolls out the updated pods in a way that is transparent to the users.

> * Should we add a checksum annotation to trigger pod restart on secret change?

This is an interesting idea, but it is out of the scope of this effort. We were not proactively restarting things even before, so this does not violate users' expectations in any way.

If we choose to implement this, we have to be careful so restart of the redis, sentinels and the connected components does not create more damage than good.

> * Can we add explicit file existence validation in init scripts?

Good point. It used to issue a warning, but I have added explicit exit to fail fast and loud.

> * Do we need to handle fsGroup for OpenShift restricted SCC?

Sorry, I cannot tell if this can be a problem. But I would expect that reasonable SCC permits reading secret files mounted to the container. I guess we will learn that no later than in the PR for gitops-operator :D

akhilnittala:
> @akhilnittala, thanks for your feedback.
> 
> > All Doubts I have:
> > 
> > * How are existing installations upgraded? Do we need backward compatibility or migration handling?
> 
> I believe this does not require any migration, because everything this is changing is happening inside pods and is fully deleted/recreated with their every restart (pod volume, container volume mount).
> 
> So what I expect to happen is that as soon as new version of operator kicks in, the reconciliation starts producing updated version of manifests (different envvar, + pod volume, + container volume mount) and rolls out the updated pods in a way that is transparent to the users.
> 
> > * Should we add a checksum annotation to trigger pod restart on secret change?
> 
> This is an interesting idea, but it is out of the scope of this effort. We were not proactively restarting things even before, so this does not violate users' expectations in any way.
> 
> If we choose to implement this, we have to be careful so restart of the redis, sentinels and the connected components does not create more damage than good.
> 
> > * Can we add explicit file existence validation in init scripts?
> 
> Good point. It used to issue a warning, but I have added explicit exit to fail fast and loud.
> 
> > * Do we need to handle fsGroup for OpenShift restricted SCC?
> 
> Sorry, I cannot tell if this can be a problem. But I would expect that reasonable SCC permits reading secret files mounted to the container. I guess we will learn that no later than in the PR for gitops-operator :D



Thanks @olivergondza  for your answers. Sounds good for me.

controllers/argoutil/redis.go
```diff
@@ -0,0 +1,342 @@
...
+	usersACL := fmt.Sprintf("user default on >%s allchannels allkeys allcommands\n", pw)
+
+	return map[string][]byte{
+		"immutable": []byte("true"),
+		// Mapping the legacy key-name, the operator customers can depend on.
+		common.ArgoCDKeyAdminPassword: redisInitialPassword,
+		// Provide ACL file content so redis-server can use file-based ACLs
+		"auth":          redisInitialPassword,
+		"auth_username": []byte("default"),
+		"users.acl":     []byte(usersACL),
```
svghadi, line 72: Is this setting exposed in upstream, or is this something we need downstream? Just trying to better understand the motivation here.

controllers/argoutil/redis.go
svghadi, line 1: What do you think about moving this into a separate `redis` file within the argocd package, similar to how we handle `applicationset`, instead of moving it to `utils`? Since most of the logic is related to the Redis deployment, it may make more sense to keep it in a dedicated file within the same package.

We can continue using capitalized function names for functions that are used across files so there is some sense of separation. It could also help in the future if we decide to refactor this into its own package.

controllers/argoutil/redis.go
olivergondza: This is consumed by redis components to configure PWD for default user. It replaces the old way that required runtime modification of config files - that is quite clunky. It is not needed by argo-cd, just more modern way to configure redis.

controllers/argoutil/redis.go
olivergondza: Yeah, I am not happy about that either, but did not find a better place to put it. The thing is that methods are called from `controllers/argocd`, `controllers/argocdagent` and `controllers/argocdagent/agent`.

olivergondza:
Tests fixed and passing...

olivergondza:
Merge conflict resolved.

@svghadi, @Mangaal, @akhilnittala, can I get github review clearly indicating if you are on board with the change or that there are changes required? Thanks!

olivergondza:
Thanks, @svghadi!

PR #2095, opened 2026-02-24
feat: Edit annotations for the argocd-server service

### LINKED ISSUE #651: Add service annotations
We need to add ability to set annotations to the server object.
I'm trying to use external-dns to point it to internal ip address and have to annotate service manually instead of using operator.

MaayanGalindez:
**What type of PR is this?**

/kind enhancement

this PR adds the ability to control the argocd-server service annotations, most of the users won't benefit from this PR but it'll help for me and my team to have a cleaner argocd setup, we use terraform to create resources in argocd which requires gRPC ingress (already a feature) but in our case we need to add some annotations to the service before using gRPC in our CNI.

I don't think there's any required documentation to be updated or created

Fixes #651

In the argocd manifest:
```
spec:
  server:
    service:
      annotations:
        test.kubernetes.io/test: test
```

and then observe if there's annotations created in the argocd-server service,
Thanks for taking the time to review this :).

## Summary by CodeRabbit

* **New Features**
  * Support for custom annotations on ArgoCD server services: specify annotations in the ArgoCD spec to apply metadata to the generated Service resources (schema and operator now accept an annotations map).

* **Bug Fixes / Behavior**
  * Reconciliation now merges and syncs user-defined service annotations so they are preserved and applied during updates.

* **Tests**
  * Added tests covering annotation reconciliation and OpenShift Auto TLS interaction to ensure annotations are preserved and applied correctly.


api/v1alpha1/argocd_types.go
```diff
@@ -589,6 +589,9 @@ type ArgoCDServerServiceSpec struct {
 	// Type is the ServiceType to use for the Service resource.
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Service Type",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:fieldGroup:Server","urn:alm:descriptor:com.tectonic.ui:text"}
 	Type corev1.ServiceType `json:"type"`
+
+	// Annotations is the map of annotations to apply to the Server Service.
+	Annotations map[string]string `json:"annotations,omitempty"`
```
akhilnittala, line 594: cant we rely on metadata.Annotations ?

MaayanGalindez: I’m thinking about renaming it to `ExtraAnnotations` and changing how it works.

I don’t want to rely on `metadata.annotations` because I’d prefer a more declarative way to add annotations to the server service. This would make it clearer that certain annotations are required for `gRPC` ingress to work with our CNI.

It should also help new team members, so they don’t have to guess that these annotations need to be added manually. In addition, it could make deploying Argo CD across multiple sites and instances easier and more consistent.

Thanks again for the review. I’d really appreciate any feedback. And if I misunderstood something or if you think this feature isn’t necessary, I’m happy to hear your opinion.

akhilnittala: IMHO, this can be achieved using the existing `metadata.annotations` field in Kubernetes. Since `metadata` is already a well-defined and structured section (like `labels`, `annotations`, etc.), it is more consistent and future-proof to rely on it instead of introducing a new custom annotations field in the CRD.

If we start adding separate fields for annotations now, we may eventually need similar fields for labels or other metadata as well. Continuously extending the CRD in this way could unnecessarily increase its length and complexity.

We should also keep in mind that Kubernetes enforces a size limit on annotations. The maximum total size of all annotations (combined keys and values) for a single object is **256 KiB (262,144 bytes)**. If this limit is exceeded, the API server will reject the object. This is an important constraint to consider in the design.

If an additional field is truly necessary, it would be better to follow the same `metadata` structure used by Kubernetes objects. That would keep the design consistent and aligned with Kubernetes conventions.

MaayanGalindez: Thanks for the detailed feedback, I definitely understand your point about keeping things aligned with the existing metadata structure and avoiding unnecessary CRD expansion. That makes sense from a consistency and long term maintenance perspective.

I’m not strongly attached to a specific implementation. The main goal here is to make it possible to add annotations to the Argo CD server Service in a way that can be defined and documented directly in the CR. It would also make this configuration automated for users who deploy the Argo CD CR via Terraform or other Argo CD instances, similar to how we currently handle ingress annotations.

If you feel this feature doesn’t provide enough value or doesn’t align with the project’s direction, I’m completely fine with closing the PR. But if there’s a preferred way to support this kind of feature, I’d really appreciate your guidance:
- How would you like to see this implemented?
- How should conflicts or overrides behave?
- Should there be additional restrictions or validation around which annotations can be set?

I’m happy to rework the proposal to better match the project’s conventions, just let me know what direction you’d prefer :).

PR #2116, opened 2026-03-05
fix: enforce K8s length limit for redis proxy service name (#2071)

### LINKED ISSUE #2071: GenerateAgentPrincipalRedisProxyServiceName can produce names too long
**Describe the bug**

From @coderabbitai in unrelated PR:

[GenerateAgentPrincipalRedisProxyServiceName](https://github.com/argoproj-labs/argocd-operator/blob/47c773497a4e0a04164b64a42e70beeb61785251/controllers/argoutil/resource.go#L204) does not use truncated CR name.

This function constructs a Kubernetes service name from the raw `crName`, which could exceed the 63-character limit. The service name format `{crName}-agent-principal-redisproxy` (with a 26-character suffix) will exceed 63 characters for any CR name longer than approximately 37 characters. Other service naming helpers in the codebase (e.g., `GetSecretNameWithSuffix` in secret.go, `nameWithSuffix` in util.go) consistently use `GetTruncatedCRName` before appending suffixes. Consider updating this function to use truncated CR names for consistency.

**To Reproduce**
N/A

**Expected behavior**

The resulting name would be shorter than 64 characters.

**Screenshots**

**Additional context**

tzprograms:
**What type of PR is this?**

/kind bug

The problem was that the `GenerateAgentPrincipalRedisProxyServiceName` function was appending a 27 character suffix to the custom resource name. For longer names, this breaks the 63 character limit for Kubernetes services.

The original ticket stated the suffix was 26 characters and suggested using `GetTruncatedCRName` or `TruncateCRName` to fix it. However, while running the unit tests, I noticed the suffix is actually 27 characters long. Since those standard helpers leave a 37 character base, the total length ends up being 64 characters, which still fails validation. Also, `GetTruncatedCRName` requires passing the entire ArgoCD object, while our function only accepts a string.

To fix this cleanly without refactoring the function signature everywhere, I just used `TruncateWithHash` directly and set the max base length to 36. This perfectly accommodates the 27 character suffix to hit exactly 63 characters.

Fixes #2071

As the function was missing any existing unit tests , I added a new unit test to explicitly enforce the service name boundary and prevent any future regressions. You can verify the fix by running:

`go test -v ./controllers/argoutil -run TestGenerateAgentPrincipalRedisProxyServiceName`

## Summary by CodeRabbit

* **Bug Fixes**
  * Service naming now automatically enforces Kubernetes length constraints (63-character maximum), preventing service creation failures.

* **Tests**
  * Added validation tests for service naming behavior and length constraint compliance.


controllers/argoutil/resource.go
```diff
@@ -202,7 +202,8 @@ func LogResourceAction(log logr.Logger, action string, object metav1.Object, exp
 }
 
 func GenerateAgentPrincipalRedisProxyServiceName(crName string) string {
-	return fmt.Sprintf("%s-agent-%s", crName, "principal-redisproxy")
+	truncatedCRName := TruncateWithHash(crName, 36)
```
akhilnittala: can we make this something meaningful name ?

tzprograms: Hi @akhilnittala  , all other functions like `NewSecretWithSuffix` , `GetSecretNameWithSuffix` use this exact variable name (`truncatedCRName`) , hence I used the same for consistency across the codebase .

olivergondza (approved):

olivergondza:
@jgwest, @jannfis, I cannot tell if eventual problems would manifest by test failures in this repo, or if some other components need to be updated as well. PTAL, thx

jparsai:
Verified changes locally, principal and agent work fine, it LGTM.  but should we have an E2E to validate it?

It seems we have similar implementation in principal and agent services, so we would face this issue for those as well. Probably we can raise another issue and fix in new PR.

olivergondza:
@jparsai, thanks for taking a look from the agent PoW. There are several E2E tests that are using this to set things up already, so we do know this does not break anything super obvious.

@tzprograms, please take a look at resolving the conflicts.

svghadi (approved):
LGTM. 

Thanks @olivergondza, @jparsai @akhilnittala for the reviews.

PR #2149, opened 2026-04-07
feat: declarative management of Argo CD webhook secrets (GitHub)

aali309:
**What type of PR is this?**

 /kind enhancement 

This PR introduces a declarative mechanism within the ArgoCD CR for managing webhook secrets (tokens) used to secure incoming Git provider webhook events. This feature removes the need to manually edit operator-managed secrets and aligns webhook secret management with GitOps principles.

**Scope**: GitHub only in this PR.

Fixes #?

1. Deploy/Run the operator (i.e `make install + make run`)
2. Create secret `kubectl create secret generic github-webhook-credentials --from-literal=secret=demo-webhook-secret -n argocd`
3. Patch the ArgoCD CR
```
kubectl patch argocd argocd -n argocd --type merge -p '
spec:
 webhookSecrets:
   github:
     secretRef:
       name: github-webhook-credentials
       key: secret
'
```
you should not get any unknown field "spec.webhookSecrets" warnings
 
4. confirm spec is stored
`kubectl get argocd argocd -n argocd -o jsonpath='{.spec.webhookSecrets}' | jq .`
you should get
```
{
  "github": {
    "secretRef": {
      "key": "secret",
      "name": "github-webhook-credentials"
    }
  }
}
```
5.  Confirm sync into argocd-secret
- `kubectl get secret github-webhook-credentials -n argocd -o jsonpath='{.data.secret}' | base64 -d && echo`
- `kubectl get secret argocd-secret -n argocd -o jsonpath='{.data.webhook\.github\.secret}' | base64 -d && echo`
Both lines should show the same value (e.g. `demo-webhook-secret`).

To confirm that the CRD knows about newl `webhookSecrets` by running 
- `kubectl explain argocd.spec.webhookSecrets --api-version=argoproj.io/v1beta1`
you should get something similat to
```
GROUP:      argoproj.io
KIND:       ArgoCD
VERSION:    v1beta1

FIELD: webhookSecrets <Object>


DESCRIPTION:
    WebhookSecrets references Kubernetes Secrets that supply webhook credentials
    per provider.
    The operator syncs values into argocd-secret under the keys Argo CD expects.
    
FIELDS:
  github	<Object>
    GitHub: Secret key reference for the webhook secret used to verify incoming
    webhook requests.
```

## Summary by CodeRabbit

* **New Features**
  * Optional spec.webhookSecrets.github.secretRef added to Argo CD CR; operator now syncs the referenced GitHub webhook secret into the Argo CD secret.

* **Tests**
  * New unit tests for webhook secret sync behavior.
  * New parallel E2E test validating GitHub webhook secret synchronization.

* **Chores**
  * CRD and bundle/CSV manifests updated (schema + timestamps).


controllers/argocd/secret.go
```diff
@@ -393,6 +398,49 @@ func (r *ReconcileArgoCD) reconcileExistingArgoSecret(cr *argoproj.ArgoCD, secre
...
+	}
+	ref := cr.Spec.WebhookSecrets.GitHub.SecretRef
+	if ref.Name == "" {
+		log.Info("skipping GitHub webhook secret sync: secretRef.name is empty")
+		return false
+	}
+	ns := ref.Namespace
+	if ns == "" {
+		ns = cr.Namespace
+	}
```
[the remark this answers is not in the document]
svghadi: Let's address this. @aali309

api/v1alpha1/argocd_types.go
```diff
@@ -898,6 +902,33 @@ type ManagedNamespaces struct {
...
+}
+
+// WebhookSecretKeySelector references one key within a Secret (mirrors SecretKeySelector-style semantics).
+// +k8s:openapi-gen=true
+type WebhookSecretKeySelector struct {
+	// Name of the Secret.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+	// Namespace of the Secret. When unset, the ArgoCD CR namespace is used.
+	Namespace string `json:"namespace,omitempty"`
```
jgwest: Strongly recommend removing `namespace` field, as ability to allowing access to Secrets across namespaces is a very easy way to break security boundaries on a cluster.

(If we ever get a request to add it from a user we can reconsider how to safely add it, but the default should be that it not be present IMHO)

api/v1beta1/argocd_types.go
```diff
@@ -1074,6 +1078,33 @@ type ManagedNamespaces struct {
...
+	SecretRef *WebhookSecretKeySelector `json:"secretRef,omitempty"`
+}
+
+// WebhookSecretKeySelector references one key within a Secret (mirrors SecretKeySelector-style semantics).
+// +k8s:openapi-gen=true
+type WebhookSecretKeySelector struct {
+	// Name of the Secret.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+	// Namespace of the Secret. When unset, the ArgoCD CR namespace is used.
```
jgwest: Ah, I left a comment on the `v1alpha1` version of the API above, but I have the same comment as coderabbit: let's not allow cross-namespace Secret access.

controllers/argocd/secret.go
```diff
@@ -393,6 +402,46 @@ func (r *ReconcileArgoCD) reconcileExistingArgoSecret(cr *argoproj.ArgoCD, secre
 	return nil
 }
 
+// applyGitHubWebhookSecretFromRef copies spec.webhookSecrets.github.secretRef into argocdSecret.Data[webhook.github.secret].
+// Returns true if argocdSecret was modified. Missing referenced Secret (NotFound) or missing key logs a warning and returns (false, nil).
+// Other Secret read errors are returned so the reconciler can retry.
+func applyGitHubWebhookSecretFromRef(ctx context.Context, c client.Client, cr *argoproj.ArgoCD, argocdSecret *corev1.Secret) (bool, error) {
+	if cr.Spec.WebhookSecrets == nil || cr.Spec.WebhookSecrets.GitHub == nil || cr.Spec.WebhookSecrets.GitHub.SecretRef == nil {
+		return false, nil
```
olivergondza: It is desirable to remove `argocdSecret.Data[common.ArgoCDKeyGitHubWebhookSecret]` when the declaration is removed, I think.

olivergondza: Applies to several other `return false, nil` clauses later as well

olivergondza (approved):
The rest looks good to me. I have left one comment to discuss.

controllers/argocd/secret.go
aali309: Good point! I agree this should be removed.

aali309: So instead of just `returning false, nil` we delete any existing value from `argocdSecret.Data` and report changed=true when a stale value was present.  And still return an error for real client read failures so reconcile can retry?

controllers/argocd/secret.go
```diff
@@ -393,6 +402,58 @@ func (r *ReconcileArgoCD) reconcileExistingArgoSecret(cr *argoproj.ArgoCD, secre
...
+		log.Info(fmt.Sprintf("warning: key %q missing or empty in Secret %s/%s, skipping GitHub webhook secret sync", ref.Key, ns, ref.Name))
+		return deleteGitHubWebhookSecret(argocdSecret), nil
+	}
+	val := make([]byte, len(raw))
+	copy(val, raw)
+
+	if argocdSecret.Data == nil {
+		argocdSecret.Data = make(map[string][]byte)
+	}
+	if existing, ok := argocdSecret.Data[common.ArgoCDKeyGitHubWebhookSecret]; ok && bytes.Equal(existing, val) {
```
olivergondza: IIUC, can be shortened as:

`if bytes.Equal(argocdSecret.Data[common.ArgoCDKeyGitHubWebhookSecret], val) {`

The missing key would return `[]byte{}` that cannot be equal to a valid new value.

tests/ginkgo/parallel/1-125_validate_github_webhook_secret_sync_test.go
```diff
@@ -0,0 +1,104 @@
...
+		var (
+			k8sClient client.Client
+			ctx       context.Context
+		)
+
+		BeforeEach(func() {
+			fixture.EnsureParallelCleanSlate()
+
+			k8sClient, _ = fixtureUtils.GetE2ETestKubeClient()
+			ctx = context.Background()
```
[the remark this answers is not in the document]
olivergondza: No errors are ignored here.

olivergondza (approved):
LGTM. Added one suggestion.

PR #2158, opened 2026-04-09
chore(tests): Align 1-120_repo_server_system_ca_trust.go with downstream

olivergondza:
**What type of PR is this?**

/kind chore

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability by centralizing resource cleanup and ensuring consistent teardown after each test.
  * Enhanced failure diagnostics by emitting debug output on test failures.
  * Strengthened certificate bundle detection with automatic fallback handling for broader system compatibility.

### DIFF
--- a/tests/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
+++ b/tests/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
@@ -55,11 +55,6 @@ import (
 )
 
 var (
-	// The differences between the upstream image using Ubuntu, and the downstream one using rhel.
-	image        = "" // argocd-operator default
-	imageVersion = "" // argocd-operator default
-	caBundlePath = "/etc/ssl/certs/ca-certificates.crt"
-
 	trustedHelmAppSource = &appv1alpha1.ApplicationSource{
 		RepoURL:        "https://stefanprodan.github.io/podinfo",
 		Chart:          "podinfo",
@@ -74,14 +69,17 @@ var (
 		Helm:           &appv1alpha1.ApplicationSourceHelm{Values: "service:\n          type: ClusterIP"},
 	}
 
-	k8sClient client.Client
-	ctx       context.Context
+	k8sClient    client.Client
+	ctx          context.Context
+	ns           *corev1.Namespace
+	cleanupNs    func()
+	image        string
+	imageVersion string
 
 	clusterSupportsClusterTrustBundles bool
 )
 
 var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
-
 	Context("1-120_repo_server_system_ca_trust", func() {
 		BeforeEach(func() {
 			fixture.EnsureSequentialCleanSlate()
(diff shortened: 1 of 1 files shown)


olivergondza:
Aligns with https://github.com/redhat-developer/gitops-operator/pull/1123

jannfis (approved):

tests/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
```diff
@@ -794,7 +786,9 @@ func getTrustedCertCount(rsPod *corev1.Pod) int {
 	command := []string{
 		"kubectl", "-n", rsPod.Namespace, "exec",
 		"-c", "argocd-repo-server", rsPod.Name, "--",
-		"cat", caBundlePath,
+		"bash", "-c",
+		// Ubuntu or RHEL location
+		"cat /etc/ssl/certs/ca-certificates.crt || cat /etc/ssl/certs/ca-bundle.crt",
```
anandf, line 791: Since the parent directory is the same `/etc/ssl/certs/`, can we not use a glob/wildcard pattern here. Something like `cat '/etc/ssl/certs/ca-*.crt'` and also do we need that `bash -c` prefix ? I was thinking something like below can work.
```
"kubectl" "-n", rsPod.Namespace, "exec",
		"-c", "argocd-repo-server", rsPod.Name, "--",
        "cat", '/etc/ssl/certs/ca-*.crt',
```

tests/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
```diff
@@ -852,6 +846,7 @@ func findRunningRepoServerPod(k8sClient client.Client, ns *corev1.Namespace) *co
 			pod = runningPods[0]
 			return nil
 		}
+
```
anandf, line 849: nitpick: This empty line can be removed IMO.

tests/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
olivergondza: Yeah, I would love to make it simpler, but in this case, it would not work:

- `cat` alone would not resolve the `*` - bash is needed.
- Using `cat` with 2 arguments (from glob, or specified explicitly), would cause the command to error with message that one of the files do not exist.
 
Here we know exactly one exists---based on the distro used---hence the use of `||`.

anandf: if `*` does not work , how about using a specific pattern `/etc/ssl/certs/ca-(bundle|certificates).crt`. If we are sure that only one of the two files exist, then it is expected to work. I haven't tried it, just throwing some ideas here to make it simpler and limiting the change.

olivergondza: Yeah, even `/etc/ssl/certs/ca-{bundle,certificates}.crt` needs bash, and there is still the missing file problem I have elaborated on earlier.

To clarify, this line does not need to differ between the operators if written as is. It takes advantage of the short-circuiting nature of `||`, if first file exists, it uses it and completes, if not, it prints the other file.

PR #2162, opened 2026-04-14
restore serving-cert annotation when tls termination is reencrypt that was stripped in older versions

alkakumari016:
**What type of PR is this?**
/kind bug 

Upgrading the OpenShift GitOps Operator specifically from v1.12.3 to v1.12.4 causes the `service.beta.openshift.io/serving-cert-secret-name` annotation to be stripped from the Argo CD server Service.

This prevents the Service-CA ClusterOperator from rotating or managing the service certificate. While the console may work initially, it will eventually return an HTTP 503 Service Unavailable error once the certificate expires or the secret is modified. Because this state is persisted in the Service object, the issue remains even if the operator is subsequently upgraded to later versions (e.g., v1.19).
This happens because in v1.12.4, default TLS termination value was set to 'passthrough' that stripped this annotation. The default value was changed back to 'reencrypt' in future versions but the stripped annotation was not added back that resulted in this regression. In this PR it is checked whether the TLS termination is reencrypt or nil (default value is reencrypt). If so then it is further checking if the cert is created by service CA. If yes then we are adding back the annotation.

Fixes #?
[GITOPS-8784
](https://redhat.atlassian.net/browse/GITOPS-8784)

## Release Notes

* **New Features**
  * Improved TLS secret validation to distinguish between system-managed and user-managed certificates

* **Bug Fixes**
  * Enhanced serving certificate annotation handling with automatic restoration when accidentally deleted

* **Tests**
  * Added E2E tests validating certificate annotation restoration across TLS configuration transitions
  * Extended unit tests for annotation management with existing TLS secrets


controllers/argocd/route.go
```diff
@@ -248,30 +248,32 @@ func (r *ReconcileArgoCD) reconcileServerRoute(cr *argoproj.ArgoCD) error {
...
+// isServingCertSecretForService reports whether secret was issued by the OpenShift
+// service certificate controller for the given Service name (originating-service-name
+// annotation plus Service ownerReference).
+func isServingCertSecretForService(secret *corev1.Secret, serviceName string) bool {
+	if secret == nil || secret.Annotations == nil {
+		return false
 	}
-
-	if !serviceAnnFound {
+	value, ok := secret.Annotations["service.beta.openshift.io/originating-service-name"]
```
akhilnittala: we can have this as a constant

akhilnittala (approved):

alkakumari016:
@coderabbitai full review

alkakumari016:
@coderabbitai full review

tests/ginkgo/parallel/cert/cert.go
```diff
@@ -0,0 +1,50 @@
...
+			CommonName:   "localhost",
+		},
+		NotBefore: time.Now(),
+		NotAfter:  time.Now().Add(time.Hour * 24),
+
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+		IPAddresses:           []net.IP{net.ParseIP("127.0.0.1")},
+	}
```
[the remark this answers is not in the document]
alkakumari016: This is just a sample cert used for just checking an annotation. It does not have an actual purpose.

tests/ginkgo/fixture/certificate/fixture.go
svghadi, line 1: Can we move this to fixtures directory?
https://github.com/alkakumari016/argocd-operator/tree/9bc3a28ec643411ec6af9cb071e623cb3a0105ca/tests/ginkgo/fixture

svghadi: See existing files for naming convention.

svghadi (changes requested):
Left a comment regarding test. Rest all looks good. Thanks.

svghadi (approved):
LGTM

PR #2163, opened 2026-04-14
feat: replace non-expiring Dex SA token with short-lived TokenRequest-based token

Rizwana777:
**What type of PR is this?**

The openshift-gitops-argocd-dex-server service account previously auto-generated a non-expiring kubernetes.io/service-account-token Secret, which violates OpenShift platform guidelines that no longer auto-generate such tokens. This PR replaces that approach with a short-lived token obtained via the Kubernetes TokenRequest API, stored in a dedicated Opaque Secret, and proactively renewed before expiry. Legacy non-expiring token Secrets are cleaned up automatically on reconcile.

Token lifetime is currently hardcoded to 1 hour (ArgoCDDexServerTokenExpirySecs = 3600) with renewal triggered at the 20-minute mark (1/3 of lifetime). Should we keep 1 hour, or would a different value be preferred? 

PR contains unit and e2e test to verify above

JIRA - https://redhat.atlassian.net/browse/GITOPS-9427

from argocd-operator 
run make install run

Create an ArgoCD instance with Dex and OpenShift OAuth enabled
   kind: ArgoCD
   metadata:
     name: example-argocd
     namespace: <your-namespace>
   spec:
     sso:
       provider: dex
       dex:
         openShiftOAuth: true

Verify no non-expiring SA token Secret was created:
   kubectl get secret example-argocd-argocd-dex-server-token -n <ns> -o jsonpath='{.data.expiry}' | base64 -d

Expected: only example-argocd-argocd-dex-server-token (Opaque type), no kubernetes.io/service-account-token type.

Verify the token Secret has valid token and expiry keys:
   kubectl get secret example-argocd-argocd-dex-server-token -n <ns> -o jsonpath='{.data.expiry}' | base64 -d

Expected: a RFC3339 timestamp ~1 hour in the future.

Verify argocd-cm uses the placeholder, not the raw token:

   kubectl get cm argocd-cm -n <ns> -o jsonpath='{.data.dex\.config}' | grep clientSecret
Expected: clientSecret: $oidc.dex.clientSecret

Verify argocd-secret has the token set
   kubectl get secret argocd-secret -n <ns> -o jsonpath='{.data.oidc\.dex\.clientSecret}' | base64 -d | cut -c1-20

Expected: a JWT token (starts with eyJ), matching the token key in the dedicated Secret.

To check dex sso login 

 kubectl get route example-argocd-server -n <ns> -o jsonpath='{.spec.host}'

Open https://<route-host> in a browser.

On the login page, click "Log in via OpenShift"

Complete the OpenShift login. If you land on the ArgoCD dashboard, Dex successfully used the short-lived token to authenticate.

## Summary by CodeRabbit

* **New Features**
  * Dex OAuth tokens are now short‑lived, stored with expiry, proactively renewed, and scheduled for timely refresh; managed token secrets use deterministic names.

* **Bug Fixes**
  * Cleanup of legacy non‑expiring service‑account token secrets and prevention of legacy token leakage into Argo CD secrets.
  * Improved error propagation during secret reconciliation.

* **Tests**
  * Added/updated tests covering renewal logic, caching, expiry parsing, legacy cleanup, and token secret behavior.

* **Chores**
  * Added configuration defaults and test helpers to support token expiry/renewal.

### DIFF
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -118,6 +118,12 @@ const (
 	// ArgoCDDefaultDexServiceAccountName is the default Service Account name for the Dex server.
 	ArgoCDDefaultDexServiceAccountName = "argocd-dex-server"
 
+	// ArgoCDDexServerTokenExpirySecs is the Dex SA token lifetime in seconds (1 hour).
+	ArgoCDDexServerTokenExpirySecs = int64(3600)
+
+	// ArgoCDDexServerTokenRenewalThresholdFraction is N in "renew when < 1/N lifetime remains".
+	ArgoCDDexServerTokenRenewalThresholdFraction = 3
+
 	// ArgoCDDefaultDexVersion is the Dex container image tag to use when not specified.
 	ArgoCDDefaultDexVersion = "sha256:b08a58c9731c693b8db02154d7afda798e1888dc76db30d34c4a0d0b8a26d913" // v2.43.0
 
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -121,8 +121,10 @@ const (
 	// ArgoCDDexServerTokenExpirySecs is the Dex SA token lifetime in seconds (1 hour).
 	ArgoCDDexServerTokenExpirySecs = int64(3600)
 
-	// ArgoCDDexServerTokenRenewalThresholdFraction is N in "renew when < 1/N lifetime remains".
-	ArgoCDDexServerTokenRenewalThresholdFraction = 3
+	// ArgoCDDexServerTokenRenewalThresholdPercent (1-99): renew the Dex token when less than this percent
+	// of ArgoCDDexServerTokenExpirySecs remains (default 33 is approximately equivalent to the last third
+	// of a 1h nominal lifetime).
+	ArgoCDDexServerTokenRenewalThresholdPercent int64 = 33
 
 	// ArgoCDDefaultDexVersion is the Dex container image tag to use when not specified.
 	ArgoCDDefaultDexVersion = "sha256:b08a58c9731c693b8db02154d7afda798e1888dc76db30d34c4a0d0b8a26d913" // v2.43.0
--- a/controllers/argocd/dex.go
+++ b/controllers/argocd/dex.go
@@ -2,6 +2,7 @@ package argocd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"strings"
@@ -46,18 +47,33 @@ func getDexServerTokenSecretName(cr *argoproj.ArgoCD) string {
 	return argoutil.GetSecretNameWithSuffix(cr, common.ArgoCDDefaultDexServiceAccountName+"-token")
 }
 
+// dexServerTokenRenewalThreshold is how much nominal lifetime may remain before we treat the Dex token
+// as due for renewal (ExpirySecs * ArgoCDDexServerTokenRenewalThresholdPercent / 100).
+func dexServerTokenRenewalThreshold() time.Duration {
+	return time.Duration(common.ArgoCDDexServerTokenExpirySecs*common.ArgoCDDexServerTokenRenewalThresholdPercent/100) * time.Second
+}
+
 // needsDexTokenRenewal returns true when the token is missing, unparseable, or within the renewal window.
 func needsDexTokenRenewal(secret *corev1.Secret) bool {
+	if secret == nil || secret.Data == nil {
+		return true
+	}
+	tokenBytes, ok := secret.Data["token"]
+	if !ok || len(tokenBytes) == 0 {
+		return true
+	}
 	expiryBytes, ok := secret.Data["expiry"]
 	if !ok {
 		return true
 	}
 	expiry, err := time.Parse(time.RFC3339, string(expiryBytes))
 	if err != nil {
+		log.Error(err, "dex token secret has unparseable expiry, renewal needed",
+			"secret", fmt.Sprintf("%s/%s", secret.Namespace, secret.Name),
+			"expiry", string(expiryBytes))
 		return true
 	}
-	renewThreshold := time.Duration(common.ArgoCDDexServerTokenExpirySecs/common.ArgoCDDexServerTokenRenewalThresholdFraction) * time.Second
-	return time.Until(expiry) < renewThreshold
+	return time.Until(expiry) < dexServerTokenRenewalThreshold()
 }
 
 // getDexOAuthClientSecret returns a time-limited Dex OAuth client token via the TokenRequest API.
@@ -80,9 +96,12 @@ func (r *ReconcileArgoCD) getDexOAuthClientSecret(cr *argoproj.ArgoCD) (*string,
 		token := string(tokenSecret.Data["token"])
 		// Schedule the next reconcile to run just before the renewal threshold so
 		// the token is proactively renewed without waiting for an external event.
-		if expiry, parseErr := time.Parse(time.RFC3339, string(tokenSecret.Data["expiry"])); parseErr == nil {
-			renewThreshold := time.Duration(common.ArgoCDDexServerTokenExpirySecs/common.ArgoCDDexServerTokenRenewalThresholdFraction) * time.Second
-			if d := time.Until(expiry) - renewThreshold; d > 0 {
+		expiry, parseErr := time.Parse(time.RFC3339, string(tokenSecret.Data["expiry"]))
+		if parseErr != nil {
+			log.Error(parseErr, "dex token secret expiry unparseable when scheduling requeue, returning cached token",
+				"secret", fmt.Sprintf("%s/%s", tokenSecret.Namespace, tokenSecret.Name))
+		} else {
+			if d := time.Until(expiry) - dexServerTokenRenewalThreshold(); d > 0 {
 				r.dexTokenRequeueAfter.Store(cr.Namespace, d)
 			}
 		}
@@ -132,8 +151,7 @@ func (r *ReconcileArgoCD) getDexOAuthClientSecret(cr *argoproj.ArgoCD) (*string,
 	}
 
 	// Schedule the next reconcile just before the renewal threshold.
-	renewThreshold := time.Duration(common.ArgoCDDexServerTokenExpirySecs/common.ArgoCDDexServerTokenRenewalThresholdFraction) * time.Second
-	if d := time.Until(tokenRequest.Status.ExpirationTimestamp.Time) - renewThreshold; d > 0 {
+	if d := time.Until(tokenRequest.Status.ExpirationTimestamp.Time) - dexServerTokenRenewalThreshold(); d > 0 {
 		r.dexTokenRequeueAfter.Store(cr.Namespace, d)
 	}
 
@@ -154,6 +172,7 @@ func (r *ReconcileArgoCD) reconcileDexLegacySATokenSecrets(cr *argoproj.ArgoCD) 
 	); err != nil {
 		return err
 	}
+	var deleteErrs []error
 	for i := range secretList.Items {
 		s := &secretList.Items[i]
 		if s.Type != corev1.SecretTypeServiceAccountToken {
@@ -162,24 +181,26 @@ func (r *ReconcileArgoCD) reconcileDexLegacySATokenSecrets(cr *argoproj.ArgoCD) 
 		if s.Annotations[corev1.ServiceAccountNameKey] != dexSAName {
 			continue
 		}
-		if !strings.HasPrefix(s.Name, "argocd-dex-server-token-") {
+
+		if !strings.HasPrefix(s.Name, dexSAName+"-token-") {
 			continue
 		}
 		argoutil.LogResourceDeletion(log, s, "removing legacy Dex service account token secret")
 		if err := r.Delete(context.TODO(), s); err != nil && !apierrors.IsNotFound(err) {
-			return err
+			deleteErrs = append(deleteErrs, fmt.Errorf("delete legacy dex token secret %s/%s: %w", s.Namespace, s.Name, err))
 		}
 	}
 	sa := newServiceAccountWithName(common.ArgoCDDefaultDexServiceAccountName, cr)
 	if err := argoutil.FetchObject(r.Client, cr.Namespace, sa.Name, sa); err != nil {
 		if apierrors.IsNotFound(err) {
-			return nil
+			return errors.Join(deleteErrs...)
 		}
-		return err
+		return errors.Join(append([]error{err}, deleteErrs...)...)
 	}
 	var filtered []corev1.ObjectReference
 	for _, ref := range sa.Secrets {
-		if strings.Contains(ref.Name, "dex-server-token") {
+		// Legacy auto token refs only (matches delete loop).
+		if strings.HasPrefix(ref.Name, dexSAName+"-token-") {
 			continue
 		}
 		filtered = append(filtered, ref)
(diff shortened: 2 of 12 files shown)


tests/ginkgo/parallel/1-095_validate_dex_clientsecret_test.go
```diff
@@ -76,46 +78,81 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
...
-			serviceAccount := &corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Name: "example-argocd-argocd-dex-server", Namespace: ns.Name}}
+			dexSAName := "example-argocd-argocd-dex-server"
+			serviceAccount := &corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Name: dexSAName, Namespace: ns.Name}}
 			Eventually(serviceAccount).Should(k8sFixture.ExistByName())
 
-			argocdCM := &corev1.ConfigMap{
-				ObjectMeta: metav1.ObjectMeta{Name: "argocd-cm", Namespace: ns.Name},
+			By("verifying no non-expiring kubernetes.io/service-account-token Secret exists for the Dex SA")
+			secretList := &corev1.SecretList{}
+			Expect(k8sClient.List(ctx, secretList, client.InNamespace(ns.Name))).To(Succeed())
```
jgwest: Can use `Consistently`, here, rather than `Eventually`, as we want to verify that at no point (in a reasonable amount of time) does the legacy token get created. WDYT?

common/defaults.go
```diff
@@ -118,6 +118,12 @@ const (
 	// ArgoCDDefaultDexServiceAccountName is the default Service Account name for the Dex server.
 	ArgoCDDefaultDexServiceAccountName = "argocd-dex-server"
 
+	// ArgoCDDexServerTokenExpirySecs is the Dex SA token lifetime in seconds (1 hour).
+	ArgoCDDexServerTokenExpirySecs = int64(3600)
+
+	// ArgoCDDexServerTokenRenewalThresholdFraction is N in "renew when < 1/N lifetime remains".
+	ArgoCDDexServerTokenRenewalThresholdFraction = 3
```
olivergondza: I am thinking if representing this as a fraction is a good idea. We cannot represent values between 50-100% of the interval.

controllers/argocd/dex.go
```diff
@@ -42,73 +41,159 @@ func UseDex(cr *argoproj.ArgoCD) bool {
...
+
+// needsDexTokenRenewal returns true when the token is missing, unparseable, or within the renewal window.
+func needsDexTokenRenewal(secret *corev1.Secret) bool {
+	expiryBytes, ok := secret.Data["expiry"]
+	if !ok {
+		return true
+	}
+	expiry, err := time.Parse(time.RFC3339, string(expiryBytes))
+	if err != nil {
+		return true
```
olivergondza, line 73: Let's log the error, as this likely describes an operator bug.

tests/ginkgo/fixture/secret/fixture.go
```diff
@@ -66,7 +66,18 @@ func HaveNonEmptyKeyValue(key string) matcher.GomegaMatcher {
...
+func HaveNonEmptyKey(key string) matcher.GomegaMatcher {
+	return fetchSecret(func(sec *corev1.Secret) bool {
+		a, exists := sec.Data[key]
+		if !exists {
+			GinkgoWriter.Println("HaveNonEmptyKey - Key:", key, "does not exist")
+			return false
+		}
+		return len(a) > 0
+	})
 }
```
jgwest, line 85: What's the difference between this function and the one above it?

tests/ginkgo/parallel/1-095_validate_dex_clientsecret_test.go
```diff
@@ -49,7 +51,7 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
 			ctx = context.Background()
 		})
 
-		It("verifies that Dex serviceaccount token secret is not leaked, and is correctly set in Argo CD argocd-secret Secret", func() {
+		It("verifies that the Dex client secret is sourced from a short-lived TokenRequest token and is correctly set in argocd-secret", func() {
```
jgwest, line 77: One of the things the PR does is cleans up after old non-expiring dex tokens. Could we validate this functionality as part of the E2E tests, here? (For example, creating a non-expiring dex token and then verifying that the operator successfully cleans it up)

tests/ginkgo/parallel/1-095_validate_dex_clientsecret_test.go
```diff
@@ -76,46 +78,81 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
...
-			secretReferencedFromServiceAccount := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: secretName, Namespace: ns.Name}}
-			Eventually(secretReferencedFromServiceAccount).Should(k8sFixture.ExistByName())
-			tokenFromSASecret := secretReferencedFromServiceAccount.Data["token"]
-			Expect(tokenFromSASecret).ToNot(BeEmpty())
+				return strings.Contains(argocdCM.Data["dex.config"], "clientSecret: $oidc.dex.clientSecret")
+			}, "2m", "5s").Should(BeTrue(), "'$oidc.dex.clientSecret' should be set. Any other value implies that the client secret is exposed via ConfigMap")
+
+			By("verifying the Dex SA has no non-expiring kubernetes.io/service-account-token Secrets in its .secrets list")
+			// The operator must clean up legacy SA token Secrets and must not auto-generate new ones.
+			Eventually(func() bool {
```
jgwest: `Consistently` rather than `Eventually`, if I am not mistaken. Any time we want to verify the absence of something, it's a good choice to use `Consistently` (allowing use to avoid cases where tests intermittently fail because controller was faster/slower than usual)

jgwest (changes requested):
Looks great, just some minor tweaks requested.

tests/ginkgo/fixture/secret/fixture.go
```diff
@@ -66,7 +66,18 @@ func HaveNonEmptyKeyValue(key string) matcher.GomegaMatcher {
...
 
+// HaveNonEmptyKey returns true if Secret has the given key and its value is non-empty.
+func HaveNonEmptyKey(key string) matcher.GomegaMatcher {
+	return fetchSecret(func(sec *corev1.Secret) bool {
+		a, exists := sec.Data[key]
+		if !exists {
+			GinkgoWriter.Println("HaveNonEmptyKey - Key:", key, "does not exist")
+			return false
+		}
+		return len(a) > 0
```
olivergondza: Not logging when `len(a)==0` can be confusing for diagnostics

olivergondza (commented):
Added some review feedback. PTAL

controllers/argocd/dex.go
```diff
@@ -42,73 +41,159 @@ func UseDex(cr *argoproj.ArgoCD) bool {
...
 
-// getDexOAuthClientSecret will return the OAuth client secret for the given ArgoCD.
+// getDexServerTokenSecretName returns the name of the Secret that stores the Dex OAuth client token.
+func getDexServerTokenSecretName(cr *argoproj.ArgoCD) string {
+	return argoutil.GetSecretNameWithSuffix(cr, common.ArgoCDDefaultDexServiceAccountName+"-token")
+}
+
+// needsDexTokenRenewal returns true when the token is missing, unparseable, or within the renewal window.
+func needsDexTokenRenewal(secret *corev1.Secret) bool {
+	expiryBytes, ok := secret.Data["expiry"]
```
akhilnittala, line 64: better to have a nil check here to avoid panic.

controllers/argocd/dex.go
```diff
@@ -42,73 +41,159 @@ func UseDex(cr *argoproj.ArgoCD) bool {
...
+	}
+	secretExists := fetchErr == nil
+
+	// Return the cached token if it is still valid.
+	if secretExists && !needsDexTokenRenewal(tokenSecret) {
+		token := string(tokenSecret.Data["token"])
+		// Schedule the next reconcile to run just before the renewal threshold so
+		// the token is proactively renewed without waiting for an external event.
+		if expiry, parseErr := time.Parse(time.RFC3339, string(tokenSecret.Data["expiry"])); parseErr == nil {
+			renewThreshold := time.Duration(common.ArgoCDDexServerTokenExpirySecs/common.ArgoCDDexServerTokenRenewalThresholdFraction) * time.Second
```
akhilnittala: looks like same code repititive 3 times, can we make it as a function ?

controllers/argocd/dex.go
```diff
@@ -42,73 +41,159 @@ func UseDex(cr *argoproj.ArgoCD) bool {
...
+	expiryBytes, ok := secret.Data["expiry"]
+	if !ok {
+		return true
+	}
+	expiry, err := time.Parse(time.RFC3339, string(expiryBytes))
+	if err != nil {
+		return true
+	}
+	renewThreshold := time.Duration(common.ArgoCDDexServerTokenExpirySecs/common.ArgoCDDexServerTokenRenewalThresholdFraction) * time.Second
+	return time.Until(expiry) < renewThreshold
```
akhilnittala: Clock skew risk (minor but real)
If:
controller clock > API server clock

controllers/argocd/dex.go
```diff
@@ -42,73 +41,159 @@ func UseDex(cr *argoproj.ArgoCD) bool {
...
+		return nil, fetchErr
+	}
+	secretExists := fetchErr == nil
+
+	// Return the cached token if it is still valid.
+	if secretExists && !needsDexTokenRenewal(tokenSecret) {
+		token := string(tokenSecret.Data["token"])
+		// Schedule the next reconcile to run just before the renewal threshold so
+		// the token is proactively renewed without waiting for an external event.
+		if expiry, parseErr := time.Parse(time.RFC3339, string(tokenSecret.Data["expiry"])); parseErr == nil {
```
akhilnittala: what if parsing fails, parseErr != nil, still we are using token I guess

controllers/argocd/dex.go
akhilnittala, line 64: I guess we need to also check for token key existence also else we will return false
needsDexTokenRenewal() → returns false (because expiry is fine)
token := string(tokenSecret.Data["token"])

Now:

If key is missing → returns ""
If empty → still ""

So you return:

return &token, nil
Caller gets an empty token with no error

akhilnittala: this might cause auth failures I guess

akhilnittala (changes requested):

common/defaults.go
Rizwana777: yes I agree its not a real fraction. The value is a whole number N, not a decimal fraction. The rule is - renew when less than 1/N of the token’s lifetime is left. So you only get a few fixed choices (e.g, last half if N=2, last third if N=3). We cannot pick every percentage in between. Here N is fixed at 3 as a simple internal default that still leaves a reasonable margin before expiry, and we are not exposing this as a user-facing setting.N=3 provides a reasonable safety margin. If we ever expose this as a user-facing setting, we should use a duration (e.g."renewBeforeExpiry: 20m") or percentage (e.g., "renewAtPercent: 66") instead.
I am happy to update the comment or constant name to say “divisor” so the intent is clearer for readers.
and also Open for any suggestions

controllers/argocd/dex.go
Rizwana777: added a nil secret/Data guard and treat missing or empty token as needing renewal so we never return a cached empty token with no error

controllers/argocd/dex.go
Rizwana777: I agree but we compare the API expiry time to time.Now() on the operator, so both sides assume their clocks are accurate(same idea as certificate expiry checks)
If the operator’s clock is fast, we renew a bit sooner which is safe with maybe a few extra renewals. If the operators clock is slow, we might cut it closer to real expiry and we already renew well before that expiry and we rely on cluster nodes having correct, aligned time (normal Kubernetes / OS setup), not on special logic in the operator. Open for suggestions

controllers/argocd/dex.go
Rizwana777: Agreed. returning ParseErr now

controllers/argocd/dex.go
Rizwana777: Done. Added dexServerTokenRenewalThreshold func

tests/ginkgo/fixture/secret/fixture.go
Rizwana777: HaveNonEmptyKey is used when we only need to check - this Secret has this key, and the value is not empty. It checks the same thing as the above helper but logs less, so we are less likely to print real secrets (like JWTs) into test output. In tokenRequest test, it was printing JWT token, so I have added this.  I am happy to remove this if it is unnecessary.

tests/ginkgo/parallel/1-095_validate_dex_clientsecret_test.go
Rizwana777: added the test

tests/ginkgo/parallel/1-095_validate_dex_clientsecret_test.go
Rizwana777: Agreed. Replaced to Consistently to assert the absence over short period of time.

tests/ginkgo/parallel/1-095_validate_dex_clientsecret_test.go
Rizwana777: I haved added Eventually and Consistently now as we want absence to hold over time and Eventually waits until the operator has cleared .secrets, then Consistently asserts the cleared state does not regress. Consistently alone would flake if the first poll ran before cleanup finished.

common/defaults.go
olivergondza: Oh, no. The intent is clear.

I am suggesting to redesign this to make this more powerful, and use percentage right away.

Rizwana777: okay got it, thank you. I have switched this to a percent of nominal lifetime ArgoCDDexServerTokenRenewalThresholdPercent - default to 33), so we renew when less than 33% of ArgoCDDexServerTokenExpirySecs remains which is approximately equivalent to old N=3 window for 1h token. PTAL

controllers/argocd/dex.go
```diff
@@ -42,73 +41,177 @@ func UseDex(cr *argoproj.ArgoCD) bool {
...
+		if s.Annotations[corev1.ServiceAccountNameKey] != dexSAName {
+			continue
+		}
+
+		if !strings.HasPrefix(s.Name, dexSAName+"-token-") {
+			continue
+		}
+		argoutil.LogResourceDeletion(log, s, "removing legacy Dex service account token secret")
+		if err := r.Delete(context.TODO(), s); err != nil && !apierrors.IsNotFound(err) {
+			return err
```
olivergondza: Is it guaranteed, that there is at most 1 secret to delete? Iff there can be _more_ such secrets, we delete some and then fail here, the successfully deleted secret is not removed from SA later.

olivergondza (commented):
Thanks for incorporating my suggestions. No have no further feedback - added one question.

controllers/argocd/dex.go
Rizwana777: yes the concern is valid and I have updated legacy cleanup to collect delete errors instead of stopping early and  run ServiceAccount ref cleanup after deletes and only update when refs change and use the same prefix check as deletes to avoid removing the opaque Dex token secret by mistake

olivergondza: Thanks, looks good

tests/ginkgo/fixture/secret/fixture.go
olivergondza: I mean, if the only concern is printing the test JWT token, I would not worry much as it is non-production one.

Rizwana777: yes that was the only concern, I have undone the changes and using HaveNonEmptyKeyValue func

jgwest (approved):
LGTM and AFAICT all feedback has been addressed. 

@olivergondza  and @akhilnittala thanks for reviewing! If you have any final feedback, feel free to provide ASAP before I merge!

olivergondza (approved):

jgwest (approved):
LGTM, thanks @Rizwana777!

PR #2174, opened 2026-04-27
feat: add shard distribution algorithm to argocd cr spec

cjcocokrisp:
**What type of PR is this?**

/kind enhancement

This PR adds the shard distribution algorithm to the spec of the Argo CD CR. It is located at `spec.controller.sharding.algorithm`. This allows users to more easily select the sharding algorithm that they want to use. Before you had to supply it as an environment variable. 

Also updated the sharding ginkgo test to make sure that it checks for the set algorithm. 

Let me know if there is anything else that I need to add/change as this is my first PR/work in the operator.

Fixes #?

1. Build custom image with this PR's addition
2. Deploy operator with the custom image
3. Create a Argo CD CR with the `spec.controller.sharding.algorithm` field set
4. Check the deployed statefulset to see if the env variable `ARGOCD_CONTROLLER_SHARDING_ALGORITHM` is set

## Summary by CodeRabbit

* **New Features**
  * Added a configurable Application Controller sharding algorithm: legacy (default), round-robin, or consistent-hashing.

* **Documentation**
  * Added reference docs describing the new sharding algorithm option and its default.

* **Tests**
  * Added validation tests to ensure the chosen sharding algorithm is applied as expected.

### DIFF
--- a/tests/ginkgo/parallel/1-048_validate_controller_sharding_test.go
+++ b/tests/ginkgo/parallel/1-048_validate_controller_sharding_test.go
@@ -95,7 +95,6 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
 			}
 			Expect(match).To(BeTrue(), "StatefulSet should have expected ARGOCD_CONTROLLER_REPLICAS")
 
-
 			By("ensuring algorithm can be set")
 			argocdFixture.Update(argoCD, func(ac *argov1beta1api.ArgoCD) {
 				ac.Spec.Controller.Sharding = argov1beta1api.ArgoCDApplicationControllerShardSpec{
--- a/tests/ginkgo/parallel/1-048_validate_controller_sharding_test.go
+++ b/tests/ginkgo/parallel/1-048_validate_controller_sharding_test.go
@@ -95,6 +95,25 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
 			}
 			Expect(match).To(BeTrue(), "StatefulSet should have expected ARGOCD_CONTROLLER_REPLICAS")
 
+
+			By("ensuring algorithm can be set")
+			argocdFixture.Update(argoCD, func(ac *argov1beta1api.ArgoCD) {
+				ac.Spec.Controller.Sharding = argov1beta1api.ArgoCDApplicationControllerShardSpec{
+					DistributionAlgorithm: "round-robin",
+				}
+			})
+
+			By("checking if ARGOCD_CONTROLLER_SHARDING_ALGORITHM env var is set in the app controller StatefulSet")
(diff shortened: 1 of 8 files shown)


cjcocokrisp:
GitHub seems to be having some stability problems right now which is causing the CI to fail. Will rerun it tomorrow to see if it builds green. 

First k3s was failing to install and now the Go / build is having a bad gateway problem when installing a package (it was green on a previous attempt).

jparsai (approved):
LGTM

tests/ginkgo/parallel/1-048_validate_controller_sharding_test.go
```diff
@@ -95,6 +95,26 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
...
+			By("checking if ARGOCD_CONTROLLER_SHARDING_ALGORITHM env var is set in the app controller StatefulSet")
+			Eventually(statefulSet).Should(k8sFixture.ExistByName())
+			match = false
+			for _, env := range statefulSet.Spec.Template.Spec.Containers[0].Env {
+				if env.Name == "ARGOCD_CONTROLLER_SHARDING_ALGORITHM" && env.Value == "round-robin" {
+					match = true
+					break
+				}
+			}
+			Expect(match).To(BeTrue(), "Statefulset should have expected ARGOCD_CONTROLLER_SHARDING_ALGORITHM")
```
jgwest: - I agree with CodeRabbit here: we can replace `Expect` with `Eventually` to avoid race condition where controller does not respond to `ArgoCD` CR update on previous lines in a timely fashion.

- You can likely use statefulsetFixture's `HaveContainerWithEnvVar` rather than writing the `Eventually` logic yourself
    - e.g. `Eventually(ss, "60s", "5s").Should(statefulset.HaveContainerWithEnvVar("FOO", "bar", 0))`

tests/ginkgo/parallel/1-048_validate_controller_sharding_test.go
```diff
@@ -95,6 +95,26 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
...
+
```
jgwest, line 108: Can add a step to the test that unsets `DistributionAlgorithm` and verifies that the env var is removed from `StatefulSet` via `Eventually`

jgwest (changes requested):

cjcocokrisp:
Feedback addressed and CI is green. Let me know if there is anything else that needs to be changed and thanks for the review.

jgwest (approved):
LGTM, thanks @cjcocokrisp, and thanks again for your patience in getting this merged (as expected, it was a trivial amount of work left on my side 😅 )

PR #2184, opened 2026-05-05
feat: Add Network policies for image updater

akhilnittala:
**What type of PR is this?**

kind enhancement

Create Network Policies for ArgoCD image updater.

Fixes #?
https://redhat.atlassian.net/browse/GITOPS-9092
Create image updater from Argocd operator
enabled will be by default for network policy, if we want delete the created network policy add spec.networkPolicy.enabled to false.
verify the network policy created for image updater

## Summary by CodeRabbit

* **New Features**
  * Configurable NetworkPolicy management for the Image Updater component (enabled by default). Operator will reconcile/create NetworkPolicies to allow metrics ingress to Image Updater and webhook traffic on the appropriate ports; when disabled, previously-created NetworkPolicies are removed.

* **Tests**
  * Automated tests added to verify creation and cleanup of Image Updater and webhook NetworkPolicy resources.

[Review Change Stack](https://app.coderabbit.ai/change-stack/argoproj-labs/argocd-operator/pull/2184)


tests/ginkgo/parallel/1-121_validate_image_updater_test.go
```diff
@@ -105,6 +107,14 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
...
 				Eventually(depl, "3m", "5s").Should(deplFixture.HaveReadyReplicas(1), depl.Name+" was not ready")
 			}
+			// verify network policy is created
+			networkPolicy := &networkingv1.NetworkPolicy{ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("%s-%s", argoCD.Name, "image-updater-network-policy"), Namespace: ns.Name}}
+			Eventually(networkPolicy).Should(k8sFixture.ExistByName())
+			var netpolDisable = false
+			argoCD.Spec.ImageUpdater.NetworkPolicy = &argov1beta1api.ImageUpdaterNetworkPolicySpec{}
+			argoCD.Spec.ImageUpdater.NetworkPolicy.Enabled = &netpolDisable
+			Expect(k8sClient.Update(ctx, argoCD)).To(Succeed())
+			Eventually(networkPolicy).Should(k8sFixture.NotExistByName())
```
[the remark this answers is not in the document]
akhilnittala: done

controllers/argocd/networkpolicies.go
```diff
@@ -54,7 +63,7 @@ func (r *ReconcileArgoCD) ReconcileNetworkPolicies(cr *argoproj.ArgoCD) error {
 		return err
 	}
 
-	if !cr.Spec.NetworkPolicy.IsEnabled() {
+	if !cr.Spec.NetworkPolicy.IsEnabled() || !cr.Spec.ImageUpdater.NetworkPolicy.IsEnabled() {
 		return r.deleteArgoCDNetworkPolicies(cr)
 	}
```
chengfang, line 67: can you look into the coderabbirai comment: "Broken control flow when only the Image Updater NP is disabled.

There are two interacting bugs in ReconcileNetworkPolicies and the new ReconcileImageUpdaterNetworkPolicy:
"

Here it looks like the config of image updater NP will affect other components.

chengfang:
when ImageUpdater NP is enabled, it will allow metrics ingress, but block any webhook traffic. Is it possible to enhance the pr to also consider ImageUpdater webhook configuration (webhook enabled or not, webhook port, etc) and also include these in the NP creation?

dkarpele:
> when ImageUpdater NP is enabled, it will allow metrics ingress, but block any webhook traffic. Is it possible to enhance the pr to also consider ImageUpdater webhook configuration (webhook enabled or not, webhook port, etc) and also include these in the NP creation?

@chengfang, I think @akhilnittala followed `NetworkPolicy` rule from Image Updater repo https://github.com/argoproj-labs/argocd-image-updater/blob/master/config/network-policy/allow-metrics-traffic.yaml

I think To allow webhook traffic we should fix it in Image Updater repo first.

api/v1beta1/argocd_types.go
```diff
@@ -338,6 +338,21 @@ type ArgoCDImageUpdaterSpec struct {
 
 	// Resources defines the Compute Resources required by the container for Argo CD Image Updater.
 	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
+	// NetworkPolicy controls whether the operator should create NetworkPolicy resources for the Image Updater. Defaults to enabled when not specified.
+	NetworkPolicy *ImageUpdaterNetworkPolicySpec `json:"networkPolicy,omitempty"`
```
dkarpele: Just trying to understand. Why do we need this API field? Can we manage ImageUpdater NetworkPolicy in the same way as other NP in the function `controllers/argocd/networkpolicies.go/ReconcileNetworkPolicies`?

akhilnittala: I added it because, currently image updater is being deployed as seperate component compared to other argocd components right, so I thought it is better to have a seperate field for network policy for image updater as well.

chengfang:
> > when ImageUpdater NP is enabled, it will allow metrics ingress, but block any webhook traffic. Is it possible to enhance the pr to also consider ImageUpdater webhook configuration (webhook enabled or not, webhook port, etc) and also include these in the NP creation?
> 
> @chengfang, I think @akhilnittala followed `NetworkPolicy` rule from Image Updater repo https://github.com/argoproj-labs/argocd-image-updater/blob/master/config/network-policy/allow-metrics-traffic.yaml
> 
> I think To allow webhook traffic we should fix it in Image Updater repo first.

Agreed we should fix in image updater upstream. But the feature in this PR does not have a dependency on upstream fix.

When using image updater outside operator, users are free to configure their NP and allow both metrics and webhook traffic. When running image updater with operator, it becomes impossible to run webhook with NP enabled in its current form, and in this sense, operator falls behind the upstream.

controllers/argocd/networkpolicies.go
```diff
@@ -1035,10 +1053,130 @@ func (r *ReconcileArgoCD) ReconcileArgoCDRepoServerNetworkPolicy(cr *argoproj.Ar
...
+				explanation += ", "
+			}
+			explanation += "ingress rules"
+			modified = true
+		}
+
+		if modified {
+			argoutil.LogResourceUpdate(log, existing, "updating", explanation)
+			if err := r.Update(context.TODO(), existing); err != nil {
+				log.Error(err, "Failed to update %s network policy in namespace %s", existing.Name, cr.Namespace)
```
chengfang: `func (l Logger) Error(err error, msg string, keysAndValues ...any)`
`with the given message and key/value pairs as context`

The message is printed without param resolution. Any additional information should be added as `keysAndValues`.

Also please check other occurrences of `log.Error`

controllers/argocd/networkpolicies.go
```diff
@@ -40,9 +40,20 @@ const (
...
+	ImageUpdaterNetworkPolicy = "image-updater-network-policy"
+	// WebhookTrafficNetworkPolicy is the name of the network policy which controls Webhook traffic
+	WebhookTrafficNetworkPolicy = "webhook-traffic-network-policy"
 )
 
 func (r *ReconcileArgoCD) ReconcileNetworkPolicies(cr *argoproj.ArgoCD) error {
+	if IsImageUpdaterAPIAvailable() {
+		if err := r.ReconcileImageUpdaterNetworkPolicy(cr); err != nil {
+			return err
+		}
```
chengfang, line 52: ImageUpdater CRD itself is available whether ImageUpdater controller is enabled or disalbed. See https://argocd-operator.readthedocs.io/en/latest/usage/image-updater/#installation

So from reading the above code, it seems ImageUpdater NP will always be created and reconciled, even if the ImageUpdater controller is disabled because the user is not using ImageUpdater.

This seems unnecessary work and creates unused objects.

chengfang:
Can you add unit tests to controllers/argocd/networkpolicies_test.go for Image updater networkpolicy?

chengfang:
Can we remove `ImageUpdaterNetworkPolicySpec`, the component-level toggle for ImageUpdater NP? Looking at other component NP, such as ApplicationSetController, and NotificationsController, neither of them have their own component-level switch. I think we should follow the same pattern.

This should make the logic simpler: we only need to create or update ImageUpdater NP only if top-level NP is enabled and ImageUpdater is enabled (.spec.imageupdater.enabled: true).

WDYT?

chengfang:
Is it possible to combine the 2 ImageUpdater policies into a single one? Since they are enabled or disable together, having 2 separate policies does not give user any more flexiblity than single one.

controllers/argocd/networkpolicies.go
akhilnittala: done

controllers/argocd/networkpolicies.go
akhilnittala: done

controllers/argocd/networkpolicies.go
```diff
@@ -1035,10 +1051,114 @@ func (r *ReconcileArgoCD) ReconcileArgoCDRepoServerNetworkPolicy(cr *argoproj.Ar
...
+		return nil
+	}
+	if err := controllerutil.SetControllerReference(cr, desired, r.Scheme); err != nil {
+		log.Error(err, "Failed to set controller reference on network policy")
+		return fmt.Errorf("failed to set controller reference on network policy. error: %w", err)
+	}
+	argoutil.LogResourceCreation(log, desired)
+	if err := r.Create(context.TODO(), desired); err != nil {
+		log.Error(err, "Failed to create %s network policy in namespace %s", desired.Name, cr.Namespace)
```
chengfang: the same `log.Error` usage issue as in line 1128, and the same fix is needed here.

controllers/argocd/networkpolicies.go
```diff
@@ -40,9 +40,20 @@ const (
...
+	// WebhookTrafficNetworkPolicy is the name of the network policy which controls Webhook traffic
+	WebhookTrafficNetworkPolicy = "webhook-traffic-network-policy"
 )
 
 func (r *ReconcileArgoCD) ReconcileNetworkPolicies(cr *argoproj.ArgoCD) error {
+	if IsImageUpdaterAPIAvailable() && cr.Spec.ImageUpdater.Enabled && cr.Spec.NetworkPolicy.IsEnabled() {
+		if err := r.ReconcileImageUpdaterNetworkPolicy(cr); err != nil {
+			return err
+		}
+	} else {
```
chengfang: I think we should omit the logging inside the else block, since very often users will not have ImageUpdater enabled as their normal operation, and seeing this message can be confusing to many users.

controllers/argocd/networkpolicies.go
```diff
@@ -40,9 +40,20 @@ const (
 	ArgoCDDexServerNetworkPolicy = "dex-server-network-policy"
 	// ArgoCDApplicationSetControllerNetworkPolicy is the name of the network policy which controls Argo CD ApplicationSet Controller traffic
 	ArgoCDApplicationSetControllerNetworkPolicy = "applicationset-controller-network-policy"
+	// ImageUpdaterNetworkPolicy is the name of the network policy which controls Image Updater traffic
+	ImageUpdaterNetworkPolicy = "image-updater-network-policy"
+	// WebhookTrafficNetworkPolicy is the name of the network policy which controls Webhook traffic
+	WebhookTrafficNetworkPolicy = "webhook-traffic-network-policy"
```
chengfang: the name `webhook-traffic-network-policy` is generic and seems to apply to all webhook traffic, but in fact it only applies to Image Updater webhook. Can we rename it to be specific to Image Updater webhook?

controllers/argocd/networkpolicies.go
```diff
@@ -1035,10 +1051,114 @@ func (r *ReconcileArgoCD) ReconcileArgoCDRepoServerNetworkPolicy(cr *argoproj.Ar
...
 }
 
-func returnNetworkPolicyHeaders(cr *argoproj.ArgoCD, NetworkPolicyName string) *networkingv1.NetworkPolicy {
+func createImageUpdaterNetworkPolicy(cr *argoproj.ArgoCD, name string, namespaceSelectorLabels map[string]string, port int32) *networkingv1.NetworkPolicy {
+	desired := returnNetworkPolicyHeaders(cr, name)
+	desired.Spec = networkingv1.NetworkPolicySpec{
+		PodSelector: metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"control-plane": "argocd-image-updater-controller",
+			},
```
chengfang, line 1063: After thinking more, I think we should be consistent with other components and also use `app.kubernetes.io/name` label:

```go
  PodSelector: metav1.LabelSelector{
      MatchLabels: map[string]string{
          "app.kubernetes.io/name": nameWithSuffix(common.ArgoCDImageUpdaterControllerComponent, cr),
      },
  },
```

controllers/argocd/networkpolicies.go
```diff
@@ -1035,10 +1051,114 @@ func (r *ReconcileArgoCD) ReconcileArgoCDRepoServerNetworkPolicy(cr *argoproj.Ar
...
+					},
+				},
+			},
+		},
+	}
+
+	return desired
+}
+
+func reconcileNetworkPolicy(r *ReconcileArgoCD, cr *argoproj.ArgoCD, desired *networkingv1.NetworkPolicy) error {
```
chengfang: can we rename this func to be ImageUpdater-specific?

controllers/argocd/networkpolicies.go
```diff
@@ -1035,10 +1051,114 @@ func (r *ReconcileArgoCD) ReconcileArgoCDRepoServerNetworkPolicy(cr *argoproj.Ar
...
+	if err := r.Create(context.TODO(), desired); err != nil {
+		log.Error(err, "Failed to create %s network policy in namespace %s", desired.Name, cr.Namespace)
+		return fmt.Errorf("failed to create %s network policy in namespace %s. error: %w", desired.Name, cr.Namespace, err)
+	}
+	return nil
+}
+
+func (r *ReconcileArgoCD) ReconcileImageUpdaterNetworkPolicy(cr *argoproj.ArgoCD) error {
+	metricsPolicy := createImageUpdaterNetworkPolicy(cr, ImageUpdaterNetworkPolicy, map[string]string{"metrics": "enabled"}, 8443)
+	webhookPolicy := createImageUpdaterNetworkPolicy(cr, WebhookTrafficNetworkPolicy, map[string]string{"webhooks": "enabled"}, 8082)
```
chengfang: NP in other components don't have additional namespace label restriction, and I think ImageUpdater should be consistent. I understand this is based on upstream ImageUpdater NP configuration, but I feel it's more important to have the same configuration pattern across argocd-operator components.

controllers/argocd/networkpolicies.go
akhilnittala: done

controllers/argocd/networkpolicies.go
akhilnittala: this is not the change as part of this PR, but yeah removed.

controllers/argocd/networkpolicies.go
akhilnittala: done

controllers/argocd/networkpolicies.go
akhilnittala: done

controllers/argocd/networkpolicies.go
akhilnittala: done

controllers/argocd/networkpolicies.go
akhilnittala: done

controllers/argocd/networkpolicies.go
```diff
@@ -40,9 +40,18 @@ const (
 	ArgoCDDexServerNetworkPolicy = "dex-server-network-policy"
 	// ArgoCDApplicationSetControllerNetworkPolicy is the name of the network policy which controls Argo CD ApplicationSet Controller traffic
 	ArgoCDApplicationSetControllerNetworkPolicy = "applicationset-controller-network-policy"
+	// ImageUpdaterNetworkPolicy is the name of the network policy which controls Image Updater traffic
+	ImageUpdaterNetworkPolicy = "image-updater-network-policy"
+	// WebhookTrafficNetworkPolicy is the name of the network policy which controls Webhook traffic
+	ImageUpdaterWebhookTrafficNetworkPolicy = "image-updater-webhook-traffic-network-policy"
```
olivergondza: nit: I would drop the word "traffic" here. Does not seem to add much information...

controllers/argocd/networkpolicies.go
```diff
@@ -89,13 +98,18 @@ func (r *ReconcileArgoCD) ReconcileNetworkPolicies(cr *argoproj.ArgoCD) error {
...
 func (r *ReconcileArgoCD) deleteArgoCDNetworkPolicies(cr *argoproj.ArgoCD) error {
-	names := []string{
-		fmt.Sprintf("%s-%s", cr.Name, ArgoCDNotificationsControllerNetworkPolicy),
-		fmt.Sprintf("%s-%s", cr.Name, ArgoCDDexServerNetworkPolicy),
-		fmt.Sprintf("%s-%s", cr.Name, ArgoCDApplicationSetControllerNetworkPolicy),
-		fmt.Sprintf("%s-%s", cr.Name, ArgoCDServerNetworkPolicy),
-		fmt.Sprintf("%s-%s", cr.Name, ArgoCDApplicationControllerNetworkPolicy),
-		fmt.Sprintf("%s-%s", cr.Name, ArgoCDRepoServerNetworkPolicy),
+	var names []string
+	if !cr.Spec.NetworkPolicy.IsEnabled() {
```
olivergondza: This is now tested in and out of the function. No need to do it twice, but most importantly deleting conditionally is somehow surprising for method named `deleteArgoCDNetworkPolicies`.

tests/ginkgo/parallel/1-121_validate_image_updater_test.go
```diff
@@ -105,6 +107,18 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
...
 			}
+			// verify network policy is created
+			networkPolicy := &networkingv1.NetworkPolicy{ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("%s-%s", argoCD.Name, "image-updater-network-policy"), Namespace: ns.Name}}
+			WebhookTrafficNetworkPolicy := &networkingv1.NetworkPolicy{ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("%s-%s", argoCD.Name, "image-updater-webhook-traffic-network-policy"), Namespace: ns.Name}}
+			Eventually(networkPolicy, "3m", "5s").Should(k8sFixture.ExistByName())
+			Eventually(WebhookTrafficNetworkPolicy, "3m", "5s").Should(k8sFixture.ExistByName())
+			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(argoCD), argoCD)).To(Succeed())
+			netpolDisable := false
+			argoCD.Spec.NetworkPolicy = argov1beta1api.ArgoCDNetworkPolicySpec{}
+			argoCD.Spec.NetworkPolicy.Enabled = &netpolDisable
```
olivergondza: It is `netpolEnable` by its meaning. Or even better `ptr.To(true)`.

controllers/argocd/networkpolicies.go
```diff
@@ -1035,10 +1049,111 @@ func (r *ReconcileArgoCD) ReconcileArgoCDRepoServerNetworkPolicy(cr *argoproj.Ar
...
+			Namespace: desired.Namespace,
+		},
+	}
+	exists, err := argoutil.IsObjectFound(r.Client, cr.Namespace, existing.Name, existing)
+	if err != nil {
+		return err
+	}
+	if exists {
+		modified := false
+		explanation := ""
```
olivergondza: Please don't use this obsolete pattern. See https://github.com/argoproj-labs/argocd-operator/blob/5eedcfe310f6baa83157e8ea8b6c1064a845ed59/controllers/argocd/secret.go#L344 for its modern incarnation.

tests/ginkgo/parallel/1-121_validate_image_updater_test.go
```diff
@@ -105,6 +107,18 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
 				Eventually(depl).Should(deplFixture.HaveReplicas(1))
 				Eventually(depl, "3m", "5s").Should(deplFixture.HaveReadyReplicas(1), depl.Name+" was not ready")
 			}
+			// verify network policy is created
+			networkPolicy := &networkingv1.NetworkPolicy{ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("%s-%s", argoCD.Name, "image-updater-network-policy"), Namespace: ns.Name}}
+			WebhookTrafficNetworkPolicy := &networkingv1.NetworkPolicy{ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("%s-%s", argoCD.Name, "image-updater-webhook-traffic-network-policy"), Namespace: ns.Name}}
```
olivergondza: nit: `webhookTrafficNetworkPolicy` (camelCase)

tests/ginkgo/parallel/1-121_validate_image_updater_test.go
```diff
@@ -105,6 +107,18 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
...
+			Eventually(networkPolicy, "3m", "5s").Should(k8sFixture.ExistByName())
```
olivergondza: Consider shortening the timeouts here. I expect the NP will reconcile much faster.

tests/ginkgo/parallel/1-121_validate_image_updater_test.go
```diff
@@ -105,6 +107,18 @@ var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
...
+			Eventually(WebhookTrafficNetworkPolicy, "3m", "5s").Should(k8sFixture.ExistByName())
+			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(argoCD), argoCD)).To(Succeed())
+			netpolDisable := false
+			argoCD.Spec.NetworkPolicy = argov1beta1api.ArgoCDNetworkPolicySpec{}
+			argoCD.Spec.NetworkPolicy.Enabled = &netpolDisable
+			Expect(k8sClient.Update(ctx, argoCD)).To(Succeed())
+			Eventually(networkPolicy, "3m", "5s").Should(k8sFixture.NotExistByName())
+			Eventually(WebhookTrafficNetworkPolicy, "3m", "5s").Should(k8sFixture.NotExistByName())
```
olivergondza: @dkarpele, @chengfang, Are some of the existing tests using the network communication? I am thinking running them with the NP on, so they verify the actual connection, even on for the positive use-case.

olivergondza (changes requested):
The code looks good to me at large, I have added multiple comments, mostly on the side of maintainability.

anandrkskd:
@akhilnittala the step to test is suggesting spec.imageUpdater.networkPolicy.enabled field is there disable NP. can you update it as well?

controllers/argocd/networkpolicies.go
akhilnittala: done

controllers/argocd/networkpolicies.go
akhilnittala: done

tests/ginkgo/parallel/1-121_validate_image_updater_test.go
akhilnittala: done

controllers/argocd/networkpolicies.go
akhilnittala: done

tests/ginkgo/parallel/1-121_validate_image_updater_test.go
akhilnittala: done

tests/ginkgo/parallel/1-121_validate_image_updater_test.go
akhilnittala: done

akhilnittala:
> @akhilnittala the step to test is suggesting spec.imageUpdater.networkPolicy.enabled field is there disable NP. can you update it as well?

done

olivergondza (approved):
The code looks good to me. I would like to see some verification of the Network Policy _actual_ function. But it seems not done for any network policies that we have.

olivergondza:
@akhilnittala, any comment on this suggestion? I do not feel strongly one way or the other, but I do wonder why we chose such granularity...

> @chengfang: Is it possible to combine the 2 ImageUpdater policies into a single one? Since they are enabled or disable together, having 2 separate policies does not give user any more flexiblity than single one.

chengfang (approved):

anandrkskd (approved):
LGTM

olivergondza:
@akhilnittala?

> @akhilnittala, any comment on this suggestion? I do not feel strongly one way or the other, but I do wonder why we chose such granularity...
> 
> > @chengfang: Is it possible to combine the 2 ImageUpdater policies into a single one? Since they are enabled or disable together, having 2 separate policies does not give user any more flexiblity than single one.

akhilnittala:
> @akhilnittala?
> 
> > @akhilnittala, any comment on this suggestion? I do not feel strongly one way or the other, but I do wonder why we chose such granularity...
> > > @chengfang: Is it possible to combine the 2 ImageUpdater policies into a single one? Since they are enabled or disable together, having 2 separate policies does not give user any more flexiblity than single one.

that is how it is in upstream repo right ?

svghadi (approved):
LGTM

olivergondza:
> > @akhilnittala?
> > > @akhilnittala, any comment on this suggestion? I do not feel strongly one way or the other, but I do wonder why we chose such granularity...
> > > > @chengfang: Is it possible to combine the 2 ImageUpdater policies into a single one? Since they are enabled or disable together, having 2 separate policies does not give user any more flexiblity than single one.
> 
> that is how it is in upstream repo right ?

Not sure what you mean by upstream repo, but to me Cheng's suggestions have their merit. Fewer resources reduce the conceptual load for humans, it is a little less resources to handle for the machines, fewer API calls, less code to maintain going forward, etc.

olivergondza (approved):
Thanks, this looks good to me!

svghadi (approved):

svghadi:
Thanks @olivergondza , @chengfang, @dkarpele, @anandrkskd for the reviews.

PR #2194, opened 2026-05-12
Enforce tokenRef strict Restrictions for ApplicationSet Any-Namespace mode

nmirasch:
**What type of PR is this?**

/kind enhancement

When ApplicationSet is allowed in namespaces beyond the Argo CD namespace, SCM/PR-style generators can use tokenRef to read Secrets there. Without strict mode, unlabeled Secrets can be accepted, which is risky in shared clusters. Argo CD’s tokenRef strict mode limits this to Secrets labeled argocd.argoproj.io/secret-type=scm-creds. The operator should turn that guardrail on by default when that mode is actually in use, with a clear opt-out. See upstream: [ApplicationSet in any namespace – tokenRef restrictions](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Appset-Any-Namespace/#tokenref-restrictions).

What this PR does:
This PR adds automatic defaulting for `applicationsetcontroller.enable.tokenref.strict.mode `in argocd-cmd-params-cm. The operator now always writes this key, setting it to `true` when `spec.applicationSet.sourceNamespaces` expands to at least one namespace in the cluster, and to `false` otherwise. If `spec.cmdParams` explicitly defines the key, that value takes precedence and is preserved across reconciliations.

The ApplicationSet controller consumes this setting through the `ARGOCD_APPLICATIONSET_CONTROLLER_TOKENREF_STRICT_MODE` environment variable. To support this, the PR adds `ArgoCDApplicationSetControllerTokenRefStrictModeCmdParamKey` and `ArgoCDApplicationSetControllerTokenRefStrictModeEnvName `constants to common/keys.go.

The defaulting logic uses `getApplicationSetSourceNamespaces`, which resolves wildcard and regex patterns defined in `spec.applicationSet.sourceNamespaces`. (It does not depend on the cluster-config namespace or on the intersection with spec.sourceNamespaces)

The operator treats the ArgoCD custom resource as the source of truth and automatically corrects any manual changes to this key in `argocd-cmd-params-cm.`

Fixes https://redhat.atlassian.net/browse/GITOPS-8628

## Summary by CodeRabbit

* **New Features**
  * ApplicationSet tokenRef strict mode now defaults to enabled for cluster-scoped Argo CD when ApplicationSet-in-any-namespace is active; can be opted out via CmdParams. Controller reads this setting from the argocd-cmd-params-cm ConfigMap into its environment.

* **Tests**
  * Added unit and E2E tests for defaulting, opt-out, enforcement, and namespace-scoped behavior.

* **Documentation**
  * CRD/schema docs updated to document the tokenRef strict-mode CmdParam and its defaulting behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/argoproj-labs/argocd-operator/pull/2194?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/api/v1beta1/argocd_types.go
+++ b/api/v1beta1/argocd_types.go
@@ -1069,8 +1069,9 @@ type ArgoCDSpec struct {
 	AggregatedClusterRoles bool `json:"aggregatedClusterRoles,omitempty"`
 
 	// CmdParams specifies command-line parameters for the Argo CD components.
-	// The only keys currently supported for this parameter are:
+	// Keys commonly used with this field include:
 	// - controller.resource.health.persist
+	// - applicationsetcontroller.enable.tokenref.strict.mode — when ApplicationSet-in-any-namespace is active, the operator defaults this to "true"
 	CmdParams map[string]string `json:"cmdParams,omitempty"`
 
 	// ArgoCDAgent defines configurations for the ArgoCD Agent component.
--- a/api/v1beta1/argocd_types.go
+++ b/api/v1beta1/argocd_types.go
@@ -1069,7 +1069,7 @@ type ArgoCDSpec struct {
 	AggregatedClusterRoles bool `json:"aggregatedClusterRoles,omitempty"`
 
 	// CmdParams specifies command-line parameters for the Argo CD components.
-	// Keys commonly used with this field include:
+	// The only keys currently supported for this parameter are:
 	// - controller.resource.health.persist
 	// - applicationsetcontroller.enable.tokenref.strict.mode — when ApplicationSet-in-any-namespace is active, the operator defaults this to "true"
 	CmdParams map[string]string `json:"cmdParams,omitempty"`
--- a/controllers/argocd/configmap.go
+++ b/controllers/argocd/configmap.go
@@ -975,17 +975,17 @@ func (r *ReconcileArgoCD) reconcileArgoCmdParamsConfigMap(cr *argoproj.ArgoCD) e
 	tokenRefStrictModeValue, hasExplicitTokenRefStrictMode := getTokenRefStrictModeCmdParamValue(cr.Spec.CmdParams)
 	if hasExplicitTokenRefStrictMode {
 		cm.Data[common.ArgoCDApplicationSetControllerTokenRefStrictModeCmdParamKey] = tokenRefStrictModeValue
-	}
-
-	// When applicationSet-in-any-namespace is effectively enabled, default tokenRef strict mode so only
-	// Secrets labeled as SCM credentials can be used with tokenRef
-	if argoutil.IsNamespaceClusterConfigNamespace(cr.Namespace) && cr.Spec.ApplicationSet != nil {
-		allowedAppSetNS, err := r.getAllowedApplicationSetSourceNamespaces(cr)
-		if err != nil {
-			return err
-		}
-		if len(allowedAppSetNS) > 0 && !hasExplicitTokenRefStrictMode {
-			cm.Data[common.ArgoCDApplicationSetControllerTokenRefStrictModeCmdParamKey] = "true"
+	} else {
+		// When applicationSet-in-any-namespace is effectively enabled, default tokenRef strict mode so only
+		// Secrets labeled as SCM credentials can be used with tokenRef
+		if argoutil.IsNamespaceClusterConfigNamespace(cr.Namespace) && cr.Spec.ApplicationSet != nil {
+			allowedAppSetNS, err := r.getAllowedApplicationSetSourceNamespaces(cr)
+			if err != nil {
+				return err
+			}
+			if len(allowedAppSetNS) > 0 {
+				cm.Data[common.ArgoCDApplicationSetControllerTokenRefStrictModeCmdParamKey] = "true"
+			}
 		}
 	}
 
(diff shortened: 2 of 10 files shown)


controllers/argocd/configmap.go
```diff
@@ -970,6 +970,25 @@ func (r *ReconcileArgoCD) reconcileArgoCmdParamsConfigMap(cr *argoproj.ArgoCD) e
...
+	// ARGOCD_APPLICATIONSET_CONTROLLER_TOKENREF_STRICT_MODE.
+	tokenRefStrictModeValue, hasExplicitTokenRefStrictMode := getTokenRefStrictModeCmdParamValue(cr.Spec.CmdParams)
+	if hasExplicitTokenRefStrictMode {
+		cm.Data[common.ArgoCDApplicationSetControllerTokenRefStrictModeCmdParamKey] = tokenRefStrictModeValue
+	}
+
+	// When applicationSet-in-any-namespace is effectively enabled, default tokenRef strict mode so only
+	// Secrets labeled as SCM credentials can be used with tokenRef
+	if argoutil.IsNamespaceClusterConfigNamespace(cr.Namespace) && cr.Spec.ApplicationSet != nil {
+		allowedAppSetNS, err := r.getAllowedApplicationSetSourceNamespaces(cr)
```
svghadi: What is the purpose of this check?

controllers/argocd/configmap.go
```diff
@@ -970,6 +970,25 @@ func (r *ReconcileArgoCD) reconcileArgoCmdParamsConfigMap(cr *argoproj.ArgoCD) e
 	const healthPersistKey = "controller.resource.health.persist"
 	cm.Data[healthPersistKey] = "true"
 
+	// TokenRef strict mode uses the upstream argocd-cmd-params-cm key consumed by
+	// ARGOCD_APPLICATIONSET_CONTROLLER_TOKENREF_STRICT_MODE.
+	tokenRefStrictModeValue, hasExplicitTokenRefStrictMode := getTokenRefStrictModeCmdParamValue(cr.Spec.CmdParams)
+	if hasExplicitTokenRefStrictMode {
+		cm.Data[common.ArgoCDApplicationSetControllerTokenRefStrictModeCmdParamKey] = tokenRefStrictModeValue
+	}
```
svghadi, line 988: We can write else here to check if sourceNamespaces for appset is enabled and set the flag accordingly.

svghadi (changes requested):

api/v1beta1/argocd_types.go
```diff
@@ -1066,8 +1066,9 @@ type ArgoCDSpec struct {
 	AggregatedClusterRoles bool `json:"aggregatedClusterRoles,omitempty"`
 
 	// CmdParams specifies command-line parameters for the Argo CD components.
-	// The only keys currently supported for this parameter are:
+	// Keys commonly used with this field include:
```
svghadi: Lets keep the old comment. Only keys specified in here are supported.

controllers/argocd/configmap.go
nmirasch: The idea was to check if the apps-in-anyNS is enabled and set TokenRefStrinctMode to true in case there is no explicit TokenRefStrictMode defined.

controllers/argocd/configmap.go
nmirasch: I have moved the check if  AppS-in-anyNS is enabled only when there is not ExplicitTokenRefStrict configuration

controllers/argocd/configmap.go
```diff
@@ -970,6 +970,25 @@ func (r *ReconcileArgoCD) reconcileArgoCmdParamsConfigMap(cr *argoproj.ArgoCD) e
...
+	// TokenRef strict mode uses the upstream argocd-cmd-params-cm key consumed by
+	// ARGOCD_APPLICATIONSET_CONTROLLER_TOKENREF_STRICT_MODE.
+	tokenRefStrictModeValue, hasExplicitTokenRefStrictMode := getTokenRefStrictModeCmdParamValue(cr.Spec.CmdParams)
+	if hasExplicitTokenRefStrictMode {
+		cm.Data[common.ArgoCDApplicationSetControllerTokenRefStrictModeCmdParamKey] = tokenRefStrictModeValue
+	} else {
+		// When applicationSet-in-any-namespace is effectively enabled, default tokenRef strict mode so only
+		// Secrets labeled as SCM credentials can be used with tokenRef
+		if argoutil.IsNamespaceClusterConfigNamespace(cr.Namespace) && cr.Spec.ApplicationSet != nil {
+			allowedAppSetNS, err := r.getAllowedApplicationSetSourceNamespaces(cr)
```
svghadi: I think we can simplify this by checking whether applicationSetSourceNamespaces has any values. If len(applicationSetSourceNamespaces) > 0, we can set tokenRef strict mode to true.

controllers/argocd/configmap.go
```diff
@@ -970,6 +970,25 @@ func (r *ReconcileArgoCD) reconcileArgoCmdParamsConfigMap(cr *argoproj.ArgoCD) e
...
+			if err != nil {
+				return err
+			}
+			if len(allowedAppSetNS) > 0 {
+				cm.Data[common.ArgoCDApplicationSetControllerTokenRefStrictModeCmdParamKey] = "true"
+			}
+		}
+	}
+
```
svghadi, line 989: Can you confirm whether this handles the case where the value was previously set to true, but later AppSet in any namespace is removed and the value is also not set via cmdparams?

It looks like reconciliation may not reset the value back to false in that case.

I think a better flow would be to default it to false first:
```
cm.Data[common.ArgoCDApplicationSetControllerTokenRefStrictModeCmdParamKey] = "false"
```
Then, if AppSet in any namespace is enabled, set it to true.

Finally, if the user has explicitly overridden the value via cmdparams, that value should take precedence.

svghadi (changes requested):

svghadi (approved):
LGTM. Thanks

PR #2210, opened 2026-05-26
test(e2e): stabilize tokenRef strict mode cmd-params tests

nmirasch:
**What type of PR is this?**

/kind failing-test

The test defaults tokenRef strict mode to false when applicationSet sourceNamespaces are empty or removed was flaky in CI. It created two ArgoCD instances in the same namespace, which both reconcile the same argocd-cmd-params-cm. That caused reconciliation contention and doubled cluster load. The second instance often never reached Available (e.g. argocd-server in CrashLoopBackOff), so the test timed out on BeAvailable() even though cmd-params logic was correct.

fix: Split the scenario into two tests: empty on create and removed after configure, with one ArgoCD per namespace (same coverage, no shared ConfigMap fight). Align all five tokenRef strict mode tests to assert argocd-cmd-params-cm via Eventually(..., "3m", "5s") instead of BeAvailable(). Cmd-params are reconciled before the instance is fully healthy; waiting for Available is unnecessary for these assertions, slower, and more fragile under load. Other 1-037 tests still use BeAvailable() where they depend on a running instance (RBAC, deployments, etc.).

Fixes #?

## Summary by CodeRabbit

* **Tests**
  * Enhanced E2E test coverage for ApplicationSet controller strict mode configuration behavior with separated test cases evaluating different sourceNamespaces scenarios, improved ConfigMap assertion timing strategies using consistent timing expectations, and refined test execution patterns to ensure reliable validation of strict mode defaults and state transitions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/argoproj-labs/argocd-operator/pull/2210?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/ginkgo/sequential/1-037_validate_applicationset_in_any_namespace_test.go
+++ b/tests/ginkgo/sequential/1-037_validate_applicationset_in_any_namespace_test.go
@@ -1346,7 +1346,6 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 				},
 			}
 			Expect(k8sClient.Create(ctx, argoCD)).To(Succeed())
-			Eventually(argoCD, "5m", "5s").Should(argocdFixture.BeAvailable())
 
 			cmdParamsCM := &corev1.ConfigMap{
 				ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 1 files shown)


akhilnittala (approved):

jgwest (approved):
LGTM, thanks @nmirasch!

PR #2211, opened 2026-05-26
fix: limit StatefulSet names to 52 chars and abbreviate application-controller to prevent label limit errors

### LINKED ISSUE #2181: Long names for Argo CD CRs causes Argo CD application controller not to start
**Describe the bug**

When testing my PR for adding sharding algorithm as a field. I went and created a a test CR with the name `argocd-sharding-algorithm-test` which is 30 characters long. When the operator creates the statefulset for the argocd instance managed by this CR it appends `-application-controller` to the name bringing the character total to 53.

Kubernetes creates a controller revision resource for statefulsets. The name of the controller revision resource seems to get stored as a label in the statefulset. Controller revisions append 11 more characters onto the name of the statefulset (a '-' then a random string of 10 characters for uniqueness). So for example in my case it becomes `argocd-sharding-algorithm-test-76fd497866`. This is 64 characters so it is over the max length for a label name. This causes the statefulset to fail to create with the below error:

```
  Type     Reason        Age                   From                    Message
  ----     ------        ----                  ----                    -------
  Warning  FailedCreate  53s (x16 over 3m37s)  statefulset-controller  Create Pod argocd-sharding-algorithm-test-application-controller-0 in StatefulSet argocd-sharding-algorithm-test-application-controller failed error: Pod "argocd-sharding-algorithm-test-application-controller-0" is invalid: metadata.labels: Invalid value: "argocd-sharding-algorithm-test-application-controller-76fd497866": must be no more than 63 bytes
```

I also tried to create a argo cd resource with over 63 characters as a name and everything there failed. When doing one around 55 characters to make sure the other argo cd components were fine which they were (they get truncated with a hash).

**To Reproduce**
1. Create a Argo CD CR with the following manifest (this one will cause the application controller stateful set to not even get created):
```yaml
kind: ArgoCD
metadata:
  name: this-name-will-push-the-char-limit
spec:
  {}
```
2. Check to see if the statefulset for the application controller is created

**Expected behavior**

When the operator creates the stateful set for the application controller, if the name is more then 29 characters it will be truncated like the rest of the components.

**Screenshots**

**Additional context**

Rizwana777:
**What type of PR is this?**
/kind bug
JIRA - https://redhat.atlassian.net/browse/GITOPS-9785

Fixes #2181 
where ArgoCD custom resources with names exceeding 29 characters caused the application controller StatefulSet to fail creation with the error "must be no more than 63 bytes". The root cause was that Kubernetes automatically appends an 11-character controller revision suffix (a hyphen plus 10-character hash) to StatefulSet names for version tracking, and the operator was not accounting for this overhead when generating StatefulSet names. When combined with long CR names and the full "application-controller" suffix (22 characters) the total length exceeded Kubernetes' 63-byte label limit.

Now I implemented a fix by adding a new NameWithSuffixForStatefulSet() function that ensures StatefulSet names stay within the Kubernetes 63-character limit while reserving space for controller revision suffixes. To make truncated names more readable, I abbreviate application-controller to app-controller before truncation avoiding awkward cutoffs. I also updated the existing NameWithSuffix() function to better handle truncation and ensure all generated resource names remain within the allowed limit.

Fixes #2181 

## Summary by CodeRabbit

* **Chores**
  * Standardized application-controller naming and introduced StatefulSet-safe name generation with legacy-name cleanup to ensure consistent selection and smooth migration.

* **Bug Fixes**
  * Updated network policies, services, and monitoring selectors to use the standardized controller names and ensured long resource names are truncated/hashed to fit Kubernetes limits.

* **Tests**
  * Added and expanded unit and E2E tests validating deterministic naming, length/abbreviation behavior, truncation, and migration/cleanup.

### DIFF
--- a/controllers/argocd/applicationset.go
+++ b/controllers/argocd/applicationset.go
@@ -1045,9 +1045,7 @@ func (r *ReconcileArgoCD) reconcileApplicationSetService(cr *argoproj.ArgoCD) er
 
 // Returns the name of the role/rolebinding for the source namespaces for applicationset-controller in the format of "argocdName-argocdNamespace-applicationset"
 func getResourceNameForApplicationSetSourceNamespaces(cr *argoproj.ArgoCD) string {
-	// For source namespace resources, we need namespace to ensure uniqueness across cluster
-	namespacedName := fmt.Sprintf("%s-%s", argoutil.TruncateCRName(cr.Name), cr.Namespace)
-	return argoutil.TruncateWithHash(fmt.Sprintf("%s-applicationset", namespacedName), 63)
+	return fmt.Sprintf("%s-%s-applicationset", cr.Name, cr.Namespace)
 }
 
 // removeUnmanagedApplicationSetSourceNamespaceResources cleansup resources from ApplicationSetSourceNamespaces if namespace is not managed by argocd instance.
--- a/controllers/argocd/applicationset.go
+++ b/controllers/argocd/applicationset.go
@@ -1045,7 +1045,9 @@ func (r *ReconcileArgoCD) reconcileApplicationSetService(cr *argoproj.ArgoCD) er
 
 // Returns the name of the role/rolebinding for the source namespaces for applicationset-controller in the format of "argocdName-argocdNamespace-applicationset"
 func getResourceNameForApplicationSetSourceNamespaces(cr *argoproj.ArgoCD) string {
-	return fmt.Sprintf("%s-%s-applicationset", cr.Name, cr.Namespace)
+	// For source namespace resources, we need namespace to ensure uniqueness across cluster
+	namespacedName := fmt.Sprintf("%s-%s", argoutil.TruncateCRName(cr.Name), cr.Namespace)
+	return argoutil.TruncateWithHash(fmt.Sprintf("%s-applicationset", namespacedName), 63)
 }
 
 // removeUnmanagedApplicationSetSourceNamespaceResources cleansup resources from ApplicationSetSourceNamespaces if namespace is not managed by argocd instance.
--- a/controllers/argocd/notifications.go
+++ b/controllers/argocd/notifications.go
@@ -153,12 +153,12 @@ func (r *ReconcileArgoCD) deleteNotificationsResources(cr *argoproj.ArgoCD) erro
 	sa := &corev1.ServiceAccount{}
 	role := &rbacv1.Role{}
 
-	if err := argoutil.FetchObject(r.Client, cr.Namespace, fmt.Sprintf("%s-%s", cr.Name, common.ArgoCDNotificationsControllerComponent), sa); err != nil {
+	if err := argoutil.FetchObject(r.Client, cr.Namespace, nameWithSuffix(common.ArgoCDNotificationsControllerComponent, cr), sa); err != nil {
 		if !apierrors.IsNotFound(err) {
 			return err
 		}
 	}
-	if err := argoutil.FetchObject(r.Client, cr.Namespace, fmt.Sprintf("%s-%s", cr.Name, common.ArgoCDNotificationsControllerComponent), role); err != nil {
+	if err := argoutil.FetchObject(r.Client, cr.Namespace, nameWithSuffix(common.ArgoCDNotificationsControllerComponent, cr), role); err != nil {
 		if !apierrors.IsNotFound(err) {
 			return err
 		}
@@ -606,9 +606,7 @@ func (r *ReconcileArgoCD) reconcileNotificationsServiceMonitor(cr *argoproj.Argo
 		return nil
 	}
 
-	name := fmt.Sprintf("%s-%s", cr.Name, "notifications-controller-metrics")
-	// Resource names must not exceed 63 characters
-	name = argoutil.TruncateWithHash(name, 63)
+	name := nameWithSuffix("notifications-controller-metrics", cr)
 	serviceMonitor := newServiceMonitorWithName(name, cr)
 	smExists, err := argoutil.IsObjectFound(r.Client, cr.Namespace, serviceMonitor.Name, serviceMonitor)
 	if err != nil {
@@ -890,7 +888,9 @@ func getNotificationsResources(cr *argoproj.ArgoCD) corev1.ResourceRequirements 
 
 // Returns the name of the role/rolebinding for the source namespaces for notifications-controller in the format of "argocdName-argocdNamespace-notifications"
 func getResourceNameForNotificationsSourceNamespaces(cr *argoproj.ArgoCD) string {
-	return fmt.Sprintf("%s-%s-notifications", cr.Name, cr.Namespace)
+	// For source namespace resources, we need namespace to ensure uniqueness across cluster
+	namespacedName := fmt.Sprintf("%s-%s", argoutil.TruncateCRName(cr.Name), cr.Namespace)
+	return argoutil.TruncateWithHash(fmt.Sprintf("%s-notifications", namespacedName), 63)
 }
 
 // setManagedNotificationSourceNamespaces populates ManagedNotificationsSourceNamespaces var with namespaces
--- a/controllers/argocd/notifications.go
+++ b/controllers/argocd/notifications.go
@@ -888,9 +888,7 @@ func getNotificationsResources(cr *argoproj.ArgoCD) corev1.ResourceRequirements 
 
 // Returns the name of the role/rolebinding for the source namespaces for notifications-controller in the format of "argocdName-argocdNamespace-notifications"
 func getResourceNameForNotificationsSourceNamespaces(cr *argoproj.ArgoCD) string {
-	// For source namespace resources, we need namespace to ensure uniqueness across cluster
-	namespacedName := fmt.Sprintf("%s-%s", argoutil.TruncateCRName(cr.Name), cr.Namespace)
-	return argoutil.TruncateWithHash(fmt.Sprintf("%s-notifications", namespacedName), 63)
+	return fmt.Sprintf("%s-%s-notifications", cr.Name, cr.Namespace)
 }
 
 // setManagedNotificationSourceNamespaces populates ManagedNotificationsSourceNamespaces var with namespaces
--- a/controllers/argocd/notifications.go
+++ b/controllers/argocd/notifications.go
@@ -607,6 +607,8 @@ func (r *ReconcileArgoCD) reconcileNotificationsServiceMonitor(cr *argoproj.Argo
 	}
 
 	name := fmt.Sprintf("%s-%s", cr.Name, "notifications-controller-metrics")
+	// Resource names must not exceed 63 characters
+	name = argoutil.TruncateWithHash(name, 63)
 	serviceMonitor := newServiceMonitorWithName(name, cr)
 	smExists, err := argoutil.IsObjectFound(r.Client, cr.Namespace, serviceMonitor.Name, serviceMonitor)
 	if err != nil {
--- a/controllers/argocd/notifications_test.go
+++ b/controllers/argocd/notifications_test.go
@@ -563,6 +563,42 @@ func TestReconcileNotifications_ServiceMonitorWithMetrics(t *testing.T) {
 	assert.Equal(t, monitoringv1.Duration("30s"), testServiceMonitor.Spec.Endpoints[0].ScrapeTimeout)
 }
 
+func TestReconcileNotifications_ServiceMonitorWithLongName(t *testing.T) {
+	// Test that ServiceMonitor names are truncated to 63 characters
+	a := makeTestArgoCD(func(a *argoproj.ArgoCD) {
+		a.Name = "this-name-will-push-the-char-limit" // 37 chars + 36 char suffix = 73 chars total
+		a.Spec.Notifications.Enabled = true
+	})
+
+	resObjs := []client.Object{a}
+	subresObjs := []client.Object{a}
+	runtimeObjs := []runtime.Object{}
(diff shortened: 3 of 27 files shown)


controllers/argocd/service.go
jparsai: This also need to be upated, as it is still using old name so it will not match any redis pod.

jparsai (changes requested):
Is it backward compatible? I think this doesn't rename/reuse existing StatefulSets, here is the scenario:
- Old  StatefulSet name is `<cr-name>-application-controller`
- After this PR new resource is created with name `<cr-name>-app-controller`
- Reconciler first checks if there is any existing resource with same name (currently it checks only for short name), but old resources exists with long name.
- Reconciler thinks there is no resource and create new with short name.
- Now we have two resources with both old and new name.

I have not tried it locally yet, but can you confirm if this scenario works (if you have validated already, ignore this comment)

I also searched for "argocd-application-controller" (ignored _test.go files), I still see few places where name is hardcoded. Do they also need to be updated?

jparsai:
It would also help to have an E2E, because this may break old deployments. E2E test  can create a resource manually and then operator should notice and reuse that (with old name only) and new resources should be created with short name.

Rizwana777:
> Is it backward compatible? I think this doesn't rename/reuse existing StatefulSets, here is the scenario:
> 
> * Old  StatefulSet name is `<cr-name>-application-controller`
> * After this PR new resource is created with name `<cr-name>-app-controller`
> * Reconciler first checks if there is any existing resource with same name (currently it checks only for short name), but old resources exists with long name.
> * Reconciler thinks there is no resource and create new with short name.
> * Now we have two resources with both old and new name.
> 
> I have not tried it locally yet, but can you confirm if this scenario works (if you have validated already, ignore this comment)
> 
> I also searched for "argocd-application-controller" (ignored _test.go files), I still see few places where name is hardcoded. Do they also need to be updated?

No, thanks for pointing out, backward compatibility was not taken care of. Now fixed backward compatibility for StatefulSet naming changes. The implementation now only abbreviates "application-controller" to "app-controller" when the full name exceeds 52 characters, preserving the original naming for short CR names and preventing unnecessary migration for existing deployments. Added owner reference validation before deleting legacy StatefulSets to prevent accidental deletion of resources owned by other ArgoCD instances

also I reviewed all hardcoded references no changes needed. Whichever remaining occurrences are internal identifiers (component constants, container names, volume names etc) that should stay hardcoded as they do not represent K8s resource names affected by length limits

controllers/argocd/service.go
Rizwana777: done

Rizwana777:
> It would also help to have an E2E, because this may break old deployments. E2E test can create a resource manually and then operator should notice and reuse that (with old name only) and new resources should be created with short name.

added e2e test

Rizwana777:
fixing e2e failures

controllers/argoutil/resource.go
```diff
@@ -130,7 +136,34 @@ func IsObjectFound(client client.Client, namespace string, name string, obj clie
...
-	return fmt.Sprintf("%s-%s", TruncateCRName(meta.Name), suffix)
+	fullName := fmt.Sprintf("%s-%s", TruncateCRName(meta.Name), suffix)
+	return TruncateWithHash(fullName, maxLabelLength)
+}
+
+// NameWithSuffixForStatefulSet returns a StatefulSet name that stays within the Kubernetes
+// label length limit after controller revision hash suffixes are applied.
+// Preserves the original suffix when possible to maintain backward compatibility.
+// Only abbreviates or truncates when the full name would exceed maxStatefulSetNameLength.
+func NameWithSuffixForStatefulSet(meta metav1.ObjectMeta, suffix string) string {
```
jparsai, line 147: Other than app controller, this function is also used for redis HA, that means redis HA will also have new name. Which is fine I guess, but we have same backward compatibility issue for redis HA as well.
Just like app controller if there is any existing resource with old name, your logic will not fetch it and reconciler will think that resource does not exists and create new with new name. This way we will have two STS controller with both old and new name and both will fight for same resource.

tests/ginkgo/parallel/1-140_validate_statefulset_naming_migration_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+	argov1beta1api "github.com/argoproj-labs/argocd-operator/api/v1beta1"
+	"github.com/argoproj-labs/argocd-operator/controllers/argoutil"
+	"github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture"
+	argocdFixture "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/argocd"
+	k8sFixture "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/k8s"
+	"github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/utils"
+)
+
+var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
```
jparsai, line 40: To ensure operator is fully backward compatible, E2E also needs to cover the scenario when "an old resource exists with long name". Will operator use that as it is and not create new name? or will it rename the old resource? or will it not detect that and create new with new name?

As redis STS is also affected with these changes so we need to cover same scenarios for that as well.

jparsai (changes requested):

controllers/argocd/notifications_test.go
```diff
@@ -563,6 +563,42 @@ func TestReconcileNotifications_ServiceMonitorWithMetrics(t *testing.T) {
...
+	err = r.List(context.TODO(), testServiceMonitorList, &client.ListOptions{Namespace: a.Namespace})
+	assert.NoError(t, err)
+	assert.Equal(t, 1, len(testServiceMonitorList.Items))
+
+	testServiceMonitor := testServiceMonitorList.Items[0]
+	// Verify the name is truncated to 63 characters
+	assert.LessOrEqual(t, len(testServiceMonitor.GetName()), 63, "ServiceMonitor name exceeds 63 character limit")
+	// Verify the ServiceMonitor was created successfully
+	assert.NotEmpty(t, testServiceMonitor.GetName())
+}
```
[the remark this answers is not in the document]
jparsai: +1 for this

jparsai:
When I tried it locally, with name `this-is-a-very-long-argocd-cr-name-for-testing`, the ha proxy and server pods didnt start because they are still looking for SA with full name.

```
Error creating: pods "this-is-a-very-long-argocd-cr-f73bfae-redis-ha-haproxy-5dcd5df487-" is forbidden: error looking up service account argocd-long-ns/this-is-a-very-long-argocd-cr-name-for-testing-argocd-redis-ha: serviceaccount "this-is-a-very-long-argocd-cr-name-for-testing-argocd-redis-ha" not found

Error creating: pods "this-is-a-very-long-argocd-cr-f73bfae-server-669b966bc8-" is forbidden: error looking up service account argocd-long-ns/this-is-a-very-long-argocd-cr-name-for-testing-argocd-server: serviceaccount "this-is-a-very-long-argocd-cr-name-for-testing-argocd-server" not found
```
`deploy.Spec.Template.Spec.ServiceAccountName` in redis, redis ha, dex, server deployments is still getting old name, hence SA reference does not align with actual SAs that are created with truncated names.

While searching for places resource name are used, I got few things which would break if instance with truncated name is created, this is existing code and worked fine because resource with long names were not working earlier, but long names are supported now (by shortening them) these places will break. 

- https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/prometheus.go#L308C178-L308C195 deployment name is set to `<CR Name>-server` but original deployment name is now `this-is-a-very-long-argocd-cr-f73bfae-server`
- in same file we have `<CR Name>-repo-server`, `<CR Name>-applicationset-controller`, `<CR Name>-dex-server`, -`<CR Name>-notifications-controller`, `<CR Name>--redis`

There are other places as well, if you search for `cr.Name`, you would find places which directly generate resource name with `cr.Name`-`<component name>`, there should be one common function to generate resource name and all of them should call that function instead of creating on spot, this will cause name drift for instance which have truncated name.
Not sure if it should be done in this PR or separate PR. @jgwest WDYT?

Rizwana777:
> When I tried it locally, with name `this-is-a-very-long-argocd-cr-name-for-testing`, the ha proxy and server pods didnt start because they are still looking for SA with full name.
> 
> ```
> Error creating: pods "this-is-a-very-long-argocd-cr-f73bfae-redis-ha-haproxy-5dcd5df487-" is forbidden: error looking up service account argocd-long-ns/this-is-a-very-long-argocd-cr-name-for-testing-argocd-redis-ha: serviceaccount "this-is-a-very-long-argocd-cr-name-for-testing-argocd-redis-ha" not found
> 
> Error creating: pods "this-is-a-very-long-argocd-cr-f73bfae-server-669b966bc8-" is forbidden: error looking up service account argocd-long-ns/this-is-a-very-long-argocd-cr-name-for-testing-argocd-server: serviceaccount "this-is-a-very-long-argocd-cr-name-for-testing-argocd-server" not found
> ```
> 
> `deploy.Spec.Template.Spec.ServiceAccountName` in redis, redis ha, dex, server deployments is still getting old name, hence SA reference does not align with actual SAs that are created with truncated names.
> 
> While searching for places resource name are used, I got few things which would break if instance with truncated name is created, this is existing code and worked fine because resource with long names were not working earlier, but long names are supported now (by shortening them) these places will break.
> 
> * https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/prometheus.go#L308C178-L308C195 deployment name is set to `<CR Name>-server` but original deployment name is now `this-is-a-very-long-argocd-cr-f73bfae-server`
> * in same file we have `<CR Name>-repo-server`, `<CR Name>-applicationset-controller`, `<CR Name>-dex-server`, -`<CR Name>-notifications-controller`, `<CR Name>--redis`
> 
> There are other places as well, if you search for `cr.Name`, you would find places which directly generate resource name with `cr.Name`-`<component name>`, there should be one common function to generate resource name and all of them should call that function instead of creating on spot, this will cause name drift for instance which have truncated name. Not sure if it should be done in this PR or separate PR. @jgwest WDYT?

As current PR introduced truncation for sts I think we should make these changes in current PR itself else it will break other things @jparsai @jgwest

controllers/argocd/role.go
```diff
@@ -44,12 +44,15 @@ func newRoleForApplicationSourceNamespaces(namespace string, rules []v1.PolicyRu
...
-	return cr.Name + "-" + argoComponentName
+	return argoutil.NameWithSuffix(cr.ObjectMeta, argoComponentName)
 }
 
 // GenerateUniqueResourceName generates unique names for cluster scoped resources
 func GenerateUniqueResourceName(argoComponentName string, cr *argoproj.ArgoCD) string {
-	return cr.Name + "-" + cr.Namespace + "-" + argoComponentName
+	// For cluster-scoped resources, we need to include namespace to ensure uniqueness
+	// First create name-namespace prefix, then add component suffix
+	namespacedName := fmt.Sprintf("%s-%s", argoutil.TruncateCRName(cr.Name), cr.Namespace)
```
jparsai: This can be reverted, because long cluster role nae does not cause any issue. Cluster role name is usually long only, because it has `CR Name + NS Name + Component Name`, which can easily be over 63 char, for example we have E2E test which generates name `example-argocd-argocd-e2e-cluster-config-application-controller`, and it works fine. So there is good probability that existing users of operator may have long cluster role name and without backward compatibility their existing cluster roles will be ignored and new ones will be created because of this change.

controllers/argocd/notifications.go
```diff
@@ -888,7 +888,9 @@ func getNotificationsResources(cr *argoproj.ArgoCD) corev1.ResourceRequirements
 
 // Returns the name of the role/rolebinding for the source namespaces for notifications-controller in the format of "argocdName-argocdNamespace-notifications"
 func getResourceNameForNotificationsSourceNamespaces(cr *argoproj.ArgoCD) string {
-	return fmt.Sprintf("%s-%s-notifications", cr.Name, cr.Namespace)
+	// For source namespace resources, we need namespace to ensure uniqueness across cluster
+	namespacedName := fmt.Sprintf("%s-%s", argoutil.TruncateCRName(cr.Name), cr.Namespace)
+	return argoutil.TruncateWithHash(fmt.Sprintf("%s-notifications", namespacedName), 63)
```
jparsai: same as https://github.com/argoproj-labs/argocd-operator/pull/2211/changes#r3372375673

controllers/argocd/applicationset.go
```diff
@@ -1045,7 +1045,9 @@ func (r *ReconcileArgoCD) reconcileApplicationSetService(cr *argoproj.ArgoCD) er
 
 // Returns the name of the role/rolebinding for the source namespaces for applicationset-controller in the format of "argocdName-argocdNamespace-applicationset"
 func getResourceNameForApplicationSetSourceNamespaces(cr *argoproj.ArgoCD) string {
-	return fmt.Sprintf("%s-%s-applicationset", cr.Name, cr.Namespace)
+	// For source namespace resources, we need namespace to ensure uniqueness across cluster
+	namespacedName := fmt.Sprintf("%s-%s", argoutil.TruncateCRName(cr.Name), cr.Namespace)
```
jparsai: same as https://github.com/argoproj-labs/argocd-operator/pull/2211/changes#r3372375673

jparsai (changes requested):
Everything else LGTM, found 3 places which actually dont need truncated names.

controllers/argocd/role.go
Rizwana777: done

jparsai (approved):
Thanks @Rizwana777 , LGTM

jgwest (approved):

PR #2235, opened 2026-06-16
fix(agent): Reconcile principal redis volume correctly

olivergondza:
**What type of PR is this?**

/kind bug

Fixes #?

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the operator’s reconciliation for the principal deployment to also monitor pod volumes and container volume mounts, ensuring Redis authentication volumes and mounts are automatically restored if removed.

* **Tests**
  * Added a sequential end-to-end test that deletes the Redis authentication volume and corresponding volume mount from the principal deployment, then verifies the operator reinstalls both within a short retry window.

### DIFF
--- a/controllers/argocdagent/deployment.go
+++ b/controllers/argocdagent/deployment.go
@@ -289,6 +289,21 @@ func updateDeploymentIfChanged(compName, saName string, cr *argoproj.ArgoCD, dep
 		deployment.Spec.Template.Spec.Containers[0].Args = buildArgs(compName)
 	}
 
+	redisAuthVolume, redisAuthMount := argoutil.MountRedisAuthToArgo(cr)
+	desiredVolumeMounts := append(buildVolumeMounts(), redisAuthMount)
+	desiredVolumes := append(buildVolumes(), redisAuthVolume)
+	if !reflect.DeepEqual(deployment.Spec.Template.Spec.Containers[0].VolumeMounts, desiredVolumeMounts) {
+		log.Info("deployment container volume mounts are being updated")
+		changed = true
+		deployment.Spec.Template.Spec.Containers[0].VolumeMounts = desiredVolumeMounts
+	}
+
+	if !reflect.DeepEqual(deployment.Spec.Template.Spec.Volumes, desiredVolumes) {
+		log.Info("deployment container volumes are being updated")
+		changed = true
+		deployment.Spec.Template.Spec.Volumes = desiredVolumes
+	}
+
 	if !reflect.DeepEqual(deployment.Spec.Template.Spec.Containers[0].SecurityContext, buildSecurityContext()) {
 		log.Info("deployment container security context is being updated")
 		changed = true
--- a/tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
+++ b/tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
@@ -19,6 +19,7 @@ package sequential
 import (
 	"context"
 	"fmt"
+	"slices"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -35,6 +36,7 @@ import (
 	argov1beta1api "github.com/argoproj-labs/argocd-operator/api/v1beta1"
 	"github.com/argoproj-labs/argocd-operator/common"
 	"github.com/argoproj-labs/argocd-operator/controllers/argocdagent"
+	"github.com/argoproj-labs/argocd-operator/controllers/argoutil"
 	"github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture"
 	agentFixture "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/agent"
 	argocdFixture "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/argocd"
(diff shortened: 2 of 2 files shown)


controllers/argocdagent/deployment.go
```diff
@@ -289,6 +289,21 @@ func updateDeploymentIfChanged(compName, saName string, cr *argoproj.ArgoCD, dep
 		deployment.Spec.Template.Spec.Containers[0].Args = buildArgs(compName)
 	}
 
+	redisAuthVolume, redisAuthMount := argoutil.MountRedisAuthToArgo(cr)
+	desiredVolumeMounts := append(buildVolumeMounts(), redisAuthMount)
+	desiredVolumes := append(buildVolumes(), redisAuthVolume)
+	if !reflect.DeepEqual(deployment.Spec.Template.Spec.Containers[0].VolumeMounts, desiredVolumeMounts) {
```
akhilnittala, line 295: we can make this align same as https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/deployment.go#L533 right ? what is your thought on this ?

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
```diff
@@ -511,6 +513,41 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 			}
 		})
 
+		It("Should restore Redis Auth volume and mount if removed from principal deployment", func() {
+			By("Creating ArgoCD instance normally")
+			Expect(k8sClient.Create(ctx, argoCD)).To(Succeed())
+			Eventually(principalDeployment).Should(k8sFixture.ExistByName())
```
jgwest, line 520: Can add this before the Deployment check to ensure ArgoCD is ready, similar to what we do for other tests
```
			Eventually(argoCD, "5m", "5s").Should(argocdFixture.BeAvailable())
```

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
```diff
@@ -511,6 +513,41 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
...
+
+			fmt.Printf("principalDeployment.Spec.Template.Spec.Volumes: %v\n", principalDeployment.Spec.Template.Spec.Volumes)
```
jgwest: Presume this is a debug only statement

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
```diff
@@ -511,6 +513,41 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
...
+			By("Removing Redis Auth volume and mount from the deployment")
+			err := k8sClient.Get(ctx, client.ObjectKey{Name: argoCDAgentPrincipalName, Namespace: ns.Name}, principalDeployment)
+			Expect(err).ToNot(HaveOccurred())
+
+			principalDeployment.Spec.Template.Spec.Volumes = slices.DeleteFunc(principalDeployment.Spec.Template.Spec.Volumes, func(v corev1.Volume) bool {
+				return v.Name == argoutil.RedisAuthVolumeName
+			})
+			principalDeployment.Spec.Template.Spec.Containers[0].VolumeMounts = slices.DeleteFunc(principalDeployment.Spec.Template.Spec.Containers[0].VolumeMounts, func(vm corev1.VolumeMount) bool {
+				return vm.Name == argoutil.RedisAuthVolumeName
+			})
```
jgwest: We can verify that the slices.DeleteFunc actually correctly matches one of the values in the slice (thus reducing its length), otherwise we can't be sure the test is testing what we think it is

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
```diff
@@ -511,6 +513,41 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
...
+			Expect(k8sClient.Update(ctx, principalDeployment)).To(Succeed())
```
jgwest: Direct call to k8sClient.Update will lead to intermittent E2E test failures due to concurrent modification of the resource. Can instead use `k8sFixture.Update(...)` [1] or `deploymentFixture.Update(...)` [2]

[1] grep for `k8sFixture.Update` for example
[2] Looks like we don't have code that calls the Update function in deployment fixture, but it looks the same as [1] but without needing the cast (https://github.com/argoproj-labs/argocd-operator/blob/fb26af7794d0eb75454a8a034498e10a9bae8e31/tests/ginkgo/fixture/deployment/fixture.go#L113)

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
```diff
@@ -511,6 +513,41 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
...
+
+			By("Redis Auth volume and mount are re-added by the operator")
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, client.ObjectKey{Name: argoCDAgentPrincipalName, Namespace: ns.Name}, principalDeployment)
+				if err != nil {
+					return false
+				}
+				volFound := slices.ContainsFunc(principalDeployment.Spec.Template.Spec.Volumes, func(v corev1.Volume) bool {
```
jgwest, line 546: Can instead use `deploymentFixture.HaveSpecTemplateSpecVolume` for this one

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
```diff
@@ -511,6 +513,41 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
...
+					return v.Name == argoutil.RedisAuthVolumeName
+				})
+				mountFound := slices.ContainsFunc(principalDeployment.Spec.Template.Spec.Containers[0].VolumeMounts, func(vm corev1.VolumeMount) bool {
```
jgwest, line 549: (There is no corresponding deploymentFixture function for volume mounts, so an Eventually function works well for this case)

controllers/argocdagent/deployment.go
```diff
@@ -289,6 +289,21 @@ func updateDeploymentIfChanged(compName, saName string, cr *argoproj.ArgoCD, dep
...
+		log.Info("deployment container volume mounts are being updated")
+		changed = true
+		deployment.Spec.Template.Spec.Containers[0].VolumeMounts = desiredVolumeMounts
+	}
+
+	if !reflect.DeepEqual(deployment.Spec.Template.Spec.Volumes, desiredVolumes) {
+		log.Info("deployment container volumes are being updated")
+		changed = true
+		deployment.Spec.Template.Spec.Volumes = desiredVolumes
+	}
```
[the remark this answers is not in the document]
jgwest: @coderabbitai can you open a separate issue for this? This issue is not just applicable here, but for other cases too. We can investigate outside of this PR.

As best as I can tell, the logic added by this PR is consistent with what we are doing in other parts of the code, and is thus 'fine for now'.

jgwest (changes requested):

controllers/argocdagent/deployment.go
olivergondza: Not sure I follow. Align in what aspect? This PR is making sure this is aligned with controllers/argocdagent/agent/deployment.go.

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
olivergondza: Nice, I was sad not finding those in `github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture`. Sure

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
olivergondza: True. Though since we need the Eventually-func for Volume Mounts anyway, I find it not worth of redoing.

tests/ginkgo/sequential/1-051_validate_argocd_agent_principal_test.go
olivergondza: See https://github.com/argoproj-labs/argocd-operator/pull/2235/changes#r3420585175

svghadi (approved):
LGTM

jgwest (approved):

olivergondza:
Thanks for the prompt review @jgwest, @svghadi.

PR #2237, opened 2026-06-16
feat: argocd agent rbac settings improvements

cjcocokrisp:
**What type of PR is this?**

/kind enhancement

This PR improves the RBAC settings for Argo CD Agent instances managed by the operator. It adds the following to both Principal and Agent components. 

Providing the environment variables `PRINCIPAL_CLUSTER_ROLE` and/or `AGENT_CLUSTER_ROLE` will disable the default role that is created and replace the components role binding with the self provided cluster role that's name is provided in those variables. This can be used to provide permissions that are not the default to the agent/principal.

If `.spec.DefaultClusterScopedRoleDisabled` is set on the Argo CD and the principal and/or agent is enabled, the default cluster role and cluster role binding will not reconcile. This allows the user to provide their own as well.

Both of these features exist for the reconciliation of Argo CD manifests so this brings that functionality to that. 

Unit tests are currently the only tests added for this. I can add E2E tests if needed so LMK.

Fixes #?

New environment variabes:
1. In the operator controller deployment add the env variable you would like to test.
2. Create the custom cluster role you would like to provide. Make sure it has the same name as what you put in the value of the env variable.
3. Create an Argo CD CR with Principal or Agent created.
4. Check to see if the role binding for the component uses the cluster role provided.

`DefaultClusterScopedRoleDisabled` behavior:
1. Create an Argo CD CR with either the Principal or Agent present along with `DefaultClusterScopedRoleDisabled` enabled
2. Check to see if the cluster role and cluster role binding was not created.

## Release Notes

* **New Features**
  * Added environment-variable support to select custom RBAC cluster roles for both principal and agent via `PRINCIPAL_CLUSTER_ROLE` and `AGENT_CLUSTER_ROLE`.
  * Added `DefaultClusterScopedRoleDisabled` to control whether cluster-scoped RBAC roles and bindings are created or removed.

* **Tests**
  * Added unit tests and new E2E coverage to verify custom role overrides and the create/delete behavior when toggling `DefaultClusterScopedRoleDisabled`.

* **Documentation**
  * Updated the environment variables guide to document `PRINCIPAL_CLUSTER_ROLE` and `AGENT_CLUSTER_ROLE`.


jgwest (changes requested):
What's in the PR LGTM, one request: 
- E2E tests are always good, but we should _definitely_ have them in this case, since the change we're making is on a security/tenancy boundary of argocd-operator

cjcocokrisp:
Added an e2e test for the scenario when `spec.DefaultClusterScopedRoleDisabled` is true and then disabled and re-enabled again to test all the cases it would be set.

jgwest (approved):
LGTM, thanks @cjcocokrisp!

PR #2242, opened 2026-06-22
chore(docs): Clean unused document links

olivergondza:
- We do not seem to use milestones/roadmap for years anymore

**What type of PR is this?**

/kind chore

Clean the clutter in the docs.

N/A

make serve-docs

## Summary by CodeRabbit

* **Documentation**
  * Enhanced Argo CD Notifications documentation with detailed configuration guidance and usage examples.
  * Clarified that notifications are disabled by default and documented the Kubernetes resources created when enabled.
  * Reorganized documentation site navigation for improved usability.
  * Removed auto-generated API reference documentation.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -51,8 +51,9 @@ nav:
   - Ingress: usage/ingress.md
   - Insights: usage/insights.md
   - Dex: usage/dex.md
-  - Notifications: usage/notifications.md
-  - Notifications in Any Namespace: usage/notifications-in-any-namespace.md
+  - Notifications:
+    - Basics: usage/notifications.md
+    - Notifications in Any Namespace: usage/notifications-in-any-namespace.md
   - Resource Management: usage/resource_management.md
   - Routes: usage/routes.md
   - Webhook secrets: usage/webhook-secrets.md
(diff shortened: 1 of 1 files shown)


svghadi:
I think we should find a way to auto generate the API documentation. I had looked into https://book-v1.book.kubebuilder.io/beyond_basics/generating_documentation a while ago, but I never got around to enabling it.

olivergondza:
> I think we should find a way to auto generate the API documentation. I had looked into https://book-v1.book.kubebuilder.io/beyond_basics/generating_documentation a while ago, but I never got around to enabling it.

What do we need it for?

I find the official reference documentation format a bit hard to navigate and use to serve as a substitute the hand-crafted ref-docs that we have. They target Go devs and not yaml editors. But I agree it would be awesome if we do not have to maintain CRD reference manually.

olivergondza:
Here are 2 alternatives to `gen-crd-api-reference-docs` that I find easier to grasp from the user perspective:

- https://doc.crds.dev/github.com/argoproj-labs/argocd-operator/argoproj.io/ArgoCD/v1beta1@v0.17.0
  - Hosts our docs right after release. Easier to consume.
 - https://docs.crossplane.io/v2.3/api/#Configuration-spec-commonAnnotations
   - Not so good for describing nested objects.

olivergondza:
Agreed offline not to drop the generated API. Will adjust the PR...

olivergondza:
@svghadi, this has now shrunk significantly. PTAL

svghadi (approved):
LGTM. Thanks

PR #2224, opened 2026-06-09
feat: Add support for Argo CD Commit Server

olivergondza:
**What type of PR is this?**

[//]: # (Uncomment only one  line, and delete the rest.)
[//]: # (For example,  would simply become: /kind bug  )

/kind enhancement

Co-authored-by: Cursor AI
All bugs are mine

**What does this PR do / why we need it**:

Introduces commit server and its selective enablement based on Source Hydrator use.

https://argo-cd.readthedocs.io/en/stable/user-guide/source-hydrator/

In detail:

- Add a new component for a `commit-server`, configured in `ArgoCD`.
- Configures new feature named `sourceHydrator`, configured in `ArgoCD`.
- Provides docs and tests for the above -^
- Includes E2E test verifying the actual source hydration
- Also adds `gitserver` fixture to setup throwaway git server/repository where tests can push commits for testing.

**Have you updated the necessary documentation?**

* [x] Documentation update is required by this PR.
* [x] Documentation has been updated.

**Which issue(s) this PR fixes**:

**How to test changes / Special notes to the reviewer**:

## Summary by CodeRabbit

* **New Features**
  * Added configurable **Commit Server** and **Source Hydrator** controller options, including opt-in enablement for Source Hydrator.
  * When enabled, the operator now provisions and reports Commit Server lifecycle status.
  * Updated CRD/UI metadata so new settings are discoverable.
* **Bug Fixes**
  * Improved enable/disable gating and Commit Server status transitions (Unknown/Pending/Running/Failed).
* **Documentation**
  * Added reference docs and YAML examples for the new controller options.
* **Tests**
  * Added unit coverage and expanded end-to-end scenarios for activation and reconciliation, including hydration over SSH/HTTPS.

PR #2252, opened 2026-06-23
feat: add per-component priorityClassName support for core ArgoCD components

aradhanay621:
> **Note**: This replaces #2250, which was accidentally closed due to an orphan branch caused by `git rebase --root` in a shallow clone during conflict resolution.

Closes #2125

Exposes a single `spec.priorityClassName` field in the `ArgoCD` Custom Resource, giving cluster administrators a uniform way to assign a Kubernetes PriorityClass to all ArgoCD component pods and prevent differential eviction under resource pressure.

Without this field, ArgoCD pods default to the cluster's global PriorityClass (or priority 0 if none is set), and cannot be protected from eviction.

## New API field

**`v1beta1` and `v1alpha1` (with conversion):**

| Field | Effect |
| --- | --- |
| `spec.priorityClassName` | Applied to all ArgoCD component pods |

## Example usage

```yaml
kind: ArgoCD
metadata:
  name: argocd-instance
spec:
  priorityClassName: high-priority
```

### API types

- `api/v1beta1/argocd_types.go` — `PriorityClassName string` added to top-level `ArgoCDSpec`
- `api/v1alpha1/argocd_types.go` — same for v1alpha1

### API conversion (v1alpha1 ↔ v1beta1)

- `api/v1alpha1/argocd_conversion.go` — `spec.PriorityClassName` mapped in `ConvertTo` / `ConvertFrom`
- `api/v1alpha1/argocd_conversion_test.go` — `TestPriorityClassNameConversion` covering both directions

### Controllers

- `controllers/argocd/statefulset.go` — Application Controller StatefulSet + Redis HA StatefulSet
- `controllers/argocd/deployment.go` — Redis Deployment, Server Deployment, Redis HA HAProxy Deployment
- `controllers/argocd/repo_server.go` — Repo Server Deployment
- `controllers/argocd/applicationset.go` — ApplicationSet Controller Deployment
- `controllers/argocd/dex.go` — Dex Deployment
- `controllers/argocd/util.go` — `reconcileDeploymentHelper` (covers Notifications + Image Updater)
- `controllers/argocdagent/deployment.go` — ArgoCD Agent principal deployment
- `controllers/argocdagent/agent/deployment.go` — ArgoCD Agent deployment

### Tests

Unit tests covering both initial creation and the update-detection (reconcile) path for all components.

### CRDs

- `config/crd/bases/`, `bundle/manifests/`, and `deploy/olm-catalog/` updated

## Testing

All unit tests pass. Codecov reports 100% coverage on all new lines.

### DIFF
--- a/api/v1alpha1/argocd_types.go
+++ b/api/v1alpha1/argocd_types.go
@@ -839,6 +839,10 @@ type ArgoCDSpec struct {
 	// NodePlacement defines NodeSelectors and Taints for Argo CD workloads
 	NodePlacement *ArgoCDNodePlacementSpec `json:"nodePlacement,omitempty"`
 
+	// PriorityClassName is the name of the PriorityClass resource to be assigned to all ArgoCD component pods.
+	// +optional
+	PriorityClassName string `json:"priorityClassName,omitempty"`
+
 	// Notifications defines whether the Argo CD Notifications controller should be installed.
 	Notifications ArgoCDNotifications `json:"notifications,omitempty"`
 
--- a/api/v1beta1/argocd_types.go
+++ b/api/v1beta1/argocd_types.go
@@ -1027,6 +1027,10 @@ type ArgoCDSpec struct {
 	// NodePlacement defines NodeSelectors and Taints for Argo CD workloads
 	NodePlacement *ArgoCDNodePlacementSpec `json:"nodePlacement,omitempty"`
 
+	// PriorityClassName is the name of the PriorityClass resource to be assigned to all ArgoCD component pods.
+	// +optional
+	PriorityClassName string `json:"priorityClassName,omitempty"`
+
 	// Notifications defines whether the Argo CD Notifications controller should be installed.
 	Notifications ArgoCDNotifications `json:"notifications,omitempty"`
 
--- a/tests/ginkgo/sequential/1-133_validate_priorityclassname_test.go
+++ b/tests/ginkgo/sequential/1-133_validate_priorityclassname_test.go
@@ -1,0 +1,206 @@
+/*
+Copyright 2025 ArgoCD Operator Developers
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sequential
+
+import (
+	"context"
+	"sort"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	schedulingv1 "k8s.io/api/scheduling/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	argoproj "github.com/argoproj-labs/argocd-operator/api/v1beta1"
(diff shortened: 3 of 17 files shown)


olivergondza:
@aradhanay621, thanks for the contribution! What is the use-case for configuring this independently for individual components?

I am concerned this leaves room for omission for `ArgoCD` admins, causing few argo pods will be evicted or scheduled late producing hard to understand outages. Contrary, I see value in the operator making sure that all the pods are using uniform priority class without the admin's responsibility to enumerate them all, and keep them in sync.

olivergondza (changes requested):
As discussed offline, adjust to configure collectively for all components.

api/v1alpha1/argocd_types.go
```diff
@@ -842,6 +847,12 @@ type ArgoCDSpec struct {
 	// Notifications defines whether the Argo CD Notifications controller should be installed.
 	Notifications ArgoCDNotifications `json:"notifications,omitempty"`
 
+	// PriorityClassName is the name of the PriorityClass resource to assign to all ArgoCD
+	// component pods (Application Controller, ApplicationSet Controller, Dex, Redis,
+	// Repo Server, and ArgoCD Server). The PriorityClass must already exist in the cluster.
```
olivergondza: Do not document the affected components. It is going to get out of sync easily.

api/v1alpha1/argocd_types.go
```diff
@@ -842,6 +847,12 @@ type ArgoCDSpec struct {
...
+	// +optional
+	PriorityClassName string `json:"priorityClassName,omitempty"`
```
olivergondza, line 844: Anybody have an idea how to represent this? Nested structure / field name? It feels somewhat arbitrary to have this on the top level. There is nothing wrong with this, just making sure we do not miss an opportunity to provide better api.

api/v1beta1/argocd_types.go
```diff
@@ -1025,6 +1030,12 @@ type ArgoCDSpec struct {
 	// Notifications defines whether the Argo CD Notifications controller should be installed.
 	Notifications ArgoCDNotifications `json:"notifications,omitempty"`
 
+	// PriorityClassName is the name of the PriorityClass resource to assign to all ArgoCD
+	// component pods (Application Controller, ApplicationSet Controller, Dex, Redis,
+	// Repo Server, and ArgoCD Server). The PriorityClass must already exist in the cluster.
```
olivergondza: Same as above

olivergondza (changes requested):
To make sure no components are overlooked, also inject in:

- `reconcileRedisHAProxyDeployment` @ `controllers/argocd/deployment.go`
- `reconcileRedisStatefulSet` @ `controllers/argocd/statefulset.go`
- `reconcileDeploymentHelper` @ `controllers/argocd/util.go`
- `argocdagent/` package
- `argocdagent/agent` package

olivergondza:
Thanks. Please add an E2E test verifying that all pods started by the operator, have the desired priority class configured. That will be a great safety net for optional components, or components added in the future (like https://github.com/argoproj-labs/argocd-operator/pull/2224). I find it easy to miss something provisioned by the operator.

tests/ginkgo/sequential/1-133_validate_priorityclassname_test.go
```diff
@@ -0,0 +1,274 @@
...
+					},
+				},
+			}
+			Expect(k8sClient.Create(ctx, argoCD)).To(Succeed())
+
+			By("waiting for ArgoCD CR to be reconciled and the instance to be ready")
+			Eventually(argoCD, "5m", "5s").Should(argocdFixture.BeAvailable())
+
+			By("verifying all Deployments in the namespace have the expected priorityClassName")
+			Eventually(func() error {
```
olivergondza: Extract this into a reusable function, so it does not need to be repeated. It can even be reused by the next `It`. Then the assertion can be written like:

```go
var allDeploymentPriorityClasses = func() {
...
}

...

Eventually(allDeploymentPriorityClasses).Should(Equal([]string{testPriorityClassName}))
```

olivergondza (approved):
Thanks a lot for implementing this!

PR #2272, opened 2026-07-08
docs: Generate apidocs for readthedocs.org

olivergondza:
**What type of PR is this?**

/kind documentation

Provides up-to-date generated CRD apidocs.

- Remove the obsolete committed apidocs, generate new ones on the fly.
  - Remove unused custom templates
- New `make apidocs-gen` to run locally, hooked into `make serve-docs`.
- Fix the reference from manually created docs, the urls have changed.
- Document its process and the development procedure.

## Summary by CodeRabbit

* **Documentation**
  * Removed the “WIP” label from the development guide and expanded ReadTheDocs development instructions.
  * Updated API documentation navigation and Notifications installation links.
  * Set the documentation site URL for ReadTheDocs.

* **Build & Docs Automation**
  * API reference documentation is now generated automatically before documentation builds and local serving.
  * Added configurable Kubernetes API version support for generated references.

* **Chores**
  * Excluded generated API documentation from version control and removed obsolete documentation templates.


olivergondza:
~Example passing build: https://app.readthedocs.org/projects/olivergondza-argocd-op/builds/33535557/
The apidocs: https://olivergondza-argocd-op.readthedocs.io/en/latest/reference/api/~

olivergondza:
@jgwest, @anandf, @svghadi, can you take a look? This re-implements the apidoc generation.

docs/usage/notifications.md
```diff
@@ -16,7 +16,8 @@ spec:
     enabled: True
 ```
 
-Users may also specify advanced configuration such as the number of replicas for the notifications controller, as well as the resource requirements for the same. The full list of available settings can be found in the [API spec](../reference/api.html.md#argoproj.io/v1alpha1.ArgoCDNotificationsSpec).
+Users may also specify advanced configuration such as the number of replicas for the notifications controller, as well as the resource requirements for the same. The full list of available settings can be 
+found in the [API spec](/reference/api/#argocdnotifications).
```
chengfang: relative path `../reference/api/#argocdnotifications` should be used above, as in the existing link.

The above link with absolute path does not work in the sample build

Makefile
```diff
@@ -385,6 +386,16 @@ ginkgo: ## Download ginkgo locally if necessary.
 update-dependencies:
 	hack/update-dependencies-script/run.sh
 
+.PHONY: apidocs-gen
+apidocs-gen: ## Generate API documentation.
+	# The config have the k8s version injected so it does not have to be updated there
+	go run github.com/elastic/crd-ref-docs@v0.3.0 \
+		--config=<(sed 's/__KUBERNETES_API_VERSION__/$(KUBERNETES_API_VERSION)/' ./docs/crd-ref-docs.config.yaml) \
+		--source-path=./api/v1beta1/ \
```
chengfang: what about api/v1alpha1-only types https://github.com/argoproj-labs/argocd-operator/tree/master/api/v1alpha1, like ArgoCDExport and NotificationsConfiguration?

docs/crd-ref-docs.config.yaml
```diff
@@ -0,0 +1,5 @@
+# Configuration for https://github.com/elastic/crd-ref-docs
+
+render:
+  # Interpolated by `make crd-ref-docs`
```
chengfang: `make apidocs-gen`

Makefile
```diff
@@ -41,7 +41,8 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
 # Set the Operator SDK version to use.
 # This is useful for CI or a project to utilize a specific version of the operator-sdk toolkit.
-OPERATOR_SDK_VERSION ?= v1.35.0
+KUBERNETES_API_VERSION ?= 1.35
+OPERATOR_SDK_VERSION ?= v$(KUBERNETES_API_VERSION).0
```
chengfang: I think KUBERNETES_API_VERSION and OPERATOR_SDK_VERSION should be independent from each other, and the existing way of declaring OPERATOR_SDK_VERSION is fine.

KUBERNETES_API_VERSION in go.mod is k8s.io/kubernetes v1.36.1, and we may want to keep them consistent.

Makefile
olivergondza: The tools cannot generate APIs of the same name in different version - it ends up overwriting one by the other. I figured we are ok with the more recent API.

If we need both - and thanks for pointing out there are types in `alpha` only - we can generate it in sibling files with 2 invocations.

Makefile
olivergondza: > I think KUBERNETES_API_VERSION and OPERATOR_SDK_VERSION should be independent from each other

Is there an intent, or just something we do not enforce to be in sync? What is the motivation for independence?

It make sense to me to keep all three (k8s.io/kubernetes in go.mod, KUBERNETES_API_VERSION and OPERATOR_SDK_VERSION) fed from the same source unless there is a well understood reason/use-case for them to differ.

On an implementation side of things, `OPERATOR_SDK_VERSION` needs the form of `X.Y.Z` while `crd-ref-docs` mandates `X.Y`. Partially a reason why I laid things out the way I did.

docs/usage/notifications.md
olivergondza: Hmm, you are spot on. Let me see if I can find a way for `make serve-docs` works more like the actual build - I changed this because local build was broken.

docs/crd-ref-docs.config.yaml
olivergondza: Thx

olivergondza:
Example passing build: https://app.readthedocs.org/projects/olivergondza-argocd-op/builds/34008406/
The apidocs:
- https://olivergondza-argocd-op.readthedocs.io/en/latest/reference/api-v1beta1/
- https://olivergondza-argocd-op.readthedocs.io/en/latest/reference/api-v1alpha1/

docs/usage/notifications.md
olivergondza: Addressed so it works both locally and on readthedocs.org

Makefile
chengfang: From what I see, KUBERNETES_API_VERSION and OPERATOR_SDK_VERSION are totally separate, so we should not base one value on the other.

For now there is no automated way to keep go.mod and Makefile in sync for shared deps, which is fine. My suggestion is basically leave OPERATOR_SDK_VERSION unchanged if not used by this pr, and add KUBERNETES_API_VERSION needed by this pr to point to the same value as in go.mod: 1.36.1

Makefile
chengfang: I think we do both to be complete. If we can add a second section or file from a 2nd invocation, that will be great.

Makefile
olivergondza: Hmm, you are correct, the versioning of the projects is separate.

Makefile
olivergondza: It is reasonably simple to do. Added

olivergondza:
Example passing build: https://app.readthedocs.org/projects/olivergondza-argocd-op/builds/34019000/
The apidocs: https://olivergondza-argocd-op.readthedocs.io/en/latest/reference/api/

mkdocs.yml
```diff
@@ -1,5 +1,5 @@
 site_name: Argo CD Operator
-site_url: ""
+site_url: https://argocd-operator.readthedocs.io/en/latest/
```
olivergondza, line 2: This turned out to be crucial so local run of `make serve-docs` runs things on `http://0.0.0.0:8000/en/latest/` ensuring links are handled and resolved the same during development and in production build.

chengfang (approved):

PR #2273, opened 2026-07-09
fix: Make Renewal Token Expiry Feature Configurable

akhilnittala:
**What type of PR is this?**
/kind bug

Making Token renewal feature configurable, feature will be implemented only when user sets spec.EnableSATokenRenewal to true in argocd CR. By default it is false, so approach is non-expiry legacy tokens.

Fixes #?
https://redhat.atlassian.net/browse/GITOPS-10364
> deploy gitops and default argocd CR will have nothing, so default behavior is vi non-expiry token legacy
> set spec.enableDexTokenExpiry to true to test expiry token renewal feature.

## Summary by CodeRabbit

* **New Features**
  * Added `spec.sso.dex.enableSATokenRenewal` to enable short-lived Dex OAuth tokens using the Kubernetes TokenRequest API; default remains the legacy non-expiring token approach.
* **Bug Fixes**
  * Improved Dex secret reconciliation to use the appropriate secret strategy based on the new setting.
  * Refined proactive renewal requeue behavior to occur only when Dex is enabled and the renewal feature is on.
  * Made legacy Dex token secret cleanup more selective to avoid disrupting valid secret references.
* **Documentation**
  * Updated the CRD and operator manifests to document `enableSATokenRenewal`.


api/v1beta1/argocd_types.go
```diff
@@ -254,6 +254,11 @@ type ArgoCDCertificateSpec struct {
 
 // ArgoCDDexSpec defines the desired state for the Dex server component.
 type ArgoCDDexSpec struct {
+	// EnableDexTokenExpiry enables the short-lived Dex token renewal feature.
+	// When true, the operator uses TokenRequest API for time-limited tokens.
+	// When false (default), the operator uses the legacy non-expiring token approach.
+	// +kubebuilder:validation:Optional
+	EnableDexTokenExpiry *bool `json:"enableDexTokenExpiry,omitempty"`
```
anandf: DexToken might be confusing. I think what we mean here is service account token for Dex workload. right ?

anandf (changes requested):

api/v1beta1/argocd_types.go
akhilnittala: dex token expiry feature enabling

svghadi: How about `enableServiceAccountTokenRenewal` or `enableSATokenRenewal`? We can omit Dex since the field is already defined under `.spec.sso.dex`.

controllers/argocd/dex.go
```diff
@@ -76,8 +76,27 @@ func needsDexTokenRenewal(secret *corev1.Secret) bool {
 	return time.Until(expiry) < dexServerTokenRenewalThreshold()
 }
 
+// isDexTokenExpiryFeatureEnabled returns true if Dex token renewal expiry is enabled.
+// Users can set enableDexTokenExpiry: false (default) in the ArgoCD CR to use the legacy non-expiring token approach.
+func isDexTokenExpiryFeatureEnabled(cr *argoproj.ArgoCD) bool {
```
svghadi: Lets rename it to `isDexSATokenExpiryFeatureEnabled`

svghadi (commented):
Can you add some more unit or e2e tests?

api/v1beta1/argocd_types.go
akhilnittala: added

controllers/argocd/dex.go
akhilnittala: done

svghadi (approved):
LGTM. Thanks

anandf (approved):

PR #2276, opened 2026-07-09
feat: add label selector support for Argo CD Agent principal and agent components

jparsai:
The hybrid architecture of Argo CD Agent requires label selectors for proper isolation of agent-managed and local workloads. This PR adds a first-class `labelSelector` field to both PrincipalSpec and AgentSpec, which gets translated to the ARGOCD_PRINCIPAL_LABEL_SELECTOR and ARGOCD_AGENT_LABEL_SELECTOR environment variables respectively.

/kind enhancement

Assisted by: Cursor

## Summary by CodeRabbit

* **New Features**
  * Added label selector settings for both the agent and principal so you can limit which resources they watch and process.
  * Updated deployment behavior to pass the configured selector into runtime settings for each component.

* **Bug Fixes**
  * Ensured label selector values are preserved when converting and validating configurations across supported versions.

* **Documentation**
  * Updated configuration schemas and examples to show the new selector option and its expected usage.

### DIFF
--- a/api/v1alpha1/argocd_conversion.go
+++ b/api/v1alpha1/argocd_conversion.go
@@ -816,6 +816,7 @@ func ConvertAlphaToBetaPrincipal(src *PrincipalSpec) *v1beta1.PrincipalSpec {
 			LogLevel:      src.LogLevel,
 			LogFormat:     src.LogFormat,
 			Image:         src.Image,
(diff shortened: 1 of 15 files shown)


jgwest (approved):
LGTM, thanks @jparsai!

PR #2281, opened 2026-07-14
feat(redis-ha-proxy): Configure Redis Ha Proxy TLS settings

akhilnittala:
**What type of PR is this?**
/kind chore

Configure TLS settings for redis ha proxy.

Fixes #?
https://redhat.atlassian.net/browse/GITOPS-10289
> deploy redis ha proxy pods with ha.enabled to true in argocd CR.

## Summary by CodeRabbit

* **Security**
  * When TLS is enabled, Redis HAProxy connections for Sentinel health checks and master selection now explicitly use TLS.
  * Generated HAProxy TLS configuration now honors centralized settings for minimum TLS version and cipher suites, including dedicated handling for TLS 1.3 cipher suites.
* **Bug Fixes**
  * Redis HAProxy config generation now correctly applies the centralized TLS profile values.
* **Tests**
  * Updated Redis HA config map expectations to match the new TLS-aware output and added unit coverage for rendered TLS directives and TLS/cipher value transformations.


controllers/argocd/configmap.go
```diff
@@ -782,10 +782,12 @@ func (r *ReconcileArgoCD) reconcileRedisHAHealthConfigMap(cr *argoproj.ArgoCD, u
 
 // reconcileRedisHAConfigMap ensures the Redis HA ConfigMap is correctly reconciled.
 func (r *ReconcileArgoCD) reconcileRedisHAConfigMap(cr *argoproj.ArgoCD, useTLSForRedis bool) error {
+	tlsMinVersion := string(r.CentralTLSConfigProfile.MinVersion)
```
anandf: Can this be nil or empty ?

build/redis/haproxy.cfg.tpl
```diff
@@ -1,6 +1,27 @@
...
+{{- if eq .tlsMinVersion "1.3"}}
+    # TLS 1.3 cipher suites
+    ssl-default-bind-ciphersuites {{.tlsCiphers}}
+    ssl-default-server-ciphersuites {{.tlsCiphers}}
+{{- else}}
+    # TLS 1.2 and below cipher lists
+    ssl-default-bind-ciphers {{.tlsCiphers}}
+    ssl-default-server-ciphers {{.tlsCiphers}}
+
+    # Also configure TLS 1.3 cipher suites when TLS 1.3 is negotiated
```
anandf, line 20: Why adding TLS 1.3 cipher suites even when minVersion is 1.2 ?

controllers/argoutil/redis.go
```diff
@@ -185,14 +185,35 @@ func GetRedisInitScript(cr *argoproj.ArgoCD, useTLSForRedis bool) string {
 	return script
 }
 
+func TLSVersionToHAProxy(tlsVersion string) string {
+	versionMap := map[string]string{
```
anandf: Why not use a switch here  instead of map. Also there is already a similar function available at `controllers/argoutil/tls.go:133` Is this function doing anything different ? Can the same function be re-used instead of creating a new one ?

anandf (changes requested):

controllers/argocd/configmap.go
akhilnittala: it can be empty on non-openshift-clusters, but is handled not to pass to command line arguments in this code 
tlsVersion := TLSVersionToHAProxy(tlsMinVersion)
	if tlsVersion != "" {
		vars["tlsMinVersion"] = tlsVersion
	}

build/redis/haproxy.cfg.tpl
akhilnittala: when minversion is set to 1.2 means minversion is 1.2 and maxversion as it is unset can be compatible for 1.3 as well,
something like if tls1.2 is minversion it is compatible for 1.2 and 1.3, but if tls1.3 is minversion then only compatible for 1.3

controllers/argoutil/redis.go
akhilnittala: removed, and used the same

build/redis/haproxy.cfg.tpl
akhilnittala: If ssl-max-ver is not set, HAProxy imposes no upper limit on the TLS version. It will negotiate the highest TLS version supported by both:

controllers/argoutil/redis.go
```diff
@@ -320,20 +330,19 @@ func GetRedisServerAddress(cr *argoproj.ArgoCD) string {
...
 		return "", fmt.Errorf("unable to parse template. error: %w", err)
 	}
 
 	buf := new(bytes.Buffer)
-	err = tmpl.Execute(buf, params)
-	if err != nil {
+	if err := tmpl.Execute(buf, params); err != nil {
 		log.Error(err, "unable to execute template")
 		return "", fmt.Errorf("unable to execute template. error: %w", err)
 	}
```
svghadi, line 343: Any particular reason to make this change or it was just a refactor?

controllers/argocd/configmap.go
```diff
@@ -782,10 +782,11 @@ func (r *ReconcileArgoCD) reconcileRedisHAHealthConfigMap(cr *argoproj.ArgoCD, u
 
 // reconcileRedisHAConfigMap ensures the Redis HA ConfigMap is correctly reconciled.
 func (r *ReconcileArgoCD) reconcileRedisHAConfigMap(cr *argoproj.ArgoCD, useTLSForRedis bool) error {
+	tlsCiphers := r.CentralTLSConfigProfile.Ciphers
 	ctx := context.TODO()
 	desired := newConfigMapWithName(common.ArgoCDRedisHAConfigMapName, cr)
 	desired.Data = map[string]string{
-		"haproxy.cfg":     argoutil.GetRedisHAProxyConfig(cr, useTLSForRedis),
+		"haproxy.cfg":     argoutil.GetRedisHAProxyConfig(cr, useTLSForRedis, r.CentralTLSConfigProfile.MinVersion, tlsCiphers),
```
svghadi:
```suggestion
		"haproxy.cfg":     argoutil.GetRedisHAProxyConfig(cr, useTLSForRedis, r.CentralTLSConfigProfile),
```
Can we directly pass the profile object?

svghadi (commented):
Looks good overall. Left a suggestion.

controllers/argoutil/redis.go
akhilnittala: to add unit test, did some refactor.

controllers/argocd/configmap.go
akhilnittala: done sid.

svghadi (approved):

controllers/argoutil/redis_test.go
```diff
@@ -0,0 +1,248 @@
...
+// are correctly rendered in the final HAProxy configuration template output
+func TestGetRedisHAProxyConfigRenderedTLSValues(t *testing.T) {
+	tests := []struct {
+		name                    string
+		useTLS                  bool
+		centralTLSConfigProfile tlsprofile.TLSConfigProfile
+		expectedInOutput        []string
+		notExpectedInOut        []string
+		validatePattern         *regexp.Regexp
+		description             string
```
olivergondza: Can we drop the field and replace with a comment? I see it is unused...

controllers/argoutil/redis_test.go
```diff
@@ -0,0 +1,248 @@
...
+			},
+			expectedInOutput: []string{
+				"ssl-default-bind-options ssl-min-ver TLSv1.3",
+				"ssl-default-server-options ssl-min-ver TLSv1.3",
+				"ssl-default-bind-ciphersuites TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256",
+				"ssl-default-server-ciphersuites TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256",
+			},
+			notExpectedInOut: []string{
+				"1.2",
+				"VersionTLS13",
```
olivergondza: What is "VersionTLS13" verifying?

controllers/argoutil/redis_test.go
```diff
@@ -0,0 +1,248 @@
...
+			var capturedVars map[string]string
+
+			original := loadTemplateFile
+			defer func() {
+				loadTemplateFile = original
+			}()
+
+			loadTemplateFile = func(path string, vars map[string]string) (string, error) {
+				capturedVars = maps.Clone(vars)
+				return renderMockTemplate(vars), nil
```
olivergondza: What is the motivation for mocking the template? I find it dangerous to test a different template than used in production.

controllers/argoutil/redis_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+			// Call the function
+			result := GetRedisHAProxyConfig(cr, tt.useTLS, tt.centralTLSConfigProfile)
+
+			// Validate captured variables exist as expected
+			t.Logf("Test: %s\nCaptured vars: %+v\nRendered output:\n%s\n", tt.name, capturedVars, result)
+
+			// Check expected values are in output
+			for _, expected := range tt.expectedInOutput {
+				if !strings.Contains(result, expected) {
```
olivergondza: Can be simplified with `assert.Contains`.

controllers/argoutil/redis_test.go
```diff
@@ -0,0 +1,248 @@
...
+					t.Errorf("expected %q to be in rendered output, but it wasn't\nOutput:\n%s", expected, result)
+				}
+			}
+
+			// Check values that should NOT be in output
+			for _, notExpected := range tt.notExpectedInOut {
+				if strings.Contains(result, notExpected) {
+					t.Errorf("did not expect %q to be in rendered output, but it was\nOutput:\n%s", notExpected, result)
```
olivergondza: Can be simplified with `assert.Contains`.

controllers/argoutil/redis_test.go
```diff
@@ -0,0 +1,248 @@
...
+			if tt.useTLS {
+				if tlsVersion := capturedVars["tlsMinVersion"]; tt.centralTLSConfigProfile.MinVersion != "" && TLSProtocolVersionString(tt.centralTLSConfigProfile.MinVersion) != "" {
+					if tlsVersion != TLSProtocolVersionString(tt.centralTLSConfigProfile.MinVersion) {
+						t.Errorf("tlsMinVersion var = %q, want %q", tlsVersion, TLSProtocolVersionString(tt.centralTLSConfigProfile.MinVersion))
+					}
+				}
+
+				if len(tt.centralTLSConfigProfile.Ciphers) > 0 {
+					expectedCiphers := strings.Join(tt.centralTLSConfigProfile.Ciphers, ":")
+					if ciphers := capturedVars["tlsCiphers"]; ciphers != expectedCiphers {
```
olivergondza: Can be simplified with `assert.Equal`.

controllers/argoutil/redis_test.go
```diff
@@ -0,0 +1,248 @@
...
+			// Check values that should NOT be in output
+			for _, notExpected := range tt.notExpectedInOut {
+				if strings.Contains(result, notExpected) {
+					t.Errorf("did not expect %q to be in rendered output, but it was\nOutput:\n%s", notExpected, result)
+				}
+			}
+
+			// Check regex pattern if provided
+			if tt.validatePattern != nil {
+				if !tt.validatePattern.MatchString(result) {
```
olivergondza: `assert.Regexp`

controllers/argoutil/redis_test.go
```diff
@@ -0,0 +1,248 @@
...
+				"ssl-default-server-options",
+				"ssl-default-bind-ciphers",
+				"ssl-default-server-ciphers",
+				"ssl-default-bind-ciphersuites",
+				"ssl-default-server-ciphersuites",
+			},
+			description: "Verify no TLS directives when TLS is disabled",
+		},
+		{
+			name:   "Multiple ciphers are colon-separated",
```
olivergondza: nit: IMO, this is verified by other tests.

controllers/argoutil/redis.go
```diff
@@ -187,12 +188,20 @@ func GetRedisInitScript(cr *argoproj.ArgoCD, useTLSForRedis bool) string {
...
-func GetRedisHAProxyConfig(cr *argoproj.ArgoCD, useTLSForRedis bool) string {
+func GetRedisHAProxyConfig(cr *argoproj.ArgoCD, useTLSForRedis bool, centralTLSConfigProfile tlsprofile.TLSConfigProfile) string {
 	path := fmt.Sprintf("%s/haproxy.cfg.tpl", getRedisConfigPath())
 	vars := map[string]string{
 		"ServiceName": NameWithSuffix(cr.ObjectMeta, "redis-ha"),
 		"UseTLS":      strconv.FormatBool(useTLSForRedis),
 	}
+	tlsVersion := TLSProtocolVersionString(centralTLSConfigProfile.MinVersion)
+	if tlsVersion != "" {
+		vars["tlsMinVersion"] = tlsVersion
```
olivergondza: nit: Stick with uppercase as the rest of the variables.

olivergondza (changes requested):
The change looks good to me on a high level. See the comments...

controllers/argoutil/redis_test.go
akhilnittala: which field you are referring to @olivergondza ?

controllers/argoutil/redis_test.go
akhilnittala: checking that that pattern of VersionTLS13 should not be there, it is something like negative test.

controllers/argoutil/redis_test.go
akhilnittala: i am using the same template, but only first few lines, not the entire template. But the content is same as haproxy config tpl.

controllers/argoutil/redis_test.go
akhilnittala: done, added

controllers/argoutil/redis_test.go
akhilnittala: done, added

controllers/argoutil/redis_test.go
akhilnittala: done,added

controllers/argoutil/redis_test.go
akhilnittala: done,added

controllers/argoutil/redis_test.go
akhilnittala: yeah, removed

controllers/argoutil/redis.go
akhilnittala: modified

controllers/argoutil/redis_test.go
olivergondza: The `description`.

controllers/argoutil/redis_test.go
olivergondza: I understand the implementation, what I am wondering about is the intent. It makes sense to check `1.2`  content is not there. But there is never string like `VersionTLS` injected in the config file, plus this is checking for 1.3

controllers/argoutil/redis_test.go
olivergondza: That is the problem, the validity of this test is dependent on the fact the two will never diverge - and there is no good way to ensure that. What is being tested is not the production code and production template, but the mocks.

controllers/argoutil/redis_test.go
akhilnittala: okay, actually we added for what the test is doing, but looks like redundant, removed it.

controllers/argoutil/redis_test.go
akhilnittala: yeah oliver

svghadi (approved):

controllers/argoutil/redis_test.go
akhilnittala: made changes.lets see the CI pipeline.

controllers/argoutil/redis_test.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+
+			var capturedVars map[string]string
+			original := loadTemplateFile
+			defer func() {
+				loadTemplateFile = original
+			}()
+			loadTemplateFile = func(path string, vars map[string]string) (string, error) {
+				capturedVars = maps.Clone(vars)
+				data, err := os.ReadFile(path)
```
olivergondza: Let's call the `return original(path, vars)` here, instead of reimplementing the function. So ultimately this will become a wrapper to capture the vars.

akhilnittala: modified.

olivergondza (approved):
Looks good to me. Thanks!

Ping @anandf, any more comments?

anandf (approved):
LGTM

PR #2284, opened 2026-07-15
feat(configmap): mask the token value annotation in the ArgoCD UI/CLI on OpenShift

dkarpele:
**What type of PR is this?**

/kind enhancement

On GitOPS operator on Openshift any secret of type "kubernetes.io/dockercfg" has the annotation "openshift.io/token-secret.value" with the token in clear that can be seen by the administrator of the gitops project and the token could be used to have access to unwanted resources, these secrets should be automatically hidden instead of requiring manual addition.

no

https://redhat.atlassian.net/browse/RFE-7499

Fixes #?

The e2e will follow in gitops operator repo

## Summary by CodeRabbit

* **New Features**
  * Added support for automatically masking sensitive `metadata.annotations` values in the Argo CD UI and CLI by honoring a new sensitive-mask annotation key on OpenShift.

* **Bug Fixes**
  * Ensures the masking entry is added or backfilled during Argo CD ConfigMap reconciliation on OpenShift.
  * Avoids duplicate entries and appends the token-secret masking value only when missing, without overwriting existing configuration.

* **Tests**
  * Added reconciliation tests covering OpenShift vs non-OpenShift behavior, backfill for existing ConfigMaps, override/append handling, and de-duplication.

### DIFF
--- a/common/keys.go
+++ b/common/keys.go
@@ -195,6 +195,9 @@ const (
 	// ArgoCDKeyUsersAnonymousEnabled is the configuration key for anonymous user access.
 	ArgoCDKeyUsersAnonymousEnabled = "users.anonymous.enabled"
 
+	// ArgoCDKeyResourceSensitiveMaskAnnotations is the key to mask metadata.annotations values in UI/CLI on Secrets.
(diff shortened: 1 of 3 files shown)


olivergondza (approved):
Please flip error assertions to use `require`. Looks good to me otherwise.

PR #2295, opened 2026-07-22
refactor: update dependency upgrade script to be modular

cjcocokrisp:
**What type of PR is this?**

/kind code-refactoring

At the current moment, to update Argo CD dependencies a Go script is used. As we add more components into the operator having this script be easy to add onto will bring a lot of benefit. 

This PR refactors the update dependency script to partition dependencies into directories where the Go script for updating that dependency lives. The root bash script that was used to run the Go script has been replaced with a script that will just update all dependencies in the directory as long as they contain a `run.sh` script. 

There is also a Makefile target for just Argo CD. 

I plan to use this new refactored version for pulling dependencies for the GitOps Promoter because there is a lot of CRDs for it so the image updater approach of just having a Makefile target that copies the file over was not enough. 

Fixes #?

Run the Makefile targets below:
`make update-dependencies`
`make update-dependencies-argocd`

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated command to update only Argo CD-related dependencies.
  * Improved dependency update execution by discovering and running component-specific update scripts with progress output.
* **Refactor**
  * Centralized shared dependency-management utilities to make update steps more consistent and uniformly handle failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -385,6 +385,11 @@ ginkgo: ## Download ginkgo locally if necessary.
 update-dependencies:
 	hack/update-dependencies-script/run.sh
 
+# Updates only Argo CD dependencies
+.PHONY: update-dependencies-argocd
+update-dependencies-argocd:
+	hack/update-dependencies-script/argocd/run.sh
+
 .PHONY: serve-docs
 serve-docs: ## Serve documentation locally using mkdocs in a container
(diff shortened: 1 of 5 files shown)


jgwest (approved):
LGTM, thanks @cjcocokrisp!

PR #2296, opened 2026-07-22
fix: prevent disallowed NamespaceManagement from breaking OpenShift  GitOps login

aali309:
**What type of PR is this?**

/kind bug 

A disallowed `NamespaceManagement` CR (points at an ArgoCD instance that does not allow managing that namespace) was returning an error from `reconcileNamespaceManagement`, which aborted the entire ArgoCD reconcile before SSO/secrets ran.

This PR records the policy denial on the `NamespaceManagement` CR status only and continues ArgoCD reconcile so Dex token renewal (and other SSO/secret work) still runs.

Fixes #?

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of disallowed namespace-management requests to record the reason in the CR’s status condition rather than failing reconciliation.
  * Prevented disallowed namespace-management cases from duplicating status updates.
  * Ensured disallowed configurations no longer block Argo CD availability or Dex token renewal.

* **Tests**
  * Updated namespace-management reconciliation expectations for “not permitted” scenarios.
  * Added a unit test to validate behavior when status updates fail.
  * Added a new sequential E2E test to confirm Dex token renewal continues even with a disallowed NamespaceManagement CR present.

### DIFF
--- a/tests/ginkgo/sequential/1-133_validate_namespacemanagement_does_not_block_dex_token_renewal_test.go
+++ b/tests/ginkgo/sequential/1-133_validate_namespacemanagement_does_not_block_dex_token_renewal_test.go
@@ -1,0 +1,230 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sequential
+
+import (
+	"context"
+	"os"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	argov1beta1api "github.com/argoproj-labs/argocd-operator/api/v1beta1"
+	"github.com/argoproj-labs/argocd-operator/common"
+	"github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture"
+	argocdFixture "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/argocd"
+	deploymentFixture "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/deployment"
+	k8sFixture "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/k8s"
+	secretFixture "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/secret"
+	fixtureUtils "github.com/argoproj-labs/argocd-operator/tests/ginkgo/fixture/utils"
+)
+
+var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
+
+	Context("1-133_validate_namespacemanagement_does_not_block_dex_token_renewal", func() {
+
+		var (
+			k8sClient            client.Client
+			ctx                  context.Context
+			cleanupArgoNamespace func()
+			cleanupManagedNS     func()
+			argoNamespace        *corev1.Namespace
+			managedNamespace     *corev1.Namespace
+			restoreOperatorEnv   func()
+		)
+
+		BeforeEach(func() {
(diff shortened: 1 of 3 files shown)


chengfang (approved):

tests/ginkgo/sequential/1-133_validate_namespacemanagement_does_not_block_dex_token_renewal_test.go
```diff
@@ -0,0 +1,230 @@
...
+			} else {
+				operatorDeployment := &appsv1.Deployment{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "argocd-operator-controller-manager",
+						Namespace: "argocd-operator-system",
+					},
+				}
+				err := k8sClient.Get(ctx, client.ObjectKeyFromObject(operatorDeployment), operatorDeployment)
+				if err != nil {
+					Skip("Operator deployment not found - test requires operator running in cluster: " + err.Error())
```
olivergondza: Skip only for "not found error", other error should fail the test. I am concerned that if this regresses, we will not notice because the test will start skipping silently.

tests/ginkgo/sequential/1-133_validate_namespacemanagement_does_not_block_dex_token_renewal_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+				originalEnvValue, _ := deploymentFixture.GetEnv(operatorDeployment, "manager", common.EnableManagedNamespace)
+				restoreOperatorEnv = func() {
+					By("restoring original operator NamespaceManagement env var")
+					if originalEnvValue != nil {
+						deploymentFixture.SetEnv(operatorDeployment, "manager", common.EnableManagedNamespace, *originalEnvValue)
+					} else {
+						deploymentFixture.RemoveEnv(operatorDeployment, "manager", common.EnableManagedNamespace)
+					}
+					time.Sleep(30 * time.Second)
```
olivergondza: Do we need the sleep, when we have `Eventually`? Why?

tests/ginkgo/sequential/1-133_validate_namespacemanagement_does_not_block_dex_token_renewal_test.go
```diff
@@ -0,0 +1,230 @@
...
+					Eventually(operatorDeployment, "3m", "5s").Should(deploymentFixture.HaveReadyReplicas(1))
+				}
+
+				deploymentFixture.SetEnv(operatorDeployment, "manager", common.EnableManagedNamespace, "true")
+				time.Sleep(30 * time.Second)
```
olivergondza: Do we need the sleep, when we have `Eventually`? Why?

tests/ginkgo/sequential/1-133_validate_namespacemanagement_does_not_block_dex_token_renewal_test.go
```diff
@@ -0,0 +1,230 @@
...
+			Eventually(tokenSecret).Should(secretFixture.HaveNonEmptyKeyValue("expiry"))
+
+			By("creating a NamespaceManagement CR that targets this ArgoCD but is not allowed")
+			nm := &argov1beta1api.NamespaceManagement{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-protect",
+					Namespace: managedNamespace.Name,
+				},
+				Spec: argov1beta1api.NamespaceManagementSpec{
+					ManagedBy: argoNamespace.Name,
```
olivergondza, line 159: Can we use some arbitrary namespace here? IIUC, it can be anything. The argocd namespace have a special handling. Even though the test is currently valid as it is, future changes to the function may surface brittleness.

tests/ginkgo/sequential/1-133_validate_namespacemanagement_does_not_block_dex_token_renewal_test.go
aali309: Agreed, the sleep looks redundant.

tests/ginkgo/sequential/1-133_validate_namespacemanagement_does_not_block_dex_token_renewal_test.go
aali309: Thanks, removing it, the sleep looks redundant.

tests/ginkgo/sequential/1-133_validate_namespacemanagement_does_not_block_dex_token_renewal_test.go
aali309: I thought that `ManagedBy` must be the namespace of the ArgoCD instance. Otherwise reconcile will be skiped?

olivergondza: Duh, you are right. Nevermind

olivergondza (approved):

olivergondza:
Can use some more eyes on this before merging. Thanks!

keithchong:
@anandf do you have spare cycles to review this?

anandf (approved):

PR #2303, opened 2026-07-27
feat: gate openshfitOAuth with openshfit check and add logs

anandrkskd:
assisted-by: claude-code

**What type of PR is this?**
/kind enhancement 

Currently we have no check in sso to warn user openshiftOAuth is only supported on Openshift. This PR adds checks that logs if users uses this field on non-openshfit platforms. Users will be notified via logs if it is unsupported.
 
Fixes #?

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to mark SSO as failed when OpenShift OAuth is enabled on a non-OpenShift cluster.
  * Invalid SSO configurations now surface clear errors and a failed SSO status.
  * OpenShift OAuth remains supported on OpenShift clusters.

* **Tests**
  * Expanded SSO coverage across OpenShift and non-OpenShift environments.
  * Added end-to-end validation for unsupported configurations and status reporting.
  * Verified default token renewal behavior remains intact.

### DIFF
--- a/controllers/argocd/sso.go
+++ b/controllers/argocd/sso.go
@@ -55,7 +55,7 @@ func (r *ReconcileArgoCD) reconcileSSO(cr *argoproj.ArgoCD, argocdStatus *argopr
 			// https://github.com/argoproj-labs/argocd-operator/pull/615 ==> conflict
 			errMsg = "must supply valid dex configuration when requested SSO provider is dex"
 			isError = true
-		} else if cr.Spec.SSO.Dex != nil && cr.Spec.SSO.Dex.OpenShiftOAuth && !IsOpenShiftCluster() {
+		} else if cr.Spec.SSO.Dex.OpenShiftOAuth && !IsOpenShiftCluster() {
 			errMsg = "OpenShift OAuth is not supported on non-OpenShift clusters"
 			isError = true
 		} else if cr.Spec.SSO.Keycloak != nil {
--- a/controllers/argocd/sso.go
+++ b/controllers/argocd/sso.go
@@ -55,12 +55,13 @@ func (r *ReconcileArgoCD) reconcileSSO(cr *argoproj.ArgoCD, argocdStatus *argopr
 			// https://github.com/argoproj-labs/argocd-operator/pull/615 ==> conflict
 			errMsg = "must supply valid dex configuration when requested SSO provider is dex"
 			isError = true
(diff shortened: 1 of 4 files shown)


controllers/argocd/sso.go
```diff
@@ -55,6 +55,9 @@ func (r *ReconcileArgoCD) reconcileSSO(cr *argoproj.ArgoCD, argocdStatus *argopr
 			// https://github.com/argoproj-labs/argocd-operator/pull/615 ==> conflict
 			errMsg = "must supply valid dex configuration when requested SSO provider is dex"
 			isError = true
```
akhilnittala, line 57: I feel this condition can be moved to first.

controllers/argocd/sso_test.go
```diff
@@ -127,10 +129,49 @@ func TestReconcile_illegalSSOConfiguration(t *testing.T) {
 			Err:                      errors.New("illegal SSO configuration: Unsupported SSO provider type. Supported provider is dex"),
 			wantSSOConfigLegalStatus: "Failed",
 		},
+		{
```
akhilnittala, line 132: are we missing test for non openshift cluster ginkgo framework ?

controllers/argocd/sso.go
anandrkskd: In the 1st if we are checking the the input validity which is more important.

controllers/argocd/sso_test.go
anandrkskd: added e2e tests.

controllers/argocd/sso.go
```diff
@@ -60,6 +60,10 @@ func (r *ReconcileArgoCD) reconcileSSO(cr *argoproj.ArgoCD, argocdStatus *argopr
...
 		}
 
+		if cr.Spec.SSO.Dex != nil && cr.Spec.SSO.Dex.OpenShiftOAuth && !IsOpenShiftCluster() {
+			log.Info(fmt.Sprintf("warning: OpenShift OAuth is not supported on non-OpenShift clusters for Argo CD %s in namespace %s", cr.Name, cr.Namespace))
+		}
+
```
svghadi: Do we want to stop the sso reconciliation at this point?

anandrkskd: We can make the sso fail if the user set OpenshiftOAuth to true, But do we want it to make it a hard failure, warning user about the miss configuration. 
WDYT?

PR #2315, opened 2026-08-03
fix: Resolve build failures from pre-existing issues and log.Error misuse

AruneshDwivedi:
Two fixes:

1. Remove non-existent EnableSATokenRenewal field from ginkgo test that causes go vet to fail
2. Convert printf-style log.Error calls to logr key-value pairs in networkpolicies.go
3. Use == 0 instead of <= 0 for len(SCMProviders) check

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error logging for network-policy operations across Argo CD components, making resource names and namespaces clearer.
  * Clarified an internal SCM provider check without changing behavior.

* **Tests**
  * Updated Dex token-renewal test coverage to rely on the default token-renewal behavior while continuing to validate OpenShift OAuth.


chengfang:
looks like all applicable changes are in another pr #2316 , I'm closing this one.

PR #2316, opened 2026-08-04
fix: convert logr printf-style to key-value pairs in networkpolicies

### LINKED ISSUE #2271: Fix logr.Logger.Error() misuse with printf-style formatting in networkpolicies.go
### Summary

`logr.Logger.Error()` does not support printf-style formatting (`%s`, etc.) in its message argument. It expects a static message string followed by structured key-value pairs. Several call sites in `controllers/argocd/networkpolicies.go` pass a format string with `%s` placeholders and extra positional arguments directly to `log.Error()`, which will not interpolate as intended and will instead be treated as (invalid/odd) key-value pairs.

### Affected file

`controllers/argocd/networkpolicies.go`

### Occurrences (as of PR #2224)

Multiple call sites use the incorrect pattern, e.g.:

```go
log.Error(err, "Failed to update %s network policy in namespace %s", existing.Name, cr.Namespace)
```

Found at (approximate) lines: 243, 257, 348, 362, 716, 730, 796, 810, 891, 905, 1033, 1047.

A couple of related call sites already work around this by pre-formatting the message with `fmt.Sprintf` (lines ~1119, ~1131), which does technically produce a correct message string but is still not idiomatic for structured logging with logr.

### Required change

Update all affected `log.Error(...)` calls to use logr's key-value pair convention, e.g.:

```go
log.Error(err, "Failed to update network policy", "name", existing.Name, "namespace", cr.Namespace)
```

This should be applied consistently across all identified call sites in `controllers/argocd/networkpolicies.go`, and the two `fmt.Sprintf`-based call sites should also be converted to the key-value style for consistency.

### Rationale

- Ensures error messages and metadata are logged correctly and are queryable/structured rather than silently mis-rendered.
- Keeps logging style consistent with the rest of the codebase's use of the `logr` API.

### Acceptance criteria

- All `log.Error()` calls in `controllers/argocd/networkpolicies.go` that currently pass `%s`/printf-style arguments are converted to use key-value pairs instead.
- No `fmt.Sprintf` is used to pre-build the message string for `log.Error()` calls in this file.

### References

- PR: https://github.com/argoproj-labs/argocd-operator/pull/2224
- Originating review comment: https://github.com/argoproj-labs/argocd-operator/pull/2224#discussion_r3540854313
- Reported by: @olivergondza (based on a catch by @chengfang)

AruneshDwivedi:
logr.Logger.Error() expects a static message string followed by structured key-value pairs. Several call sites in controllers/argocd/networkpolicies.go were passing printf-style format strings with %s placeholders and extra positional arguments directly to log.Error(), which would not interpolate as intended and instead be treated as invalid key-value pairs.

Converted all 14 affected call sites to use logr's key-value pair convention, and replaced the two fmt.Sprintf-based workarounds with the same pattern for consistency.

Fixes #2271

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network policy error logging with clearer, structured resource names and namespaces.
  * No changes to error handling or application behavior.

### DIFF
--- a/controllers/argocd/networkpolicies.go
+++ b/controllers/argocd/networkpolicies.go
@@ -233,7 +233,7 @@ func (r *ReconcileArgoCD) ReconcileDexServerNetworkPolicy(cr *argoproj.ArgoCD) e
 		if modified {
 			argoutil.LogResourceUpdate(log, existing, "updating", explanation)
 			if err := r.Update(context.TODO(), existing); err != nil {
-				log.Error(err, "Failed to update %s network policy in namespace %s", existing.Name, cr.Namespace)
+				log.Error(err, "Failed to update network policy", "name", existing.Name, "namespace", cr.Namespace)
 				return fmt.Errorf("failed to update %s network policy in namespace %s. error: %w", existing.Name, cr.Namespace, err)
 			}
 		}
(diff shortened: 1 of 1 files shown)


chengfang:
@AruneshDwivedi pls rebase and fix DCO check failure.

olivergondza (approved):

AruneshDwivedi:
Rebased onto latest master and added DCO sign-off. The underlying fix is unchanged.

PR #1735, opened 2025-05-15
Rename the volume name from 'extensions' to 'rollout-extension'

Rizwana777:
https://issues.redhat.com/browse/GITOPS-6662 - JIRA

PR #2329, opened 2026-08-14
feat: add support for the gitops-promoter

cjcocokrisp:
**What type of PR is this?**

/kind enhancement

This PR adds in support for the [GitOps Promoter](https://github.com/argoproj-labs/gitops-promoter) project into the operator. This project is used for environment promotions between branches. 

Features:
- Reconciliation of the following components for the Promoter
  - Controller and its associated resources. This is the component that actually does the promotions and such.
  - API Server and its associated resources. The API Server uses [Kubernetes API Aggregation](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/) and is meant to be interacted with through the GitOps Promoter's CLI binary. Can also be disabled.
  - Argo CD UI extension. Will show up if you deploy promoter resources through an application. 
- Certs for the API Server. Certs need to be manually provided via a secret reference at the moment. These certs are not client facing and are only the API aggregation connection.
- Webhook enablement through a configurable setting that will deploy its exposure as a service.
- Added a script that will pull all of the CRDs for the Promoter to keep them updated.
- Added E2E tests that test the life cycle of the promoter deployment.

Future Work:
- At the moment, multi tenancy will be hard because the API Server uses Kubernetes API Aggregation. Need to do more research on this and potentially propose changing to not use that in the upstream if we decide that we want multi tenancy.
- AutoTLS for the API Server.
- Support for the checksum for the Argo CD UI Extension download. At the moment when the extension installer downloads the checksum file it includes two checksums: one for the actual extension and one for the SBOM of the extension. The installer concats the two which causes the check to fail so I need to push a fix for that upstream after investigating more.
- Add E2E tests for actual promotion process and tests that test it with the Source Hydrator

(The above list is subject to change through the review process and as I look into things more just wanted to throw a dump of some things I want to improve on this in future PRs)

(Will add them at the beginning of next week)

Fixes #?

- Build a custom image of the operator.
- Deploy it to a cluster
- Create an `ArgoCD` with fields for the promoter enabled
- See if it works correctly
  - The tutorials in the GitOps Promoter docs are good
  - I also tested with the operator deployed source hydrator as well.

## Summary by CodeRabbit

* **New Features**
  * Added GitOps Promoter configuration for Argo CD, including controller and API server deployment, resources, images, environment variables, TLS, webhooks, service types, and UI extension support.
  * Added lifecycle management for Promoter services, permissions, API services, and configuration resources.
  * Added Kubernetes resources for repositories, SCM providers, promotion strategies, commit statuses, pull requests, scheduled and timed checks, web requests, and change transfers.
* **Documentation**
  * Documented environment variables for configuring shared Promoter cluster roles.
* **Chores**
  * Added tooling to update GitOps Promoter resources and dependencies.


cjcocokrisp:
@coderabbitai full review

api/v1beta1/argocd_types.go
```diff
@@ -1781,3 +1777,59 @@ func (r *ArgoCDRouteSpec) UseExternalCertificate() bool {
...
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// APIServer defines the configuration for the promoter's API server
+	APIServer *PromoterAPIServerSpec `json:"apiserver,omitempty"`
+
+	// WebhookEnabled controls whether the controllers webhook functionality is enabled
+	WebhookEnabled bool `json:"webhookEnabled,omitempty"`
+
+	// WebhookServiceType defines what service type the webhook service will be. If none is provided defaults to ClusterIP
+	WebhookServiceType string `json:"webhookServiceType,omitempty"`
```
olivergondza: Can we introduce a new object for webhook-related config rather than have two scalar keys in the top level? If we are reasonably certain there will not be more keys needed in the future, maybe we can treat `nil` here as "webhook is disabled".

controllers/gitopspromoter/apiservice.go
```diff
@@ -0,0 +1,154 @@
+// Copyright 2026 ArgoCD Operator Developers
```
olivergondza, line 1: Since the gitops promoter is newly written, it would be great to structure it based on function and not the underlying openshift kind. This is a oddity we hope to address in the rest of the codebase at some point.

api/v1beta1/argocd_types.go
cjcocokrisp: Yeah that sounds good, will get to that along with the docs and fixing all the CI errors tomorrow.

controllers/gitopspromoter/apiservice.go
cjcocokrisp: Need to do some more research on how the promoter works and seeing if proposing instead of using the API Aggregation just making it a CRD to the upstream. This is just going to make true multi-tenancy impossible for the Promoter. 

There also could be an Openshiftism that would solve this too but I do not know of any ATM.

common/defaults.go
```diff
@@ -324,6 +324,12 @@ vs-ssh.visualstudio.com ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC7Hr1oTWqNqOlzGJOf
 	ArgoCDDefaultClusterDomain = "cluster.local"
 	// ArgoCDDefaultWebTerminalEnabled is the default web terminal enabled switch.
 	ArgoCDDefaultWebTerminalEnabled = "false"
+
+	// GitOpsPromoterDefaultImageName is the default image name for the GitOps Promoter
+	GitOpsPromoterDefaultImageName = "quay.io/argoprojlabs/gitops-promoter:v0.35.0"
+
+	// GitopsPromoterExtensionInstallerName is the default url for the GitOps Promoter extension
+	GitopsPromoterExtensionURL = "https://github.com/argoproj-labs/gitops-promoter/releases/download/v0.35.0/gitops-promoter-argocd-extension.tar.gz"
```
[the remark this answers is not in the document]
cjcocokrisp: At the current moment, the checksum file for the GitOps promoter includes multiple entries with the extension file name in it. The first is for the actual extension tar and the second is the SBOM for it. Because they get concatenated the check always fails. A fix will need to be submitted to the extension repo and then the check sums can be implemented here and for the Argo Rollouts extension as well.

config/crd/bases/promoter.argoproj.io_revertcommits.yaml
```diff
@@ -0,0 +1,54 @@
...
+            description: RevertCommitSpec defines the desired state of RevertCommit
+            properties:
+              foo:
+                description: Foo is an example field of RevertCommit. Edit revertcommit_types.go
+                  to remove/update
+                type: string
+            type: object
+          status:
+            description: RevertCommitStatus defines the observed state of RevertCommit
+            type: object
```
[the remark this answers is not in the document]
cjcocokrisp: I think that this is fine to keep for now as an example. I'm assuming that compiling your own RevertCommit spec would require the user to upload and manage a custom image for the promoter. I think that we might want to also ship an opinionated RevertCommit CRD at some point once we learn more about what they do because the docs do not have much ATM.

controllers/gitopspromoter/deployment.go
```diff
@@ -0,0 +1,477 @@
...
+			deployment.Spec.Template.Spec.Volumes = cfg.volumes
+			changed = true
+		}
+
+		actualVolumeMounts := sliceEmptyToNil(deployment.Spec.Template.Spec.Containers[0].VolumeMounts)
+		desiredVolumeMounts := sliceEmptyToNil(cfg.volumeMounts)
+		if !reflect.DeepEqual(actualVolumeMounts, desiredVolumeMounts) {
+			deployment.Spec.Template.Spec.Containers[0].VolumeMounts = cfg.volumeMounts
+			changed = true
+		}
```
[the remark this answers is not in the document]
cjcocokrisp: This is intended. To allow the user to provide and change the args to what they desire. A lot of the promoter's options are not accessible via environment variables so this is what will be needed for now.

api/v1beta1/argocd_types.go
```diff
@@ -1781,3 +1777,69 @@ func (r *ArgoCDRouteSpec) UseExternalCertificate() bool {
 	}
 	return false
 }
+
+// PromoterSpec defines the desired state for the GitOps Promoter
+type PromoterSpec struct {
+	// Enabled defines whether gitops promoter controller should be deployed or not
```
jgwest: Can add a comment on saying the default is to be disabled if nil (presuming this is true)

And courtesy of Claude:
```
  6. PromoterSpec.Enabled missing omitempty (argocd_types.go:1782). All other *bool fields use json:"...,omitempty". This one doesn't, making it inconsistent and effectively required in
     JSON.
```

api/v1beta1/argocd_types.go
```diff
@@ -1781,3 +1777,69 @@ func (r *ArgoCDRouteSpec) UseExternalCertificate() bool {
...
+	// CABundleSecretName is the name of the secret holding the CA bundle for the API Server's API Service
+	CABundleSecretName string `json:"caSecretName,omitempty"`
+
+	// CABundleSecretKey is the name of the key that holds the CA bundle for the API Server's API Service (defaults to "ca.crt")
+	CABundleSecretKey string `json:"caSecretKey,omitempty"`
+}
+
+// PromoterControllerWebhookSpec defines the Webhook options for the GitOps Promoter's controller
+type PromoterControllerWebhookSpec struct {
+	// Enabled defines whether the webhook is enabled for the Promoter's controller
```
jgwest: As above, can add comment indicating default

bundle/manifests/argocd-operator.clusterserviceversion.yaml
```diff
@@ -1878,8 +1920,13 @@ spec:
           resources:
           - apiservices
           verbs:
+          - create
```
jgwest: JGW: what is an apiservice?

bundle/manifests/argocd-operator.clusterserviceversion.yaml
```diff
@@ -1990,6 +2037,21 @@ spec:
           - patch
           - update
           - watch
+        - apiGroups:
```
jgwest: JGW: double check these are consistent with upstream

bundle/manifests/promoter.argoproj.io_clusterscmproviders.yaml
```diff
@@ -0,0 +1,265 @@
+apiVersion: apiextensions.k8s.io/v1
```
jgwest: JGW: Verify there is a good mechanism for updating these

config/rbac/role.yaml
```diff
@@ -152,6 +157,21 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
```
jgwest: JGW: check upstream

controllers/argocd/deployment.go
```diff
@@ -1102,25 +1101,37 @@ func (r *ReconcileArgoCD) reconcileServerDeployment(cr *argoproj.ArgoCD, useTLSF
...
+	if rolloutsUIEnabled {
 		deploy.Spec.Template.Spec.InitContainers = append(deploy.Spec.Template.Spec.InitContainers, getRolloutInitContainer()...)
+	} else {
+		deploy.Spec.Template.Spec.InitContainers = removeInitContainer(deploy.Spec.Template.Spec.InitContainers, "rollout-extension")
+	}
+
+	if promoterUIEnabled {
+		deploy.Spec.Template.Spec.InitContainers = append(deploy.Spec.Template.Spec.InitContainers, getPromoterInitContainer()...)
+	} else {
+		deploy.Spec.Template.Spec.InitContainers = removeInitContainer(deploy.Spec.Template.Spec.InitContainers, "promoter-extension")
```
jgwest: JGW: magic constant

controllers/argocd/deployment.go
```diff
@@ -1102,25 +1101,37 @@ func (r *ReconcileArgoCD) reconcileServerDeployment(cr *argoproj.ArgoCD, useTLSF
 
 	deploy.Spec.Template.Spec.Volumes = serverVolumes
 
-	const rolloutsVolumeName = "rollout-extensions"
-	if cr.Spec.Server.EnableRolloutsUI {
+	rolloutsUIEnabled := cr.Spec.Server.EnableRolloutsUI
+	promoterUIEnabled := cr.Spec.Promoter.IsEnabled() && cr.Spec.Promoter.ArgoCDUIExtensionEnabled
+
+	const extensionsVolumeName = "extensions"
```
jgwest: Shouldn't use generic volume name of `extensions`; we [used to use that name](https://github.com/argoproj-labs/argocd-operator/pull/1735) but stopped as it tended to clash with users also using that name. 

Just need something _slightly_ less generic, e.g. off the top of my head, `operator-extensions`, `argo-cd-operator-extensions`, etc etc.

controllers/argocd/deployment.go
```diff
@@ -1366,7 +1376,7 @@ func getRolloutInitContainer() []corev1.Container {
 			Name: "rollout-extension",
 			VolumeMounts: []corev1.VolumeMount{
 				{
-					Name:      "rollout-extensions",
+					Name:      "extensions",
```
jgwest: JGW: why rename

controllers/argocd/deployment.go
```diff
@@ -1378,6 +1388,8 @@ func getRolloutInitContainer() []corev1.Container {
 		},
 	}
 
+	containers[0].SecurityContext.RunAsUser = ptr.To(int64(1000))
```
jgwest: In other cases in the code where we are setting RunAsUser to 1000, we are explicitly ensuring that we are NOT running on OpenShift (implying this won't work there), see for example use of RunAsUser in `reconcileRedisDeployment` in the same file.

controllers/argocd/deployment.go
```diff
@@ -1386,15 +1398,51 @@ func getRolloutInitContainer() []corev1.Container {
...
+				{
+					Name:      "tmp",
+					MountPath: "/tmp",
+				},
+			},
+			SecurityContext: argoutil.DefaultSecurityContext(),
+		},
+	}
+
+	containers[0].SecurityContext.RunAsUser = ptr.To(int64(1000))
```
jgwest: As above, likely non-OpenShift

controllers/argocd/deployment.go
```diff
@@ -1386,15 +1398,51 @@ func getRolloutInitContainer() []corev1.Container {
...
+	} else {
+		containers[0].Image = common.ArgoCDExtensionInstallerImage
+		containers[0].Env = []corev1.EnvVar{
+			{
+				Name:  "EXTENSION_URL",
+				Value: common.GitopsPromoterExtensionURL,
+			},
+		}
 	}
+	// TODO: look into adding the checksums for extensions installation. Could even add it to the rollouts one too.
```
jgwest: Rather than (or in addition to) a TODO, it's likely better to open a GitHub issue for this.  (Presuming this was not just a reminder to handle this in the context of THIS PR)

controllers/argocd/deployment_test.go
```diff
@@ -1702,25 +1706,25 @@ func TestReconcileServer_RolloutUI(t *testing.T) {
 	assert.Equal(t, "rollout-extension", deployment.Spec.Template.Spec.InitContainers[0].Name)
 	assert.Equal(t, common.ArgoCDExtensionInstallerImage, deployment.Spec.Template.Spec.InitContainers[0].Image)
 
-	// assert that rollout-extensions volume is mounted at /tmp/extensions for both the initContainer and container
+	// assert that extensions volume is mounted at /tmp/extensions for both the initContainer and container
```
jgwest: JGW-TODO: hmm

controllers/argocd/deployment_test.go
```diff
@@ -2096,7 +2100,7 @@ func TestReconcileArgoCD_reconcileRedisDeployment_testImageUpgrade(t *testing.T)
 	resObjs := []client.Object{cr}
 	subresObjs := []client.Object{cr}
 	runtimeObjs := []runtime.Object{}
-	sch := makeTestReconcilerScheme(argoproj.AddToScheme)
+	sch := makeTestReconcilerScheme(argoproj.AddToScheme, promoter.AddToScheme, apiregistrationv1.AddToScheme)
```
jgwest: JGW: I don't suppose there is an easier way than to add this addtoscheme everywhere

controllers/argocd/localusers_test.go
```diff
@@ -1,5 +1,4 @@
-// Copyright 2025 ArgoCD Operator Developers
-//
+// Copyright 2025 ArgoCD Operator Developers//
```
jgwest: Looks like second line got backspaced into first, here

go.mod
```diff
@@ -43,48 +46,49 @@ require (
...
+	google.golang.org/genproto v0.0.0-20250303144028-a0af3efb3deb // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20260803160001-6ac0973c030d // indirect
+	k8s.io/streaming v0.36.2 // indirect
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.36.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v6 v6.4.2 // indirect
 )
 
 require (
-	github.com/argoproj-labs/argocd-image-updater v1.3.0
+	github.com/argoproj-labs/argocd-image-updater v1.2.1
```
jgwest: Downgrade?

go.mod
```diff
@@ -21,20 +21,23 @@ require (
...
+	k8s.io/client-go v0.36.2
 	k8s.io/kube-aggregator v0.36.1
-	k8s.io/utils v0.0.0-20260210185600-b8788abfbbc2
+	k8s.io/utils v0.0.0-20260707023825-cf1189d6abe3
 	sigs.k8s.io/controller-runtime v0.24.1
 	sigs.k8s.io/yaml v1.6.0
 )
 
-require github.com/openshift/library-go v0.0.0-20260703081820-c6cd1a243d2d
+require (
```
jgwest: JGW: Hmm

tests/ginkgo/fixture/agent/fixture.go
```diff
@@ -393,62 +392,6 @@ func VerifyLogs(deploymentName, namespace string, requiredMessages []string) {
 	}, "120s", "5s").Should(BeTrue(), "Agent should process cluster cache updates")
 }
 
-type certificateRequest struct {
```
jgwest: JGW: hmm

controllers/gitopspromoter/rolebinding.go
```diff
@@ -0,0 +1,279 @@
...
+			clusterRoleBinding.RoleRef = expectedRoleRef
+
+			argoutil.LogResourceUpdate(log, clusterRoleBinding, fmt.Sprintf("promoter cluster role binding %s has the wrong subject or role ref", bindingName))
+			if err := client.Update(context.Background(), clusterRoleBinding); err != nil {
+				return nil, fmt.Errorf("failed to update promoter cluster role %s: %v", clusterRoleBinding.Name, err)
+			}
+			return clusterRoleBinding, nil
+		}
+
+		// RoleRef field is immutable so a deletion and recreation is required
```
jgwest, line 122: Flagged by Claude, likewise flagged  by Code Rabbit above it looks like:
```
RoleRef immutability handling is dead code (rolebinding.go:132). In ReconcilePromoterClusterRoleBinding, the first if block (line 119) checks subjects OR roleRef differ and does an
     Update — but Kubernetes rejects RoleRef updates (it's immutable). The delete+recreate block at line 132 is unreachable because the first block already returns. Fix: check subjects and
     roleRef separately — update for subjects, delete+recreate for roleRef.
```

controllers/gitopspromoter/rolebinding.go
```diff
@@ -0,0 +1,279 @@
...
+			roleBinding.Subjects = expectedSubjects
+
+			argoutil.LogResourceUpdate(log, roleBinding, fmt.Sprintf("promoter cluster role binding %s has the wrong subject or role ref", bindingName))
+			if err := client.Update(context.Background(), roleBinding); err != nil {
+				return nil, fmt.Errorf("failed to update promoter cluster role %s: %v", roleBinding.Name, err)
+			}
+			return roleBinding, nil
+		}
+
+		// RoleRef field is immutable so a deletion and recreation is required
```
jgwest, line 191: Flagged by claude, same as above:
```Same pattern in ReconcilePromoterRoleBinding (rolebinding.go:202). If both subjects and roleRef differ, only subjects get updated; the roleRef change is silently dropped.```

File: controllers/argocd/deployment.go [hunk 26]
```diff
@@ -1386,15 +1398,51 @@ func getRolloutInitContainer() []corev1.Container {
...
+			SecurityContext: argoutil.DefaultSecurityContext(),
+		},
+	}
+
+	containers[0].SecurityContext.RunAsUser = ptr.To(int64(1000))
+
+	if value, exists := os.LookupEnv(common.ArgoCDExtensionImageEnvName); exists {
+		containers[0].Image = value
+	} else {
+		containers[0].Image = common.ArgoCDExtensionInstallerImage
```
jgwest, line 1466: If I'm not mistaken, these two are incorrectly referring to the argo cd extension, and instead should be referring to new env var/constant for gitops promoter image. (Or if this is intentionally, add a comment explaining why)

Courtesy of claude:
```
Shared env var for extension images (deployment.go:1411). Both getPromoterInitContainer and getRolloutInitContainer read ARGOCD_EXTENSION_INSTALLER_IMAGE. Setting it for one extension
     breaks the other. The promoter should have its own env var (e.g., GITOPS_PROMOTER_EXTENSION_IMAGE).
```

controllers/gitopspromoter/deployment.go
```diff
@@ -0,0 +1,478 @@
...
+// ReconcilePromoterAPIServerDeployment reconciles the Promoter's API Server Deployment
+// Calls the generic ReconcilePromoterDeployment function with the settings for the API Server
+func ReconcilePromoterAPIServerDeployment(client client.Client, compName string, sa *corev1.ServiceAccount, cr *argoproj.ArgoCD, scheme *runtime.Scheme) (*appsv1.Deployment, error) {
+	cfg := createAPIServerConfig()
+	enabled := cr.Spec.Promoter == nil || cr.Spec.Promoter.APIServer.IsEnabled()
+	if cr.Spec.Promoter != nil && cr.Spec.Promoter.APIServer != nil && cr.Spec.Promoter.APIServer.TLS != nil && cr.Spec.Promoter.APIServer.TLS.CertSecretName != "" {
+		cfg.volumes = buildAPIServerVolumes(cr)
+		cfg.volumeMounts = buildAPIServerVolumeMounts()
+	} else {
+		log.Info("Warning: no TLS cert for the API Server specified, api server may fail to start.")
```
jgwest, line 106: Courtesy of claude:
```
 TLS warning on every reconcile (deployment.go:106). The log.Info("Warning: no TLS cert...") fires every cycle, even when the promoter is disabled. Gate it with an IsEnabled check.```

File: common/keys.go [hunk 28]
```diff
@@ -262,6 +262,12 @@ const (
 	// ArgoCDAgentClusterRoleEnvName is an environment variable to specify a custom cluster role for the agent component of Argo CD Agent
 	ArgoCDAgentClusterRoleEnvName = "AGENT_CLUSTER_ROLE"
 
+	// GitOpsPromoterControllerClusterRoleEnvNmae is an environment variable to specify a custom cluster role for the GitOps Promoter Controller
```
jgwest: `GitOpsPromoterControllerClusterRoleEnvNmae` -> `GitOpsPromoterControllerClusterRoleEnvName` (`Nmae`)

tests/ginkgo/sequential/1-134_validate_gitops_promoter_test.go
```diff
@@ -0,0 +1,479 @@
...
+	promoterControllerResourceName              = "test-promoter-controller-manager"
+	promoterAPIServerResourceName               = "test-promoter-apiserver"
+	promoterControllerResourceNameWithNamespace = "test-%s-promoter-controller-manager"
+	promoterAPIServerResourceNameWithNamespace  = "test-%s-promoter-apiserver"
+	apiServerTLSSecretName                      = "test-tls-secret"
+	apiServerCABundleSecretName                 = "test-cabundle-secret"
+	apiServerCABundleSecretKey                  = "ca.crt"
+)
+
+var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
```
jgwest: Should be `Sequential` to match the package

tests/ginkgo/sequential/1-134_validate_gitops_promoter_test.go
```diff
@@ -0,0 +1,479 @@
...
+			apiServerDeployment              *appsv1.Deployment
+			apiServerService                 *corev1.Service
+			apiServerClusterRoleBindingNames []string
+			apiServerClusterRoleNames        []string
+			apiServerAPIService              *apiregistrationv1.APIService
+			apiServerRoleBinding             *rbacv1.RoleBinding
+		)
+
+		BeforeEach(func() {
+			fixture.EnsureParallelCleanSlate()
```
jgwest: Should be `SequentialCleanSlate` to match the package

(slightly disappointed neither claude nor code rabbit noticed this, probably at some point we can write some improved review instructions for the reviewing agents...)

controllers/gitopspromoter/policyrules.go
```diff
@@ -0,0 +1,336 @@
...
+			name:        fmt.Sprintf("%s-%s", generatePromoterResourceNameWithNamespace(compName, cr), "promotionstrategydetails-viewer"),
+			roleRefName: fmt.Sprintf("%s-%s", generatePromoterResourceNameWithNamespace(compName, cr), "promotionstrategydetails-viewer"),
+			policyRule:  buildPolicyRuleForAPIServerPromotionStrategyDetailsViewer(),
+		},
+	}
+}
+
+// buildPolicyRuleForControllerClusterRole creates the default policy rule for the Controller's cluster role
+func buildPolicyRuleForControllerClusterRole() []rbacv1.PolicyRule {
+	return []rbacv1.PolicyRule{
```
jgwest, line 79: For security reasons, cluster scoped resources should only be created by cluster-scoped Argo CD instances.

Cluster-scoped Argo CD instances are designated via the `ARGOCD_CLUSTER_CONFIG_NAMESPACES` env var within the Subscription, which is then checked within the operator via `IsNamespaceClusterConfigNamespace`.

Without these types of check (which we use for other cluster-scoped functionality within the codebase), a user could:
- 1) Create a Namespace-scoped Argo CD instance
- 2) Enable gitops promoter feature (which would grant clusterrole/rolebinding to the local instance)
- 3) Ride the `ServiceAccount` from 1)  to elevated privileges using the new privileges from 2).

As part of this PR, we should thus ensure:
- A) gitops promoter can only be enabled for cluster-scoped Argo CD instances, and will only create those cluster-scoped resources if cluster-scoped
- B) We clean up any cluster-scoped resources exist where there is not a corresponding cluster-scoped argo cd instance.

controllers/gitopspromoter/apiservice.go
```diff
@@ -0,0 +1,161 @@
...
+			key = cr.Spec.Promoter.APIServer.TLS.CABundleSecretKey
+		}
+
+		if val, ok := caSecret.Data[key]; ok {
+			apiSvc.CABundle = val
+		} else {
+			return apiregistrationv1.APIServiceSpec{}, fmt.Errorf("ca bundle not found in secret %s at key %s, API Server may not work correctly", cr.Spec.Promoter.APIServer.TLS.CABundleSecretName, key)
+		}
+	} else {
+		apiSvc.InsecureSkipTLSVerify = true
```
jgwest: We likely don't want to enable this setting (or any `Insecure`-prefixed setting) without explicit user intent. As Claude notes: `When no CA bundle is configured, the aggregated API endpoint accepts unverified TLS, enabling MITM attacks against the Kubernetes API aggregation layer.`

controllers/argocd/deployment.go
```diff
@@ -1102,25 +1101,37 @@ func (r *ReconcileArgoCD) reconcileServerDeployment(cr *argoproj.ArgoCD, useTLSF
...
+		deploy.Spec.Template.Spec.InitContainers = removeInitContainer(deploy.Spec.Template.Spec.InitContainers, "rollout-extension")
+	}
+
+	if promoterUIEnabled {
+		deploy.Spec.Template.Spec.InitContainers = append(deploy.Spec.Template.Spec.InitContainers, getPromoterInitContainer()...)
+	} else {
+		deploy.Spec.Template.Spec.InitContainers = removeInitContainer(deploy.Spec.Template.Spec.InitContainers, "promoter-extension")
+	}
 
+	if rolloutsUIEnabled || promoterUIEnabled {
```
jgwest: Courtesy of Claude:

  4. Duplicate volume/volumeMount accumulation — The extensions volume is appended without dedup on each reconciliation, eventually causing Kubernetes validation failures.


```
● The duplicate volume/volumeMount accumulation occurs in controllers/argocd/deployment.go:1120-1135.

  The function reconcileServerDeployment is called on every reconciliation loop. When it runs, it fetches the existing deployment from the cluster (line ~960), then mutates it in place. The
  problem is at line 1120:

  if rolloutsUIEnabled || promoterUIEnabled {
      deploy.Spec.Template.Spec.Containers[0].VolumeMounts = append(deploy.Spec.Template.Spec.Containers[0].VolumeMounts, corev1.VolumeMount{
          Name:      extensionsVolumeName,
          MountPath: "/tmp/extensions/",
      })

      deploy.Spec.Template.Spec.Volumes = append(deploy.Spec.Template.Spec.Volumes, corev1.Volume{
          Name: extensionsVolumeName,
          VolumeSource: corev1.VolumeSource{
              EmptyDir: &corev1.EmptyDirVolumeSource{},
          },
      })
  }

  On the first reconciliation, the deployment has no extensions volume, so the append adds one — correct. On the second reconciliation, the code fetches the deployment that already has the
  extensions volume from the previous run, then appends another one. There's no check like "does this volume already exist?" or a corresponding removeVolume/removeVolumeMount call before the
  append.

  The same issue applies to the init containers at lines 1108-1115 — getRolloutInitContainer() and getPromoterInitContainer() are appended without first removing any existing ones with those
  names.

  Contrast this with the else branch at line 1132, which correctly uses removeVolume and removeVolumeMount to clean up. The fix would be to either remove-then-append, or check for existence
  before appending.
```

controllers/gitopspromoter/deployment.go
```diff
@@ -0,0 +1,478 @@
...
+
+	if image := os.Getenv(EnvGitOpsPromoterImage); image != "" {
+		return image
+	}
+
+	return common.GitOpsPromoterDefaultImageName
+}
+
+// buildControllerSecurityContext builds the SecurityContext for the Controller's container within its Deployment
+func buildControllerSecurityContext() *corev1.SecurityContext {
```
jgwest, line 317: Courtesy of Claude:

Missing RunAsNonRoot/seccomp on the promoter controller container (inconsistent with all other components).

```
● The missing RunAsNonRoot and seccomp profile is in controllers/gitopspromoter/deployment.go:315-321:

  func buildControllerSecurityContext() *corev1.SecurityContext {
      return &corev1.SecurityContext{
          AllowPrivilegeEscalation: ptr.To(false),
          Capabilities: &corev1.Capabilities{
              Drop: []corev1.Capability{"ALL"},
          },
      }
  }

  Compare this to the API server's security context just below it at line 325:

  func buildAPIServerSecurityContext() *corev1.SecurityContext {
      return &corev1.SecurityContext{
          RunAsNonRoot: ptr.To(true),
          SeccompProfile: &corev1.SeccompProfile{
              Type: corev1.SeccompProfileTypeRuntimeDefault,
          },
      }
  }

  And to the baseline used by all other ArgoCD components (DefaultSecurityContext(), which the promoter init containers at line 1421 do use):

  // Sets: AllowPrivilegeEscalation=false, ReadOnlyRootFilesystem=true,
  //        RunAsNonRoot=true, SeccompProfile=RuntimeDefault, Drop ALL caps

  So the controller is missing RunAsNonRoot, SeccompProfile, and ReadOnlyRootFilesystem. The API server is missing AllowPrivilegeEscalation, Capabilities.Drop, and ReadOnlyRootFilesystem.
  Neither matches the project's own DefaultSecurityContext() baseline — each omits a different subset of fields. Ideally both would use DefaultSecurityContext() (or a superset of it) rather
  than hand-rolling partial security contexts.

```

controllers/gitopspromoter/apiservice.go
```diff
@@ -0,0 +1,161 @@
...
+	argoutil.LogResourceCreation(log, apiSvc)
+	if err := client.Create(context.Background(), apiSvc); err != nil {
+		return nil, fmt.Errorf("failed to create promoter apiserver apiservice %s: %v", apiSvc.Name, err)
+	}
+	return apiSvc, nil
+}
+
+// buildAPIService creates the basic object for the API Service
+func buildAPIService(compName string, cr *argoproj.ArgoCD) *apiregistrationv1.APIService {
+	// FIXME: this is also a hard coded resource like the controller configuration.
```
jgwest: As mentioned for TODO elsewhere, having a GitHub issue to track this would also be advantageous.

controllers/gitopspromoter/controllerconfiguration.go
```diff
@@ -0,0 +1,161 @@
...
+				return controllerConfiguration, nil
+			}
+			argoutil.LogResourceDeletion(log, controllerConfiguration, "promoter controller configuration is being deleted due to being disabled")
+			if err := client.Delete(context.Background(), controllerConfiguration); err != nil {
+				return nil, fmt.Errorf("failed to delete controller configuration %s: %v", controllerConfiguration.Name, err)
+			}
+			return controllerConfiguration, nil
+		}
+
+		// TODO:: For now do not do any reconcilation based on the spec of the controller configuration to allow for edits
```
jgwest: Likewise for other TODOs, can't hurt to track this with an issue (if we feel it's worth doing).

jgwest (changes requested):

jgwest:
Additional review comments, meant to include these in the review comment above :sweat_smile: 

### 1) TODOs in the code	
I've requested you convert some of the TODOs/FIXMEs to GitHub issues. You can either:
A) Leave them as is. (see 'Philosophically:' below)
B) Open GH issues for them (IMHO a best practice). I think Code Rabbit can be coaxed to do this for you.
C) Convert them from a TODO/FIXME into a general comment (e.g. if you feel they are a 'nice to have' task that is best left as a comment, rather than something we specifically need to track)

(Philosophically: I'm generally against adding TODOs into the code, _without_ also adding a tracking GitHub issue, becaused IMHO code TODOs are where tasks go to die; usually TODOS are left in the code and never fixed. BUT, opinions differ on this, even between e.g. Jann and I. So I defer to your own judgement on what is worth doing here).

### 2) Doc updates

Could use some doc updates for this feature within this repo. (I see you noted this in the PR description).

The doc doesn't need to be detailed: It only needs to mention the installation steps (via CR), and then point users at [upstream docs](https://gitops-promoter.readthedocs.io/en/latest/getting-started/) for any additonal configuration they may want to do.

I'm fine with merging this PR without doc updates for now (in the interest of time and code/feature freeze), _as long as those are contributed by a subsequent PR_.


### 3) Cluster scoped Argo CD instances only

Note my 'For security reasons, cluster scoped resources should only be created by cluster-scoped Argo CD instances' comment below to ensure we don't allow privilege escalation via this feature.

go.mod
cjcocokrisp: Probably just a mistake in my rebase. 😅

controllers/argocd/deployment.go
cjcocokrisp: Yeah I will open issues for the TODOs and FIXMEs I left. This one has a dependency on fixing an issue in the extension installer due to it concating multiple checksums (the promoter extension tar and the sbom tar for the extension were being concatenated causing the check to fail, argo rollouts extension does not seem to have a checksum as well so something else to be done for that too).

controllers/argocd/deployment.go
cjcocokrisp: I'm curious if the way the custom value is set had a bug before this. I interpreted the `ARGOCD_EXTENSION_IMAGE` env var to just be a way to select the image of [the argocd-extension-installer](https://github.com/argoproj-labs/argocd-extension-installer) that is used when installing the rollouts extension.

But looking at it again, seeing that env is not set in the spec for the init container it makes me think that the env variable is supposed to just point at an image that has the extension in it? Not sure if that would make sense because an image like that would need to download the extension and would be a lot for a user to do just to use a custom extension. I would think it would be more like you can provide an environment variable or even just have it in the Argo CD CR spec to just give the URL to the extension version you want and not have to go through the work of building and distributing a container image for it.

The `ARGOCD_EXTENSION_IMAGE` env variable is also not mentioned in the environment variable docs as well. https://argocd-operator.readthedocs.io/en/latest/usage/environment_variables/

As I had a TODO for the checksums for the extensions I think there could be a lot of enhancements we could make in the operator for the lifecycle of extensions. Not sure if this is possible at the moment but it also could be nice to allow users to specify a list of extensions to download because I'm sure there's more then just the ones for Rollouts and the Promoter. Will open more refined issues soon.

tests/ginkgo/sequential/1-134_validate_gitops_promoter_test.go
cjcocokrisp: Same especially since this was just a copy and paste mistake you'd think it would pick up on that easily. It's good that this caught the gap.

jgwest (approved):
LGTM, thanks @cjcocokrisp!

PR #2337, opened 2026-08-21
feat: implement wildcard support for IMAGE_UPDATER_WATCH_NAMESPACES

dkarpele:
**What type of PR is this?**

/kind enhancement

see GITOPS-10660
Update ImageUpdater in ArgoCD Operator to accept wildcard patterns in IMAGE_UPDATER_WATCH_NAMESPACES, in addition to explicit comma-separated namespaces. Ensure namespace selection follows the same pattern semantics used by other OpenShift GitOps components, including common tenant patterns such as *-argocd.

The implementation should follow applicationSet.sourceNamespaces

Fixes #?
GITOPS-10660

## Summary by CodeRabbit

- **New Features**
  - Image Updater supports exact namespace names, glob patterns, and regular expressions for watch scope.
  - Matching namespaces are reflected in deployment configuration and access permissions.
  - Watch-scope changes automatically update permissions and deployment settings as namespaces match or stop matching.
  - Creating or removing namespaces triggers watch-scope updates.
  - Image pull secret references are updated automatically on supported platforms.

- **Bug Fixes**
  - Terminating namespaces are excluded from watch-scope matching.
  - Invalid, overly broad, or conflicting patterns are rejected.

- **Documentation**
  - Updated watch-scope guidance and examples.

- **Tests**
  - Added coverage for pattern matching, updates, and stale permission cleanup.

### DIFF
--- a/controllers/argocd/image_updater.go
+++ b/controllers/argocd/image_updater.go
@@ -768,14 +768,28 @@ func (r *ReconcileArgoCD) reconcileImageUpdaterDeployment(cr *argoproj.ArgoCD, s
 	return r.reconcileDeploymentHelper(cr, desiredDeployment, "image updater", cr.Spec.ImageUpdater.Enabled)
 }
 
+// splitAndFilterPatterns splits a comma-separated pattern string, trims whitespace from
+// each token, and discards empty entries that result from trailing/leading commas or
+// consecutive commas. An empty pattern would match every namespace under glob semantics,
+// which could silently grant cluster-wide RBAC access.
+func splitAndFilterPatterns(raw string) []string {
+	var patterns []string
+	for _, p := range strings.Split(raw, ",") {
+		if t := strings.TrimSpace(p); t != "" {
+			patterns = append(patterns, t)
+		}
+	}
+	return patterns
+}
+
 // expandImageUpdaterWatchNamespaces lists all cluster namespaces and returns those that
 // match any pattern in the comma-separated watchNamespaces value. Each pattern may be a
 // glob-style wildcard or a full regular expression (the same semantics used by
 // .spec.applicationSet.sourceNamespaces). The returned slice is sorted for determinism.
 func (r *ReconcileArgoCD) expandImageUpdaterWatchNamespaces(watchNamespaces string) ([]string, error) {
-	patterns := strings.Split(watchNamespaces, ",")
-	for i := range patterns {
-		patterns[i] = strings.TrimSpace(patterns[i])
+	patterns := splitAndFilterPatterns(watchNamespaces)
+	if len(patterns) == 0 {
+		return nil, nil
 	}
 
 	clusterNamespaces := &corev1.NamespaceList{}
--- a/controllers/argocd/image_updater.go
+++ b/controllers/argocd/image_updater.go
@@ -839,6 +839,8 @@ func splitAndFilterPatterns(raw string) []string {
 // match any pattern in the comma-separated watchNamespaces value. Each pattern may be a
 // glob-style wildcard or a full regular expression (the same semantics used by
 // .spec.applicationSet.sourceNamespaces). The returned slice is sorted for determinism.
+// Namespaces in a Terminating state (DeletionTimestamp != nil) are excluded from both
+// the match result and the total count, as creating content in such namespaces is forbidden.
 func (r *ReconcileArgoCD) expandImageUpdaterWatchNamespaces(watchNamespaces string) ([]string, error) {
 	patterns := splitAndFilterPatterns(watchNamespaces)
 	if len(patterns) == 0 {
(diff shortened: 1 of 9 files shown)


controllers/argocd/image_updater.go
```diff
@@ -697,6 +732,31 @@ func (r *ReconcileArgoCD) reconcileImageUpdaterDeployment(cr *argoproj.ArgoCD, s
...
+
+	var matched []string
+	for _, ns := range clusterNamespaces.Items {
+		if glob.MatchStringInList(patterns, ns.Name, glob.REGEXP) {
+			matched = append(matched, ns.Name)
+		}
+	}
+	sort.Strings(matched)
+	return matched, nil
+}
```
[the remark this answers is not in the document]
dkarpele: I think it's a rare case. If admin wants to use new namespace and trigger reconciliation they will add new ns to spec.sourceNamespaces. This will also trigger Image Updater

controllers/argocd/image_updater.go
```diff
@@ -546,6 +561,26 @@ func (r *ReconcileArgoCD) reconcileImageUpdaterDeployment(cr *argoproj.ArgoCD, s
...
+			expanded, err := r.expandImageUpdaterWatchNamespaces(rawValue)
+			if err != nil {
+				return err
+			}
+			if len(expanded) > 0 {
+				imageUpdaterEnv = argoutil.EnvSet(imageUpdaterEnv, corev1.EnvVar{
+					Name:  "IMAGE_UPDATER_WATCH_NAMESPACES",
+					Value: strings.Join(expanded, ","),
+				})
+			}
```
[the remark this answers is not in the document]
dkarpele: fixed

controllers/argocd/image_updater.go
```diff
@@ -72,6 +137,11 @@ func (r *ReconcileArgoCD) reconcileImageUpdaterControllerEnabled(cr *argoproj.Ar
 		}
 	}
 
+	// desiredNamespaces tracks which per-namespace Roles/RoleBindings should exist after this
+	// reconcile. It is populated only in the pattern-list case; for "*" and "" it stays empty,
+	// which causes pruneImageUpdaterNamespaceRBAC to remove any stale per-namespace objects.
+	desiredNamespaces := map[string]struct{}{}
```
olivergondza: nit: Can be shortened to `map[string]any`

controllers/argocd/image_updater_test.go
```diff
@@ -736,6 +793,22 @@ func TestReconcileImageUpdater_WatchNamespacesMode(t *testing.T) {
...
+			}
+			for _, ns := range tt.clusterNamespaces {
+				if _, expected := allMatched[ns]; expected {
+					continue
+				}
+				err := r.Get(context.TODO(), types.NamespacedName{
+					Name:      getRoleNameForApplicationSourceNamespaces(ns, a),
+					Namespace: ns,
+				}, &rbacv1.Role{})
+				assert.True(t, errors.IsNotFound(err), "namespace %s should not have a manager role", ns)
```
olivergondza: For the sake of defect localization, incorporate the actual error in the message.

olivergondza (approved):

dkarpele:
@olivergondza 
release-0.20 branch was created so I think we can move forward with this PR

controllers/argocd/image_updater_test.go
```diff
@@ -968,6 +1051,272 @@ func TestReconcileImageUpdaterControllerEnabled_PrunesStaleNamespaceRBAC(t *test
...
+			wantEnvValue:    "team-a,team-b",
+		},
+		{
+			// Exact name alongside an unmatched glob: exact name expands normally,
+			// unmatched glob is dropped from the env var.
+			name:            "exact name expands, unmatched glob dropped",
+			clusterNS:       []string{"ns1"},
+			watchNamespaces: "ns1,app-*",
+			wantEnvValue:    "ns1",
+		},
```
chengfang, line 1122: I tested with the following wildcard-like values:
```go
		{
			name:            "double star",
			clusterNS:       []string{"ns1, argocd, kube-system, image-updater"},
			watchNamespaces: "**",
			wantEnvValue:    "*", // actual: ns1, argocd, kube-system, image-updater
		},
		{
			name:            "question star",
			clusterNS:       []string{"ns1, argocd, kube-system, image-updater"},
			watchNamespaces: "?*",
			wantEnvValue:    "*", // actual: ns1, argocd, kube-system, image-updater
		},
		{
			name:            "regex star",
			clusterNS:       []string{"ns1, argocd, kube-system, image-updater"},
			watchNamespaces: "/.*/",
			wantEnvValue:    "*", // actual: ns1, argocd, kube-system, image-updater
		},
        {
			name:            "three star",
			clusterNS:       []string{"ns1, argocd, kube-system, image-updater"},
			watchNamespaces: "**/*",
			wantEnvValue:    "*", // actual: ns1, argocd, kube-system, image-updater
		},
```

We currently don't consider them as '*', but they all expaned to all namespaces and will probalby cause unwanted priviledge escalation.

When normalizing, we only catches * and its variants, and these regex-style values will pass as non-wildcard, list-style values.

When expanding to concrete namespaces in `expandImageUpdaterWatchNamespaces`, we allow both glob and regex patterns and they expanded to all namespaces.

dkarpele: Great catch, thanks!

controllers/argocd/image_updater.go
chengfang, line 638: `imageUpdaterEnv` is initialized to the raw user input value, and then reset to teh normalized expanded namespaces for non-empty and non-* case.  

For empty case, the raw value may be something like ',,,'; for wildcard case, raw value can be '*,', but in these 2 cases, `imageUpdaterEnv` is not reset to normalized values, and directly passed to the deployment below.

I think it may be clearer to switch `watchNamespaces` on "", "*", and default, and reset or remove `imageUpdaterEnv` accordingly.

controllers/argocd/image_updater.go
```diff
@@ -845,6 +845,20 @@ func (r *ReconcileArgoCD) expandImageUpdaterWatchNamespaces(watchNamespaces stri
...
+	// enforces. Returning an error here prevents unintended privilege escalation and
+	// directs the user to the correct, explicit cluster-scoped configuration.
+	if len(clusterNamespaces.Items) > 0 && len(matched) == len(clusterNamespaces.Items) {
+		return nil, fmt.Errorf(
+			"IMAGE_UPDATER_WATCH_NAMESPACES %q matches all %d cluster namespaces; "+
+				"use \"*\" for cluster-scoped mode",
+			watchNamespaces, len(clusterNamespaces.Items),
+		)
+	}
+
```
chengfang, line 884: Reference https://github.com/argoproj-labs/argocd-operator/blob/c354222740c1d67f31fd2845d25ff0ee4e1fc6f6/controllers/argocd/rolebinding.go#L244

When matching namespace, consider following the above handling in rolebinding.go to filter out namespaces being deleted, and exclude them from image updater watch namespaces.

Something like:
```go
      // Namespaces in Terminating state reject object creation with 403 (not NotFound), which
      // would abort the entire reconcile. Exclude them from the match set — and from the
      // all-namespaces comparison below, so a terminating namespace cannot mask a
      // match-everything pattern.
      activeCount := 0
      var matched []string
      for _, ns := range clusterNamespaces.Items {
              if ns.DeletionTimestamp != nil {
                      log.Info("skipping terminating namespace for Image Updater RBAC", "namespace", ns.Name)
                      continue
              }
              activeCount++
              if glob.MatchStringInList(patterns, ns.Name, glob.REGEXP) {
                      matched = append(matched, ns.Name)
              }
      }

      if activeCount > 0 && len(matched) == activeCount {
              return nil, fmt.Errorf(
                      "IMAGE_UPDATER_WATCH_NAMESPACES %q matches all %d active cluster namespaces; "+
                              "use \"*\" for cluster-scoped mode",
                      watchNamespaces, activeCount,
              )
      }
```

controllers/argocd/image_updater.go
chengfang, line 439: `newRoleForApplicationSourceNamespaces` was previously called by other controllers like application controller to create a role object with a fixed name.

When we reuse this function for image updater, we will also create another role object of the same fixed name but different rules, if the `namespace` is the same. This can happen when users also specify the app namespace as image updater watch namespaces. There will be interference between the 2 controllers and regression in the other component.

Can we use a unique name?

chengfang, line 452: same as Role above.

dkarpele: I think this is out of scope for the PR

dkarpele: I think this is out of scope for the PR

dkarpele:
@coderabbitai resume

chengfang (approved):

PR #2339, opened 2026-08-25
feat(dex): Add support for kubernetes storage for dex to avoid session invalidation on dex ServiceAccount token rotation

anandf:
**What type of PR is this?**
/kind bug

Assisted by: Google Gemini (for generating the custom bootstrap script for dex server)
Assisted by: Claude Code Opus 4.6 (for generating unit tests)

Due to recent change for improving the security, dex service account tokens gets rotated  every 1hour. When dex service account rotates, the dex config and the client secret changes, causing the existing logged in sessions to be invalidated. This causes existing logged-in users to be forced to re-login. The root cause of this issue is because Argo CD hard codes the storage to `memory` and there is no option provided to configure this specific field https://github.com/argoproj/argo-cd/blob/aa4629e50c2ae8c138b9333e4add48a7124d2aff/util/dex/config.go#L31

To fix the issue, the signing keys needs to be stored in a persistent storage and dex provides the following options for persistent storage :
1. Kubernetes (recommended for kubernetes based environment)
2. etcd v3
3. Postgres/MySQL
4. SQLLite (not recommended for production environment)

There is an upstream PR in argo-cd https://github.com/argoproj/argo-cd/pull/28624 created to make this hardcoded storage setting to a configurable value. The PR is proposed as an enhancement and may not be available for back-porting. It may be available for 3.6 or 3.7 timelines based on when the PR would be accepted and merged. For bug fixes, this timelines and lack of backporting option makes this approach unacceptable, Due to the reason explained, a fix in the operator code is the preferred approach.

** Summary of changes done in this PR.
1. Unless Kubernetes storage is explicitly disabled (via an environment variable in `Subscription` object), use the kubernetes storage 
2. The ArgoCD Operator will create the required CRDs at runtime when kubernetes storage is enabled. If the operator does not have permissions to create a CRD, it will fallback to memory based storage.
3. When kubernetes storage is used,  the startup script of the dex container image is modified to use a custom script instead of the default `argocd-dex rundex` which is valid for in-memory storage.
4. The custom script does the following steps. These steps are the same steps done inside the `rundex` subcommand, but the script provides a mechanism to change the storage config before the server is started.
i. Create a self signed TLS certificate using `openssl` command.
ii. Generate the base Dex Config file using `argocd-dex gendexcfg`
iii. Use the `awk` command to update the storage configuration in the generated base configuration and use that for the dex server.
iv. Poll for any changes done to the dex config in `argocd-cm` configmap. The poll interval is 15s. The change detection is done via `gendexcfg` command and comparing it with the previous configuration via `sha256sum`. 
v. When the SA token rotates, the dex config is modified and within 15s this change is detected and the process gets restarted via the TERM signal.
 5. The PR is also using `go build tags` to inject custom values for dex server token expiry and threshold. This is to improve the testing process and reduce the wait time to simulate and actual token rotation event. This code will be included only when the code is compiled with the `-tag debug` option. Otherwise this custom code is not included in production images.
 
Leaking go-routines is caused due to the process being handled within the go runtime. `rundex` command uses internal go runtime's `os.Exec` command, which I suspect is not cleaning up some of the pending go-routines that could be left causing a leak. With a shell based approach, and passing  the SIGTERM signal to the underlying dex server process directly, the cleanup of pending go-routines, must be addressed as well. This assumption needs to be verified and requires an environment where the issue can be reproduced.

Fixes
 [GITOPS-10429](https://redhat.atlassian.net/browse/GITOPS-10429)
 [GITOPS-10364](https://redhat.atlassian.net/browse/GITOPS-10364)
```
oc apply -f - <<EOF
kind: CatalogSource
metadata:
spec:
  displayName: "GITOPS DEVEL"
  image: quay.io/anjoseph/gitops-operator-catalog:v1.21.3-2
  publisher: "GITOPS DEVEL"
  sourceType: grpc
EOF
```
7. Install the OpenShift GitOps Operator
6.1 Create the namespace and `OperatorGroup` resources.

```
```
```
oc apply -f - <<EOF
kind: OperatorGroup
metadata:
spec:
  upgradeStrategy: Default
EOF
```
2.2 Create the `Subscription` resource with the required env overrides.
```
oc apply -f - <<EOF
kind: Subscription
metadata:
  labels:
  name: gitops-operator
spec:
  channel: latest
  name: gitops-operator
  config:
    env:
      - name: ARGOCD_DEX_SERVER_TOKEN_EXPIRY_SECONDS
        value: "600"
      - name: ARGOCD_DEX_SERVER_TOKEN_RENEWAL_THRESHOLD
        value: "70"
      - name: ARGOCD_DEX_STORAGE_ENV_OVERRIDE
        value: "true"
EOF
```
8. Login to the session using SSO and wait for the service account to be rotated. With the env overrides for token expiry seconds and threshold, it would take maximum of 17 minutes for the token to be rotated.
```
# Get the expiry time
oc get secret openshift-gitops-argocd-dex-server-token -o jsonpath='{.data.expiry}'| base64 -d

# Get the current time
date --utc

# Wait for 7 mins after the current time has passed the expiry time. The token and expiry time would get updated.

# Check the logs that the dex server has restarted due to change detected in argocd-cm/argocd-secret.
oc logs <openshift-dex-server-pod> -n openshift-gitops
```
10. Ensure that the user is not being forced to login when the service account was rotated.

## Summary by CodeRabbit

- **New Features**
  - Dex can use Kubernetes-backed storage when enabled, improving persistence and deployment flexibility.
  - Dex startup now supports mounted TLS certificates and can generate self-signed certificates when needed.
  - Dex automatically creates and manages the required Kubernetes custom resources.

- **Configuration**
  - Token expiration and renewal thresholds can be customized in debug builds.

- **Bug Fixes**
  - Improved cleanup of operator-created Dex service-account token secrets while preserving unrelated secrets.

- **Build and Deployment**
  - Optional build tags are now supported for local and container builds.


controllers/argocd/dex.go
```diff
@@ -717,5 +758,110 @@ func (r *ReconcileArgoCD) reconcileDexResources(cr *argoproj.ArgoCD) error {
...
+}
+
+// reconcileDexServerClusterRole reconciles required clusterrole for dex server controller when ArgoCD is cluster-scoped
+func (r *ReconcileArgoCD) reconcileDexServerClusterRole(cr *argoproj.ArgoCD) (*v1.ClusterRole, error) {
+
+	policyRules := []v1.PolicyRule{
+		// ApplicationSet
+		{
+			APIGroups: []string{"apiextensions.k8s.io"},
+			Resources: []string{"customresourcedefinitions"},
```
akhilnittala: cant we add these on reconciler as kubebuilder annotations, just a thought.
https://github.com/argoproj-labs/argocd-operator/blob/master/controllers/argocd/argocd_controller.go#L119

controllers/argocd/dex.go
```diff
@@ -717,5 +758,110 @@ func (r *ReconcileArgoCD) reconcileDexResources(cr *argoproj.ArgoCD) error {
...
+	if !reflect.DeepEqual(existingClusterRB.Subjects, clusterRB.Subjects) {
+		existingClusterRB.Subjects = clusterRB.Subjects
+		argoutil.LogResourceUpdate(log, existingClusterRB, "updating subjects")
+		if err := r.Update(context.TODO(), existingClusterRB); err != nil {
+			return err
+		}
+	} else if !reflect.DeepEqual(existingClusterRB.RoleRef, clusterRB.RoleRef) {
+		// RoleRef can't be updated, delete the rolebinding so that it gets recreated
+		argoutil.LogResourceDeletion(log, existingClusterRB, "roleref changed, deleting rolebinding so it gets recreated")
+		_ = r.Delete(context.TODO(), existingClusterRB)
```
akhilnittala: ignoring the error here ?

controllers/argocd/dex.go
```diff
@@ -717,5 +758,110 @@ func (r *ReconcileArgoCD) reconcileDexResources(cr *argoproj.ArgoCD) error {
...
+		}
+		argoutil.LogResourceCreation(log, clusterRB)
+		return r.Create(context.TODO(), clusterRB)
+	}
+
+	// if subj differ, update the rolebinding
+	if !reflect.DeepEqual(existingClusterRB.Subjects, clusterRB.Subjects) {
+		existingClusterRB.Subjects = clusterRB.Subjects
+		argoutil.LogResourceUpdate(log, existingClusterRB, "updating subjects")
+		if err := r.Update(context.TODO(), existingClusterRB); err != nil {
```
akhilnittala: instead of multiple context.TODO(), we can have context.Background() defined in single place and use that context for entire logic.

controllers/argocd/dex.go
anandf: Cluster roles are removed. CRDs are now created by the operator during installation via OLM. Added the CRD definitions in the `bundle/manifests` directory. So these cluster level policies are no longer required.

controllers/argocd/dex.go
anandf: removed this code

controllers/argocd/dex.go
```diff
@@ -720,5 +734,110 @@ func (r *ReconcileArgoCD) reconcileDexResources(cr *argoproj.ArgoCD) error {
...
+		{
+			APIGroups: []string{"apiextensions.k8s.io"},
+			Resources: []string{"customresourcedefinitions"},
+			Verbs:     []string{"create"},
+		},
+		{
+			APIGroups: []string{"dex.coreos.com"},
+			Resources: []string{"*"},
+			Verbs:     []string{"*"},
+		},
```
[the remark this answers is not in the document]
anandf: removed this code

controllers/argocd/dex.go
```diff
@@ -468,6 +477,32 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoproj.ArgoCD) error {
 		VolumeMounts:    dexVolumeMounts,
 	}}
 
+	// if dex kubernetes storage override is enabled, generate a base config,
+	// override the storage config and use that to server the dex server
+	if argoutil.IsDexKubernetesStorageEnabled() {
+		deploy.Spec.Template.Spec.Containers[0].Command = []string{"/bin/sh", "-c"}
+		deploy.Spec.Template.Spec.Containers[0].Args = []string{
+			`# 1. Handle TLS cert generation (mimicking rundex)
```
akhilnittala: is this for testing purpose ?

anandf: No. This is the actual custom startup script that will be used.

anandf: I will remove the comments and make it shorter.

hack/update-dependencies-script/argocd/main.go
```diff
@@ -12,7 +12,10 @@ import (
 	"github.com/argoproj-labs/argocd-operator/dependency-upgrade/utils"
 )
 
-const ArgoCDGitHubRepoURL = "https://github.com/argoproj/argo-cd"
+const (
+	ArgoCDGitHubRepoURL = "https://github.com/argoproj/argo-cd"
+	DexGitHubRepoURL    = "https://github.com/dexidp/dex"
```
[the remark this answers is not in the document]
anandf: This change is no longer required and reverted. CRDs are created and managed via the operator code directly instead of being added to the bundle and being owned by OLM. This is to avoid default aggregate to edit and aggregate to admin annotations being added. These CRDs are required only by the dex server and not by any other user in the kubernetes cluster.

controllers/argocd/dex.go
```diff
@@ -378,6 +379,19 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoproj.ArgoCD) error {
...
+	// use the kubernetes storage to store the signing keys and the auth codes by
+	// overriding the storage config through env variables
+	if argoutil.IsDexStorageOverrideEnabled() {
+		dexEnv = append(dexEnv, corev1.EnvVar{
+			Name:  "DEX_STORAGE_TYPE",
+			Value: "kubernetes",
+		}, corev1.EnvVar{
+			Name:  "DEX_STORAGE_CONFIG",
+			Value: "{\"inCluster\": true}",
+		})
```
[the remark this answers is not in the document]
anandf: These changes are removed and not longer required.

controllers/argocd/dex.go
```diff
@@ -468,6 +461,22 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoproj.ArgoCD) error {
...
+	// override the storage config and use that to server the dex server, if not continue using the in-memory storage.
+	if argoutil.IsDexKubernetesStorageEnabled() && canCreateDexCRDs {
+		deploy.Spec.Template.Spec.Containers[0].Command = []string{"/bin/sh", "-c"}
+		deploy.Spec.Template.Spec.Containers[0].Args = argoutil.DexServerCustomStartupScript()
+		err := argoutil.EnsureDexCRDs(context.TODO(), r.Client)
+		if err != nil {
+			return err
+		}
+	}
+
```
chengfang, line 479: further below in this func, we merge fields from `deploy` to existing dex deployment, if `deplExists` is true. Should we also copy these new fields (`Command`, `Args`) if the same condition is met?

anandf: Yes, thanks for pointing it out. I had missed it. That section of code will be invoked during an upgrade and the changes needed there as well.

anandf: I have addressed the comment in a recent change PTAL. Thanks.

controllers/argocd/dex.go
```diff
@@ -468,6 +461,22 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoproj.ArgoCD) error {
...
 	}}
 
+	canCreateDexCRDs, err := argoutil.CanCreateDexCRDs(context.TODO(), r.Client)
+	if err != nil {
+		log.Error(err, "unable to check if dex CRDs can be created")
+		canCreateDexCRDs = false
+	}
+	// if dex kubernetes storage override is enabled, and if operator can create CRDs, generate a base config,
+	// override the storage config and use that to server the dex server, if not continue using the in-memory storage.
+	if argoutil.IsDexKubernetesStorageEnabled() && canCreateDexCRDs {
```
chengfang, line 471: since this feature is on by default, the above condition can be true in most cases, even if `UseDex()` is false. So when sso/dex is off, we will still check the CRD creation permission, and create them for no gain. How about first check `UseDex()`, and do the perm check and creation if actually needed?

controllers/argocd/policyrule.go
```diff
@@ -147,6 +148,16 @@ func policyRuleForDexServer() []v1.PolicyRule {
...
 		},
 	}
+	// if dex storage override is enabled, add policy rules for storing dex related custom resources
+	// in kubernetes storage.
+	if argoutil.IsDexKubernetesStorageEnabled() {
+		dexPolicyRules = append(dexPolicyRules, v1.PolicyRule{
+			APIGroups: []string{"dex.coreos.com"},
+			Resources: []string{"*"},
+			Verbs:     []string{"*"},
+		})
```
chengfang, line 158: that's pretty wide permissions granted when `IsDexKubernetesStorageEnabled` is on, which is true by default, even if `UseDex()` is false. Can we also add `UseDex` to the gate to reduce its scope?

controllers/argocd/dex_config_debug.go
```diff
@@ -0,0 +1,48 @@
...
+const (
+	EnvKeyDexServerTokenExpirySecs       = "ARGOCD_DEX_SERVER_TOKEN_EXPIRY_SECONDS"
+	EnvKeyDexServerTokenRenewalThreshold = "ARGOCD_DEX_SERVER_TOKEN_RENEWAL_THRESHOLD"
+)
+
+// dexServerTokenRenewalThreshold is how much nominal lifetime may remain before we treat the Dex token
+// values can be overridden with environment variables when debug build tag is enabled.
+func dexServerTokenRenewalThreshold() time.Duration {
+	dexServerTokenRenewalThresholdEnv := os.Getenv(EnvKeyDexServerTokenRenewalThreshold)
+	var threshold int64
```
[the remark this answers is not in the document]
chengfang: Looks a good suggestion. Can you evaluate it?

controllers/argocd/policyrule.go
anandf: I have corrected this method in my latest change adding these policy only if dex is enabled.

controllers/argocd/dex.go
anandf: Added the check to see if `UseDex()` returns true. Making these checks only if the SSO provider is dex makes sense. I have added this change in the latest code changes.

chengfang (approved):

akhilnittala:
we need to hold this changes right @anandf ?

akhilnittala (changes requested):
hold the changes as storage type is getting changed to etcd , changes are https://github.com/argoproj-labs/argocd-operator/pull/2367/changes

PR #2367, opened 2026-09-09
feat(dex): Add support for etcd based storage for dex

anandf:
**What type of PR is this?**
/kind bug

Assisted by: Google Gemini (for generating the custom bootstrap script for dex server)
Assisted by: Claude Code Opus 4.6 (for generating unit tests)

Due to recent change for improving the security, dex service account tokens gets rotated  every 1hour. When dex service account rotates, the dex config and the client secret changes, causing the existing logged in sessions to be invalidated. This causes existing logged-in users to be forced to re-login. The root cause of this issue is because Argo CD hard codes the storage to `memory` and there is no option provided to configure this specific field https://github.com/argoproj/argo-cd/blob/aa4629e50c2ae8c138b9333e4add48a7124d2aff/util/dex/config.go#L31

To fix the issue, the signing keys needs to be stored in a persistent storage and dex provides the following options for persistent storage :
1. Kubernetes (recommended for kubernetes based environment)
2. etcd v3
3. Postgres/MySQL
4. SQLLite (not recommended for production environment)

There is an upstream PR in argo-cd https://github.com/argoproj/argo-cd/pull/28624 created to make this hardcoded storage setting to a configurable value. The PR is proposed as an enhancement and may not be available for back-porting. It may be available for 3.6 or 3.7 timelines based on when the PR would be accepted and merged. For bug fixes, this timelines and lack of backporting option makes this approach unacceptable, Due to the reason explained, a fix in the operator code is the preferred approach.

** Summary of changes done in this PR.
1. Unless etcd storage is explicitly disabled (via an environment variable in `Subscription` object), use the etcd storage 
2. When etcd storage is used,  the startup script of the dex container image is modified to use a custom script instead of the default `argocd-dex rundex` which is valid for in-memory storage.
3. The custom script does the following steps. These steps are the same steps done inside the `rundex` subcommand, but the script provides a mechanism to change the storage config before the server is started.
i. Create a self signed TLS certificate using `openssl` command.
ii. Generate the base Dex Config file using `argocd-dex gendexcfg`
iii. Use the `awk` command to update the storage configuration in the generated base configuration and use that for the dex server.
iv. Poll for any changes done to the dex config in `argocd-cm` configmap. The poll interval is 15s. The change detection is done via `gendexcfg` command and comparing it with the previous configuration via `sha256sum`. 
v. When the SA token rotates, the dex config is modified and within 15s this change is detected and the process gets restarted via the TERM signal.
 5. The PR is also using `go build tags` to inject custom values for dex server token expiry and threshold. This is to improve the testing process and reduce the wait time to simulate and actual token rotation event. This code will be included only when the code is compiled with the `-tag debug` option. Otherwise this custom code is not included in production images.
 
Leaking go-routines is caused due to the process being handled within the go runtime. `rundex` command uses internal go runtime's `os.Exec` command, which I suspect is not cleaning up some of the pending go-routines that could be left causing a leak. With a shell based approach, and passing  the SIGTERM signal to the underlying dex server process directly, the cleanup of pending go-routines, must be addressed as well. This assumption needs to be verified and requires an environment where the issue can be reproduced.

Fixes
 [GITOPS-10429](https://redhat.atlassian.net/browse/GITOPS-10429)
 [GITOPS-10364](https://redhat.atlassian.net/browse/GITOPS-10364)
```
oc apply -f - <<EOF
kind: CatalogSource
metadata:
spec:
  displayName: "GITOPS DEVEL"
  image: quay.io/anjoseph/gitops-operator-catalog:v1.21.3-4
  publisher: "GITOPS DEVEL"
  sourceType: grpc
EOF
```
7. Install the OpenShift GitOps Operator
6.1 Create the namespace and `OperatorGroup` resources.

```
```
```
oc apply -f - <<EOF
kind: OperatorGroup
metadata:
spec:
  upgradeStrategy: Default
EOF
```
2.2 Create the `Subscription` resource with the required env overrides.
```
oc apply -f - <<EOF
kind: Subscription
metadata:
  labels:
  name: gitops-operator
spec:
  channel: latest
  name: gitops-operator
  config:
    env:
      - name: ARGOCD_DEX_SERVER_TOKEN_EXPIRY_SECONDS
        value: "600"
      - name: ARGOCD_DEX_SERVER_TOKEN_RENEWAL_THRESHOLD
        value: "70"
      - name: ARGOCD_DEX_STORAGE_ENV_OVERRIDE
        value: "true"
EOF
```
8. Login to the session using SSO and wait for the service account to be rotated. With the env overrides for token expiry seconds and threshold, it would take maximum of 17 minutes for the token to be rotated.
```
# Get the expiry time
oc get secret openshift-gitops-argocd-dex-server-token -o jsonpath='{.data.expiry}'| base64 -d

# Get the current time
date --utc

# Wait for 7 mins after the current time has passed the expiry time. The token and expiry time would get updated.

# Check the logs that the dex server has restarted due to change detected in argocd-cm/argocd-secret.
oc logs <openshift-dex-server-pod> -n openshift-gitops
```
10. Ensure that the user is not being forced to login when the service account was rotated.

## Summary by CodeRabbit

* **New Features**
  * Dex now supports etcd-backed storage with an automatically managed etcd sidecar.
  * Dex storage can be configured for etcd or in-memory operation.
  * Debug builds support configuring Dex token expiry and renewal thresholds through environment variables.
  * The etcd sidecar image can be customized through an environment variable.
  * Go and Docker builds support optional build tags.

* **Bug Fixes**
  * Improved Dex deployment reconciliation to keep container and sidecar settings synchronized.
  * Added startup checks to ensure etcd is ready before Dex starts.


controllers/argocd/dex.go
```diff
@@ -609,6 +630,44 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoproj.ArgoCD) error {
...
+			}
+			if !reflect.DeepEqual(deploy.Spec.Template.Spec.Containers[1].Command, existing.Spec.Template.Spec.Containers[1].Command) {
+				existing.Spec.Template.Spec.Containers[1].Command = deploy.Spec.Template.Spec.Containers[1].Command
+				changes = append(changes, "etcd container command")
+			}
+			if !reflect.DeepEqual(deploy.Spec.Template.Spec.Containers[1].ImagePullPolicy, existing.Spec.Template.Spec.Containers[1].ImagePullPolicy) {
+				existing.Spec.Template.Spec.Containers[1].ImagePullPolicy = deploy.Spec.Template.Spec.Containers[1].ImagePullPolicy
+				changes = append(changes, "etcd container image pull policy")
+			}
+		}
```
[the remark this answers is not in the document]
anandf: resolved

controllers/argoutil/dex.go
```diff
@@ -0,0 +1,78 @@
...
+	return []string{
+		`set -eo pipefail
+trap 'kill -TERM $DEX_PID 2>/dev/null; exit 0' INT TERM
+
+EXTRA_ARGS=""
+if [ -s /tls/tls.crt ] && [ -s /tls/tls.key ]; then
+cp /tls/tls.crt /tmp/tls.crt
+cp /tls/tls.key /tmp/tls.key
+elif command -v openssl >/dev/null 2>&1; then
+  openssl req -x509 -newkey rsa:2048 -nodes \
```
chengfang, line 34: openssl may not be available in dex image.

controllers/argoutil/dex.go
```diff
@@ -0,0 +1,78 @@
...
+else
+EXTRA_ARGS="--disable-tls"
+fi
+# run in a loop and restart the dex server process if there is a change in dex config.
+while true; do
+  /shared/argocd-dex gendexcfg ${EXTRA_ARGS} -o /tmp/base.yaml
+  awk '/^storage:/ { print "storage:\n  type: etcd\n  config:\n    endpoints:\n    - \"http://127.0.0.1:2379\"\n    namespace: dex"; skip=1; next } skip && /^[a-zA-Z0-9_-]+:/ { skip=0 } !skip' /tmp/base.yaml > /tmp/dex.yaml
+  
+  echo "waiting for etcd to be ready..."
+  until curl -sf http://127.0.0.1:2381/health > /dev/null 2>&1; do sleep 1; done
```
chengfang: curl may not be available in dex image.

Makefile
```diff
@@ -212,7 +218,7 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 
 
 start-e2e: install-prometheus-crds ## Start operator for E2E tests (installs required CRDs if needed)
-	ARGOCD_CLUSTER_CONFIG_NAMESPACES="argocd-e2e-cluster-config, argocd-test-impersonation-1-046, argocd-agent-principal-1-051, argocd-agent-agent-1-052, appset-argocd, appset-old-ns, appset-new-ns, appset-argocd-clusterrole, ns-hosting-principal, ns-hosting-managed-agent, ns-hosting-autonomous-agent, gitops-promoter-1-134" make run
+	ARGOCD_CLUSTER_CONFIG_NAMESPACES="argocd-e2e-cluster-config, argocd-test-impersonation-1-046, argocd-agent-principal-1-051, argocd-agent-agent-1-052, appset-argocd, appset-old-ns, appset-new-ns, appset-argocd-clusterrole, ns-hosting-principal, ns-hosting-managed-agent, ns-hosting-autonomous-agent, gitops-promoter-1-134" ARGOCD_DEX_ETCD_STORAGE_ENABLED="false" make run
```
chengfang, line 221: Dex's storage.type enum is memory | kubernetes | etcd | sqlite3 | postgres | mysql. How about changing our feature flag to align with dex, something like: ARGOCD_DEX_STORAGE_TYPE = memory | kubernetes | etcd | sqlite3 | postgres | mysql, though we currently only need first 3.

anandf: Yes, that makes sense. It will help in future switch over where we can support additional configuration types.

controllers/argoutil/dex.go
```diff
@@ -0,0 +1,68 @@
...
+    sleep 15
+    /shared/argocd-dex gendexcfg ${EXTRA_ARGS} -o /tmp/check_base.yaml 2>/dev/null || continue
+    if [ "$(sha256sum < /tmp/base.yaml)" != "$(sha256sum < /tmp/check_base.yaml)" ]; then
+      echo "Configuration change detected in argocd-cm/argocd-secret. Restarting Dex process..."
+      kill -TERM $DEX_PID
+      wait $DEX_PID 2>/dev/null || true
+      break
+    fi
+  done
+done`,
```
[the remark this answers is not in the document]
anandf: This is addressed by adding the suggested change with `kill -0`

controllers/argoutil/dex.go
anandf: yes. it is not available in the upstream image. If `openssl` is not available, we will disable tls. Openssl availabillity is checked via `elif command -v openssl >/dev/null 2>&1; then`, if not available tls is disabled.

anandf: Downstream images of dex that we build, has openssl and curl available.

controllers/argoutil/dex.go
anandf: curl is available in the downstream dex image. I will add a check to skip the curl command if upstream dex does not have it.

Makefile
```diff
@@ -212,7 +218,7 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
...
```
chengfang, line 222: `ARGOCD_DEX_ETCD_STORAGE_ENABLED` is the old name and should be changed to the new one.

controllers/argoutil/dex.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+// getDexStorageType returns the storage type that needs to be used for dex.
+func getDexStorageType() string {
+	if env, found := os.LookupEnv("ARGOCD_DEX_STORAGE_TYPE"); found {
+		return env
+	}
+	return DefaultDexStorageType
+}
```
chengfang, line 101: We could use some validation of the raw env var value.

controllers/argocd/dex.go
```diff
@@ -468,6 +461,76 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoproj.ArgoCD) error {
 		VolumeMounts:    dexVolumeMounts,
 	}}
 
+	if UseDex(cr) && argoutil.IsDexEtcdStorageEnabled() {
+		deploy.Spec.Template.Spec.Containers[0].Command = []string{"/bin/sh", "-c"}
+		deploy.Spec.Template.Spec.Containers[0].Args = argoutil.DexServerCustomStartupScript()
+		deploy.Spec.Template.Spec.Containers = append(deploy.Spec.Template.Spec.Containers, corev1.Container{
```
chengfang, line 463: can we make etcd an init container with restartPolicy: Always, to let kubernetes manage the lifecycle and ordering of etcd and dex containers? So etcd will starts before dex and shutdown after dex?

controllers/argocd/dex.go
```diff
@@ -468,6 +461,76 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoproj.ArgoCD) error {
...
+				TimeoutSeconds:      3,
+				FailureThreshold:    3,
+			},
+			ReadinessProbe: &corev1.Probe{
+				ProbeHandler: corev1.ProbeHandler{
+					HTTPGet: &corev1.HTTPGetAction{
+						Path:   "/readyz",
+						Port:   intstr.FromInt32(2381),
+						Scheme: corev1.URISchemeHTTP,
+					},
```
chengfang, line 517: do we need to add the above http probes to dex networkpolicy to allow them? OCP will likely allow them but will be good for other platforms.

PR #2377, opened 2026-09-14
fix(dex): Added support to rotate serviceaccount token as projected volume mount and not requiring a dex server restart

anandf:
**What type of PR is this?**
/kind bug

Dex server pod restarts the dex server process whenever the static clientSecret changes. This causes existing token verification to fail, as the signing tokens stored in memory gets deleted. To overcome this limitation, external storage options (kubernetes/etcd) was evaluated, though it solved the problem of loosing the dex data, it requires additional care and encryption mechanism to store the sensitive data securely.

Using projected tokens, helps in token rotation directly in the disk by the kubelet and by making an enhancement in dex to read these short-lived tokens directly from the file, helps to avoid the dex server process restarts. It also simplifies the current token rotation mechanism which requires the operator to continuosly look at the service account secrets for any expiry and use the `TokenRequest` API to generate the token stored in a kubernetes `Secret` object. 

Fixes 
https://redhat.atlassian.net/browse/GITOPS-10364
https://redhat.atlassian.net/browse/GITOPS-10429

```
oc apply -f - <<EOF
kind: CatalogSource
metadata:
spec:
  displayName: "GITOPS DEVEL"
  image: quay.io/anjoseph/gitops-operator-catalog:v1.21.3-4
  publisher: "GITOPS DEVEL"
  sourceType: grpc
EOF
```
2. Install the OpenShift GitOps Operator
2.1 Create the namespace and `OperatorGroup` resources.

```
```
2.2 Create the Operator Group
```
oc apply -f - <<EOF
kind: OperatorGroup
metadata:
spec:
  upgradeStrategy: Default
EOF
```
2.3 Create the `Subscription` resource with the required env overrides.
```
oc apply -f - <<EOF
kind: Subscription
metadata:
  labels:
  name: gitops-operator
spec:
  channel: latest
  name: gitops-operator
EOF
```
3. Login to the session using SSO and wait for the service account to be rotated. It will take around 10 mins for the rotation to happen. Keep running the below command every 10 mins to confirm that the rotation is happening directly in the project SA token by the kubelet.
Get the expiry time
```
date -d @$(kubectl exec -it openshift-gitops-dex-server-74fcd9ddf5-dqzqf -c dex -- sh -c 'cat /var/run/secrets/kubernetes.io/serviceaccount/token | cut -d "." -f 2 | base64 -d 2>/dev/null' | jq )
```

Wait for 10 mins after the current time has passed the expiry time. The token and expiry time would get updated.

4. Ensure that the dex server process in the `argocd-dex-server` pod is not being restarted and the user is not being forced to re-login when the service account was rotated.

5. Ensure that no secret named `argocd-dex-server-token` gets created to hold the secret. Since the SA token is rotated directly in file using projected volume mount.

## Summary by CodeRabbit

* **Changes**
  * Dex authentication now uses projected service account tokens instead of manually managed token Secrets.
  * Dex tokens expire after 10 minutes by default.
  * Dex deployments securely mount projected token, CA certificate, and namespace information.
  * Automatic Dex token renewal scheduling and storage of Dex OAuth client secrets in the Argo CD Secret have been removed.


controllers/argocd/dexUtil.go
```diff
@@ -11,6 +11,53 @@ import (
 	"github.com/argoproj-labs/argocd-operator/controllers/argoutil"
 )
 
+const (
+	defaultDexStorageType         = "memory"
+	defaultNoopScriptlet          = "cp /tmp/base.yaml /tmp/dex.yaml"
+	customBootstrapScriptTemplate = `set -eo pipefail
```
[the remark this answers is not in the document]
anandf: This shell script based change is removed now. Using `/shared/argocd-dex rundex` command to run the process with the recent change.

controllers/argocd/dex.go
```diff
@@ -288,12 +169,12 @@ func (r *ReconcileArgoCD) getOpenShiftDexConfig(cr *argoproj.ArgoCD) (string, er
...
-			"clientSecret": "$oidc.dex.clientSecret",
-			"redirectURI":  redirectURI,
-			"insecureCA":   true, // TODO: Configure for openshift CA,
-			"groups":       groups,
+			"issuer":                  "https://kubernetes.default.svc", // TODO: Should this be hard-coded?
+			"clientID":                getDexOAuthClientID(cr),
+			"redirectURI":             redirectURI,
+			"insecureCA":              true, // TODO: Configure for openshift CA,
+			"groups":                  groups,
+			"serviceAccountTokenFile": "/var/run/secrets/kubernetes.io/serviceaccount/token",
```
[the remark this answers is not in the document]
anandf: `serviceAccountTokenFile` is the new field we are trying to add to dex openshift connector configuration via PR https://github.com/dexidp/dex/pull/5023. This change depends on the dex PR being approved and merged.

Rizwana777:
I think this [test](https://github.com/argoproj-labs/argocd-operator/blob/master/tests/ginkgo/parallel/1-095_validate_dex_clientsecret_test.go#L61) will fail because clientSecret: $oidc.dex.clientSecret field has been removed  and token Secret exists with token + expiry which no longer getting created, this test needs to be updated with current changes @anandf

chengfang:
> I think this [test](https://github.com/argoproj-labs/argocd-operator/blob/master/tests/ginkgo/parallel/1-095_validate_dex_clientsecret_test.go#L61) will fail because clientSecret: $oidc.dex.clientSecret field has been removed and token Secret exists with token + expiry which no longer getting created, this test needs to be updated with current changes @anandf

also check `tests/ginkgo/sequential/1-133_validate_namespacemanagement_does_not_block_dex_token_renewal_test.go`, which has the same problem.

controllers/argocd/dex.go
```diff
@@ -388,6 +268,47 @@ func (r *ReconcileArgoCD) reconcileDexDeployment(cr *argoproj.ArgoCD) error {
 				EmptyDir: &corev1.EmptyDirVolumeSource{},
 			},
 		},
+		{
+			Name: "sa-token-volume",
+			VolumeSource: corev1.VolumeSource{
+				Projected: &corev1.ProjectedVolumeSource{
```
chengfang, line 294: similar to `expirationSeconds`, consider explicitly setting `DefaultMode: &projectedDefaultMode,` to make sure its value is consistent between read and write to the api server, and to avoid api server setting to its non-nil default value. This can make sure the `reflect.deepEqual` comparison gives the expected result.

```go
+       // Set explicitly to the value the API server would default to, otherwise the
+       // desired volume never compares equal to the stored one and every reconcile
+       // reports a spurious "volumes" change.
+       projectedDefaultMode := corev1.ProjectedVolumeSourceDefaultMode
```

anandf: Added the `DefaultMode` for projected volumes.

chengfang (approved):

controllers/argocd/dex.go
```diff
@@ -90,6 +90,23 @@ func (r *ReconcileArgoCD) reconcileDexLegacySATokenSecrets(cr *argoproj.ArgoCD)
 			return errors.Join(append([]error{err}, deleteErrs...)...)
 		}
 	}
+	// Bounded short-lived token secrets generated through TokenRequest API must also be removed.
```
chengfang: can you also add this doc line to the func docstring?

controllers/argocd/dex.go
chengfang, line 94: if the SA is not found, the func will return rearly here and the new block of code for secret cleanup will be skipped. Is this intended?

anandf: Correct. I moved the code changes and kept all secret removal together, before proceeding to remove the SA stale reference. That way, this early return should not impact the secret removal.

controllers/argocd/dex.go
anandf: Yes, updated the doc comments too to explain what exactly this function does.

controllers/argocd/dex.go
```diff
@@ -90,6 +90,23 @@ func (r *ReconcileArgoCD) reconcileDexLegacySATokenSecrets(cr *argoproj.ArgoCD)
...
+
+		argoutil.LogResourceDeletion(log, &secret, "removing bound short-lived dex token secret")
+		if err := r.Delete(context.TODO(), &secret); err != nil {
+			if !apierrors.IsNotFound(err) {
+				deleteErrs = append(deleteErrs, fmt.Errorf("delete bound short-lived dex token secret %s/%s: %w", secret.Namespace, secret.Name, err))
+				continue
+			}
+		}
+		deletedSecretNames[secret.Name] = struct{}{}
+	}
```
Rizwana777, line 109: LGTM thanks @anandf

chengfang (approved):

svghadi (approved):
LGTM. Thanks

